PR #3, opened 2015-11-20
datastore: make a simple CLI task list demo

okdave:


okdave:
@broady can you take a look?

datastore/tasks/tasks.go
```diff
@@ -0,0 +1,222 @@
...
+	"strings"
+	"text/tabwriter"
+	"time"
+
+	"golang.org/x/net/context"
+	"google.golang.org/cloud/datastore"
+)
+
+func main() {
+	projID := os.Getenv("DATASTORE_PROJECT_ID")
```
broady: `GCLOUD_DATASET_ID`

broady:
LGTM! I'll merge after these two comments ^

broady:
You actually don't need the region tags if it's just wrapping a single function. Our code include widget has a `go_function` parameter that'll extract a func and its comment.

datastore/tasks/tasks.go
okdave: I'm trying to keep consistent with the other examples (py for example). The team is having this conversation too. Ideally, once it's settled, the datastore.NewClient will "just work™" and we won't need to do this magic here at all.

okdave: @WalterHub any opinion on what I should do here?

okdave:
@WalterHub fixed the end tags

WalterHub:
I don't think the go_function technique will work consistently with the way other languages have been set up (to use a common region tag).

datastore/tasks/tasks.go
WalterHub: Checking.

WalterHub: For v1beta3, DATASTORE_PROJECT_ID needs to be used since it is consistent with what the emulator expects and what the other v1beta3 client lib samples / docs are using.

For post-v1beta3 releases, it can be revisited, but whatever env var is used, it should use _PROJECT_ instead of _DATASET_ (moving away from "dataset").

broady: Ah yes, I found the wrong one. I see GCLOUD_DATASET_ID used in a couple code samples. They should probably be changed to DATASTORE_PROJECT_ID.

broady:
OK. Ready for me to merge?

WalterHub:
Yes from my view.

okdave:
Lock it in.

broady:
locked

PR #5, opened 2015-11-26
datastore/tasks: fix new lines and formatting

okdave:
Accidentally used `Println` when I should have been used `Printf`. :-/


broady:
Merged: bf7694a8754aaac4702b455dbd403a7df1fdda78

(changed author to dgd@golang.org)

PR #19, opened 2016-04-04
docs: add sql/listinstances example

broady:
@adg

### DIFF
--- a/docs/sql/listinstances/listinstances.go
+++ b/docs/sql/listinstances/listinstances.go
@@ -1,0 +1,32 @@
+// Copyright 2015 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package listinstances
+
+import (
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/sqladmin/v1beta4"
+)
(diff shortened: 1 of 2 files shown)


docs/sql/listinstances/listinstances.go
```diff
@@ -0,0 +1,30 @@
+package listinstances
```
adg: Need copyright

docs/sql/listinstances/listinstances.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// Create the Google Cloud SQL service.
+	service, err := sqladmin.New(client)
+	if err != nil {
+		return nil, err
+	}
+	// Create the request to list instances for the project ID.
+	request := service.Instances.List(projectId)
```
adg: make this a one-liner with the next statement?

docs/sql/listinstances/listinstances.go
broady: Done (and for _test)

docs/sql/listinstances/listinstances.go
broady: Done. And cleaned up a bunch of other stuff (like using the scope constant, variable names)

PR #21, opened 2016-04-04
golang-samples: add test for license headers

broady:
@adg

### DIFF
--- a/license_test.go
+++ b/license_test.go
@@ -1,0 +1,52 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package samples
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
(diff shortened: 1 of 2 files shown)


license_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// Break up string so the current file is also tested.
+const sentinel = "Google Inc. All rights reserved." + "\n" +
```
adg: why `" + "\n"` and not just `\n"` ?

broady: Mo `+` , mo problems.

license_test.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+
+		src, err := ioutil.ReadFile(path)
+		if err != nil {
+			return nil
+		}
+
+		// Find license
+		if !bytes.Contains(src, []byte(sentinel)) {
+			t.Errorf("License header not found in %q", path)
```
adg: maybe:

`"%v: license header not present"`

and ditto the next one

PR #27, opened 2016-04-07
travis: add cache configuration for $GOPATH/{src,pkg}

broady:


broady:
Should speed up tests by about 40 seconds.

.travis.yml
```diff
@@ -11,6 +11,15 @@ go:
 - 1.6
 #- tip # NOTE: disabled - see https://github.com/travis-ci/gimme/issues/38
 
+before_cache:
+  - rm -rf $GOPATH/src/github.com/GoogleCloudPlatform/golang-samples
```
adg: why is there a space before these bullets, but not down by 'install:'

adg:
Happy for you to merge this.

.travis.yml
broady: I'm not sure. To be honest, I don't really understand the YAML syntax. I'll try removing them.

PR #29, opened 2016-04-12
Fixes comment about fetching entities.

WalterHub:
Instead of fetching queries.


WalterHub:
I'm a Googler, no need to sign the CLA according to the site.

Thanks,
Walter

broady:
@walterhub you need to update your username in http://github/

PR #32, opened 2016-04-26
docs/appengine: add warmup sample

broady:
PTAL @adg 

Replaces [go/warmupgosample](https://goto.google.com/warmupgosample)


warmup/app.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func init() {
+	http.HandleFunc("/_ah/warmup", warmupHandler)
+}
+
+func warmupHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := appengine.NewContext(r)
+
+	// Do things using ctx.
```
adg: This is an oblique comment. Is the "using ctx" important? It seems to be the focus, when what you really mean is something like

```
// Perform warmup tasks (eg, blah blah).
```

warmup/app.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package sample
```
adg: Should this have a package comment?

warmup/app.go
broady: I was trying to think of what you'd want to do in a warmup request, though. Don't they all involve using a ctx? Otherwise you could just do it in init.

Will still make the comment more generic.

adg: Maybe mention accessing App Engine APIs specifically. 

// Perform warmup tasks that require a context, such as retrieving data from Datastore.

warmup/app.go
broady: done

warmup/app.go
broady: done

adg:
LGTM

PR #33, opened 2016-04-26
docs: move error-reporting/fluent sample

broady:
Moved from [go/errorreportinggosample](https://goto.google.com/errorreportinggosample)

### DIFF
--- a/docs/error-reporting/fluent/main.go
+++ b/docs/error-reporting/fluent/main.go
@@ -1,0 +1,71 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"log"
+	"net/http"
+	"runtime"
+
+	"github.com/fluent/fluent-logger-golang/fluent"
+)
+
+var logger *fluent.Fluent
+
+func init() {
+	// Initializes a Fluentd logger.
+	var err error
+	logger, err = fluent.New(
+		fluent.Config{FluentPort: 24224, FluentHost: "localhost"})
+	if err != nil {
+		panic(err)
+	}
+
+	// Registers a handler.
+	http.HandleFunc("/demo", demoHandler)
+}
+
+func report(stackTrace string, r *http.Request) {
+	payload := map[string]interface{}{
+		"serviceContext": map[string]interface{}{
+			"service": "myapp",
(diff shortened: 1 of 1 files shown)


broady:
@adg

docs/error-reporting/fluent/main.go
```diff
@@ -0,0 +1,71 @@
...
+				"url":       r.URL.String(),
+				"userAgent": r.UserAgent(),
+				"referrer":  r.Referer(),
+				"remoteIp":  r.RemoteAddr,
+			},
+		},
+	}
+	err := logger.Post("myapp.errors", payload)
+	if err != nil {
+		log.Fatal(err)
```
adg: is this the right thing to do? bring down your app because logging failed?

docs/error-reporting/fluent/main.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+// Handler for the incomming requests.
+func demoHandler(w http.ResponseWriter, r *http.Request) {
+	// How to handle an error.
+	defer func() {
```
adg: this is how to handle a *panic*

docs/error-reporting/fluent/main.go
```diff
@@ -0,0 +1,71 @@
...
+		if e := recover(); e != nil {
+			stack := make([]byte, 1<<16)
+			stackSize := runtime.Stack(stack, true)
+			report(string(stack[:stackSize]), r)
+		}
+	}()
+
+	// Error is generated here.
```
adg: panic is raised here

docs/error-reporting/fluent/main.go
```diff
@@ -0,0 +1,71 @@
...
+	"log"
+	"net/http"
+	"runtime"
+
+	"github.com/fluent/fluent-logger-golang/fluent"
+)
+
+var logger *fluent.Fluent
+
+func init() {
```
adg: I'd put all this in func main

broady:
I don't want to change any of this code. I'm just moving it. I think it's a pretty poor example, and I need to talk to the error reporting team to come up with a better one.

adg:
Okay, LGTM but please file a bug to fix this example.

broady:
Did some style changes to make it somewhat respectable, and added a TODO at the top.

Removed that log.Fatal. That's pretty bad.

PR #37, opened 2016-04-27
datastore/tasks: add system tests

broady:


broady:
@okdave 
@adg 

PTAL

datastore/tasks/tasks_test.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+	"log"
+	"os"
+	"testing"
+	"time"
+
+	"golang.org/x/net/context"
+	"google.golang.org/cloud/datastore"
+
```
adg: var client ...

adg:
LGTM modulo my naming suggestion

datastore/tasks/tasks_test.go
broady: Done

PR #43, opened 2016-04-27
monitoring: add system tests, clean up code

broady:


broady:
@waprin PTAL

broady:
Split diff makes this much easier to view:
https://github.com/GoogleCloudPlatform/golang-samples/pull/43/files?diff=split

waprin:
code lgtm. tests seem a little lazy, in the other samples I verified some of the expected output, and for custom metric i wrote a value and read it back and asserted it. however we can always look back on improving them after CDCP.

broady:
well, the functions don't return any values at the moment, so it's not possible without changing the structure of the code.

I don't see a a huge amount of value in reading back the values. These tests might be lazy, but at least they assert that the sample code is making well-formed requests.

broady:
@googlebot go home, you're drunk

PR #47, opened 2016-04-29
README: add index of samples, doc strings for all samples.

broady:


broady:
You can see the new readme here:
https://github.com/GoogleCloudPlatform/golang-samples/blob/index/README.md

PR #50, opened 2016-05-11
getting-started/bookshelf: update to new pubsub package

broady:


broady:
PTAL @mcgreevy @adg

adg:
LGTM

getting-started/bookshelf/pubsub_worker/worker.go
```diff
@@ -29,27 +28,29 @@ var (
...
 	if err != nil {
 		log.Fatalf("could not access Google Books API: %v", err)
 	}
 
 	// ignore returned errors, which will be "already exists". If they're fatal
 	// errors, then following calls (e.g. in the subscribe function) will also fail.
-	pubsub.CreateTopic(pubSubCtx, bookshelf.PubSubTopic)
-	pubsub.CreateSub(pubSubCtx, subName, bookshelf.PubSubTopic, 0, "")
+	bookshelf.PubsubClient.NewTopic(ctx, bookshelf.PubsubTopicID)
+	bookshelf.PubsubClient.NewSubscription(ctx, subName, bookshelf.PubsubClient.Topic(bookshelf.PubsubTopicID), 0, nil)
```
mcgreevy: I think it would be cleaner to pass in the TopicHandle which is returned from NewTopic here.

Of course, that would mean that you should also check the error returned from NewTopic, but I think that would be good practice anyway.

getting-started/bookshelf/pubsub_worker/worker.go
```diff
@@ -29,27 +28,29 @@ var (
...
+
+	subscription = bookshelf.PubsubClient.Subscription(subName)
```
mcgreevy: here, too, just use the value returned from NewSubscription.

getting-started/bookshelf/pubsub_worker/worker.go
broady:
```
 	// ignore returned errors, which will be "already exists". If they're fatal
 	// errors, then following calls (e.g. in the subscribe function) will also fail.
```

I want something like "EnsureExists"

broady: also I'd expect 

topicHandle.NewSubscription(...) since the TopicHandle is required anyway. What do you think?

getting-started/bookshelf/pubsub_worker/worker.go
```diff
@@ -29,27 +28,29 @@ var (
 	countMu sync.Mutex
 	count   int
 
-	booksClient *books.Service
-	pubSubCtx   context.Context
+	booksClient  *books.Service
+	subscription *pubsub.SubscriptionHandle
```
mcgreevy: what's with the global?

getting-started/bookshelf/pubsub_worker/worker.go
```diff
@@ -71,39 +72,37 @@ func main() {
 }
 
 func subscribe() {
```
mcgreevy: This name doesn't really reflect what it does.  Why not something like:

func process(sub *SubscriptionHandle)

?

getting-started/bookshelf/pubsub_worker/worker.go
```diff
@@ -71,39 +72,37 @@ func main() {
...
-			var id int64
-			if err := json.Unmarshal(msg.Data, &id); err != nil {
-				log.Printf("could not decode message data: %#v", msg)
-				go pubsub.Ack(pubSubCtx, subName, msg.AckID)
-				continue
+		log.Printf("[ID %d] Processing.", id)
+		go func() {
+			if err := update(id); err != nil {
+				log.Printf("[ID %d] could not update: %v", id, err)
+				return
```
mcgreevy: You need to call msg.Done here (you can pass it false to NACK).

getting-started/bookshelf/pubsub_worker/worker.go
mcgreevy: The topic handle is required, but a subscription may be in a different project to the topic.

So topicHandle.NewSubscription would need to take a "subscriptionProjID", even though in many cases it would be the same as the topic project id, and no other functions (except NewClient) take a project id.

mcgreevy: > I want something like "EnsureExists"

you mean that is your intention with the existing code, or you want "EnsureExists" to be in the library API?

broady: Both!

getting-started/bookshelf/pubsub_worker/worker.go
broady: Done

PR #51, opened 2016-05-11
Updating/replacing references to "module" with "service"

RichieEscarez:
This code sample is outdated and does not align with the terminology changes that have been published to this app engine topic: https://cloud.google.com/appengine/docs/go/modules/converting


broady:
True. Why do the docs say this, though?

> Services were previously called "modules", and services are still declared in app.yaml files as modules.

broady:
Closing so we don't accidentally merge. I'll re-open once appcfg supports this config flag.

PR #61, opened 2016-06-17
Add files to deploy bookshelf to Container Engine.

jeffmendoza:


broady:
@googlebot yes, it's ok

(how does this thing even work?!)

PR #63, opened 2016-06-19
getting-started/bookshelf: handle MySQL connections better

### LINKED ISSUE #55: getting-started/bookshelf: mysql reconnects not handled
The sample runs "use library" only once when the app starts. If the sql package at any point returns a new connection, the app won't work properly.

I think the fix is to call "sql.Open" again after verifying that the DB exists, specifying the DB name in the DSN.

broady:
* Use one connection to ensure database and table exists, then use a
new DSN for future connections.
* Fix db.Close() calling itself (would cause an infinite, recursive loop)

Fixes #55.

### DIFF
--- a/getting-started/bookshelf/db_mysql.go
+++ b/getting-started/bookshelf/db_mysql.go
@@ -54,7 +54,7 @@ type MySQLConfig struct {
 }
 
 // dataStoreName returns a connection string suitable for sql.Open.
-func (c MySQLConfig) dataStoreName() string {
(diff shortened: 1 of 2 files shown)


broady:
@laixer does this look reasonable?

broady:
@laixer does this look reasonable?

/cc @dgryski, who appears to know the sql package better than me.

laixer:
LGTM Thank you :)

PR #59, opened 2016-06-15
docs/.../sendgrid: update to sendgrid v3 API

broady:

PR #66, opened 2016-06-27
docs/.../sendgrid: pin to v2 sendgrid API

broady:


jonparrott:
LGTM, but how would you feel about just using their API directly?

broady:
@jonparrott I wrote that, too, but it seemed worse.

https://github.com/GoogleCloudPlatform/golang-samples/pull/59/files

jonparrott:
Ah I see, well, carry on.

broady:
Maybe for later.

PR #71, opened 2016-07-19
monitoring: use pointer for typed value

broady:
The API client was changed:
https://code-review.googlesource.com/5259


broady:
@waprin @mcgreevy

waprin:
lgtm

monitoring/custommetric/custommetric.go
```diff
@@ -88,7 +89,7 @@ func writeTimeSeriesValue(s *monitoring.Service, projectID, metricType string) e
 					EndTime:   now,
 				},
 				Value: &monitoring.TypedValue{
-					Int64Value: rand.Int63n(10),
+					Int64Value: &randVal,
```
mcgreevy: Alternatively, this could be:

Int64Value: googleapi.Int64(rand.Int63n(10))

broady: Yeah, I figured. I didn't want to import another package. But I guess that's why it's there.

PR #75, opened 2016-08-03
natural_language: add analyze command

broady:


broady:
Maybe this should be `golang-samples/language/analyze` (drop the natural, it's cleaner)

language/analyze/README.md
```diff
@@ -0,0 +1,67 @@
...
+  `GOOGLE_APPLICATION_CREDENTIALS` environment variable:
+
+  ```bash
+  export GOOGLE_APPLICATION_CREDENTIALS=/path/to/your-project-credentials.json
+  ```
+
+[cloud-console]: https://console.cloud.google.com
+[project]: https://cloud.google.com/natural-language/docs/getting-started#set_up_your_project
+
+## Run the sample
```
rakyll: Should we suggest them to go get the repo and run the installed binary instead?

language/analyze/analyze.go
```diff
@@ -0,0 +1,104 @@
...
+
+func printResp(v interface{}, err error) {
+	if err != nil {
+		log.Fatal(err)
+	}
+	b, err := json.MarshalIndent(v, "", "    ")
+	if err != nil {
+		log.Fatal(err)
+	}
+	os.Stdout.Write(b)
```
rakyll: Why not fmt.Println?

language/analyze/analyze_test.go
```diff
@@ -0,0 +1,84 @@
...
+	c := newClient(t)
+
+	res, err := analyzeEntities(c, "Homer Simpson likes donuts.")
+	if err != nil {
+		t.Fatalf("got %v, want nil err", err)
+	}
+	found := false
+	for _, e := range res.Entities {
+		if e.Name == "Homer Simpson" {
+			found = true
```
rakyll: Can we just return here? Then, there is no need for the found variable.

language/analyze/analyze_test.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		t.Fatalf("got %v, want nil err", err)
+	}
+
+	found := false
+	for _, tok := range res.Tokens {
+		if tok.Lemma == "gopher" {
+			found = true
+			if tok.PartOfSpeech.Tag != "NOUN" {
+				t.Errorf("PartOfSpeech: got %+v, want NOUN", tok.PartOfSpeech.Tag)
```
rakyll: same here

rakyll:
Left a few comments, otherwise LGTM.

language/analyze/README.md
broady: Yes, I think so. I want to add that for all of the other samples, too, though, and to do it consistently. So I'll add that later.

All of the docs on cloud.google.com should suggest `go get`, though.

language/analyze/analyze.go
broady: `fmt.Println(string(b))`?

language/analyze/analyze_test.go
broady: Done

language/analyze/analyze_test.go
broady: Ah yeah, that's nicer.

language/analyze/analyze.go
rakyll: It is ugly but I'd make sure that there is a new line in the end when command prints the JSON and terminates.

broady: `MarshalIndent` uses output with a newline at the end.

language/README.md
```diff
@@ -0,0 +1,11 @@
+# Cloud Natural Language API Go samples
+
+This directory contains [Cloud Natural Language API](https://cloud.google.com/language/) Go samples.
```
jerjou: /natural-language/
I think it's either a branding thing or an seo thing. I think the idea is, for urls and such, use the full name, but for code (eg package names) it's okay to just use 'language'

language/analyze/analyze.go
```diff
@@ -0,0 +1,104 @@
...
+	os.Exit(2)
+}
+
+func analyzeEntities(s *language.Service, text string) (*language.AnalyzeEntitiesResponse, error) {
+	req := s.Documents.AnalyzeEntities(&language.AnalyzeEntitiesRequest{
+		Document: &language.Document{
+			Content: text,
+			Type:    "PLAIN_TEXT",
+		},
+		EncodingType: "UTF8",
```
jerjou: Will command-line arguments necessarily be encoded in utf8, or is it platform-dependent?

language/README.md
broady: Oops!

language/analyze/analyze.go
broady: All strings in Go are UTF-8.

jerjou: It's confusing :)
Take a look at the [tests](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/language/api/analyze_test.py#L63) in python-docs-samples. I tried to demonstrate what passing in different encodings does there.

Basically, it's so that the api can calculate offsets into the string for a detected entity/whatever. However, over the wire everything is encoded in utf8, so if - when you get it back as a string - your original string was encoded in another encoding, then utf8 offsets wouldn't work for your string. So you have to tell the api what your string is encoded in, so it can calculate offsets for your string.

broady: Got it. So it's not a character offset, it's a byte offset?

jerjou: ...almost. That's what I thought initially too, but that'd be too simple to explain ;)
Conceptually, it's an index into the 'character'. But because of variable-width encodings (and how indexing works in the languages), things get hairy.
So, for UTF8 - yes. Because the lowest unit of 'character' is a byte, it's an index into the byte string.
For UTF16 (eg in javascript), the lowest unit of 'character' is 2 bytes, so it's an index into an array where each slot in the array is 2-bytes. eg if I have the string `var str = "😶abc";`, where `😶` takes up 4 bytes, and `a` takes up 2, the index for `a` would be 2, since `😶` takes up slots 0 and 1.

PR #76, opened 2016-08-04
all: update to new cloud.google.com package path

broady:


broady:
Also fixes an incorrect import path in an appengine sample (it imported cloud/datastore instead of appengine/datastore)

docs/appengine/datastore/consistency/consistency.go
```diff
@@ -6,7 +6,7 @@ package sample
 
 import (
 	"golang.org/x/net/context"
-	"google.golang.org/cloud/datastore"
+	"google.golang.org/appengine/datastore"
```
rakyll: Why is this importing appengine now?

broady: It was wrong before (see comment above)

rakyll: Ack

rakyll:
LGTM

broady:
Waiting for build.

PR #80, opened 2016-08-18
speech/transcript: initial commit

rakyll:
The program will be upgraded to use the streaming API and send content in smaller chunks.

/cc @broady


speech/transcript/transcript.go
```diff
@@ -0,0 +1,78 @@
+// Copyright 2015 Google Inc. All rights reserved.
```
broady: 2016

speech/transcript/transcript.go
```diff
@@ -0,0 +1,78 @@
...
+	if len(os.Args) < 2 {
+		fmt.Fprintln(os.Stderr, usage)
+		os.Exit(2)
+	}
+
+	ctx := context.Background()
+	conn, err := transport.DialGRPC(ctx,
+		option.WithEndpoint("speech.googleapis.com:443"),
+		option.WithScopes("https://www.googleapis.com/auth/cloud-platform"),
+	)
```
broady: Hmmm - `transport` isn't really meant to be a public package.

Maybe it should be, though. It looks pretty good here.

speech/transcript/transcript.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		panic(err)
```
broady: `log.Fatal`

speech/transcript/transcript.go
```diff
@@ -0,0 +1,78 @@
...
+			Encoding:   speech.InitialRecognizeRequest_LINEAR16,
+			SampleRate: 16000,
+		},
+		AudioRequest: &speech.AudioRequest{Content: data},
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	// print the results
```
broady: Full sentences for comments (capitalize first word, period at the end).

speech/transcript/transcript.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, resp := range rresp.Responses {
+		if resp.Error != nil {
+			fmt.Fprintf(os.Stderr, "error in recognize response: %v\n", resp.Error)
```
broady: Hmmm, what actually is this error?

How come there are multiple responses?

speech/transcript/transcript.go
rakyll: It is so useful that I thought it is designed to be public.

broady:
LGTM otherwise. I suggest making a function like

`func Recognize(ctx context.Context, client speech.SpeechClient, *data []byte) (*NonStreamingRecognizeResponse, error)`

speech/transcript/transcript.go
broady: Actually, it seems pretty fair to use it here until we have the generated gRPC clients. The huge plus is handling ADC.

rakyll:
Done with all and added tests.

speech/transcript/transcript_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"io/ioutil"
+	"testing"
+
+	"golang.org/x/net/context"
+	"google.golang.org/api/option"
+	"google.golang.org/api/transport"
+	speech "google.golang.org/genproto/googleapis/cloud/speech/v1"
+)
+
+func TestRecognize(t *testing.T) {
```
broady: `testutil.SystemTest(t)`

speech/transcript/transcript_test.go
```diff
@@ -0,0 +1,56 @@
...
+		t.Fatalf("got error from response: %v", err)
+	}
+	if len(resp.Results) < 1 {
+		t.Fatal("want results; got none")
+	}
+	if len(resp.Results[0].Alternatives) < 1 {
+		t.Fatal("want alternatives; got none")
+	}
+	if got, want := resp.Results[0].Alternatives[0].Transcript, "quit"; got != want {
+		t.Errorf("want transcript: %q; got %q", want, got)
```
broady: nit: `got; want` rather than `want; got`

(I agree with you, though, I prefer the way you did it, but apparently this is the preferred way)

broady: Also maybe print out the whole of the resp, it'll be good for debugging.

broady:
:+1:

PR #88, opened 2016-08-23
bigquery: add synchronous query sample

tswast:
@broady Please take a look.

I would have added an "asynchronous query" like https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/bigquery/api/async_query.py but I couldn't figure out how to use the "jobs" API with the bigquery client. Filed: https://github.com/GoogleCloudPlatform/gcloud-golang/issues/327

Also, I wanted to support standard SQL syntax instead of or in addition to legacy SQL syntax (the default), but I couldn't find a parameter for it. Filed https://github.com/GoogleCloudPlatform/gcloud-golang/issues/326

### DIFF
--- a/bigquery/syncquery/syncquery.go
+++ b/bigquery/syncquery/syncquery.go
@@ -1,0 +1,64 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Command syncquery queries a Google BigQuery dataset.
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"cloud.google.com/go/bigquery"
+
+	"golang.org/x/net/context"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Println("usage: ", os.Args[0], "'query text'")
+		os.Exit(1)
+	}
+	proj := os.Getenv("GCLOUD_PROJECT")
+	if proj == "" {
+		fmt.Println("GCLOUD_PROJECT environment variable must be set.")
+		os.Exit(1)
+	}
+
+	rows, err := Query(proj, os.Args[1])
+	if err != nil {
+		log.Fatalln(err)
+	}
+	for _, row := range rows {
+		fmt.Println(row)
+	}
+}
+
+// Query returns a slice of the reults of a query.
+func Query(proj, q string) ([]bigquery.ValueList, error) {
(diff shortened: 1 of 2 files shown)


broady:
Can you move this to just `bigquery` in the root? I want to drop the `docs` directory eventually.

docs/bigquery/syncquery/syncquery_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	rows, err := Query(tc.ProjectID, "SELECT corpus FROM publicdata:samples.shakespeare GROUP BY corpus;")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	found := false
+	for _, row := range rows {
+		if row[0] == "romeoandjuliet" {
+			found = true
```
broady: I like this, too, but you can also use `return` and then just `t.Errorf` at the end, to avoid the flag.

docs/bigquery/syncquery/syncquery.go
```diff
@@ -0,0 +1,64 @@
...
+		if err := iter.Get(&row); err != nil {
+			return nil, err
+		}
+		rows = append(rows, row)
+	}
+
+	return rows, iter.Err()
+}
+
+func main() {
```
broady: Move main to the top

docs/bigquery/syncquery/syncquery.go
```diff
@@ -0,0 +1,64 @@
...
+	if len(os.Args) < 2 {
+		fmt.Println("usage: ", os.Args[0], "'query text'")
+		os.Exit(1)
+	}
+	proj := os.Getenv("GCLOUD_PROJECT")
```
broady: What do you think of `BIGQUERY_PROJECT_ID`?
I'm really not sure what we should do about this, generally.
I'm also fine with `GCLOUD_PROJECT`.

broady:
LGTM

docs/bigquery/syncquery/syncquery_test.go
tswast: > use return and then just t.Errorf at the end
done.

docs/bigquery/syncquery/syncquery.go
tswast: > Move main to the top

Done.

broady:
go away @googlebot, you're drunk

broady:
👏 👏 👏  thx @tswast

PR #95, opened 2016-08-30
speech/transcript: rename to caption

rakyll:
So it also aligns with the streaming streaming sample, livecaption.

/cc @broady


broady:
LGTM, but also change `README.md`

rakyll:
Ok, done. Waiting for the build to finish before merging.

PR #98, opened 2016-08-31
logging: add simplelog sample

broady:


broady:
cc @jmdobry @jonparrott @pfritzsche

broady:
Tests coming now. Take a look and see if this is reasonable first, though?

logging/simplelog/simplelog.go
```diff
@@ -0,0 +1,134 @@
...
+		// For example, if Flush() failed.
+		log.Printf("client.OnError: %v", err)
+	}
+	// [END setup]
+
+	log.Print("Writing some log entries.")
+	writeEntry(client)
+	structuredWrite(client)
+
+	// It takes some time for the log entries to appear.
```
jonparrott: In Python we did these as separate functions/commands, which avoids the necessity of a sleep in the code.

broady: I saw. I like it. This seems OK, too. You just "go run" and it does everything.

jonparrott: Sure - either way. I'm neurotic about sleeps in code, but it's fine if you're cool with. :)

logging/simplelog/simplelog.go
```diff
@@ -0,0 +1,134 @@
+// iample simplelog writes some entries, lists them, then deletes the log.
```
rakyll: Sample simplelog...

logging/simplelog/simplelog.go
```diff
@@ -0,0 +1,134 @@
...
+	"cloud.google.com/go/preview/logging"
+
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+	// [END imports]
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Println("Usage: simplelog <project-id>")
```
rakyll: We use an env variable for the project name in some references, an argument in some other. Why not using GCLOUD_PROJECT env var everywhere?

logging/simplelog/simplelog.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+func writeEntry(client *logging.Client) {
+	// [START write_log_entry]
+	const name = "log-example"
+	logger := client.Logger(name)
+	defer logger.Flush() // Ensure the entry is written.
+
+	stdlog := logger.StandardLogger(logging.Info)
```
rakyll: s/stdlog/infolog would be nicer. We are expecting users to construct multiple loggers for each log level.

logging/simplelog/simplelog.go
```diff
@@ -0,0 +1,134 @@
...
+			entry.Payload)
+	}
+
+	log.Print("Deleting log.")
+	if err := deleteLog(client); err != nil {
+		log.Fatalf("Could not delete log: %v", err)
+	}
+}
+
+func writeEntry(client *logging.Client) {
```
rakyll: Should we call this writeWithStdLogger something?

logging/simplelog/simplelog.go
```diff
@@ -0,0 +1,134 @@
...
+	// [END write_log_entry]
+}
+
+func deleteLog(client *logging.Client) error {
+	ctx := context.Background()
+
+	// [START delete_log]
+	const name = "log-example"
+	if err := client.DeleteLog(ctx, name); err != nil {
+		return err
```
rakyll: Wanna log.Fatal here and not returning error from deleteLog?

logging/simplelog/simplelog.go
broady: Done

logging/simplelog/simplelog.go
broady: It's for a snippet, so I thought `return err` might be better without any context. What do you think?

logging/simplelog/simplelog.go
broady: Done. Changed the message a little, too.

logging/simplelog/simplelog.go
broady: Done

logging/simplelog/simplelog.go
rakyll: yea, ok, we don't know what user want to do with the error.

logging/simplelog/simplelog.go
```diff
@@ -0,0 +1,152 @@
+// Sample simplelog writes some entries, lists them, then deletes the log.
```
rakyll: Missing license header.

logging/simplelog/simplelog.go
```diff
@@ -0,0 +1,156 @@
...
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+	// [END imports]
+)
+
+func main() {
+	if len(os.Args) != 3 {
+		if len(os.Args) == 2 {
+			usage("Missing command.")
+		} else {
```
rakyll: You don't need the else. usage exists.

if len(os.Args) == 2 {
   usage("...")
}

usage("")

rakyll: Actually, let's make this easier.

if len(os.Args) == 2 {
   usage("...")
}

if len(os.Args) != 3 {
    usage("")
}

rakyll:
LGTM with a nits.

logging/simplelog/simplelog.go
broady: Done

logging/simplelog/simplelog.go
broady: Yes, the second one is better. It isn't obvious `usage` exits, so I added the else block.

broady:
Added tests. Have a look?

They'll be quite flaky, unfortunately. Let's see how bad it is in practice. Might be OK for now (like pubsub).

logging/simplelog/simplelog_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	time.Sleep(5 * time.Second)
+
+	entries, err := getEntries(client, tc.ProjectID)
+	if err != nil {
+		t.Fatalf("getEntries: %v", err)
+	}
+
+	if got, want := len(entries), 2; got != want {
+		t.Fatalf("len(entries): got %d; want %d", got, want)
```
rakyll: len(entries) = %d; want %d

No need for got when you don't have to explain what you got in English.

logging/simplelog/simplelog_test.go
```diff
@@ -0,0 +1,71 @@
...
+	wantContain := map[string]*logging.Entry{
+		"Anything":                            entries[0],
+		"The payload can be any type!":        entries[0],
+		"infolog is a standard Go log.Logger": entries[1],
+	}
+
+	for want, entry := range wantContain {
+		msg := fmt.Sprintf("%s", entry.Payload)
+		if !strings.Contains(msg, want) {
+			t.Errorf("want %q to contain %q", msg, want)
```
rakyll: Is entry.Payload encoded in JSON? Why can't we assert the JSON body rather than a strings.Contains?

rakyll:
LGTM with a few nits.

logging/simplelog/simplelog_test.go
broady: Nope, it comes back as a proto struct (also, undocumented).

PR #100, opened 2016-09-01
pubsub: Add subscriptions

rakyll:
/cc @broady

### DIFF
--- a/pubsub/subscriptions/main.go
+++ b/pubsub/subscriptions/main.go
@@ -1,0 +1,138 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Command subscriptions is a tool to manage Google Cloud Pub/Sub subscriptions by using the Pub/Sub API.
+// See more about Google Cloud Pub/Sub at https://cloud.google.com/pubsub/docs/overview.
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"time"
+
+	// [START imports]
+	"golang.org/x/net/context"
+
+	"cloud.google.com/go/pubsub"
+	// [END imports]
+)
+
+func main() {
+	ctx := context.Background()
+	// [START auth]
+	proj := os.Getenv("GOOGLE_CLOUD_PROJECT")
+	if proj == "" {
+		fmt.Fprintf(os.Stderr, "GOOGLE_CLOUD_PROJECT environment variable must be set.\n")
+		os.Exit(1)
+	}
+	client, err := pubsub.NewClient(ctx, proj)
+	if err != nil {
+		log.Fatalf("Could not create pubsub Client: %v", err)
+	}
+	// [END auth]
+
+	// Print all the subscriptions in the project.
+	fmt.Println("Listing all subscriptions from the project:")
+	for _, sub := range list(client) {
+		fmt.Printf("%v\n", sub.Name())
+	}
+
+	const topic = "example-topic"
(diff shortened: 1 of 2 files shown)


pubsub/subscriptions/main.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	fmt.Printf("Created subscription: %v\n", sub)
+	// [END create_subscription]
+	return nil
+}
+
+func delete(c *pubsub.Client, name string) error {
+	ctx := context.Background()
+	// [START delete_subscription]
+	sub := c.Subscription(name)
```
broady: s/c/client/g

pubsub/subscriptions/main.go
```diff
@@ -0,0 +1,150 @@
...
+	if err := sub.Delete(ctx); err != nil {
+		return fmt.Errorf("Failed to delete the subscription: %v\n", err)
```
broady: I don't think we need to wrap the errors for the samples. Also, don't put newlines at the end of wrapper errors.

pubsub/subscriptions/main.go
```diff
@@ -0,0 +1,150 @@
...
+	it, err := sub.Pull(ctx)
+	if err != nil {
+		return fmt.Errorf("Failed to pull from subsription: %v", err)
+	}
+	defer it.Stop()
+
+	// Consume 10 messages.
+	for i := 0; i < n; i++ {
+		msg, err := it.Next()
+		if err == pubsub.Done {
```
broady: pubsub.Done will never be returned here.

pubsub/subscriptions/main.go
```diff
@@ -0,0 +1,150 @@
...
+	// [START pull_messages]
+	sub := c.Subscription(name)
+	it, err := sub.Pull(ctx)
+	if err != nil {
+		return fmt.Errorf("Failed to pull from subsription: %v", err)
+	}
+	defer it.Stop()
+
+	// Consume 10 messages.
+	for i := 0; i < n; i++ {
```
broady: n is outside the scope of the region tag

pubsub/subscriptions/main.go
rakyll: Done

pubsub/subscriptions/main.go
rakyll: Done

pubsub/subscriptions/main.go
rakyll: Unless another client is also consuming the topic.

We should keep it for having a reference how to use the iterator.

pubsub/subscriptions/main.go
rakyll: Hardcoded. Not ideal but these are snippets.

rakyll:
PTAL

Also added tests.

pubsub/subscriptions/main.go
broady: It's only returned after Stop is called. It's actually pretty strange that `pubsub.Done` exists at all.

pubsub/subscriptions/main.go
```diff
@@ -0,0 +1,149 @@
...
+		fmt.Printf("%v\n", sub.Name())
+	}
+
+	const topic = "example-topic"
+	// Create a topic to subscribe to.
+	t, err := client.NewTopic(ctx, topic)
+	if err != nil {
+		log.Fatalf("Failed to create the topic: %v", err)
+	}
+	defer t.Delete(ctx) // cleanup when finished using t.
```
broady: defer is not called when `log.Fatal` is called

broady:
LGTM

Issue #104, opened 2016-09-02
pubsub: flaky test, bad error message


broady:
```
main_test.go:59: got [0xc8204e2440 0xc8204e2460]; want the list with topic = "golang-samples-topic-example""
```

broady:
```
main_test.go:94: got []; want a list with subscription "golang-samples-subscription"
```

rakyll:
What is running against this project test account's pubsub?

We probably need isolation of names for these tests. I am not thinking of any other issue why they are flaky.

broady:
Just these tests. But they might run concurrently.

rakyll:
I will try to namespace everything and delete stuff I create only in the lifecycle of these tests.

PR #111, opened 2016-09-09
pubsub/subscriptions: make list less flaky

broady:
Updates #104.

### DIFF
--- a/pubsub/subscriptions/main_test.go
+++ b/pubsub/subscriptions/main_test.go
@@ -81,7 +81,8 @@ func TestList(t *testing.T) {
 
 	var ok bool
 	var subs []*pubsub.Subscription
-	for attempts := 0; !ok && attempts < 5; attempts++ {
+outer:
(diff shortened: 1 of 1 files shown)


pubsub/subscriptions/main_test.go
```diff
@@ -78,20 +79,29 @@ func TestCreate(t *testing.T) {
...
-		t.Fatalf("failed to list subscriptions: %v", err)
-	}
 	var ok bool
-	s := c.Subscription(subName)
-	for _, sub := range subs {
-		if s.Name() == sub.Name() {
-			ok = true
-			break
+	var subs []*pubsub.Subscription
+	for attempts := 0; !ok && attempts < 5; attempts++ {
```
rakyll: Do we need `!ok`? We are breaking when we find it.

rakyll:
LGTM with a nit.

pubsub/subscriptions/main_test.go
broady: we don't. removing

PR #115, opened 2016-09-14
pubsub: fix the broken build

rakyll:
Related to the change at https://code-review.googlesource.com/#/c/7478.


rakyll:
FYI, @broady.

PR #116, opened 2016-09-14
all: add retry support for flaky tests

broady:
Test output is only written for the final attempt.

A similar interface to *testing.T is provided to each attempt.

```
$ go test -v ./internal/testutil/
=== RUN   TestRetry
--- PASS: TestRetry (0.00s)
=== RUN   TestRetryAttempts
--- PASS: TestRetryAttempts (0.00s)
    retry.go:29: Success after 5 attempts:
        retry_test.go:24: This line should appear only once.
        retry_test.go:25: attempt=5
PASS
ok      github.com/GoogleCloudPlatform/golang-samples/internal/testutil 0.023s
```

### DIFF
--- a/internal/testutil/retry.go
+++ b/internal/testutil/retry.go
@@ -1,0 +1,108 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package testutil
+
+import (
+	"bytes"
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"testing"
+	"time"
+)
+
+func Flaky(t *testing.T, max int, sleep time.Duration, f func(r *R)) bool {
+	for attempt := 1; attempt <= max; attempt++ {
+		r := &R{t: t, Attempt: attempt, log: &bytes.Buffer{}}
+
+		last := attempt == max
+		done := make(chan bool)
+		go func() {
+			defer close(done)
+			f(r)
+		}()
+		<-done
+		success := !r.retry && !r.fail
+
+		if success || last || r.fail {
+			state := "FAIL"
+			if success {
+				state = "SUCCESS"
+				if r.log.Len() == 0 {
+					return true
+				}
+			}
+			t.Logf("Attempt %d: %s%s", attempt, state, r.log.String())
+			if !success {
+				t.Fail()
+			}
+		}
+		if success {
+			return true
+		}
+		if !r.retry {
+			break
+		}
+		time.Sleep(sleep)
+	}
+	return false
+}
+
+type R struct {
+	Attempt int
+
+	t     *testing.T
+	retry bool
+	fail  bool
+	log   *bytes.Buffer
+}
+
+func (r *R) Retry() {
+	r.retry = true
+}
+
+func (r *R) RetryNow() {
+	r.retry = true
+	runtime.Goexit()
+}
+
+func (r *R) FailNow() {
+	r.fail = true
+	runtime.Goexit()
+}
+
+func (r *R) Fail() {
+	r.fail = true
+}
+
+func (r *R) Retryf(s string, v ...interface{}) {
+	r.logf(s, v...)
+	r.Retry()
+}
+
+func (r *R) Fatalf(s string, v ...interface{}) {
+	r.logf(s, v...)
+	r.Fail()
+	runtime.Goexit()
+}
+
+func (r *R) Logf(s string, v ...interface{}) {
+	r.logf(s, v...)
+}
+
+func (r *R) logf(s string, v ...interface{}) {
+	fmt.Fprint(r.log, "\n")
+	fmt.Fprint(r.log, lineNumber())
+	fmt.Fprintf(r.log, s, v...)
+}
+
+func lineNumber() string {
+	_, file, line, ok := runtime.Caller(3) // logf, public func, user function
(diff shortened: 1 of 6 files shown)


internal/testutil/retry.go
```diff
@@ -0,0 +1,108 @@
...
+	"bytes"
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"testing"
+	"time"
+)
+
+func Flaky(t *testing.T, max int, sleep time.Duration, f func(r *R)) bool {
```
rakyll: sleep or some exponential backoff options?

internal/testutil/retry.go
```diff
@@ -0,0 +1,108 @@
...
+	log   *bytes.Buffer
+}
+
+func (r *R) Retry() {
+	r.retry = true
+}
+
+func (r *R) RetryNow() {
+	r.retry = true
+	runtime.Goexit()
```
rakyll: do you want to exit here? why?

internal/testutil/retry.go
```diff
@@ -0,0 +1,108 @@
...
+func (r *R) FailNow() {
+	r.fail = true
+	runtime.Goexit()
+}
+
+func (r *R) Fail() {
+	r.fail = true
+}
+
+func (r *R) Retryf(s string, v ...interface{}) {
```
rakyll: This makes things a bit confusing. Not sure if the convince is worth it.

internal/testutil/retry.go
broady: Changed to prefer `return` in the calling code.

internal/testutil/retry.go
broady: It just sleeps between.

No need for exponential backoff - that's for production code. These are tests.

internal/testutil/retry.go
broady: Changed to `Logf` and `Failf`.

broady:
@rakyll PTAL
@okdave any comments on the syntax?

`testutil.Flaky(retries, sleep, func)`

internal/testutil/retry.go
```diff
@@ -0,0 +1,81 @@
...
+	log    *bytes.Buffer
+}
+
+// Fail marks the run as failed, and will retry once the function returns.
+func (r *R) Fail() {
+	r.failed = true
+}
+
+// Failf is equivalent to Logf followed by Fail.
+func (r *R) Failf(s string, v ...interface{}) {
```
okdave: Why not `Errorf` like T?

It doesn't seem like a future `FailNow` and `Fatalf` would be be unrealistic?

internal/testutil/retry.go
```diff
@@ -0,0 +1,81 @@
...
+
+		if !r.failed {
+			if r.log.Len() != 0 {
+				t.Logf("Success after %d attempts:%s", attempt, r.log.String())
+			}
+			return true
+		}
+
+		if attempt == maxAttempts {
+			t.Logf("FAILED after %d attempts:%s", attempt, r.log.String())
```
okdave: Should you keep the original log lines (as a slice?) so you can actually replay these as errors/logs directly on T?

internal/testutil/retry.go
broady: Indeed. I originally had `FailNow`. Removed at the request of @rakyll to simplify.

`Errorf` seems fine, I wanted to make it clear that it would trigger a retry, and not simply fail. I originally differentiated between failing hard (e.g, some part of the function that isn't flaky) and failing soft (intermittent errors).

internal/testutil/retry.go
broady: Replaying onto T loses the context.

By printing the whole thing in one call to Logf, you get the nicer indented output (see the top post on this pull request).

internal/testutil/retry.go
jba: I'm worried that people will think Failf is magic like Fatalf (they look similar), and exits the test function without terminating the program.

I think Errorf is better because everyone will know that control continues from that point, and they must return.

broady: I agree. I can implement Errorf, Fatalf, and FailNow.

Can also do Log/Fatal, but we hardly use those and using Logf/Fatalf is pretty close.

zombiezen: +1 for making `*R` implement similar methods to `*testing.T`.  There's a reason interfaces like `testing.TB` exist.

internal/testutil/retry.go
```diff
@@ -0,0 +1,81 @@
...
+	"runtime"
+	"strconv"
+	"testing"
+	"time"
+)
+
+// Flaky runs function f for up to maxAttempts times until f returns successfully, and reports whether f was run successfully.
+// It will sleep for the given period between invocations of f.
+// Use the provided *testutil.R instead of a *testing.T from the function.
+func Flaky(t *testing.T, maxAttempts int, sleep time.Duration, f func(r *R)) bool {
```
rakyll: What about calling this just Retry?

It is a general purpose retrying thing.

internal/testutil/retry.go
rakyll: +1 for being consistent with testing.T.

internal/testutil/retry.go
broady: Do we need to retry anything other than flaky tests? If not, I'd rather call it Flaky. It makes the intent clearer.

PR #119, opened 2016-09-16
vision/label: use cloud.google.com/go/vision package

broady:


broady:
@jerjou once merged, will need to update the docs - I added region tags for ease.

vision/label/label.go
```diff
@@ -6,59 +6,53 @@
...
-	res, err := service.Images.Annotate(batch).Do()
+	// Perform the request.
+	annotations, err := client.DetectLabels(ctx, image, 10)
 	if err != nil {
 		return nil, err
 	}
-
+	// [END request]
+	// [START transform]
 	labels := make([]string, 0)
```
jba: I would prefer we write

  var labels []string

everywhere. It's shorter and more idiomatic.

broady: Good catch. Done.

PR #120, opened 2016-09-16
speech/caption: use cloud.google.com/go/speech/apiv1beta1 package

broady:


broady:
Also adds region tags, simplifies test code a bit, reorg code a bit to better match the Vision sample.

speech/caption/caption.go
```diff
@@ -12,10 +12,12 @@ import (
 	"log"
 	"os"
 
+	// [START imports]
 	"golang.org/x/net/context"
-	"google.golang.org/api/option"
-	"google.golang.org/api/transport"
-	speech "google.golang.org/genproto/googleapis/cloud/speech/v1beta1"
+
+	speech "cloud.google.com/go/speech/apiv1beta1"
```
jerjou: `apiv1beta1`....😒

broady: Yeah. It's not pretty, but it's the best we've got. I can explain why over chat.

PR #125, opened 2016-09-20
speech: add speech API sample with the async req/resp

rakyll:


speech/captionasync/captionasync.go
```diff
@@ -0,0 +1,120 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Command captionasync sends an audio data to the Google Speech API
```
jba: delete "an"

speech/captionasync/captionasync.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+const usage = `Usage: captionasync <audiofile>
+
+Audio file is required to be 16-bit signed little-endian encoded
+with a sample rate of 16000.
+`
+
+func main() {
+	flag.Parse()
```
jba: Why, if there are no flags?

speech/captionasync/captionasync.go
```diff
@@ -0,0 +1,120 @@
...
+	// Print the results.
+	for _, result := range resp.Results {
+		for _, alt := range result.Alternatives {
+			fmt.Printf("\"%v\" (confidence=%3f)\n", alt.Transcript, alt.Confidence)
+		}
+	}
+}
+
+func send(client *speech.Client, filename string) (string, error) {
+	ctx := context.Background()
```
jba: This should take a context as first arg. It doesn't really matter in a small program like this, but people will be looking to these for good style, and we should be encouraging passing of contexts. They are also likely to cut and paste this function for use in their own programs, where it may be important to pass the context.

speech/captionasync/captionasync.go
```diff
@@ -0,0 +1,120 @@
...
+
+	op, err := client.AsyncRecognize(ctx, req)
+	if err != nil {
+		return "", err
+	}
+	return op.Name, nil
+}
+
+func wait(client *speech.Client, opName string) (*speechpb.AsyncRecognizeResponse, error) {
+	ctx := context.Background()
```
jba: ditto

speech/captionasync/captionasync.go
```diff
@@ -0,0 +1,120 @@
...
+	var err error
+	for {
+		op, err = opClient.GetOperation(ctx, &longrunningpb.GetOperationRequest{
+			Name: opName,
+		})
+		if err != nil {
+			return nil, err
+		}
+		if op.Done {
+			break // operation done
```
jba: I can think of many places in this code where I wish there were comments, but this is not one of them!

I think the async nature of this API, and the use of long-running ops, are more unusual and worthy of comment.

Not that you should remove this -- it's OK.

speech/captionasync/captionasync.go
```diff
@@ -0,0 +1,120 @@
...
+	case op.GetResponse() != nil:
+		var resp speechpb.AsyncRecognizeResponse
+		if err := proto.Unmarshal(op.GetResponse().Value, &resp); err != nil {
+			return nil, err
+		}
+		return &resp, nil
+	}
+
+	// should never happen.
+	panic("unreachable")
```
jba: You could avoid this by writing

if err := op.GetError(); err != nil {
    return ...
}
[proceed here assuming op.GetResponse() != nil]

I'd prefer that. ymmv

speech/captionasync/captionasync.go
rakyll: Removed!

speech/captionasync/captionasync.go
rakyll: It is so hard to get this right because the send block will later be tagged and embedded in the docs on cloud site. I am not sure whether we should keep the ctx initialization in the tagged snippet or out of it. If yes, then ctx shouldn't be passed as an argument.

We only partition these samples to functions for the sake of testability. Otherwise, the entire sample ideally can live in one main function.

speech/captionasync/captionasync.go
rakyll: Removed.

speech/captionasync/captionasync.go
jba: OK, I don't understand the tagging part so I'll leave this to your judgment.

speech/captionasync/captionasync.go
jba: Up to you.

broady:
Tests don't pass.

rakyll:
Fixed in https://github.com/GoogleCloudPlatform/golang-samples/commit/9021d0a457ecbdc8cf6cb596eab191f890218d5d.

PR #126, opened 2016-09-22
storage: initial commit of the objects program

rakyll:


storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Sample objects creates, list, deletes objects and runs
+// other similar operations on them by using the Google Storage API.
+// More documentation is available at
+// https://cloud.google.com/storage/docs/json_api/v1/.
```
broady: Remove `json_api/v1/` from URL here

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Sample objects creates, list, deletes objects and runs
```
broady: Sample listobjects creates an object, lists objects from a bucket, and deletes that object.

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+	// [START download_file]
+	o := client.Bucket(bucket).Object(object)
+	if err := o.Delete(ctx); err != nil {
+		return err
+	}
+	// [END download_file]
+	return nil
+}
+
+func createBucketIfNotexits(ctx context.Context, client *storage.Client, projectID, bucket string) error {
```
broady: Have the user pass in an existing bucket?

Otherwise, roll this into the bucketmgr sample?

broady: Function name needs work.

`createBucketIfNotExists` or `ensureBucketExists`

I prefer the second one.

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	if err := createBucketIfNotexits(ctx, client, projectID, bucketName); err != nil {
+		log.Fatal(err)
+	}
+	// create another bucket later to be used to copy objects from the prev one.
+	if err := createBucketIfNotexits(ctx, client, projectID, dstBucketName); err != nil {
+		log.Fatal(err)
+	}
+	name := fmt.Sprintf("gs://%v/%v", bucketName, objectName)
```
broady: This is only used in the Fatalf statement. Move it in there.

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+func move(client *storage.Client, bucket, object string) (string, error) {
+	ctx := context.Background()
+	// [START move_file]
+	dstName := object + "-rename"
+
+	src := client.Bucket(bucket).Object(object)
+	dst := client.Bucket(bucket).Object(dstName)
+
+	if _, err := src.CopyTo(ctx, dst, nil); err != nil {
+		return "", err
```
broady: Weird to have return statements in extracted snippets.

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+
+	attrs, err := attrs(client, bucketName, objectName)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("object attributes of %v: %+v\n", name, attrs)
+
+	err = makePublic(client, bucketName, objectName)
+	if err != nil {
+		log.Fatalf("failed to make the object public: %v", err)
```
broady: Full, capitalized, sentences for Fatalf calls.

storage/objects/main.go
broady: There is no listing in this sample.

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+
+const (
+	bucketName    = "golang-example-objects-bucket"
+	dstBucketName = "golang-example-objects-bucket-dst"
+)
+
+func main() {
+	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
+	if projectID == "" {
+		fmt.Fprintf(os.Stderr, "GOOGLE_CLOUD_PROJECT environment variable must be set.\n")
```
jba: log.Fatal?

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if err := createBucketIfNotexits(ctx, client, projectID, bucketName); err != nil {
+		log.Fatal(err)
+	}
+	// create another bucket later to be used to copy objects from the prev one.
+	if err := createBucketIfNotexits(ctx, client, projectID, dstBucketName); err != nil {
```
jba: s/exits/Exists/

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+		fmt.Fprintf(os.Stderr, "GOOGLE_CLOUD_PROJECT environment variable must be set.\n")
+		os.Exit(1)
+	}
+
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
+
```
jba: bucket := client.Bucket(bucketName)

That's how this client should be used: create a BucketHandle ASAP and everything else should be easier.

storage/objects/main.go
jba: s/name/url/

or uri.

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+	if err := createBucketIfNotexits(ctx, client, projectID, bucketName); err != nil {
+		log.Fatal(err)
+	}
+	// create another bucket later to be used to copy objects from the prev one.
+	if err := createBucketIfNotexits(ctx, client, projectID, dstBucketName); err != nil {
+		log.Fatal(err)
+	}
+	name := fmt.Sprintf("gs://%v/%v", bucketName, objectName)
+
```
jba: Again, make the ObjectHandle here and use throughout.

storage/objects/main.go
jba: Actually, do we want to use standard Go style:

"Making object public: %v"

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+	// [START upload_file]
+	f, err := os.Open("notes.txt")
+	if err != nil {
+		return err
+	}
+	wc := client.Bucket(bucket).Object(object).NewWriter(ctx)
+	if _, err = io.Copy(wc, f); err != nil {
+		return err
+	}
+	if err := f.Close(); err != nil {
```
jba: Since you're reading f, you can defer f.Close() after checking that Open succeeded.

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+	// [START download_file]
+	rc, err := client.Bucket(bucket).Object(object).NewReader(ctx)
+	if err != nil {
+		return nil, err
+	}
+	data, err := ioutil.ReadAll(rc)
+	if err != nil {
+		return nil, err
+	}
+	if err := rc.Close(); err != nil {
```
jba: ditto: defer OK for reads

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+		return nil, err
+	}
+	return data, nil
+	// [END download_file]
+}
+
+func attrs(client *storage.Client, bucket, object string) (*storage.ObjectAttrs, error) {
```
jba: Not worth it's own function: if you have the ObjectHandle, it's just one call.

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+		return nil, err
+	}
+	return attrs, nil
+	// [END get_metadata]
+}
+
+func makePublic(client *storage.Client, bucket, object string) error {
+	ctx := context.Background()
+	// [START public]
+	acl := client.Bucket(bucket).Object(object).ACL()
```
jba: If you had the ObjectHandle, you could chain:

`obj.ACL().Set(...)`

and then this wouldn't be worth its own function.

storage/objects/main.go
```diff
@@ -0,0 +1,196 @@
...
+	dst := client.Bucket(dstBucket).Object(srcObject + "-copy")
+
+	if _, err := src.CopyTo(ctx, dst, nil); err != nil {
+		return err
+	}
+	// [END copy_file]
+	return nil
+}
+
+func delete(client *storage.Client, bucket, object string) error {
```
jba: Not worth its own function.

storage/objects/main.go
rakyll: I don't think the motivation behind these samples is to distribute them as actual tools. If so, they are terribly structured and poorly documented. I assume, we won't use any of these other than providing snippets. And generating a bucket automatically makes this a non-tool.

storage/objects/main.go
rakyll: fmt for humans, log for machines. I understand why we should capitalize fmts but not sure about logs. Doing them both.

storage/objects/main.go
rakyll: What do you mean? It is your idea that the snippets need to be in their functions so can be testable.

This is generally a poor sample though. Because storage API doesn't support moving and you need to handle the case where deleting the original fails.

storage/objects/main.go
rakyll: It is too long.

storage/objects/main.go
rakyll: Done

storage/objects/main.go
rakyll: Done

storage/objects/main.go
rakyll: Removed these.

storage/objects/main.go
rakyll: Removed these.

storage/objects/main.go
rakyll: Reamed

rakyll:
Closed in the favor of https://github.com/GoogleCloudPlatform/golang-samples/pull/129.

PR #128, opened 2016-09-28
logging: add exportlogs sample

rakyll:


rakyll:
This is hard to test against the prod service given you need an actual log to be exported. Other languages are using mocks or not testing these calls at all.

/cc @broady @jba

logging/exportlogs/exportlogs.go
```diff
@@ -0,0 +1,125 @@
...
+	it := client.Sinks(ctx)
+	for {
+		sink, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		sinks = append(sinks, sink)
```
broady: `sinks = append(sinks, sink.ID)` perhaps?

return a `[]string`

Demonstrates a small amount of manipulation

logging/exportlogs/exportlogs.go
```diff
@@ -0,0 +1,125 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Sample exportlogs lists, creates, updates, and deletes log sinks.
+package main
+
+import (
+	"context"
```
broady: wrong context

jba:
There is a fake in logging/internal/testing. I could pull it out if it's useful. But a mock is OK by me.

broady:
Probably not worth it for testing this.

The only thing worth testing here is the format of the sink destination (`storage...../bucket`)

rakyll:
Done.

> There is a fake in logging/internal/testing

Let's wait until we decide whether to move these snippets to the actual repo. Then, we can share testing libs without having to export the internal stuff.

PR #129, opened 2016-09-29
storage: initial commit of objects

rakyll:


rakyll:
/cc @broady @jba 

Rewrote the sample with your feedback. Closing the other PR.

storage/objects/main.go
```diff
@@ -0,0 +1,198 @@
...
+	if len(os.Args) < 2 {
+		usage("missing subcommand")
+	}
+
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
+
```
jba: I would make the BucketHandle and ObjectHandle here and pass them around. That's the way we should be teaching people to use the storage API. That way the read function, for example, can just take one arg.

storage/objects/main.go
```diff
@@ -0,0 +1,198 @@
...
+	if _, err := src.CopyTo(ctx, dst, nil); err != nil {
+		return err
+	}
+	// [END copy_file]
+	return nil
+}
+
+func delete(client *storage.Client, bucket, object string) error {
+	ctx := context.Background()
+	// [START download_file]
```
jba: download_file?

storage/objects/main.go
rakyll: The tutorials prepared for these snippets are very independent. Without giving enough context about how BucketHandle and ObjectHandle are constructed, it is very hard for the developers to follow. That's why every snippet is beginning at the client level.

rakyll:
@jba, I initially opened https://github.com/GoogleCloudPlatform/golang-samples/pull/126 that was merging a similar program but decided to rewrite it and abandoned the previous PR.

PR #139, opened 2016-10-20
bigquery: update example for new bigquery client surface

jba:
Change iterator to new methods.

Also, return nil instead of rows so far on error, to be idiomatic.


jba:
@broady @rakyll

broady:
@dpebot merge when green

PR #148, opened 2016-10-31
bigquery: add simpleapp example

tswast:
This is to get included in the docs here: https://cloud.google.com/bigquery/create-simple-app-api

### DIFF
--- a/bigquery/simpleapp/simpleapp.go
+++ b/bigquery/simpleapp/simpleapp.go
@@ -3,7 +3,6 @@
 // license that can be found in the LICENSE file.
 
 // Command simpleapp queries the Shakespeare sample dataset in Google BigQuery.
-// [START bigquery_simple_app_all]
 package main
 
 import (
@@ -25,19 +24,17 @@ func main() {
 		os.Exit(1)
 	}
 
-	rows, err := Query(proj)
+	rows, err := query(proj)
 	if err != nil {
-		log.Fatalln(err)
+		log.Fatal(err)
 	}
-	err = PrintResults(os.Stdout, rows)
-	if err != nil {
-		log.Fatalln(err)
+	if err := printResults(os.Stdout, rows); err != nil {
+		log.Fatal(err)
 	}
 }
 
-// Query returns a slice of the results of a query.
-// [START bigquery_simple_app_query]
-func Query(proj string) (*bigquery.RowIterator, error) {
+// query returns a slice of the results of a query.
+func query(proj string) (*bigquery.RowIterator, error) {
 	ctx := context.Background()
 
 	client, err := bigquery.NewClient(ctx, proj)
@@ -46,21 +43,18 @@ func Query(proj string) (*bigquery.RowIterator, error) {
 	}
 
 	query := client.Query(
-		"SELECT " +
-			"APPROX_TOP_COUNT(corpus, 10) as title, " +
-			"COUNT(*) as unique_words " +
-			"FROM `publicdata.samples.shakespeare`;")
+		`SELECT
+		 APPROX_TOP_COUNT(corpus, 10) as title,
+		 COUNT(*) as unique_words
+		 FROM ` + "`publicdata.samples.shakespeare`;")
 	// Use standard SQL syntax for queries.
 	// See: https://cloud.google.com/bigquery/sql-reference/
 	query.QueryConfig.UseStandardSQL = true
 	return query.Read(ctx)
 }
 
-// [END bigquery_simple_app_query]
-
-// PrintResults prints results of a query to the Shakespeare dataset.
-// [START bigquery_simple_app_print]
-func PrintResults(w io.Writer, iter *bigquery.RowIterator) error {
+// printResults prints results from a query to the Shakespeare dataset.
+func printResults(w io.Writer, iter *bigquery.RowIterator) error {
 	for {
 		var row bigquery.ValueList
 		err := iter.Next(&row)
@@ -71,8 +65,6 @@ func PrintResults(w io.Writer, iter *bigquery.RowIterator) error {
 			return err
 		}
 
-		// TODO: use ValueLoader to load a struct instead of using the ValueList directly.
-		// See: https://github.com/GoogleCloudPlatform/google-cloud-go/issues/399
 		fmt.Fprintln(w, "titles:")
 		ts := row[0].([]bigquery.Value)
 		for _, t := range ts {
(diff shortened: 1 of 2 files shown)


tswast:
Looks like it's the pubsub sample that's failing

bigquery/simpleapp/simpleapp.go
```diff
@@ -0,0 +1,91 @@
...
+	// See: https://cloud.google.com/bigquery/sql-reference/
+	query.QueryConfig.UseStandardSQL = true
+	return query.Read(ctx)
+}
+
+// [END bigquery_simple_app_query]
+
+// PrintResults prints results of a query to the Shakespeare dataset.
+// [START bigquery_simple_app_print]
+func PrintResults(w io.Writer, iter *bigquery.RowIterator) error {
```
rakyll: The existing samples avoid having func signatures and the context between the tags, see https://github.com/GoogleCloudPlatform/golang-samples/blob/master/pubsub/topics/main.go#L64. We should probably keep doing it for now even thought having signatures visible in the snippers are easier to follow.

/cc @broady

bigquery/simpleapp/simpleapp.go
```diff
@@ -0,0 +1,91 @@
...
+		if err == iterator.Done {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+
+		// TODO: use ValueLoader to load a struct instead of using the ValueList directly.
+		// See: https://github.com/GoogleCloudPlatform/google-cloud-go/issues/399
+		fmt.Fprintln(w, "titles:")
```
rakyll: shouldn't we need a \n after "titles:"?

bigquery/simpleapp/simpleapp.go
```diff
@@ -0,0 +1,91 @@
...
+		os.Exit(1)
+	}
+
+	rows, err := Query(proj)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	err = PrintResults(os.Stdout, rows)
+	if err != nil {
+		log.Fatalln(err)
```
broady: `log.Fatal(err)` is fine here.

bigquery/simpleapp/simpleapp.go
```diff
@@ -0,0 +1,91 @@
...
+	client, err := bigquery.NewClient(ctx, proj)
+	if err != nil {
+		return nil, err
+	}
+
+	query := client.Query(
+		"SELECT " +
+			"APPROX_TOP_COUNT(corpus, 10) as title, " +
+			"COUNT(*) as unique_words " +
+			"FROM `publicdata.samples.shakespeare`;")
```
broady: You might want to use Go's literal strings here?

A bit awkward with the backticks in the dataset name, though :(

bigquery/simpleapp/simpleapp.go
```diff
@@ -0,0 +1,91 @@
...
+	// Use standard SQL syntax for queries.
+	// See: https://cloud.google.com/bigquery/sql-reference/
+	query.QueryConfig.UseStandardSQL = true
+	return query.Read(ctx)
+}
+
+// [END bigquery_simple_app_query]
+
+// PrintResults prints results of a query to the Shakespeare dataset.
```
broady: "results of a query of"? "results from a query to" is probably even better.
ambiguous otherwise. the result is not being printed to a dataset.

bigquery/simpleapp/simpleapp_test.go
```diff
@@ -0,0 +1,32 @@
...
+	tc := testutil.SystemTest(t)
+
+	rows, err := Query(tc.ProjectID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var b bytes.Buffer
+	err = PrintResults(&b, rows)
+	if err != nil {
```
broady: compact this

if err := ...; err != nil

bigquery/simpleapp/simpleapp.go
broady: Actually you can also just extract the Go function (with its doc) with the `go_function` parameter in the github include widget. Keeps the code clean.

bigquery/simpleapp/simpleapp.go
tswast: Done.

bigquery/simpleapp/simpleapp.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	err = PrintResults(os.Stdout, rows)
+	if err != nil {
+		log.Fatalln(err)
+	}
+}
+
+// Query returns a slice of the results of a query.
+// [START bigquery_simple_app_query]
+func Query(proj string) (*bigquery.RowIterator, error) {
```
tswast: Done.

bigquery/simpleapp/simpleapp.go
```diff
@@ -0,0 +1,91 @@
...
+	ctx := context.Background()
+
+	client, err := bigquery.NewClient(ctx, proj)
+	if err != nil {
+		return nil, err
+	}
+
+	query := client.Query(
+		"SELECT " +
```
tswast: Done. Though I did have to concat in the dataset name, since standard SQL syntax uses backticks for table names.

bigquery/simpleapp/simpleapp.go
tswast: Done.

You're right, that was ambiguous.

bigquery/simpleapp/simpleapp.go
tswast: Done.

tswast: Done. I've removed all the region tags.

bigquery/simpleapp/simpleapp.go
```diff
@@ -0,0 +1,91 @@
...
+		err := iter.Next(&row)
+		if err == iterator.Done {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+
+		// TODO: use ValueLoader to load a struct instead of using the ValueList directly.
+		// See: https://github.com/GoogleCloudPlatform/google-cloud-go/issues/399
```
tswast: Removed.

bigquery/simpleapp/simpleapp.go
tswast: I'm using Fprintln here.

bigquery/simpleapp/simpleapp_test.go
tswast: Done.

PR #150, opened 2016-11-03
copy modified Endpoints sample to new location

tangiel:
This change copies the sample from docs/managed_vms/endpoints/ to a new
endpoints/getting-started directory. It also generalizes the sample to
work on all GCP environments by adding a custom Dockerfile for use on
GCE and GKE, as well as creating a separate main function for non-App
Engine environments.


broady:
A few questions:

* Why is this being copied, and not moved?
* The separate main functions are unnecessary. To support both MVM/Flex and GCE/GKE, do the following:
   * Default to port 8080.
   * Override the port number using the `PORT` environment variable.
   * For the GCE/GKE dockerfile, add a line with `ENV PORT 8081`.

```
port := "8080"
if p := os.Getenv("PORT") != "" {
  port = "8081"
}
log.Fatal(http.ListenAndServe(":"+port, nil))
```

Also, there are no docs for `Dockerfile.custom`?

It would be helpful if you added separate commits so the delta is clear.

tangiel:
* It's being copied because the cloud docs aren't updated yet. Once they are, the old sample will be removed.
* But `appengine.Main` adds health checks in the case of Flex and so on. It's not just the port that's being customized.
* Lack of docs for Dockerfile.custom is intentional, they will be added to the cloud docs where we have quickstarts for using Docker. The goal here is to not have duplicate information in the README, which links to the quickstarts.

broady:
The health checks will work just fine. A non-502/503/504 response is deemed healthy.

tangiel:
Added a flag instead of using an environment variable. Is that acceptable?

broady:
Nope, you need to use the `PORT` env variable for Flex. Also, move the main function back to where it was originally.

tangiel:
What do you mean by "originally"? There was no main originally, only an init. If you mean to collapse it back into one file that makes sense.

broady:
Original: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/docs/managed_vms/endpoints/app.go#L19-L36

tangiel:
Oops, guess I changed it from main to init. Thought it was originally all in init. Pushed changes, testing them now.

tangiel:
Verified that the docker image works and Flex deployment works.

PR #155, opened 2016-11-11
Update app.yaml in go sample to support env:flex

mazloft:


broady:
Can you explain why? Are you a Googler? If so, please contact me on Hangouts.

PR #156, opened 2016-11-12
datastore, bookshelf: update to new datastore Key accessors

broady:


broady:
/cc @jba @zombiezen
we'll need to update the docs on cloud.google.com, too.

broady:
@dpebot merge when green

PR #158, opened 2016-11-14
docs/managed_vms/cloudsql: use FUSE connection string

broady:
The TCP proxy is not bound to the loopback address. The FUSE proxy
works, though.

Add some other instructions on what the connection string should look
like and what the TCP proxy address should be when testing locally.

### DIFF
--- a/docs/managed_vms/cloudsql/app.yaml
+++ b/docs/managed_vms/cloudsql/app.yaml
@@ -7,12 +7,21 @@ automatic_scaling:
 #[START env]
 env_variables:
   # See https://github.com/go-sql-driver/mysql
-  MYSQL_CONNECTION: user:password@tcp(127.0.0.1:3306)/dbname
+  #
+  # Replace INSTANCE_CONNECTION_NAME with the same value as in the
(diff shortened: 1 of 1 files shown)


docs/managed_vms/cloudsql/app.yaml
```diff
@@ -7,12 +7,21 @@ automatic_scaling:
...
+  # instead use the "tcp" dialer by setting the environment variable:
+  # MYSQL_CONNECTION=user:password@tcp(127.0.0.1:3306)/dbname
 #[END env]
 
 #[START cloudsql_settings]
-# Replace INSTANCE_CONNECTION_NAME with the value obtained  when configuring your
+# Replace INSTANCE_CONNECTION_NAME with the value obtained when configuring your
 # Cloud SQL instance, available from the Google Cloud Console or from the Cloud SDK.
+# For SQL v2 instances, this should be in the form of "project:region:instance".
+# SQL v1 instances are not currently supported.
```
jonparrott: Remove `currently`? It implies that there may be support in the future.

PR #159, opened 2016-11-14
appengine_flexible: copy samples from managed_vms, change vm:true to env:flex

broady:
This merely copies all of the samples across. The e2e tests have not been updated.


broady:
@dpebot merge when green

PR #162, opened 2016-11-15
language: update to v1 from v1beta1

broady:
This has one significant change for Sentiment Analysis:
polarity is dropped and replaced by score.


broady:
Dependent on https://code-review.googlesource.com/#/c/9310/

PR #163, opened 2016-11-15
translate: add snippets

broady:
Add snippets for

* Creating a client with an API key
* Translating text
* Language detection
* Listing supported languages


rakyll:
LGTM

PR #165, opened 2016-11-16
Update ports for endpoints sample.

jeffmendoza:
Need to switch ports of containers to match other languages.

`gcr.io/google-samples/echo-go:1.0` is updated and pushed to listen on 8080.


broady:
hopefully the new commit message is accurate

PR #166, opened 2016-11-17
pubsub: add IAM and IAM policy samples

broady:


broady:
/cc @jba @rakyll 

These tests will fail - gotta set up the right auth for the Travis service account. PTAL though.

pubsub/subscriptions/main.go
```diff
@@ -163,3 +165,56 @@ func createTopicIfNotExists(c *pubsub.Client) *pubsub.Topic {
...
+	sub := c.Subscription(subName)
+	policy, err := sub.IAM().Policy(ctx)
+	if err != nil {
+		log.Fatalf("GetPolicy: %v", err)
+	}
+	// Other valid prefixes are "serviceAccount:", "user:"
+	// See the documentation for more values.
+	policy.Add(iam.AllUsers, iam.Viewer)
+	policy.Add("group:cloud-logs@google.com", iam.Editor)
+	if err := sub.IAM().SetPolicy(ctx, policy); err != nil {
```
jba: I don't know if you want to deal with this in a sample, but this read-modify-write will check etags, and thus will fail if there is a concurrent modification. So really, you should put this in a loop.

broady: Yeah, not for this sample, I think. I'll add a comment that describes this possibility.

broady:
```
+       // NOTE: It may be necessary to retry this operation if IAM policies are
+       // being modified concurrently. SetPolicy will return an error if the policy
+       // was modified since it was retrieved.
```

broady:
For future reference, since tests are sharded across projects, here's what I ran to get this working:

```
gcloud projects add-iam-policy-binding golang-samples-tests --member=serviceAccount:travis@golang-samples-tests.iam.gserviceaccount.com --role=roles/pubsub.admin;

for n in $(seq 10); do echo $n;
gcloud projects add-iam-policy-binding golang-samples-tests-$n --member=serviceAccount:travis@golang-samples-tests.iam.gserviceaccount.com --role=roles/pubsub.admin;
done
```

PR #178, opened 2016-11-23
internal/testutil: manage GCP resource creation and cleanup

rakyll:


broady:
```
Step #2: storage/buckets/main_test.go:18: undefined: testutil.Namespace
Step #2: FAIL	github.com/GoogleCloudPlatform/golang-samples/storage/buckets [build failed]
```

bigquery/snippets/snippet_test.go
```diff
@@ -9,16 +9,14 @@ import (
...
 
 	"cloud.google.com/go/bigquery"
-	"cloud.google.com/go/storage"
 	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
 	"golang.org/x/net/context"
-	"golang.org/x/oauth2/google"
-	rawbq "google.golang.org/api/bigquery/v2"
 )
 
+const ns = "snippets"
```
broady: this needs to be unique across all tests

`const ns = "bigquery-snippets`

internal/testutil/resources.go
```diff
@@ -0,0 +1,222 @@
...
+
+// NextObject will return a new unique object name for the given namespace.
+func NextObject(namespace string) string {
+	atomic.AddUint64(&idCounter, 1)
+	return fmt.Sprintf("%s_%s_object%d_%d", appPrefix, namespace, idCounter, time.Now().Unix())
+}
+
+// CleanupDatasets deletes the given datasets and all expired
+// datasets belongs to the tests.
+func CleanupDatasets(t *testing.T, name ...string) {
```
broady: Why doesn't it take a namespace?

internal/testutil/resources.go
```diff
@@ -0,0 +1,222 @@
...
+	idCounter uint64
+	expiry    = time.Hour
+)
+
+const appPrefix = "golang_samples"
+
+// NextDataset returns the next unique dataset ID generated for the given namespace.
+func NextDataset(namespace string) string {
+	atomic.AddUint64(&idCounter, 1)
+	return fmt.Sprintf("%s_%s_dataset%d_%d", appPrefix, namespace, idCounter, time.Now().Unix())
```
broady: This is racy.
```
id := atomic.AddUint64(&idCounter, 1)
return fmt.Sprintf("%s_%s_dataset%d_%d", appPrefix, namespace, id, time.Now().Unix())
```

internal/testutil/resources.go
```diff
@@ -0,0 +1,222 @@
...
+var (
+	idCounter uint64
+	expiry    = time.Hour
+)
+
+const appPrefix = "golang_samples"
+
+// NextDataset returns the next unique dataset ID generated for the given namespace.
+func NextDataset(namespace string) string {
+	atomic.AddUint64(&idCounter, 1)
```
jba: I think mutexes are fine unless performance is really an issue.

internal/testutil/resources.go
```diff
@@ -0,0 +1,222 @@
...
+	return fmt.Sprintf("%s_%s_dataset%d_%d", appPrefix, namespace, idCounter, time.Now().Unix())
+}
+
+// NextTable returns the next unique table ID generated for the given namespace
+func NextTable(namespace string) string {
```
jba: It seems like these could all be one function.

Or at least they should call one more general, unexported function.

internal/testutil/resources.go
```diff
@@ -0,0 +1,222 @@
...
+
+	hc, err := google.DefaultClient(ctx, rawbq.CloudPlatformScope)
+	if err != nil {
+		t.Errorf("DefaultClient: %v", err)
+	}
+	s, err := rawbq.New(hc)
+	if err != nil {
+		t.Errorf("bigquery.New: %v", err)
+	}
+	call := s.Datasets.Delete(tc.ProjectID, datasetID)
```
jba: This is now available in cloud.google.com/go/bigquery.

broady:
I think this PR is obsolete, in light of the testing project pool (that is, no need to think about concurrent builds). Resource IDs can stay the same between test runs, as long as they incorporate the project ID in their name.

PR #181, opened 2016-11-29
docs/appengine/cloudsql: add instructions for Cloud SQL v2

broady:


broady:
Supersedes #172 #171

broady:
@piaxc can you take a look? I copied the default values and instructions from the Flexible instructions.

I think root is a reasonable default user, and an empty password also reasonable. What do you think?

broady:
@piaxc nope, specifying database is up to the user. this sample lists the databases, so no database name is needed.

amended the comments per your suggestions. No slash needed. I confirmed this works.

broady:
@dpebot merge when green

PR #184, opened 2016-11-30
bigquery: update samples to use []Value

jba:
ValueList has been removed. []Value is used instead.


jba:
@broady @rakyll

broady:
lgtm, will be merged once the test passes

broady:
i guess the automerger didn't work. oh well.

PR #185, opened 2016-12-08
Update Endpoints samples:

sepehre:
1. Change default service names to the new cloud.goog format.
2. Change app.yaml beta_settings to endpoints_api_service.

### DIFF
--- a/endpoints/getting-started/app.yaml
+++ b/endpoints/getting-started/app.yaml
@@ -2,5 +2,7 @@ runtime: go
 env: flex
 
 endpoints_api_service:
-  name: ENDPOINTS SERVICE-NAME # To be replaced.
-  config_id: ENDPOINTS CONFIG-ID # To be replaced.
+  # The following values are to be replaced by information from the output of
+  # 'gcloud service-management deploy openapi.yaml' command.
+  name: ENDPOINTS SERVICE-NAME
+  config_id: ENDPOINTS CONFIG-ID
--- a/endpoints/getting-started/app.yaml
+++ b/endpoints/getting-started/app.yaml
@@ -1,8 +1,6 @@
 runtime: go
 env: flex
 
-beta_settings:
-  # Enable Google Cloud Endpoints API management.
-  use_endpoints_api_management: true
-  # Specify the Open API specification.
-  endpoints_swagger_spec_file: openapi.yaml
+endpoints_api_service:
+  name: ENDPOINTS SERVICE-NAME # To be replaced.
+  config_id: ENDPOINTS CONFIG-ID # To be replaced.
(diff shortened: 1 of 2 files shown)


sepehre:
Please do not merge until documentation updated have been coordinated.

endpoints/getting-started/app.yaml
```diff
@@ -1,8 +1,6 @@
...
 env: flex
 
-beta_settings:
-  # Enable Google Cloud Endpoints API management.
-  use_endpoints_api_management: true
-  # Specify the Open API specification.
-  endpoints_swagger_spec_file: openapi.yaml
+endpoints_api_service:
+  name: ENDPOINTS SERVICE-NAME # To be replaced.
+  config_id: ENDPOINTS CONFIG-ID # To be replaced.
```
broady: To be replaced by what?

Should these be `ENDPOINTS-SERVICE-NAME` to be consistent with `YOUR-PROJECT-ID`?

endpoints/getting-started/app.yaml
```diff
@@ -1,8 +1,6 @@
 runtime: go
 env: flex
 
-beta_settings:
-  # Enable Google Cloud Endpoints API management.
-  use_endpoints_api_management: true
-  # Specify the Open API specification.
-  endpoints_swagger_spec_file: openapi.yaml
```
broady: How is the Open API spec referenced now?

sepehre: It is not referenced in app.yaml any more. Users would run "gcloud service-management deploy openapi.yaml" first, and then take the output of that command (service name and config_id) and update app.yaml only when they need to update their API Specification. This is consistent with the GCE/GKE model, and avoid re-deploying the Open API spec on every GAE Flex deployment.

endpoints/getting-started/app.yaml
sepehre: The commend could probably be more descriptive. The intention is have the users go through the Quickstart, run "gcloud service-management deploy openapi.yaml" first, and then take the output of that command (service name and config_id) and update app.yaml 

I left the space (ENDPOINTS" "SERVICE-NAME) in there intentionally because I want the "gcloud app deploy" command to fail if the users (who should be going through the Quickstart) do not update their app.yaml for whatever reason. The command would fail due to the AppInfo regex validation.

We are in the process of adding a better client side validation to make sure the name and config_id provided are actually valid, but unfortunately adding such a validation causes other concerns, so we are still working on it. In the meanwhile, if you provide an invalid (but syntactically correct) config_id, the deployment would proceed and fail much later after spinning for 20 minutes... which is horrible.

Let me know if you'd like to discuss the details.

sepehre:
Please merge.

PR #188, opened 2017-01-06
appengine_flexible/endpoints: add Google ID example

rthellend:
Add an example for Google ID authentication.

### DIFF
--- a/appengine_flexible/endpoints/README.md
+++ b/appengine_flexible/endpoints/README.md
@@ -45,5 +45,15 @@ First, [download a Service Account JSON key file](https://developers.google.com/
 Then, run:
 
 ```
-go run client/main.go -host=https://my-app.appspot.com -service-account=path_to_service_account.json
+go run client/main.go -host=https://my-app.appspot.com -jwt -service-account=path_to_service_account.json
+```
+
+### Send a request using Google ID authentication
+
+First, [download a Service Account JSON key file](https://developers.google.com/identity/protocols/OAuth2ServiceAccount#creatinganaccount).
+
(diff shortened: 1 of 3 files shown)


rthellend:
I'm not sure why googlebot isn't seeing my google.com email address.

broady:
@rthellend you've got to set it with `git config --global user.email foo@google.com`

then `git commit --amend --reset-author`

and/or set your GitHub username at http://github/

broady:
A few questions before I review the code...

Can you make the update here, instead?
https://github.com/GoogleCloudPlatform/golang-samples/tree/master/endpoints/getting-started

(I *think* appengine_flexible/endpoints should be removed, let me confirm)

Did you make similar changes to the openapi.yaml for other languages? They should be consistent. e.g., the name of the audience.

rthellend:
I created a new pull request: https://github.com/GoogleCloudPlatform/golang-samples/pull/189 . I'll close this one after this comment.

Moved to endpoints/getting-started.

I did not make any changes for other languages. I only made it here to be consistent between JWT and Google ID. In the case of Google ID, the token that's generated by the google api always has https://. I'm happy to change it back for JWT if you prefer.

broady:
@rthellend Yeah, best to keep it consistent.

Could you change the others, too? At the least, changing the `x-` keys would be good. If not, I can get the owners of those samples to do the updates.

rthellend:
I'd rather let the owners do the updates in the other repos, if you don't mind too much.

I can change the x-google keys back if you prefer. The old names are deprecated, but they still work.

Thanks.

PR #189, opened 2017-01-09
endpoints/getting-started: Add Google ID

rthellend:
Add an example for Google ID authentication.

### DIFF
--- a/endpoints/getting-started/README.md
+++ b/endpoints/getting-started/README.md
@@ -32,7 +32,17 @@ First, [download a Service Account JSON key file](https://developers.google.com/
 Then, run:
 
 ```bash
-go run client/main.go -host=https://my-app.appspot.com -service-account=path_to_service_account.json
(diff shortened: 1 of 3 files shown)


odeke-em:
Thanks @rthellend, you might need to rebase and resolve some merge conflicts.
/cc @broady @adams-sarah

odeke-em:
Alos @rthellend, would you mind retitling this PR, seems like it is more of using JWT than just the Google ID if am not mistaken? And also IMO it could be used to address https://github.com/GoogleCloudPlatform/golang-samples/issues/242. If it does, could you please update the commit message like this
```shell
endpoints/getting-started: add sample of using JWT tokens and GoogleID

Fixes #242

[...Other content you'd like to talk about...]
```

tbpg:
Thank you for sending this! I'm cleaning up some old PRs - this is almost 2 years old, so I'm going to close it. But, comment/reopen if you'd like to see it through. An important bit is there needs to be a doc the sample will be embedded in.

PR #190, opened 2017-01-12
Add write timeseries region tags

waprin:


broady:
No need, just use `go_function=writeTimeSeriesValue` instead of a region tag

PR #195, opened 2017-02-08
kms: add crypter sample program

bshaffer:

PR #196, opened 2017-02-08
testing: add environment variables for Cloud KMS tests

broady:
golang-samples-tests{,2-10} all have Cloud KMS enabled and a key at
global/ring1/key1.


broady:
Merged into #195.

PR #201, opened 2017-03-03
vision: use new "ImageFromURI" function

broady:


broady:
@dpebot merge when green. thx!

PR #207, opened 2017-03-15
pubsub: switch to the new API

rakyll:
Related to https://github.com/GoogleCloudPlatform/google-cloud-go/commit/f4eafa8cbb48e8ef5e23b542c3cb51cd1cbce787.

### DIFF
--- a/pubsub/subscriptions/main.go
+++ b/pubsub/subscriptions/main.go
@@ -13,13 +13,11 @@ import (
 	"time"
 
 	// [START imports]
-	"golang.org/x/net/context"
-
+	"cloud.google.com/go/iam"
 	"cloud.google.com/go/pubsub"
+	"golang.org/x/net/context"
 	"google.golang.org/api/iterator"
 	// [END imports]
-
-	"cloud.google.com/go/iam"
 )
 
 func main() {
@@ -88,10 +86,20 @@ func pullMsgs(client *pubsub.Client, name string, topic *pubsub.Topic) error {
 	ctx := context.Background()
 
 	// publish 10 messages on the topic.
+	var results []*pubsub.PublishResult
 	for i := 0; i < 10; i++ {
-		topic.Publish(ctx, &pubsub.Message{
+		res := topic.Publish(ctx, &pubsub.Message{
 			Data: []byte(fmt.Sprintf("hello world #%d", i)),
 		})
+		results = append(results, res)
+	}
+
+	// Check if all messages are published.
+	for _, r := range results {
+		_, err := r.Get(ctx)
+		if err != nil {
+			log.Printf("Publish error: %v", err)
+		}
 	}
 
 	// [START pull_messages]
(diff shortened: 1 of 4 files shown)


pubsub/subscriptions/main.go
```diff
@@ -89,12 +89,9 @@ func pullMsgs(client *pubsub.Client, name string, topic *pubsub.Topic) error {
 
 	// publish 10 messages on the topic.
 	for i := 0; i < 10; i++ {
-		_, err := topic.Publish(ctx, &pubsub.Message{
+		topic.Publish(ctx, &pubsub.Message{
```
rakyll: @jba, what's our best practice here? Do we allow fire and forget?

broady: Save the future result and check for errors after all of the messages are sent.

pubsub/topics/main.go
```diff
@@ -111,15 +111,14 @@ func publish(client *pubsub.Client, topic, msg string) error {
 	ctx := context.Background()
 	// [START publish]
 	t := client.Topic(topic)
-	msgIDs, err := t.Publish(ctx, &pubsub.Message{
+	result := t.Publish(ctx, &pubsub.Message{
 		Data: []byte(msg),
 	})
+	id, err := result.Get(ctx)
```
broady: Add a comment above describing what this is doing.

rakyll: Done

pubsub/subscriptions/main.go
rakyll: Done, but it is awkward to do it in the same block. The developer would question why there is no blocking publish and what the merits of non-blocking publish has here. It is not quite a good experience having to write all the boilerplate to achieve a simple blocking publish.

pubsub/subscriptions/main.go
```diff
@@ -13,13 +13,11 @@ import (
 	"time"
 
 	// [START imports]
-	"golang.org/x/net/context"
-
+	"cloud.google.com/go/iam"
 	"cloud.google.com/go/pubsub"
+	"golang.org/x/net/context"
 	"google.golang.org/api/iterator"
 	// [END imports]
```
broady: Revert this re-ordering

pubsub/subscriptions/main.go
```diff
@@ -88,12 +87,19 @@ func pullMsgs(client *pubsub.Client, name string, topic *pubsub.Topic) error {
...
 		})
+		results = append(results, res)
+	}
+
+	// Check if all messages are published.
+	for _, r := range results {
+		_, err := r.Get(ctx)
 		if err != nil {
-			return err
+			log.Printf("Publish error: %v", err)
```
broady: This should return the error

broady:
other samples need updating, too

```
Step #2: getting-started/bookshelf/app/app.go:301: assignment count mismatch: 2 = 1
Step #2: appengine_flexible/pubsub/pubsub.go:101: assignment count mismatch: 2 = 1
```

pubsub/subscriptions/main.go
broady: I did it (pushed to your branch)

rakyll:
LGTM

PR #209, opened 2017-03-17
pubsub: use new Subscription.Receive interface

broady:


pubsub/subscriptions/main.go
```diff
@@ -104,25 +105,24 @@ func pullMsgs(client *pubsub.Client, name string, topic *pubsub.Topic) error {
...
-		if err == iterator.Done {
-			break
-		}
-		if err != nil {
-			return err
+	var mu sync.Mutex
+	received := 0
+	sub := client.Subscription(name)
+	receiveCtx, cancel := context.WithCancel(ctx)
+	sub.Receive(receiveCtx, func(ctx context.Context, msg *pubsub.Message) {
```
jba: You should still check for error. It could be that the service failed, and you'd want to know.

broady: Done

PR #214, opened 2017-03-27
bigtable: move bigtable examples to golang-samples

rakyll:
Move the bigtable examples to golang-samples as they are getting deleted from cloud.google.com/go. See https://code-review.googlesource.com/c/11656/.


rakyll:
The current CI failure is coming from a flaky racy test case in pubsub, I am going to merge this PR.

PR #218, opened 2017-03-31
Add a snippet for how to detect if running in the development server

devlance:
We don't currently demonstrate in code how to do this type of check in the docs.


broady:
thanks, Brett!

PR #220, opened 2017-04-03
pubsub: deflake tests

### LINKED ISSUE #219: pubsub: tests are flaky
Tests don't use a unique topic/subscription name, so they're flaky.

They need to incorporate the project ID in them to ensure they don't conflict with tests run in parallel.

broady:
Add the Project ID as a prefix to ensure tests can run in parallel.

Fixes #219.

### DIFF
--- a/pubsub/topics/main_test.go
+++ b/pubsub/topics/main_test.go
@@ -15,11 +15,13 @@ import (
 	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
 )
 
-const topicID = "golang-samples-topic-example"
+var topicID string
 
 func setup(t *testing.T) *pubsub.Client {
 	ctx := context.Background()
 	tc := testutil.SystemTest(t)
+
+	topicID = tc.ProjectID + "-test-topic"
 
 	client, err := pubsub.NewClient(ctx, tc.ProjectID)
 	if err != nil {
--- a/pubsub/subscriptions/main_test.go
+++ b/pubsub/subscriptions/main_test.go
@@ -17,15 +17,12 @@ import (
 )
 
 var topic *pubsub.Topic
-
-const (
-	subID   = "golang-samples-subscription"
-	topicID = "golang-samples-topic"
-)
+var subID string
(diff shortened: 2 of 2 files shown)


broady:
Also seeing this, which seems to be correlated to pub/sub IAM tests:

```
=== RUN   TestIAM
2017/04/03 18:48:16 rpc error: code = Unavailable desc = there is no address available
FAIL	github.com/GoogleCloudPlatform/golang-samples/pubsub/subscriptions	3.068s
```

pubsub/topics/main_test.go
```diff
@@ -77,20 +79,34 @@ func TestPublish(t *testing.T) {
...
-	if role, member := iam.Viewer, iam.AllUsers; !policy.HasRole(member, role) {
-		t.Fatalf("want %q as viewer, got %v", member, policy)
-	}
+	testutil.Retry(t, 10, time.Second, func(r *testutil.R) {
+		policy, err := getPolicy(c, topicID)
+		if err != nil {
+			r.Errorf("getPolicy: %v", err)
+		}
+		if role, member := iam.Editor, "group:cloud-logs@google.com"; !policy.HasRole(member, role) {
+			r.Errorf("want %q as viewer, got %v", member, policy)
```
rakyll: got first, want second: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

pubsub/topics/main_test.go
```diff
@@ -77,20 +79,34 @@ func TestPublish(t *testing.T) {
...
+		}
+		if role, member := iam.Viewer, iam.AllUsers; !policy.HasRole(member, role) {
+			r.Errorf("want %q as viewer, got %v", member, policy)
```
rakyll: ditto

pubsub/topics/main_test.go
broady: The values aren't being checked for equality.

It's more like `got HasRole(role, member)=false; want true; policy = %v`

PR #221, opened 2017-04-04
vision/detect: retry tests, show better error messages

broady:


broady:
Whoops, didn't actually push the commit that added retries, but let's see how this goes.

Maybe we'll see a pattern in a particular image that fails.

PR #222, opened 2017-04-04
bigquery/snippets: move ImportExport to an E2E test

broady:
This test takes a long time sometimes, and doesn't print any
intermediary output.

Print a log message very 5 minutes to work Travis's 10 minute timeout.

Move it to an E2E test, only run nightly.


broady:
FYI @tswast

PR #225, opened 2017-04-04
speech: update to apiv1 from apiv1beta1

broady:


broady:
R= @gguuss

broady:
Need `LanguageCode`.

PR #230, opened 2017-04-05
storage/objects: use a single bucket per project, improve tests

broady:
Add tests for listing by prefix (to validate the behaviour present in
the comments).

Use a simplified mechanism to create and clean up test buckets.
This change means that given a single test project, integration tests cannot
run concurrently. Our tests are sharded across projects, so this
simplifies setup and cleanup.


broady:
R= @jba @rakyll

rakyll:
This is great, we can tweak the size of the project pool depending on the flakiness of the tests.

PR #232, opened 2017-04-10
speech/captionasync: use GCS URI, inline audio is not allowed

broady:
"Inline audio is not allowed with LongRunningRecognize. Please use a GCS
URI."


broady:
R= @gguuss

PR #241, opened 2017-04-14
getting-started/bookshelf: use golang:alpine to build for GKE

broady:
* use golang:alpine as base image
* consolidate both commands into one image
* "go get" from the Dockerfile (no need to use aedeploy)

### DIFF
--- a/getting-started/bookshelf/gke_deployment/Dockerfile
+++ b/getting-started/bookshelf/gke_deployment/Dockerfile
@@ -1,6 +1,18 @@
-# Dockerfile extending the generic Go image with application files for a
-# single application.
-FROM gcr.io/google_appengine/golang
+FROM golang:alpine
 
-COPY . /go/src/app
-RUN go-wrapper install -tags appenginevm
+RUN apk add --no-cache ca-certificates
+
+COPY . /go/src/bookshelf
+
+RUN set -ex \
+      && apk add --no-cache --virtual .build-deps \
+              git \
+      && go get -v bookshelf/... \
+      && apk del .build-deps
+
+RUN go build -o /app bookshelf/app
+RUN go build -o /pubsub_worker bookshelf/pubsub_worker
+
+WORKDIR /go/src/bookshelf/app
+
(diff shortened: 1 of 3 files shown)


getting-started/bookshelf/gke_deployment/Dockerfile
```diff
@@ -1,6 +1,18 @@
...
+
+COPY . /go/src/bookshelf
+
+RUN set -ex \
+      && apk add --no-cache --virtual .build-deps \
+              git \
+      && go get -v bookshelf/... \
+      && apk del .build-deps
+
+RUN go build -o /app bookshelf/app
```
ahmetb: Try: 

    go install ./...

this should load all the packages once and hence build faster and build to /go/bin.

getting-started/bookshelf/gke_deployment/Dockerfile
```diff
@@ -1,6 +1,18 @@
...
+RUN go build -o /pubsub_worker bookshelf/pubsub_worker
+
+WORKDIR /go/src/bookshelf/app
```
ahmetb: Instead of this, try `ENV PATH $PATH:/go/bin`. This way you don't need CMD or ENTRYPOINT, the pod spec can just define the cmd you want to invoke directly.

getting-started/bookshelf/gke_deployment/Dockerfile
broady: go install doesn't "go get".

"go build" is equivalent in this case.

getting-started/bookshelf/gke_deployment/Dockerfile
broady: This is a better argument for using "go install".

getting-started/bookshelf/gke_deployment/Dockerfile
ahmetb: You can add a `RUN go get ./...` ?

ahmetb:
LGTM

PR #251, opened 2017-05-09
speech/livecaption: remove the unused connection

rakyll:
This is a left over from the pre-autogenerated client days
where we were using a gRPC client to talk to the API.


adams-sarah:
LGTM

PR #253, opened 2017-05-11
pubsub: add samples for PushConfig, PublishSettings, and ReceiveSettings

rakyll:


pubsub/topics/main.go
```diff
@@ -125,6 +147,34 @@ func publish(client *pubsub.Client, topic, msg string) error {
...
+func publishWithSettings(client *pubsub.Client, topic string, msg []byte) error {
+	ctx := context.Background()
+	// [START publish_settings]
+	t := client.Topic(topic)
+	t.PublishSettings = pubsub.PublishSettings{
+		ByteThreshold:  5000,
+		CountThreshold: 10,
+		DelayThreshold: 100 * time.Millisecond,
+	}
+	result := t.Publish(ctx, &pubsub.Message{Data: msg})
```
adams-sarah: I'm not sure if this is typical for these samples, but perhaps giving some idea of what a `result` is would be nice? Like a comment or something? or use it maybe?

pubsub/topics/main.go
```diff
@@ -125,6 +147,34 @@ func publish(client *pubsub.Client, topic, msg string) error {
...
+func publishSingleGoroutine(client *pubsub.Client, topic string, msg []byte) error {
+	ctx := context.Background()
+	// [START publish_single_goroutine]
+	t := client.Topic(topic)
+	t.PublishSettings = pubsub.PublishSettings{
+		NumGoroutines: 1,
+	}
+	result := t.Publish(ctx, &pubsub.Message{Data: msg})
+	// [END publish_single_goroutine]
+	_ = result
```
adams-sarah: same question

pubsub/topics/main.go
adams-sarah: also, if comments are a thing in these samples, maybe a comment that this kicks of many goroutines?
or do you think that's an impl detail?

rakyll: I avoided to repeat the `result` handling section previously because the previous Receive snippet documented it, but gonna add that section for both of the new samples to make it easier to copy/paste.

re: num of goroutines. It is too much detail in this context. That's why we will only mention it after the document opens a section about concurrency related settings.

pubsub/topics/main.go
rakyll: Done.

adams-sarah:
LGTM!

PR #257, opened 2017-05-18
vision: switch to the generated surface

rakyll:
Remove the dependency to cloud.google.com/go/vision and use
the new autogenated helpers.

### DIFF
--- a/vision/detect/detect.go
+++ b/vision/detect/detect.go
@@ -14,22 +14,24 @@ import (
 	"io"
 	"os"
 
-	"cloud.google.com/go/vision"
+	"cloud.google.com/go/vision/apiv1"
 	"golang.org/x/net/context"
+	pb "google.golang.org/genproto/googleapis/cloud/vision/v1"
 )
 
 func init() {
 	// Refer to these functions so that goimports is happy before boilerplate is inserted.
 	_ = context.Background()
-	_ = vision.NewClient
+	_ = vision.ImageAnnotatorClient{}
 	_ = os.Open
+	_ = pb.AnnotateImageRequest{}
 }
 
 // detectFaces gets faces from the Vision API for an image at the given file path.
 func detectFaces(w io.Writer, file string) error {
 	ctx := context.Background()
 
-	client, err := vision.NewClient(ctx)
+	client, err := vision.NewImageAnnotatorClient(ctx)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


vision/detect/detect.go
```diff
@@ -14,22 +14,24 @@ import (
 	"io"
 	"os"
 
-	"cloud.google.com/go/vision"
+	"cloud.google.com/go/vision/apiv1"
```
jba: Can we use an import rename of "vision" so people don't get confused when they see the `vision.` below?

broady: +1, good catch. it's what goimports does by default.

rakyll: The type names are different, goimports will match cloud.google.com/go/vision/apiv1. I am suggesting to keep it this way.

rakyll: Correcting myself, NewImageFromURI and NewImageFromReader stay the same. goimports may pick the wrong package but renaming doesn't save us from this wrong picking. I also really don't have a good idea what a better name replacement should be. It is a part of a core doc, whatever we name it, it will end up at people's codebases given most users just copy/paste.

broady:
goimports adds the named import. Remove the line and see what it does.

rakyll:
Renames are not picked up when you are in a new package, see http://imgur.com/a/jizUo. But it is better than the wrong package is picked up.

vision/detect/detect.go
```diff
@@ -14,22 +14,24 @@ import (
 	"io"
 	"os"
 
-	"cloud.google.com/go/vision"
+	vs "cloud.google.com/go/vision/apiv1"
```
broady: `vs "cloud.google.com/go/vision/apiv1"`

vision/detect/detect.go
```diff
@@ -14,22 +14,24 @@ import (
...
 	"golang.org/x/net/context"
+	pb "google.golang.org/genproto/googleapis/cloud/vision/v1"
```
broady: `visionpb`, perhaps?

vision/detect/detect.go
rakyll: cloud.google.com/go/vision/apiv1 package name is already vision. If we keep the snippets with "vision", users are likely to import the wrong package when they copy paste from the docs: http://imgur.com/a/jizUo.  Use of "vs" disables goimports and make them explicitly import the package by looking at snippet's import path.

rakyll:
proto import is removed, it is never used now.

rakyll:
I am merging this if no one disagrees. In the future, we can revert the import name to "vision" once the handcrafted vision package is removed and we are confident that user's local caches don't contain it anymore :)

broady:
I don't understand, when the import wasn't named, it was imported as "vision".

Other samples import with the same name if the last part of the import path doesn't match the package name. It's what goimports does.

vision/label/label.go
```diff
@@ -12,7 +12,7 @@ import (
 	"path/filepath"
 
 	// [START imports]
-	"cloud.google.com/go/vision"
+	"cloud.google.com/go/vision/apiv1"
```
broady: This one too

vision/vision_quickstart/main.go
```diff
@@ -12,15 +12,15 @@ import (
 	"os"
 
 	// Imports the Google Cloud Vision API client package.
-	"cloud.google.com/go/vision"
+	"cloud.google.com/go/vision/apiv1"
```
broady: And in the quickstart

broady:
LGTM

PR #260, opened 2017-05-24
fix off-by-one typo

ptone:
10 messages that are published are zero-indexed.

This sample never runs to completion as is

cc @broady


broady:
I think it's more straightforward to move the `received++` statement above that if statement.

What do you think?

(Thanks for the PR!)

ptone:
yeah, makes sense.  Never had to update a PR with GH editor, didn't even have a local branch handy :)

PR #269, opened 2017-06-15
mustGetenv is undefined, use os.Getenv instead

LawrenceMok:
mustGetenv is undefined, use os.Getenv instead


broady:
It's not undefined? Look at the bottom of the file.

PR #270, opened 2017-06-16
kms: fix casing of "plaintext" and "ciphertext" in variable names

russ-:
When dealing with encryption, "plaintext" and "ciphertext" are terms of art with
particular meaning. Plaintext is data used as input for encryption, whereas
ciphertext is the output of encryption. Plaintext need not be "plain text" -- it
could be an image, for example.

The example code for other languages are already cased/spelled correctly.

### DIFF
--- a/kms/crypter/crypter.go
+++ b/kms/crypter/crypter.go
@@ -56,7 +56,7 @@ func main() {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


broady:
Thank you.

PR #273, opened 2017-06-28
spanner: encourage bind parameters in statements

broady:
Use parameter binding in the SQL statement.

### DIFF
--- a/spanner/spanner_snippets/snippet.go
+++ b/spanner/spanner_snippets/snippet.go
@@ -253,7 +253,12 @@ func queryUsingIndex(ctx context.Context, w io.Writer, client *spanner.Client) e
 	stmt := spanner.Statement{
 		SQL: `SELECT AlbumId, AlbumTitle, MarketingBudget
 			FROM Albums@{FORCE_INDEX=AlbumsByAlbumTitle}
-			WHERE AlbumTitle >= 'Aardvark' AND AlbumTitle < 'Goo'`}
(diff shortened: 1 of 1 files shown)


spanner/spanner_snippets/snippet.go
```diff
@@ -253,7 +253,12 @@ func queryUsingIndex(ctx context.Context, w io.Writer, client *spanner.Client) e
 	stmt := spanner.Statement{
 		SQL: `SELECT AlbumId, AlbumTitle, MarketingBudget
 			FROM Albums@{FORCE_INDEX=AlbumsByAlbumTitle}
-			WHERE AlbumTitle >= 'Aardvark' AND AlbumTitle < 'Goo'`}
+			WHERE AlbumTitle >= @start_title AND AlbumTitle < @end_title`,
+		Params: map[string]interface{}{
+			"start_title": "Aardvark",
```
vkedia: Can you have the user provide the values for these params with these being the defaults? That is how we have done it in PHP.

broady: I don't think it'll look very good.

```
func queryUsingIndex(ctx context.Context, w io.Writer, client *spanner.Client, queryStart, queryEnd string) error {
	if queryStart == "" {
		queryStart = "Aardvark"
	}
	if queryEnd == "" {
		queryEnd = "Goo"
	}
	// Rest of sample
}
```

vkedia: I dont think its too bad. If it really concerns you, I am ok with not having default value at all and have the user always specify the value.

PR #274, opened 2017-06-29
datastore: add "distinct on" sample

broady:


broady:
@WalterHub @DimaNovik PTAL

PR #275, opened 2017-07-01
pubsub: Subscription.ModifyPushConfig is now Subscription.Update

rakyll:
Build broken due to changes in https://code-review.googlesource.com/c/14311.


broady:
suggest changing commit title to

`pubsub: Subscription.ModifyPushConfig is now Subscription.Update`

LGTM otherwise. Merge when build passes.

PR #279, opened 2017-07-24
Add openapi-appengine.yaml file for App Engine deployments.

p-buse:
After conversation with the Endpoints team, this was the most effective solution we came up with to guide users through setting up Endpoints on App Engine. Having users uncomment code proved to be difficult. If you have another idea that could work better, please let me know. Otherwise, this will have to do.


broady:
Seems overly repetitive, but sure.

PR #282, opened 2017-07-31
Add WithBlock option to NewClient.

80nashi:
Add WithBlock option to NewClient to block.
Without this, calling testPermissions immediately after NewClient can fail with the following error.

2017/07/31 17:20:15 testPermissions failed: rpc error: code = Unavailable desc = there is no address available

### DIFF
--- a/pubsub/topics/main.go
+++ b/pubsub/topics/main.go
@@ -18,6 +18,8 @@ import (
 	"cloud.google.com/go/iam"
 	"cloud.google.com/go/pubsub"
 	"google.golang.org/api/iterator"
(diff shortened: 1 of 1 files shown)


80nashi:
Ah, so, TestPermissions is supposed to run successfully without the option?
I filed https://github.com/GoogleCloudPlatform/google-cloud-go/issues/713 for this issue.

So, we should fix the library instaed of golang-samples.
Please feel free to reject this PR, then.  Thanks!

broady:
It should, yeah. There shouldn't be any need to set options like that.

Thank you for filing that bug!

PR #285, opened 2017-08-01
auth: add implicit auth snippet

rakyll:


broady:
This is only good for those style libraries. Do we not care about google.golang.org/api/... packages at all?

This sample may actually add to the confusion.

auth/snippets.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Package snippets contain Google Cloud authentication snippets.
+package snippets
```
broady: auth_snippets?

auth/snippets.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Package snippets contain Google Cloud authentication snippets.
```
broady: s/contain/contains/

auth/snippets.go
```diff
@@ -0,0 +1,41 @@
...
+	"cloud.google.com/go/storage"
+	"golang.org/x/net/context"
+	"google.golang.org/api/iterator"
+)
+
+func adc() {
+	ctx := context.Background()
+
+	// [START auth_cloud_implicit]
+	// Instantiate a client. If you don't specify credentials
```
broady: Get rid of "instantiate". We're not Java.

auth/snippets.go
rakyll: underscore in package names is a lint error, used authsnippets instead.

auth/snippets.go
rakyll: Done

auth/snippets.go
rakyll: Done

PR #298, opened 2017-08-22
docs/storage: remove listbuckets

broady:
Obsolete sample, previously used for auth samples, but now those are in
/auth and the storage list buckets sample is under /storage/buckets.


rakyll:
LGTM

PR #299, opened 2017-08-22
logging/stdlogging: use a package-level var for logger

broady:
Use a package-level variable for a logger, and set up the logger in init.

Show that closing the logger at the end of the program (end of `main()` is the way to flush any pending messages).

### DIFF
--- a/logging/stdlogging/main.go
+++ b/logging/stdlogging/main.go
@@ -15,7 +15,9 @@ import (
 	"golang.org/x/net/context"
 )
 
-var logger log.Logger
+// logger is a Go standard library logger.
+var logger *log.Logger
+var logClient *logging.Client
 
 func init() {
 	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


logging/stdlogging/main.go
```diff
@@ -24,18 +26,22 @@ func main() {
...
+}
 
+func main() {
+	// Close flushes any pending messages and closes the client.
+	defer logger.Close()
+	
 	// Logs "hello world", log entry is visible at
 	// Stackdriver Logs.
-	logger.Println("hello world")
+	logger.Println("hello world")
```
rakyll: Is this file not gofmt'ed?

broady:
PTAL @rakyll 

Not completely happy with this now. Elevated logging.Client up to a package var. :/

tbpg:
Going through old PRs. It's unclear if we want to do this. I've seen some guidance to just pass around loggers as struct fields or arguments, in which case you wouldn't want the global. So, going to close this to clean up the current state, but feel free to reopen if you want to go through with the update.

PR #302, opened 2017-09-07
spanner: fix error handling in snippet code

madari:
The proposal fixes the error handling in the Go sample code for Spanner.

In particular, the sample code contains code such as this:
`if err := op.Wait(); err == nil { print(ok) }; return err`

In cases such as this, errors from op.Wait will not be correctly propagated thus causing the program to always exit with "success".


broady:
Subtle!

Thank you. I have a strong preference for this code style, and this is a good reminder of why.

/cc @vkedia

PR #304, opened 2017-09-10
Fix error message for mysql transactions

mousavian:
Fixing the error message raising from mysql but stating "memorydb" wrongly.


mousavian:
I signed it!

broady:
thanks!

PR #305, opened 2017-09-11
vision: fix broken demos

danicat:
The path to the `cat.jpg` file is broken in the vision_quickstart demo. This pull request fixes it.


broady:
thanks!

Issue #235, opened 2017-04-12
Add READMEs for speech


danaharon:
1) Readme should contain instructions that explain how to get the sample to compile and how to run it.
2) Readme should include the "go get -u cloud.google.com/go/speech/apiv1" command so that users will know that they need to run it

odeke-em:
@danaharon would you like to submit a PR for this? Your issue description makes for great requirements for a PR, IMHO.

danicat:
Seems like nobody is working on this right now... it's ok if I do it?

adams-sarah:
yas! that would be wonderful, thank you @danicat

adams-sarah:
@danicat, can't seem to assign you. but this is yours if you want it :)

danicat:
@adams-sarah probably only collaborators can be assigned, but let me handle this. I'm doing some cleaning up in the examples for all the machine learning APIs. I've used some of them in one of my presentations last week so I want to do this while it's fresh. :)

broady:
thanks, @danicat 

Check out other language repos, too.

https://github.com/GoogleCloudPlatform/python-docs-samples
https://github.com/GoogleCloudPlatform/nodejs-docs-samples

danicat:
Hey @broady, I had a look at the other repos... it seems that every repo has it's own way of organizing stuff, even though the content should be pretty much the same. I wonder if it wouldn't be possible to set a standard for everybody to follow, at least regarding the documentation and repo structure.

danicat:
May we close this one? Or there's anything else that needs to be done?

broady:
Seems done to me. If there are any improvements, someone can open a new issue.

PR #310, opened 2017-09-12
speech: add a readme for each example

danicat:
I've added a README.md file for each Speech API example. I've also moved the raw audio files to a testdata directory just like the Vision API examples to avoid duplication. Solves issue #235.


speech/speech_quickstart/main.go
```diff
@@ -29,7 +29,7 @@ func main() {
 	}
 
 	// Sets the name of the audio file to transcribe.
-	filename := "/path/to/audio.raw"
```
broady: I think these are fine, actually. For the quickstarts, we do expect people to modify the programs. I'd rather not make them too specific to this repo.

broady:
READMEs look good, but I don't think the deletion/moving of the raw audio files is an improvement. Could you revert that?

speech/speech_quickstart/main.go
danicat: ok! I'll revert to the original

danicat:
I will, but I wonder what would be the benefit of having several copies of the same file all over the place? The same thing (having a testdata folder) was done for the vision API and I thought it was a good idea.

broady:
Hm, OK. If they're all exactly the same, sounds good to me.

danicat:
Yes, they are all the same. 

And, yes, *now* you can assume the integration tests are ok. 

I didn't know this limitation about Travis CI. I'm still learning how to contribute to this repo, sorry about the trouble!

broady:
Thank you! I can add you as a contributor so you can push to branches in this repo rather than to your fork. Then the integration tests will run.

danicat:
Oh, great! Thank you!

PR #315, opened 2017-09-22
storage: add example for set a CORS

sanleon:
add the example for setting a cors on the bucket.
I can't found a method that set a cors on cloud.google.com/go/storage.
So I'm using google.golang.org/api/storage/v1


broady:
I think we should just add this to the library rather than adding this code sample.

Thanks, though.

See https://github.com/GoogleCloudPlatform/google-cloud-go/issues/889 for the feature request.

Issue #246, opened 2017-04-28
speech: fix API example deadline to 240s, to avoid GRPC deadline too short error


mattetti:
The sample code wasn't working for me due to a GRPC setting error:

```
2017/04/27 18:23:36 Could not recognize: code:3 message:"Client GRPC deadline too short. Should be at least: 3 * audio-duration + 5 seconds. Current deadline is: 199.99940906430129 second(s). Required at least: 204 second(s)." 
```

I fixed that by extending the deadline in the context:

```golang
        bgctx := context.Background()
	ctx, _ := context.WithDeadline(bgctx, time.Now().Add(240*time.Second))
	conn, err := transport.DialGRPC(ctx,
		option.WithEndpoint("speech.googleapis.com:443"),
		option.WithScopes("https://www.googleapis.com/auth/cloud-platform"),
	)
```

broady:
Which sample? Shouldn't need to make any modification like this.

mattetti:
Here is the sample I had issues with https://github.com/GoogleCloudPlatform/golang-samples/blob/01f58b1df950edfbe1ff28eff9e870fabe6790fc/speech/livecaption/livecaption.go

broady:
Found out the problem.

Speech API supports only 60 seconds of streaming recognition:
https://cloud.google.com/speech/limits#content

https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1894

We need a better error and/or docs. Either the Speech client or underlying Speech API should return a better error.

odeke-em:
Thank you for reporting this @mattetti, I too encountered the same error, sometime late
last year but things were too noisy for me to file a bug here.

With what @broady deduced in https://github.com/GoogleCloudPlatform/golang-samples/issues/246#issuecomment-319887146, @mattetti might you be interested or have time to send a PR to fix this? If not, no biggie, it might picked up.

mattetti:
@odeke-em unfortunately I moved on and I won't be able to, but thanks a lot for offering.

danicat:
@broady I'll take this one, ok? The new README already states briefly about the limits, but I could make it more explicit in this example. Also, I'll work on a better error message.

broady:
Sounds good. Thanks.

Not sure if you can do anything about the error message, my first thought is it should be fixed in the backend.

danicat:
Ok, I've finally managed to reproduce the error... there are actually two possible scenarios:
[...]

PR #316, opened 2017-09-22
speech: fix livecaption README and error message

danicat:
Fixes issue #246.

I've updated the example README to include a bit more information about the limits and usage. I've also updated the code to filter the errors 3 and 11 to print a more user friendly message regarding the 60 second limit.

### DIFF
--- a/speech/livecaption/README.md
+++ b/speech/livecaption/README.md
@@ -31,8 +31,57 @@ go build
 cat ../testdata/audio.raw | livecaption
 ```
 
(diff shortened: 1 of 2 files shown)


danicat:
@broady I've changed the implementation to just print an additional warning instead of over writing the error message. If you still think it is not ok I can revert the changes to the code and leave just the changes to the docs.

mowsmith:
Thank you for the doc changes. It helped me get my mic from gstreamer working.

tbpg:
Looks good to me! I rebased and will merge.

PR #317, opened 2017-09-26
bigquery: add new parameter for Dataset.Create

broady:
The BigQuery Go package changed to require a metadata parameter.

### DIFF
--- a/bigquery/bigquery_quickstart/main.go
+++ b/bigquery/bigquery_quickstart/main.go
@@ -30,11 +30,8 @@ func main() {
 	// Sets the name for the new dataset.
 	datasetName := "my_new_dataset"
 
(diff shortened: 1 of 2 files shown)


bigquery/bigquery_quickstart/main.go
```diff
@@ -30,11 +30,8 @@ func main() {
 	// Sets the name for the new dataset.
 	datasetName := "my_new_dataset"
 
-	// Creates a Dataset instance.
-	dataset := client.Dataset(datasetName)
```
broady: note: removed this. what do you think?

the comment was quite poorly worded. can change it back and remove the comment?

broady:
Merging now to green the build.

PR #319, opened 2017-09-27
storage/gcsupload: don't close client and return handles

broady:
Handles aren't valid once the client is closed.

See: https://code-review.googlesource.com/16950


odeke-em:
PS: I couldn't merge early because the CI hadn't run yet, but thanks for doing that ;)

PR #321, opened 2017-09-27
bigquery: fix tests

broady:
The default changed from legacy SQL to standard SQL, so this SELECT
statement must be updated.


broady:
@rakyll this should fix those tests that were failing on your other PR

PR #323, opened 2017-10-05
travis: drop 1.6.x, add 1.9.x

broady:


broady:
RIP 1.6.

gRPC dropped support for it yesterday.

1.8 is available on GAE, and will hopefully go GA soon.

odeke-em:
Nice! Finally 1.8 on GAE, thus we can also finally replace "golang.org/x/context" --> "context"

jba:
@broady Can you point me to the gRPC announcement or whatever? This seems a little premature to me, since 1.8 is not yet GA on AppEngine.

@odeke-em I don't think we can make the change to `context` until we can depend on type aliases, which means we have to wait for 1.10 to come out, and for GAE to be on 1.9.

odeke-em:
@jba ah I see, makes sense. Drat, double drat.

PR #327, opened 2017-10-31
travis: test with most recent minor go version

adams-sarah:
Just doing this for consistency with other cloud clients (which i'm updating atm to add 1.9).
Feel free to reject if you disagree.


broady:
Thanks! This is better. Could you add 1.6.x back in while you're there?

PR #328, opened 2017-11-06
language: add snippets for text classification

broady:


language/analyze/analyze_test.go
```diff
@@ -85,6 +85,17 @@ func TestSyntax(t *testing.T) {
...
 
+func TestClassify(t *testing.T) {
+	testutil.SystemTest(t)
+	ctx, c := newClient(t)
+
+	res, err := analyzeSyntax(ctx, c, "If you bend the gopher, his belly folds.")
+	if err != nil {
+		t.Fatalf("got %v, want nil err", err)
+	}
+	printResp(res, err)
```
odeke-em: @broady this test always prints out something to stdout eh? Perhaps could we make this
```go
      if testing.Verbose() {
             printResp(res, err)
      }
```
what do you think?

broady: Thanks. I forgot to actually write the test. PTAL.

PR #329, opened 2017-11-07
Check callback session state

huikang:
Session is invalid if it is a new session.

Signed-off-by: Hui Kang <kangh@us.ibm.com>

### DIFF
--- a/getting-started/bookshelf/app/auth.go
+++ b/getting-started/bookshelf/app/auth.go
@@ -95,6 +95,10 @@ func oauthCallbackHandler(w http.ResponseWriter, r *http.Request) *appError {
 		return appErrorf(err, "invalid state parameter. try logging in again.")
 	}
 
+	if oauthFlowSession.IsNew {
+		return appErrorf(err, "invalid session state. try logging in again.")
+	}
+
 	redirectURL, ok := oauthFlowSession.Values[oauthFlowRedirectKey].(string)
 	// Validate this callback request came from the app.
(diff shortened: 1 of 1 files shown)


broady:
Why? If it's new, then ok will be false here:

```
	redirectURL, ok := oauthFlowSession.Values[oauthFlowRedirectKey].(string)
	// Validate this callback request came from the app.
	if !ok {
		return appErrorf(err, "invalid state parameter. try logging in again.")
	}
```

huikang:
@broady if the season is new, that means this session is not callback session from previous login.

The ok tests if the redirect URL is a string type. Am I right?

broady:
Yes, and it'll only be a string if it's not a new session.

It'll be nil if it's a new session, and fail the type check.

huikang:
On Tue, Nov 7, 2017 at 4:06 PM, Chris Broadfoot <notifications@github.com>
wrote:

> Yes, and it'll only be a string if it's not a new session.
>
> It'll be nil if it's a new session, and fail the type check.
>
Thanks for your confirming. Shall I further make change to the PR or do you
think it is good?

broady:
All good. Thanks!

huikang:
@broady Does the PR fix the problem? I thought it should be merged to fix the problem of invalid session. No?

broady:
I thought I explained why there's no problem. The "IsNew" check is redundant.

Can you explain how you came to an issue? Can you provide steps to reproduce?

PR #330, opened 2017-11-08
docs/appengine: add cloudsql_postgres sample

broady:


docs/appengine/cloudsql_postgres/cloudsql.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2016 Google Inc. All rights reserved.
```
odeke-em: s/2016/2017/g

broady: done!

docs/appengine/cloudsql_postgres/cloudsql.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"bytes"
+	"database/sql"
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+
+	_ "github.com/broady/gae-postgres"
```
odeke-em: Nice repo :)

docs/appengine/cloudsql_postgres/cloudsql.go
```diff
@@ -0,0 +1,67 @@
...
+		return
+	}
+
+	connectionName := mustGetenv("CLOUDSQL_CONNECTION_NAME")
+	user := mustGetenv("CLOUDSQL_USER")
+	password := os.Getenv("CLOUDSQL_PASSWORD") // NOTE: password may be empty
+
+	w.Header().Set("Content-Type", "text/plain")
+
+	db, err := sql.Open("gae-postgres", fmt.Sprintf("user=%s password=%s cloudsql=%s", user, password, connectionName))
```
odeke-em: @broady I wonder if we should instead open this outside of the handler so that the sample won't have to open a new connection each time. However, if making a global DB connection is undesirable, then this is works.

broady: Huh! Weird. When I wrote this sample (a long time ago), I assumed you'd need to pass in a ctx somewhere.

It seems obvious now that GAE would let you connect to the database without the context of an incoming request. Cool.

Will leave this PR for now, and update the MySQL sample tomorrow first.

odeke-em: Awesome, cool. Catch you later!

PR #332, opened 2017-11-10
videointelligence: add quickstart

broady:


videointelligence/video_quickstart/main.go
```diff
@@ -0,0 +1,65 @@
...
+		log.Fatalf("Failed to start annotation job: %v", err)
+	}
+
+	resp, err := op.Wait(ctx)
+	if err != nil {
+		log.Fatalf("Failed to annotate: %v", err)
+	}
+
+	// Only one video was processed, so get the first result.
+	result := resp.GetAnnotationResults()[0]
```
odeke-em: How about:
```
annotations := resp.GetAnnoationResults()
if len(annotations) == 0 {
   log.Fatalf("No annotations found")
}
// Only one video was processed, so get the first
results := annotations[0]
```

broady: I think this is guaranteed to be len==1.

broady:
R= @remi @tswast

videointelligence/video_quickstart/main.go
odeke-em: Oh okay, gotcha, thanks for that update. In that case, you've got my vote Mr. President!

broady: I'll wait for @remi (Cloud ML API expert) to confirm.

videointelligence/video_quickstart/main.go
```diff
@@ -0,0 +1,65 @@
...
+	ctx := context.Background()
+
+	// Creates a client.
+	client, err := video.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("Failed to create client: %v", err)
+	}
+
+	op, err := client.AnnotateVideo(ctx, &videopb.AnnotateVideoRequest{
+		InputUri: "gs://demomaker/cat.mp4",
```
tswast: I created a `cloud-storage-samples` bucket that might make sense to use for samples like this.

broady:
```
AccessDeniedException: 403 cbro@google.com does not have storage.objects.list access to cloud-storage-samples.
```

tswast: Oops, I forgot to make it public. I've made you owner of the `storage-samples` project that owns that bucket. I'd make it public myself but am WFH without a corp cert right now.

tswast: Ah, I can't make you an owner with the app, either. @frankyn might be able to help.

frankyn: You are an admin now @broady!

broady: Merging now without this change. Please file a bug if this URI isn't appropriate for the quickstart.

PR #333, opened 2017-11-10
videointelligence: add video_analyze program

broady:
Includes:
* Label detection
* Shot analysis
* Explicit content detection

Use either a local file or GCS URI.

PR #334, opened 2017-11-10
Same as #333 except using github.com/broady/preprocess

broady:


broady:
@odeke-em what do you think of this template/preprocessing language instead?

prototyped: github.com/broady/preprocess

odeke-em:
Hello @broady, sorry for the late reply, I was away from my computer for a bit. Nice!! Let me take a look at more samples of it.

videointelligence/video_analyze/gen/template.go
```diff
@@ -0,0 +1,127 @@
...
+func boilerplate() { //# omit
+	//# def dorequest
+	ctx := context.Background()
+	client, err := video.NewClient(ctx)
+	if err != nil {
+		return err
+	}
+	//# if !gcs
+
+	f, err := os.Open(file)
```
odeke-em: We could replace the 5 lines below with:
```go
fileBytes, err := ioutil.ReadFile(file)
```

broady: Great call. Done.

PR #335, opened 2017-11-14
docs/appengine: remove cloudsql_postgres sample

broady:
This call path is unsupported (but happens to work).


aarti:
I see this example for a standard environment is deleted - is cloudsql/postgres unsupported in the standard environment, Do I have to switch to flexible?

broady:
It's supported for Flex, and I expect will be supported for Standard
eventually (but likely through a different mechanism).

It's technically possible to connect to it on standard via the Socket API,
but you'd need to create a new database connection for each incoming
request.

Flex is certainly the easiest option today.

PR #339, opened 2017-11-27
Update README.md

navinger:
I removed the links to the quickstarts because they use the bookstore-grpc sample and not this one.

### DIFF
--- a/endpoints/getting-started-grpc/README.md
+++ b/endpoints/getting-started-grpc/README.md
@@ -2,11 +2,6 @@
 
 This sample demonstrates how to use Google Cloud Endpoints using Go and gRPC.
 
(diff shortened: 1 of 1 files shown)


broady:
Is this sample just completely unused? BTW, the first link doesn't even work (it 404's)

/cc @p-buse

navinger:
I know the link is broken. The Doc Tracker report found it, which is how I
found it.

I have links to this getting started with gRPC sample in the docs. There
was a misunderstanding earlier in the year. I thought the bookstore gRPC
sample was going to be ported to all the languages (currently it is in
Python and Java).

I’ll talk to buser to figure out what to do for a long-term solution.

PR #349, opened 2017-12-13
added region tags

angelafunk:


appengine/gophers/gophers-3/index.html
```diff
@@ -8,14 +8,14 @@
 <h1>The Gopher Network</h1>
 
 <img id="logo" src="/static/gcp-gopher.svg" width="100">
-
+// [START html_template]
```
broady: These have to be HTML comments

PR #351, opened 2017-12-13
update app.yaml

angelafunk:
+added region tags
+added upload of static index.html file
+changed static_dir name to static-files, so user knows that the url and static_dir name doesn't have to be the same.
+added comments

### DIFF
--- a/appengine/gophers/gophers-2/app.yaml
+++ b/appengine/gophers/gophers-2/app.yaml
@@ -2,7 +2,15 @@ runtime: go
 api_version: go1
 
 handlers:
+// [START static_handlers]
+# If the path is empty, show static file index.html
+- url: /
+  static_files: index.html
+  upload: index.html
+  
+# Otherwise, find file in static-files directory
 - url: /static
-  static_dir: static
+  static_dir: static-files
+// [END static_handlers]  
 - url: /.*
   script: _go_app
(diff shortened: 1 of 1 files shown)


appengine/gophers/gophers-2/app.yaml
```diff
@@ -2,7 +2,15 @@ runtime: go
 api_version: go1
 
 handlers:
+// [START static_handlers]
```
broady: Need to be `#` comments for YAML files, I think.

appengine/gophers/gophers-2/app.yaml
```diff
@@ -2,7 +2,15 @@ runtime: go
...
+# If the path is empty, show static file index.html
+- url: /
+  static_files: index.html
+  upload: index.html
+  
+# Otherwise, find file in static-files directory
 - url: /static
-  static_dir: static
+  static_dir: static-files
```
broady: I think this was correct before?

angelafunk: Thinking about changing the folder name to static-files so users know that the folder name doesn't have to be the same as the URL. WDYT?
Also, took out the template variable in gophers-2/main.go, so I can introduce templates all at once in next page of tutorial.

broady: Keeping it to "static" makes the most sense. They can play around with it if they need to, but keeping the directory the same as the eventual path is going to be what most people want.

PR #353, opened 2017-12-13
update main.go

angelafunk:
not using templates yet, so deleted template variable

### DIFF
--- a/appengine/gophers/gophers-2/main.go
+++ b/appengine/gophers/gophers-2/main.go
@@ -6,11 +6,6 @@ import (
 
 	"google.golang.org/appengine"
 )
-
-var (
-	indexTemplate = template.Must(template.ParseFiles("index.html"))
-)
-
 func main() {
 	http.HandleFunc("/", indexHandler)
 	appengine.Main()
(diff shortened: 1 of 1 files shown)


appengine/gophers/gophers-2/main.go
```diff
@@ -21,6 +16,4 @@ func indexHandler(w http.ResponseWriter, r *http.Request) {
 		http.Redirect(w, r, "/", http.StatusFound)
 		return
 	}
-
-	indexTemplate.Execute(w, nil)
```
broady: This means the page will be blank. We still want to render the template, even though there aren't any parameters.

It's the same as writing out the index file.

broady: Or is this because you added a static handler for the index page?

angelafunk: static handler! WDYT?

broady: Sounds good.

This looks OK.

appengine/gophers/gophers-2/main.go
```diff
@@ -6,11 +6,6 @@ import (
 
 	"google.golang.org/appengine"
 )
-
-var (
-	indexTemplate = template.Must(template.ParseFiles("index.html"))
-)
-
```
broady: Add back one of these newlines

PR #361, opened 2017-12-28
Endpoints example with JWT authentication.

kryzthov:
- Add flag to configure the interface used by the server.
 - Add flag to specify an authentication bearer token in the client.
 - Add service configuration for authentication through Google service
   account.
 - Update client to optionally authenticate request using a JWT token
   signed by a Google service account.
 - Add documentation for service account authentication.
 - `gcloud service-management` -> `gcloud endpoints services`

@p-buse @navinger Please take a look. This change adds an example of authentication of a request using a JWT token signed by a Google service account.

### DIFF
--- a/endpoints/getting-started-grpc/client/main.go
+++ b/endpoints/getting-started-grpc/client/main.go
@@ -83,6 +83,8 @@ func main() {
 			log.Fatalf("Unable to generate JWT token: %v", err)
 		}
 		*token = jwt.AccessToken
+		// Note: the generated JWT token has a 1h TTL.
+		// Make sure to refresh the token before it expires (see jwt.Expiry).
 	}
 
 	ctx := context.Background()
--- a/endpoints/getting-started-grpc/client/main.go
+++ b/endpoints/getting-started-grpc/client/main.go
@@ -40,11 +40,10 @@ import (
 	"log"
 
 	"golang.org/x/net/context"
+	"golang.org/x/oauth2/google"
 	"google.golang.org/grpc"
 	pb "google.golang.org/grpc/examples/helloworld/helloworld"
 	"google.golang.org/grpc/metadata"
-
-	googleoauth2 "golang.org/x/oauth2/google"
 )
 
 const defaultName = "world"
@@ -70,12 +69,12 @@ func main() {
 
 	if *keyfile != "" {
 		log.Printf("Authenticating using Google service account key in %s", *keyfile)
-		sakey, err := ioutil.ReadFile(*keyfile)
+		keyBytes, err := ioutil.ReadFile(*keyfile)
 		if err != nil {
-			log.Fatalf("Unable to read service account key file %s: %s", *keyfile, err)
+			log.Fatalf("Unable to read service account key file %s: %v", *keyfile, err)
 		}
 
-		tokenSource, err := googleoauth2.JWTAccessTokenSourceFromJSON(sakey, *audience)
+		tokenSource, err := google.JWTAccessTokenSourceFromJSON(keyBytes, *audience)
 		if err != nil {
 			log.Fatalf("Error building JWT access token source: %v", err)
 		}
--- a/endpoints/getting-started-grpc/client/main.go
+++ b/endpoints/getting-started-grpc/client/main.go
@@ -83,8 +83,9 @@ func main() {
 			log.Fatalf("Unable to generate JWT token: %v", err)
 		}
 		*token = jwt.AccessToken
-		// Note: the generated JWT token has a 1h TTL.
-		// Make sure to refresh the token before it expires (see jwt.Expiry).
+		// NOTE: the generated JWT token has a 1h TTL.
+		// Make sure to refresh the token before it expires by calling TokenSource.Token() for each outgoing requests.
+		// Calls to this particular implementation of TokenSource.Token() are cheap.
 	}
 
 	ctx := context.Background()
--- a/endpoints/getting-started-grpc/client/main.go
+++ b/endpoints/getting-started-grpc/client/main.go
@@ -35,19 +35,26 @@ package main
 
 import (
 	"flag"
+	"fmt"
+	"io/ioutil"
 	"log"
 
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	pb "google.golang.org/grpc/examples/helloworld/helloworld"
 	"google.golang.org/grpc/metadata"
+
+	googleoauth2 "golang.org/x/oauth2/google"
 )
 
 const defaultName = "world"
 
 var (
-	addr = flag.String("addr", "127.0.0.1:50051", "Address of grpc server.")
-	key  = flag.String("api-key", "", "API key.")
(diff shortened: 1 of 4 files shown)


kryzthov:
@broady Would you be able to take a look, when you have a chance? Thanks!

endpoints/getting-started-grpc/client/main.go
```diff
@@ -35,19 +35,26 @@ package main
...
+	"fmt"
+	"io/ioutil"
 	"log"
 
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	pb "google.golang.org/grpc/examples/helloworld/helloworld"
 	"google.golang.org/grpc/metadata"
+
+	googleoauth2 "golang.org/x/oauth2/google"
```
broady: please leave this as it was. just "google"

endpoints/getting-started-grpc/client/main.go
```diff
@@ -61,8 +68,33 @@ func main() {
 	defer conn.Close()
 	c := pb.NewGreeterClient(conn)
 
+	if *keyfile != "" {
+		log.Printf("Authenticating using Google service account key in %s", *keyfile)
+		sakey, err := ioutil.ReadFile(*keyfile)
+		if err != nil {
+			log.Fatalf("Unable to read service account key file %s: %s", *keyfile, err)
```
broady: `%v` for the second formatting directive (for err)

endpoints/getting-started-grpc/client/main.go
```diff
@@ -61,8 +68,33 @@ func main() {
 	defer conn.Close()
 	c := pb.NewGreeterClient(conn)
 
+	if *keyfile != "" {
+		log.Printf("Authenticating using Google service account key in %s", *keyfile)
+		sakey, err := ioutil.ReadFile(*keyfile)
```
broady: keyContents, perhaps? (rather than sakey)

or keyBytes

endpoints/getting-started-grpc/client/main.go
```diff
@@ -61,8 +68,33 @@ func main() {
...
+
+		tokenSource, err := googleoauth2.JWTAccessTokenSourceFromJSON(sakey, *audience)
+		if err != nil {
+			log.Fatalf("Error building JWT access token source: %v", err)
+		}
+		jwt, err := tokenSource.Token()
+		if err != nil {
+			log.Fatalf("Unable to generate JWT token: %v", err)
+		}
+		*token = jwt.AccessToken
```
broady: This is OK for this program, but should probably check token validation before using it later on L96.

endpoints/getting-started-grpc/server/main.go
```diff
@@ -43,20 +44,23 @@ import (
 	"google.golang.org/grpc/reflection"
 )
 
-const (
-	port = ":50051"
+var (
+	listen = flag.String("listen", ":50051", "Network interface:port to listen on for gRPC connections.")
```
broady: host:port

endpoints/getting-started-grpc/server/main.go
```diff
@@ -43,20 +44,23 @@ import (
...
 )
 
 // server is used to implement helloworld.GreeterServer.
 type server struct{}
 
 // SayHello implements helloworld.GreeterServer
 func (s *server) SayHello(ctx context.Context, in *pb.HelloRequest) (*pb.HelloReply, error) {
+	log.Printf("Handling SayHello request [%s] with context %s", in, ctx)
```
broady: %v rather than %s? (for both)

endpoints/getting-started-grpc/server/main.go
broady: call this addr, rather than listen.

broady:
Thank you very much for doing this. Looks generally good. Just a bunch of nits.

endpoints/getting-started-grpc/client/main.go
kryzthov: Done

endpoints/getting-started-grpc/client/main.go
kryzthov: Done

endpoints/getting-started-grpc/client/main.go
kryzthov: Done

endpoints/getting-started-grpc/server/main.go
kryzthov: Done

endpoints/getting-started-grpc/server/main.go
kryzthov: Done

endpoints/getting-started-grpc/server/main.go
kryzthov: Done

endpoints/getting-started-grpc/client/main.go
kryzthov: What sort of validation are you thinking of?

kryzthov:
Fixed comments, please take another look.

endpoints/getting-started-grpc/client/main.go
broady: Expiry, mostly. But it looks like these tokens will never expire.

Still, the call to tokenSource.Token is cheap, and you'd generally want to call it for every outgoing request.

You'd need to restructure this code a bit to accommodate this suggestion. But it's just a suggestion.

broady:
LGTM, let me know if you want to change the handling of Token(). Otherwise, I'll submit.

endpoints/getting-started-grpc/client/main.go
kryzthov: Ah, makes sense. Across all implementations I've seen so far, JWT tokens are generated with a TTL of 1h. The Go implementation hard-codes 1h (https://github.com/golang/oauth2/blob/master/google/jwt.go#L56).

kryzthov:
I believe this is a good first step.
I'll follow up with other examples. In particular, it would be nice to have an example that demonstrates how to use an ID Token (although I suspect the oauth2 library needs to be adapted for that first).

broady:
Right. I remember doing some work with id_token in the past but can't remember if I got it working or not.

Maybe you could just add a comment saying that it has a 1h ttl and in production code, you'd want to call Token() for each request?

kryzthov:
Yes, I just added a note about the TTL and `jwt.Expiry`.

endpoints/getting-started-grpc/client/main.go
broady: Could the comment instead say:

```
// NOTE: the generated JWT token has a 1h TTL.
// Make sure to refresh the token before it expires by calling TokenSource.Token() for each outgoing request.
// Calls to TokenSource.Token() are cheap.
```

(all caps on NOTE, too)

kryzthov: I updated the note as you suggested, although I wanted to make clear that `TokenSource.Token()` calls are not cheap in general. This particular implementation is reasonably cheap but not negligible, and other implementations (eg. for 2-legged OAuth or ID token) might involve additional RPCs.

PR #363, opened 2018-01-03
Prevent nil deref when pubsub topic or sub already exists

jhillyerd:
Pubsub Create funcs now return nil topic and subscription upon error,
necessating this fix to prevent segfault on subsequent runs.

### DIFF
--- a/getting-started/bookshelf/pubsub_worker/worker.go
+++ b/getting-started/bookshelf/pubsub_worker/worker.go
@@ -51,7 +51,7 @@ func main() {
 	topic := bookshelf.PubsubClient.Topic(bookshelf.PubsubTopicID)
 	exists, err := topic.Exists(ctx)
 	if err != nil {
-		log.Fatal("Error checking for topic: %v", bookshelf.PubsubTopicID, err)
+		log.Fatalf("Error checking for topic: %v", err)
 	}
 	if !exists {
 		if _, err := bookshelf.PubsubClient.CreateTopic(ctx, bookshelf.PubsubTopicID); err != nil {
@@ -63,7 +63,7 @@ func main() {
 	subscription = bookshelf.PubsubClient.Subscription(subName)
 	exists, err = subscription.Exists(ctx)
 	if err != nil {
-		log.Fatal("Error checking for subscription: %v", subName, err)
+		log.Fatalf("Error checking for subscription: %v", err)
 	}
 	if !exists {
 		if _, err = bookshelf.PubsubClient.CreateSubscription(
--- a/getting-started/bookshelf/pubsub_worker/worker.go
+++ b/getting-started/bookshelf/pubsub_worker/worker.go
@@ -50,7 +50,13 @@ func main() {
 	// ignore returned errors, which will be "already exists". If they're fatal
 	// errors, then following calls (e.g. in the subscribe function) will also fail.
 	topic, _ := bookshelf.PubsubClient.CreateTopic(ctx, bookshelf.PubsubTopicID)
+	if topic == nil {
+		topic = bookshelf.PubsubClient.Topic(bookshelf.PubsubTopicID)
+	}
 	subscription, _ = bookshelf.PubsubClient.CreateSubscription(ctx, subName, pubsub.SubscriptionConfig{Topic: topic})
+	if subscription == nil {
+		subscription = bookshelf.PubsubClient.Subscription(subName)
+	}
 
 	// Start worker goroutine.
 	go subscribe()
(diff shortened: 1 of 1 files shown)


jhillyerd:
Please re-run tests, travis timed out.

Breaking change: https://github.com/GoogleCloudPlatform/google-cloud-go/commit/776cf82fdbc9cb47b6b0eaeaf8932880bd666bad

broady:
I don't think this change to the sample code is correct. 

This seems like a bug (unintentional breaking change) in the client library. We should maintain the old behavior and add a test.

/cc @jba

jba:
This may have been a breaking change, but only because it was a bugfix. Go functions should never return a non-zero value with an error unless there is a very compelling reason.

The sample code isn't written correctly. It should never have ignored the errors from CreateXXX in the first place. How do you know they're AlreadyExists? Maybe you have bad creds, the service is unavailable, or you're out of quota, or maybe you spelled something wrong. 

The right code is something like:
```
topic, err := client.CreateTopic(...)
if status.Code(err) == codes.AlreadyExists {
    topic = client.Topic(...)
} else if err != nil {
    log.Fatal(err) // or whatever
}
```

broady:
I think it was my laziness in parsing the error. (see the comment above those calls)

We should probably have a sentinel error for AlreadyExists in the pubsub package (rather than having to go out to the transport-dependent error package).

jba:
> We should probably have a sentinel error for AlreadyExists in the pubsub package (rather than having to go out to the transport-dependent error package).

That's a tough one. Do we do this for every client? And for all the codes, or just some of them? It's a slippery slope. I think the most pragmatic solution is to let the transport layer leak through.

broady:
I think this code predates the "AlreadyExists" checks. It's probably better to just use those (even though that's a bit racy).

jhillyerd:
Updated to use Exists methods.

getting-started/bookshelf/pubsub_worker/worker.go
```diff
@@ -47,10 +47,33 @@ func main() {
...
+	subscription = bookshelf.PubsubClient.Subscription(subName)
+	exists, err = subscription.Exists(ctx)
+	if err != nil {
+		log.Fatalf("Error checking for subscription: %v", err)
+	}
+	if !exists {
+		if _, err = bookshelf.PubsubClient.CreateSubscription(
+			ctx,
+			subName,
+			pubsub.SubscriptionConfig{Topic: topic},
```
broady: nit: put this back on one line. you can split the SubscriptionConfig out onto a new line if you want.

getting-started/bookshelf/pubsub_worker/worker.go
```diff
@@ -47,10 +47,33 @@ func main() {
...
+	if err != nil {
+		log.Fatalf("Error checking for topic: %v", err)
+	}
+	if !exists {
+		if _, err := bookshelf.PubsubClient.CreateTopic(ctx, bookshelf.PubsubTopicID); err != nil {
+			log.Fatalf("Failed to create topic: %v", err)
+		}
+	}
+
+	// Create topic subscription if it does not yet exist
```
broady: nit: full stop after comment (full sentences).

(on both)

broady:
LGTM

getting-started/bookshelf/pubsub_worker/worker.go
jhillyerd: Does that include the err != nil portion?  In otherwords, A:

		if _, err = bookshelf.PubsubClient.CreateSubscription(ctx, subName,
			pubsub.SubscriptionConfig{Topic: topic},
		); err != nil {
			log.Fatalf("Failed to create subscription: %v", err)
		}

or B (I find this harder to read):

		if _, err = bookshelf.PubsubClient.CreateSubscription(ctx, subName,
			pubsub.SubscriptionConfig{Topic: topic}); err != nil {
			log.Fatalf("Failed to create subscription: %v", err)
		}

broady:
```
	if _, err := bookshelf.PubsubClient.CreateSubscription(ctx, subName, pubsub.SubscriptionConfig{Topic: topic}); err != nil {
		log.Fatalf("Failed to create subscription: %v", err)
	}
```

or

```
	if _, err := bookshelf.PubsubClient.CreateSubscription(ctx, subName, pubsub.SubscriptionConfig{
		Topic: topic,
	}); err != nil {
		log.Fatalf("Failed to create subscription: %v", err)
	}
```

but the latter isn't great, either. the long line just seems better.

jhillyerd:
Long line it is.

jhillyerd:
I don't have write access, so someone else will need to merge.

PR #367, opened 2018-01-08
all: add uuid.Must calls to satori/go.uuid

broady:
The uuid package changed to return an error.


broady:
@jadekler FYI

jadekler:
@broady Yeah, noticed that late last week. Should've connected the dots and checked this repo - will do so next time

PR #368, opened 2018-01-08
bigquery/bigquery_datatransfer_quickstart: fix license, nits

broady:
* License fails the license checker (license checker and headers should be
updated to work with Google LLC).
* Fix godoc strings (region tag was part of the godoc).
* Re-order imports to be prettier.


broady:
@jadekler FYI

PR #370, opened 2018-01-16
Add sample program to demonstrate how to enable debugging tools

rakyll:
This sample will be used as a reference for the guide at
https://github.com/GoogleCloudPlatform/google-cloud-go/wiki/OpenCensus.

### DIFF
--- a/opencensus/spanner_quickstart/main.go
+++ b/opencensus/spanner_quickstart/main.go
@@ -1,0 +1,60 @@
+// Copyright 2018 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Sample spanner_quickstart contains a sample application that
+// uses Google Spanner Go client, and reports metrics
+// and traces for the outgoing requests.
+package main
+
+import (
+	"log"
+
+	"cloud.google.com/go/spanner"
+	ss "go.opencensus.io/exporter/stats/stackdriver"
+	ts "go.opencensus.io/exporter/trace/stackdriver"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/trace"
+	"golang.org/x/net/context"
+)
+
+func main() {
+	ctx := context.Background()
+
+	// Enable OpenCensus exporters to export traces and metrics
+	// to Stackdriver Monitoring and Tracing.
+	// Exporters use Application Default Credentials to authenticate.
+	// See https://developers.google.com/identity/protocols/application-default-credentials
+	// for more details.
+	statsExporter, err := ss.NewExporter(ss.Options{ProjectID: "your-project-id"})
+	if err != nil {
+		log.Fatal(err)
+	}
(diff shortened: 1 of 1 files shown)


opencensus/spanner/main.go
```diff
@@ -0,0 +1,60 @@
...
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Sample spanner contains a sample application that
+// uses Google Spanner Go client, and reports metrics
+// and traces for the outgoing requests.
+package main
+
+import (
+	"context"
```
broady: We're still using x/net/context across this repo. Can you change it?

broady:
This looks like it's in the quickstart format. Can you change the binary name to be consistent w/ the other quickstarts?

opencensus/spanner/main.go
```diff
@@ -0,0 +1,60 @@
...
+	// This database must exist.
+	databaseName := "projects/your-project-id/instances/your-instance-id/databases/your-database-id"
+
+	client, err := spanner.NewClient(ctx, databaseName)
+	if err != nil {
+		log.Fatalf("Failed to create client %v", err)
+	}
+	defer client.Close()
+
+	_, err = client.Apply(ctx, []*spanner.Mutation{
```
broady: InsertOrUpdateMap for simplicity?

(Fine if you don't want to)

opencensus/spanner/main.go
```diff
@@ -0,0 +1,60 @@
...
+// and traces for the outgoing requests.
+package main
+
+import (
+	"context"
+	"log"
+
+	"cloud.google.com/go/spanner"
+	ss "go.opencensus.io/exporter/stats/stackdriver"
+	ts "go.opencensus.io/exporter/trace/stackdriver"
```
broady: These package names aren't ideal. Is there some better one?

opencensus/spanner/main.go
rakyll: Done

opencensus/spanner/main.go
rakyll: This quickstart is not demonstrating how to use the Spanner client. It is demonstrating why it is useful to have out-of-the-box tracing for fat client libraries. Apply is a great method because it does a lot internally and the instrumentation of the client with custom spans allows users to see all those different steps without them having to learn about the internals of the client.

opencensus/spanner/main.go
rakyll: I wouldn't focus too much on this, we are merging these packages into one package soon. https://github.com/census-instrumentation/opencensus-go/issues/312.

rakyll:
Do we require 1.6 for the samples? I can publish this as a gist, it doesn't have be here.

broady:
opencensus_quickstart, maybe? or opencensus_spanner_quickstart?

It's OK to be here. Up to you. You get the CI for free (at least checks that it continues to compile).

rakyll:
Followed up with https://github.com/GoogleCloudPlatform/golang-samples/pull/371.

PR #376, opened 2018-01-24
Close the client.

panesargoog:


monitoring/monitoring_quickstart/main.go
```diff
@@ -72,6 +72,11 @@ func main() {
 		log.Fatalf("Failed to write time series data: %v", err)
 	}
 
+	// Closes the client and flushes the data to StackDriver.
```
broady: nit: Stackdriver, not StackDriver.

But I think you can just leave this comment off entirely.

broady:
Thank you!

PR #383, opened 2018-01-29
Update blobstore.go

angelafunk:
https://cloud.google.com/appengine/docs/go/blobstore/

err = rootTemplate.Execute(w, uploadURL)

is called before the rootTemplate is introduced in the next snippet


broady:
I think the implication in the docs is that you add the template code at the package level.

Deleting it doesn't seem right.

PR #384, opened 2018-01-29
helloworld app.yaml outside of guestbook branches

angelafunk:
Create new folder so that go get can be used in app engine standard quickstart example


broady:
I think we should have the actual hello world code here, too?

Something simple?

```
package main

import (
    "fmt"
    "net/http"
)

func main() {
    http.HandleFunc("/", handle)
    appengine.Main()
}

func handle(w http.ResponseWriter, r *http.Request) {
    fmt.Fprintln(w, "Hello, world!")
}
```

hello.go
```diff
@@ -0,0 +1,15 @@
+package main
```
broady: This file's in the wrong directory. Should be in appengine/helloworld

angelafunk:
Can we get this in the main branch?

broady:
Done!

PR #385, opened 2018-01-29
helloworld hello.go file created

angelafunk:
Create new folder so that go get can be used in app engine standard quickstart example


broady:
Can you roll this into the other pull request?

Also, please let's encourage people to use package main/func main

PR #387, opened 2018-02-06
BigQuery JSON Snippets

shollyman:
Adds two BigQuery snippets for loading newline-delimited JSON from Cloud Storage.
One snippets defines the input schema explicitly, the other leverages
schema autodetection.

The samples leverage WriteEmpty as the write disposition to avoid appending
data to an existing table (the default is WriteAppend).


bigquery/snippets/snippet.go
```diff
@@ -305,3 +305,56 @@ func exportToGCS(client *bigquery.Client, datasetID, tableID, gcsURI string) err
...
+	// [START bigquery_load_table_gcs_json]
+	gcsRef := bigquery.NewGCSReference("gs://cloud-samples-data/bigquery/us-states/us-states.json")
+	gcsRef.SourceFormat = bigquery.JSON
+	gcsRef.Schema = bigquery.Schema{
+		{Name: "name", Type: bigquery.StringFieldType},
+		{Name: "post_abbr", Type: bigquery.StringFieldType},
+	}
+	loader := client.Dataset(datasetID).Table(tableID).LoaderFrom(gcsRef)
+
+	loader.WriteDisposition = bigquery.WriteEmpty
```
broady: nit: move to the previous line. the newline after is fine.

explanation: the WriteDisposition is part creating the loader, so it should belong with the code that created it. It doesn't seem to be part of the call to loader.Run, so shouldn't be next to that piece of code.

bigquery/snippets/snippet.go
```diff
@@ -305,3 +305,56 @@ func exportToGCS(client *bigquery.Client, datasetID, tableID, gcsURI string) err
...
+	if err != nil {
+		return err
+	}
+	status, err := job.Wait(ctx)
+	if err != nil {
+		return err
+	}
+
+	if status.Err() != nil {
+		fmt.Printf("Job completed with error: %v", status.Err())
```
broady: return an error, like the other error handling code.

You might want to just change this to `return fmt.Errorf(...)`

bigquery/snippets/snippet.go
```diff
@@ -305,3 +305,56 @@ func exportToGCS(client *bigquery.Client, datasetID, tableID, gcsURI string) err
...
```
broady: Same comments here.

broady:
LGTM

PR #388, opened 2018-02-07
New features from latest Vision API update

nkashy1:
Cloud Vision API has just received an update and some new features have become generally available. This PR shows the new functionality in the Vision API `detect` sample.

Features:

1. Best guess labels for web entity detection

2. Confidence scores in OCR annotations

3. `Racy` axis in safe search.

- - -

To test:

1. `cd vision/detect`

1. `go generate`

1. `go test`

You can also try running on an image that you have available locally by:

```go build && ./detect $PATH_TO_IMAGE```

### DIFF
--- a/vision/detect/detect.go
+++ b/vision/detect/detect.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	vision "cloud.google.com/go/vision/apiv1"
 	"golang.org/x/net/context"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: vision/detect/detect.go [hunk 1]
```diff
@@ -197,8 +198,30 @@ func detectDocumentText(w io.Writer, file string) error {
...
+				fmt.Fprintln(w, "\t\tParagraphs:")
+				for _, paragraph := range block.Paragraphs {
+					fmt.Fprintf(w, "\t\t\tConfidence: %f", paragraph.Confidence)
+					fmt.Fprintln(w, "\t\t\tWords:")
+					for _, word := range paragraph.Words {
+						symbols := make([]string, len(word.Symbols))
+						for i, s := range word.Symbols {
+							symbols[i] = s.Text
+						}
+						wordText := strings.Join(symbols, "")
```
odeke-em: How come we are mushing up the symbols without any spaces? Is this intentional?

nkashy1: Yep, those represent the individual letters in a word.

nkashy1: (For example, the `word.Symbols` would be `["l", "o", "l"]` to represent the "word" `"lol"`)

nkashy1:
@odeke-em , @broady : Looking for guidance on addition of region tags to the generated code.

I need to add region tags (`[START x]` and `[END x]` comments as in [this code](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/vision/cloud-client/detect/beta_snippets.py)) to the new samples, but it is not clear if I should be modifying the code generator or just doing it by hand in `detect.go`.

vision/detect/detect.go
```diff
@@ -170,6 +174,7 @@ func detectText(w io.Writer, file string) error {
 	return nil
 }
 
+// [START vision_detect_document]
```
broady: insert a newline after this region tag. otherwise it becomes part of the godoc comment.

vision/detect/detect.go
```diff
@@ -301,12 +331,16 @@ func detectSafeSearch(w io.Writer, file string) error {
...
 	fmt.Fprintln(w, "Spoofed:", props.Spoof)
 	fmt.Fprintln(w, "Violence:", props.Violence)
 
 	return nil
 }
 
+// [END vision_detect_safe_search]
+
+// [START vision_detect_web]
 // detectWeb gets image properties from the Vision API for an image at the given file path.
```
broady: Same for this one. Newline after the region tag.

broady:
LGTM. just a nit about spacing for the region tag.

vision/detect/detect.go
nkashy1: @broady : Fixed (and with subsequent region tags, as well)

nkashy1:
@broady : Didn't understand about `go_function`, could you link to a sample that uses it (or GitHub screenshot)?

broady:
Sent via IM. (internal to Google)

nkashy1:
@broady : Thanks. Can we merge this in for now? I will talk to the technical writer working on the documentation corresponding to these samples and see if we can switch over to `go_function` once I make the hand-off.

odeke-em:
>Edit 2: Never mind, just swallowed my trepidation and modified the code generator. It went better than expected. @odeke-em - could you please take another look?

Sorry for the late reply, I missed this notification as I had stepped away from work to go grab dinner :) Thanks for the updates, I'll def be using these sometime soon.

PR #395, opened 2018-02-19
Change getDeviceStates region tag to "iot_get_device_states".

noerog:


noerog:
I don't believe the build failures are due to this change?

broady:
Need to change the end tag, too.

noerog:
Done, thanks.

PR #405, opened 2018-03-01
bookshelf: fix format of appErrorf call

### LINKED ISSUE #404: bookshelf: template.go#Execute incorrectly calls appErrorf
```go
if err := tmpl.t.Execute(w, d); err != nil {
		return appErrorf(err, "could not write template: %v")
	}
```

PR coming up.

tbpg:
Fixes #404.

Without this change, the underlying template error is not printed when the template fails to Execute.


broady:
LGTM

PR #407, opened 2018-03-01
Add 1.10.x to Travis build versions

tbpg:
`1.10` would be interpreted as a float, but I think `1.10.x` is OK without quotes. Will confirm with the auto-Travis run on this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ branches:
 matrix:
   include:
     - go: 1.6.x
-    - go: 1.7.x
     - go: 1.8.x
     - go: 1.9.x
     - go: 1.10.x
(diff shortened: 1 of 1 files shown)


broady:
I guess we need to "go get" before "go vet" now.

broady:
1.10.x lgtm though

broady:
#412 addresses the vet warnings. PTAL.

broady:
actually, @tbpg, could you drop 1.7.x? we don't support it anymore, and it's not available on GAE.

tbpg:
There is 1.6, but not 1.7?

broady:
Yeah, 1.6 for Standard.

tbpg:
Closing this in favor of Kokoro (#440).

PR #411, opened 2018-03-05
Change location of time package in list of imports

angelafunk:


appengine/gophers/gophers-4/main.go
```diff
@@ -8,10 +8,11 @@ import (
 	"fmt"
 	"html/template"
 	"net/http"
-	"time"
```
tbpg: What if we move the START to above `"time"` here? That would mean the tutorial says to add `"google.golang.org/appengine"` when it's already been used in the earlier parts, but it's good to keep the group of standard library imports.

angelafunk: sgtm!

broady:
Did some rearrangement and ran gofmt. see https://github.com/GoogleCloudPlatform/golang-samples/pull/411/commits/838b11d9512e6c5ff45c0d394dfcf7b5693857a7

PR #412, opened 2018-03-05
datastore: fix vet errors for go1.10

broady:
datastore/snippets/snippet_test.go:30: ExampleNewIncompleteKey refers to unknown identifier: NewIncompleteKey
... (41 more)

Renamed functions in snippet_test to fix this class of vet errors:
sed -i -e 's/func Example/func Snippet/g' snippet_test.go

datastore/tasks/tasks.go:202: Println arg list ends with redundant newline
Changed to fmt.Print to silence this error.

### DIFF
--- a/datastore/snippets/snippet_test.go
+++ b/datastore/snippets/snippet_test.go
@@ -27,21 +27,21 @@ type Task struct {
 	Collaborators   []string
 }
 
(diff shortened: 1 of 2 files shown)


tbpg:
Thank you!

PR #424, opened 2018-03-21
trace: replace the obsolete client with OpenCensus

rakyll:


rakyll:
Do not merge yet.

broady:
Is this the right thing for the Trace docs?

Should we have a separate OC quickstart?

/cc @tbpg @SurferJeffAtGoogle

rakyll:
This is for the Trace docs. We replaced cloud.google.com/go with OpenCensus for Go but the documentation still pointing to the old library.

We are going to move to OpenCensus for all other languages as well. I am working on it with @mtwo.

broady:
merge whenever you're ready 👍

PR #425, opened 2018-03-23
appengine/gophers: alert if Firebase JS not setup

tbpg:
It's easy to forget to configure Firebase. So, this checks it's been configured and alerts the user if not.

`main.go` also needs to be configured. But, I'm not sure what the right way to alert for that is, if at all.

cc @angelafunk

### DIFF
--- a/appengine/gophers/gophers-5/index.html
+++ b/appengine/gophers/gophers-5/index.html
@@ -9,10 +9,16 @@
   <!-- [START init_firebase] -->
   <!-- *******************************************************************************************
        * TODO(DEVELOPER): Paste the initialization snippet from:
-       * Firebase Console > Overview > Add Firebase to your web app.
+       * http://console.firebase.google.com > Overview > Add Firebase to your web app.
        ***************************************************************************************** -->
   <!-- [END init_firebase] -->
-
+  <script>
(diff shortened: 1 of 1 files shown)


broady:
my first thought was putting this where the "firebase" variable is used would be better:
https://github.com/GoogleCloudPlatform/golang-samples/blob/4b715cefe70f2d2aa11eb4362969b84b8e7a54b5/appengine/gophers/gophers-5/static/index.js#L1-L21

but on more thought, having it right next to where the user should enter the code is also quite nice.

just presenting another option if you think index.js is a better place. if not, LGTM.

appengine/gophers/gophers-5/index.html
```diff
@@ -9,10 +9,16 @@
   <!-- [START init_firebase] -->
   <!-- *******************************************************************************************
        * TODO(DEVELOPER): Paste the initialization snippet from:
-       * Firebase Console > Overview > Add Firebase to your web app.
+       * http://console.firebase.google.com > Overview > Add Firebase to your web app.
```
broady: https

tbpg: Done.

PR #435, opened 2018-03-31
typo in app.yaml blocking deployment to app engine

dimitropoulos:
results in:

```
------------------------------------ STDERR ------------------------------------
2018/03/31 16:27:35 invalid api_version value 1
--------------------------------------------------------------------------------
```

otherwise


tbpg:
Thank you! I squashed the commits and changed the message a bit.

broady:
See also #447

PR #437, opened 2018-04-02
Added regional tag

angelafunk:


tbpg:
I added space above the `var` declaration to avoid confusing GoDoc, ran `gofmt`, and updated the commit message to start with `appengine/gophers/5`.

Issue #440, opened 2018-04-03
kokoro: move from Travis to Kokoro


tbpg:
Will need a .kokoro directory describing how to build/test the project then some Google-internal configuration to start running it.

tbpg:
See #451 for getting tests running.

PR #441, opened 2018-04-04
kokoro: add basic config and build script

tbpg:
I'll expand this to get secrets and run tests after builds are working.

Part of #440.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -e
+set -x
+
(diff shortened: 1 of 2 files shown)


broady:
Can this live under "testing"?

You may want to re-use this bash script:

https://github.com/GoogleCloudPlatform/golang-samples/blob/master/testing/cloudbuild/run.bash

tbpg:
Yes, I'll move it in a followup PR once secrets are ready. Thanks for the link - I'll reuse it.

Issue #200, opened 2017-02-17
Spanner Array Examples


leadinglocally:
Please make some example that includes use of arrays.

I tried 'unmarshalling' the row element type []string into *[]string and []string, but it fails with this:

type []string cannot be used for decoding ARRAY[STRING]

leadinglocally:
Also some examples for the ToStruct method for rows would be nice.

type Provider struct {
  Name string               `spanner:"name"`
  Address string            `spanner:"address"`
  Country string            `spanner:"country"`
  Lat float64               `spanner:"lat"`
  Lng float64               `spanner:"lng"`
  Tags []string             `spanner:"tags"`
}

something like that?

Ulexus:
For others who run into this problem, the issue appears to be that `ARRAY` types can be null, so you must use `spanner.NullString` as the receiving type for `ARRAY[STRING]` return values.

Ulexus:
Example:

```go
type Country struct {
  Name string
  Cities []spanner.NullString
}

func load(ctx context.Context, tx *spanner.ReadOnlyTransaction, name string) (*Country,error) {
  it := tx.Query(ctx, spanner.NewStatement(`
    SELECT a.Name, ARRAY(
        SELECT b.Name FROM cities b WHERE a.Country = b.Country
     ) FROM countries a 
     WHERE 
        a.Name = 'USA'
   `))
   defer it.Stop()

   row, err := it.Next()
   if err == iterator.Done {
      return nil, errors.New("not found")
   }
   if err != nil {
      return nil, err
   }

   var country Country
   err = row.ToStruct(&country)
   return &country, err
}
```

odeke-em:
@Ulexus thank you for the response/example. Would you like to make that into a PR that we can review and that can be your credited contribution?

jabubake:
@jsimonweb FYI

PR #442, opened 2018-04-05
Add example of using Spanner arrays in a query

Ulexus:
Implements #200


spanner/spanner_arrays/main.go
```diff
@@ -0,0 +1,180 @@
...
+		SELECT a.Name AS Name, ARRAY(
+			SELECT b.Name FROM Cities b WHERE a.CountryId = b.CountryId
+		) AS Cities FROM Countries a
+	`))
+	defer it.Stop()
+
+	for {
+		row, err := it.Next()
+		if err == iterator.Done {
+			return
```
broady: maybe `break` rather than `return`

spanner/spanner_arrays/main.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		if err != nil {
+			log.Fatalf("failed to read results: %v", err)
+		}
+
+		var country Country
+		err = row.ToStruct(&country)
```
broady: use compact error form:

`if err := row.ToStruct(&country); err != nil {`

spanner/spanner_arrays/main.go
```diff
@@ -0,0 +1,180 @@
...
+		if err != nil {
+			log.Fatalf("failed to read row into Country struct: %s", err)
```
broady: `%v` formatter directive for errors

spanner/spanner_arrays/main.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func createDatabase(ctx context.Context, adminClient *database.DatabaseAdminClient, db string) error {
+	matches := regexp.MustCompile("^(.*)/databases/(.*)$").FindStringSubmatch(db)
+	if matches == nil || len(matches) != 3 {
+		log.Fatalf("Invalid database id %s", db)
+	}
+
+	op, err := adminClient.CreateDatabase(ctx, &adminpb.CreateDatabaseRequest{
+		Parent:          matches[1],
```
broady: probably good to create variables for the matched elements.

i assume this is something like
```
var (
   projectID    = matches[1]
   databaseName = matches[2]
)

File: spanner/spanner_arrays/main.go [hunk 5]
```diff
@@ -0,0 +1,180 @@
...
+	if _, err := op.Wait(ctx); err == nil {
+		log.Printf("Created database [%s]", db)
+	}
+	return err
+}
+
+func removeDatabase(ctx context.Context, adminClient *database.DatabaseAdminClient, db string) error {
+	err := adminClient.DropDatabase(ctx, &adminpb.DropDatabaseRequest{Database: db})
+	if err != nil {
+		log.Fatalf("Failed to remove database [%s]: %v", db, err)
```
broady: log.Fatal is fairly odd here, I'd really expect this code to return the error and the caller to handle the error (by calling log.Fatal)

broady:
Oh! I just noticed you're not a Googler. Thanks for the submission! Overall, it looks very good.

Let me talk to our Spanner folks to see whether this kind of sample is useful for other languages, or whether we can incorporate it into our documentation.

spanner/spanner_arrays/main.go
Ulexus: Yeah; because this is just a named defer, returning an error is relatively useless; now handling it locally without returning error.  I'm still not totally sanguine about it philosophically; maybe it would be best to just handle it as an inline function in the defer.

spanner/spanner_arrays/main.go
Ulexus: Fine; throw my own pet peeve back at me.  Thanks!

spanner/spanner_arrays/main.go
Ulexus: Fixed; thanks

spanner/spanner_arrays/main.go
Ulexus: I rarely like to use the compact form, especially since it likes to cause vet-shadow errors, but I'm happy to accede to any norms in this case.

spanner/spanner_arrays/main.go
Ulexus: Right; thanks

Ulexus:
I don't think it warrants a full example, especially since there are many other, equally obtuse or at least unexpected gotchas with the Go <-> Spanner interactions, but I don't think I have any access to modify the documentation.  Someone on the old issue I originally commented on suggested a PR, so this is it.

Feel free to throw it away, if you'd like, but the documentation could definitely use clarification here.

Ulexus:
I signed it!

broady:
Looks like the signing didn't work properly. You might need to check that the e-mail addresses match.

spanner/spanner_arrays/main.go
```diff
@@ -0,0 +1,183 @@
...
+		log.Fatalf("Invalid database id %s", db)
+	}
+
+	var (
+		projectID    = matches[1]
+		databaseName = matches[2]
+	)
+
+	op, err := adminClient.CreateDatabase(ctx, &adminpb.CreateDatabaseRequest{
+		Parent:          databaseName,
```
jsimonweb: Update to be:

Parent:          projectID,

spanner/spanner_arrays/main.go
```diff
@@ -0,0 +1,183 @@
...
+		CreateStatement: fmt.Sprintf("CREATE DATABASE `%s`", projectID),
```
jsimonweb: Update to be:

CreateStatement: fmt.Sprintf("CREATE DATABASE `%s`", databaseName),

hostirosti:
Sean, can you add an example for an array column as well?
e.g.    cities ARRAY<STRING(1000)>

Ulexus:
@hostirosti I could do so, but does that substantially change the client API?   My purpose wasn't to demonstrate _Spanner_ so much as the curious and unexpected machinations necessary to interact with it using the golang client.  The crux of that, really, being the `[]spanner.NullString` type.  

I don't expect that would be any different with an in-table array.   I haven't tried, though.  If there is a significant difference, it should definitely be illustrated.

Ulexus:
On the other hand, I wouldn't want anyone to think that this contrived example was an indication that for something as simple as this you _should_ use a separate table rather than an in-table array.  I just don't want to use some terribly complicated example for illustration.

Ulexus:
Fair enough... just need to contrive something else for an in-table array... colors! :)

spanner/spanner_arrays/README.md
```diff
@@ -0,0 +1,6 @@
+This is a demonstration of the use of Cloud Spanner arrays using the golang
```
tbpg: Nit: "Go client library"

spanner/spanner_arrays/README.md
```diff
@@ -0,0 +1,6 @@
...
+client library.  The key takeaway here is that even if your fields are not
```
tbpg: "your string fields"?

spanner/spanner_arrays/README.md
```diff
@@ -0,0 +1,6 @@
...
+nullable, you still _must_ use the `spanner.NullString` type to receive them.
+Populating them can be done with either `spanner.NullString` _or_ plain
+`string`.
```
tbpg: Add a comment saying why? The string field may not be set in the row, so the value can be nil.

spanner/spanner_arrays/main.go
```diff
@@ -0,0 +1,199 @@
+// Copyright 2017 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Sample spanner_arrays is a basic program that queries Google's Cloud Spanner
```
tbpg: "Sample spanner_arrays queries Google's Cloud Spanner and returns an array within the results."

The "basic" part may be discouraging if the sample isn't immediately understood.

spanner/spanner_arrays/main.go
```diff
@@ -0,0 +1,199 @@
...
+	Cities  []spanner.NullString
+}
+
+func main() {
+	ctx := context.Background()
+
+	dsn := flag.String("database", "projects/your-project-id/instances/your-instance-id/databases/your-database-id", "Cloud Spanner database name")
+	flag.Parse()
+
+	// Connect to the Spanner Admin API
```
tbpg: Nit: period at the end.

spanner/spanner_arrays/main.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	defer admin.Close()
+
+	err = createDatabase(ctx, admin, *dsn)
+	if err != nil {
+		log.Fatalf("failed to create database: %v", err)
+	}
+	defer removeDatabase(ctx, admin, *dsn)
+
+	// Connect to database
```
tbpg: Nit: `// Connect to the database.`

spanner/spanner_arrays/README.md
Ulexus: Unfortunately, I don't _know_ why.  That's what makes this odd.  The field is set, for instance, with the `NOT NULL` restriction, but `spanner.NullString` is still required if that string field is part of an array.  Basically, if the _array_ is nullable, so must be the underlying scalar type (it would seem).

spanner/spanner_arrays/main.go
Ulexus: Good point; thanks

PR #451, opened 2018-04-11
kokoro: run tests with Trampoline

tbpg:
This also moves the Kokoro configuration into the testing directory.

system_tests.sh is based on testing/cloudbuild/run.bash with some
modifications to make it work with Kokoro. In particular, the service
account key is stored in KOKORO_KEYSTORE_DIR and the repo is checked out
to a `github` subdirectory.

Builds are not cached and every project uses the same source image (from
the golang-samples-tests project). This could be improved in the future,
but I'm not sure the best way to dynamically use different source images
in Trampoline (which starts the image).


testing/kokoro/system_tests.sh
```diff
@@ -0,0 +1,50 @@
...
+
+set -e
+
+export GOOGLE_APPLICATION_CREDENTIALS=$KOKORO_KEYSTORE_DIR/71386_golang-samples-kokoro-service-account
+export GOLANG_SAMPLES_KMS_KEYRING=ring1
+export GOLANG_SAMPLES_KMS_CRYPTOKEY=key1
+
+curl https://storage.googleapis.com/gimme-proj/linux_amd64/gimmeproj > /bin/gimmeproj && chmod +x /bin/gimmeproj;
+gimmeproj version;
+export GOLANG_SAMPLES_PROJECT_ID=$(gimmeproj -project golang-samples-tests lease 12m);
```
broady: nit: can you change this to 22m? (since the test timeout got bumped to 20m)

broady:
lgtm

PR #455, opened 2018-04-12
pubsub/subscriptions: Ack all 10 messages

### LINKED ISSUE #402: pubsub/subscriptions: final message Nack'd instead of Ack'd
Hi there, in the subscriptions example, when it receives the final (10th) message, it calls `cancel()` to cancel the context, as expected, but then it calls `Nack()` on the message, which seems strange.

It also doesn't print out the final message, so only 9/10 messages get displayed.

The subscription is deleted afterwards, so it probably doesn't matter much, but it seems a bit odd.

tbpg:
Fixes #402.

The mutex should prevent skipping over 10 and never calling `cancel()` and `return`.

### DIFF
--- a/pubsub/subscriptions/main.go
+++ b/pubsub/subscriptions/main.go
@@ -113,14 +113,13 @@ func pullMsgs(client *pubsub.Client, name string, topic *pubsub.Topic) error {
 	err := sub.Receive(cctx, func(ctx context.Context, msg *pubsub.Message) {
 		mu.Lock()
 		defer mu.Unlock()
+		msg.Ack()
+		fmt.Printf("Got message: %q\n", string(msg.Data))
 		received++
-		if received >= 10 {
+		if received == 10 {
 			cancel()
-			msg.Nack()
 			return
 		}
-		fmt.Printf("Got message: %q\n", string(msg.Data))
-		msg.Ack()
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


broady:
Not sure if this code is correct. I don't know if the mutex entirely protects against a race.

/cc @jadekler @jba

jba:
It looks fine, except I would put the lock just above the receive++. We don't want people to think that the Ack or printf need to be locked.

broady:
I'm pretty sure if there are more than 10 messages, the goroutine to process message 11, 12, etc will still get spawned while 10 is processing (and cancelling the context). That's why "received > 10" is important. Here's the patch I came up with (didn't send it for review, because it's so ugly):

```diff
diff --git a/pubsub/subscriptions/main.go b/pubsub/subscriptions/main.go
index d2c435a..826599c 100644
--- a/pubsub/subscriptions/main.go
+++ b/pubsub/subscriptions/main.go
@@ -113,12 +113,16 @@ func pullMsgs(client *pubsub.Client, name string, topic *pubsub.Topic) error {
        err := sub.Receive(cctx, func(ctx context.Context, msg *pubsub.Message) {
                mu.Lock()
                defer mu.Unlock()
+
                received++
                if received >= 10 {
-                       cancel()
-                       msg.Nack()
-                       return
+                       defer cancel()
+                       if received > 10 {
+                               msg.Nack()
+                               return
+                       }
                }
+
                fmt.Printf("Got message: %q\n", string(msg.Data))
                msg.Ack()
        })
```

broady:
Yeah, maybe we don't care.

These nuances really highlight how problematic this sample is. "Consume 10 messages" isn't a very common use case.

PR #457, opened 2018-04-12
storage/buckets: use ProjectID for test buckets

tbpg:
If two different projects happen to use the same bucket name, race
conditions may cause tests to fail.

This also deletes any buckets before using them for tests. If the
buckets already exist, they will cause TestCreate to fail and may have
the wrong permissions.

I don't particularly like using `TestMain`, but it seems like the cleanest way to delete the buckets before starting.

### DIFF
--- a/storage/buckets/main_test.go
+++ b/storage/buckets/main_test.go
@@ -41,7 +41,7 @@ func TestCreateWithAttrs(t *testing.T) {
 	tc := testutil.SystemTest(t)
 	name := bucketName + "-attrs"
 	// Clean up bucket before running test.
-	deleteBucket(storageClient, bucketName+"-attrs")
+	deleteBucket(storageClient, name)
 	if err := createWithAttrs(storageClient, tc.ProjectID, name); err != nil {
 		t.Fatalf("failed to create bucket (%q): %v", name, err)
 	}
@@ -73,6 +73,7 @@ outer:
 }
 
 func TestIAM(t *testing.T) {
+	testutil.SystemTest(t)
 	if _, err := getPolicy(storageClient, bucketName); err != nil {
 		t.Errorf("getPolicy: %#v", err)
 	}
@@ -85,6 +86,7 @@ func TestIAM(t *testing.T) {
 }
 
 func TestRequesterPays(t *testing.T) {
+	testutil.SystemTest(t)
 	if err := enableRequesterPays(storageClient, bucketName); err != nil {
 		t.Errorf("enableRequesterPay: %#v", err)
 	}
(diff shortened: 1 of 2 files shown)


tbpg:
Updated to use a new bucket name.

storage/buckets/main_test.go
```diff
@@ -15,41 +16,54 @@ import (
...
+	tc, ok := testutil.ContextMain(m)
+	if !ok {
+		log.Println("Skipping tests: GOLANG_SAMPLES_PROJECT_ID must be set")
+		return
+	}
+
+	bucketName = tc.ProjectID + "-storage-tests"
+	// Clean up buckets before running tests.
+	delete(c, bucketName)
+	delete(c, bucketName+"-attrs")
```
broady: delete is a built-in function, can we find a better name?

broady: i think this cleanup can be moved into TestCreate, removing the need for TestMain?

storage/buckets/main_test.go
```diff
@@ -15,41 +16,54 @@ import (
 	"golang.org/x/net/context"
 )
 
-var bucketName = fmt.Sprintf("golang-example-buckets-%d", time.Now().Unix())
+var (
+	c          *storage.Client
```
broady: s/c/storageClient/

Variables names should approximately scale with the scope they're used in

storage/buckets/main_test.go
```diff
@@ -15,41 +16,54 @@ import (
...
 	}
-	return client
+
+	tc, ok := testutil.ContextMain(m)
+	if !ok {
+		log.Println("Skipping tests: GOLANG_SAMPLES_PROJECT_ID must be set")
+		return
+	}
+
+	bucketName = tc.ProjectID + "-storage-tests"
```
broady: this might conflict with some other storage tests. "-storage-buckets-tests" maybe?

not sure if that ends up being too long (there's a limit on bucket name length)

broady:
LGTM overall

storage/buckets/main_test.go
tbpg: Changed to `deleteBucket`. Sample*? Yes.

tbpg: Done.

storage/buckets/main_test.go
tbpg: Done.

storage/buckets/main_test.go
tbpg: Done.

tbpg:
Updated to call SystemTest in all tests (otherwise they fail if the `GOLANG_SAMPLES_PROJECT_ID` environment variable is not set.

PR #460, opened 2018-04-16
testing/kokoro: add builds for Go 1.6, 1.8, & 1.9

tbpg:
1.10 will be added later since it fails the gofmt check right now. Part of issue #440. See #406 for 1.10.


broady:
nice! did you create images for those? it's probably time to add some better docs/build script for those images

tbpg:
Yes, I created images. But, I may have mis-configured them, though (caching the wrong packages, for example).

Is there a programmatic way to update the Go version of an image? Relatively straightforward once it's downloaded, but I don't know where to download the latest of each major version from.

PR #465, opened 2018-04-18
opencensus: add comment about default sampling

tbpg:


opencensus/opencensus_spanner_quickstart/main.go
```diff
@@ -36,6 +36,11 @@ func main() {
 	view.RegisterExporter(exporter)
 	trace.RegisterExporter(exporter)
 
+	// Use trace.AlwaysSample() to always record traces. The
+	// default sampler skips some traces to conserve resources,
+	// but can make it hard to debug test traffic.
+	// trace.SetDefaultSampler(trace.AlwaysSample())
```
rakyll: Maybe you can enable this line and explain the user that they should remove it instead.

tbpg: Done.

PR #475, opened 2018-05-01
Add missing region tag

davidcavazos:
Added `dialogflow_list_intents` region tag


dialogflow/intent_management/intent_management.go
```diff
@@ -109,6 +109,7 @@ func main() {
 	}
 }
 
+// [START dialogflow_list_intents]
```
tbpg: Would you mind adding a newline between this tag and the function declaration to avoid having the tag show up in GoDoc?

PR #479, opened 2018-05-07
testing/kokoro: upload XML with test results

tbpg:
This will allow fine-grained test results (rather than only the top
level pass/fail).


tbpg:
Go 1.10 failure caused by flake (but, was correctly parsed by Sponge).

PR #485, opened 2018-05-11
appengine/gophers/5 align w gophers-6

angelafunk:


appengine/gophers/gophers-5/static/index.js
```diff
@@ -27,7 +27,7 @@ window.addEventListener('load', function () {
       // User is signed in.
       document.getElementById('sign-out').hidden = false;
       document.getElementById('post-form').hidden = false;
-      document.getElementById('account-details').textContent =
+      var account = document.getElementById('account-details').textContent =
```
tbpg: We should delete `var account =` from gophers-6, rather than add it to gophers-5. It's not needed.

PR #486, opened 2018-05-12
Update main.go

plcstpierre:
As of Go 1.7 golang.org/x/net/context package is available in the standard library under the name context. https://golang.org/pkg/context.


plcstpierre:
I signed it!

tbpg:
Thank you for the PR! Unfortunately, we need to continue to support Go 1.6 for now, which only has the `golang.org/x/net/context` package.

PR #487, opened 2018-05-14
respond the latest satori/go.uuid

shinoshu:
Respond the latest satori/go.uuid.
Fix because function of uuid.Must changed args.


shinoshu:
@googlebot I signed it!

broady:
Sorry, this code doesn't compile.

NewV4() returns (UUID, error)

and Must() takes a (UUID, error)

That means that combining the functions works:

`Must(uuid.NewV4())`

PR #490, opened 2018-05-15
dlp: add field name to struct initializer

tbpg:
Fields were added and the sample no longer compiles.


tbpg:
Kokoro failure caused by logger test flake.

PR #492, opened 2018-05-16
translate: Add region tags for Translate samples

beccca:


translate/snippets/snippet.go
```diff
@@ -122,3 +136,5 @@ func translateTextWithModel(targetLanguage, text, model string) (string, error)
 	}
 	return resp[0].Text, nil
 }
+
+// [END translate_text_with_model]
```
tbpg: I think the space at the end of this line is causing `gofmt` to complain.

tbpg:
Test failure is unrelated (DLP flake).

PR #493, opened 2018-05-16
[DO NOT MERGE] Updating all the region tags to have a prefix of datastore

billyjacobson:
[DO NOT MERGE] Updating all the region tags to have a prefix of datastore

### DIFF
--- a/datastore/tasks/tasks.go
+++ b/datastore/tasks/tasks.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by the Apache 2.0
 // license that can be found in the LICENSE file.
 
-// [START all]
+// [START datastore_all]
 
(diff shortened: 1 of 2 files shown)


datastore/tasks/tasks.go
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by the Apache 2.0
 // license that can be found in the LICENSE file.
 
-// [START all]
+// [START datastore_all]
```
tbpg: Is this the canonical name of this region?

billyjacobson: I'm not sure if this is used at all, but am fixing it now incase it is used in the future.

jmdobry:
I'll ping you when we're ready to merge.

PR #494, opened 2018-05-16
Update Video Detect region tags.

kurtisvg:


videointelligence/video_analyze/video_analyze.go
```diff
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/net/context"
 )
 
+// [START video_analyze_labels_local]
```
tbpg: Would you mind adding a newline between the region tag comments and the function declaration? As is, the region tag comments will end up in the function GoDoc.

kurtisvg: How does that look?

tbpg: LGTM. I ran `gofmt` since it was complaining.

tbpg:
Holding off on merging for now.

PR #497, opened 2018-05-18
oauth: add explicit snippets

tbpg:
The snippets correspond to this page: https://cloud.google.com/docs/authentication/production.

I moved the tags out of the implicit function to match most of the other snippets on that page.

### DIFF
--- a/auth/snippets.go
+++ b/auth/snippets.go
@@ -14,12 +14,14 @@ import (
 	"golang.org/x/oauth2/google"
 	cloudkms "google.golang.org/api/cloudkms/v1"
 	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
 )
 
-func adc() {
-	ctx := context.Background()
+// [START auth_cloud_implicit]
 
-	// [START auth_cloud_implicit]
+// implicit uses Application Default Credentials to authenticate.
+func implicit() {
+	ctx := context.Background()
 
 	// For API packages whose import path is starting with "cloud.google.com/go",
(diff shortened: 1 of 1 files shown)


auth/snippets.go
```diff
@@ -54,6 +56,64 @@ func adc() {
...
+// [END auth_cloud_explicit]
+
+// [START auth_cloud_explicit_compute_engine]
+// [START auth_cloud_explicit_app_engine]
+
+// explicitDefault finds the default credentials.
+func explicitDefault(projectID string) {
+	ctx := context.Background()
+
+	creds, err := google.FindDefaultCredentials(ctx, storage.ScopeReadOnly)
```
jba: Why would you need to do this explicitly? `storage.NewClient` with no options will do exactly this, I believe. (I think you answered this in your IM to me, but I didn't understand it.)

tbpg: I'm not aware of why you would need `creds` directly. This happens to use the same storage API to demonstrate, but this isn't the recommended way to do it. Maybe we should add a comment like:

> It is very uncommon to need to explicitly get the default credentials. Most of the time, client libraries can use Application Default Credentials without having to pass the credentials in directly. See implicit above.

What do you think? I think it's better to have a Go entry in the doc, even if it says this is not recommended.

jba: Oh, I get it, this is part of a multi-language piece of advice. Yes, please add that comment.

tbpg: Done.

Issue #501, opened 2018-05-21
bigquery: avoid relying on time.Now().Unix() in tests


tbpg:
Sometimes, relying on `time.Now().Unix()` leads to collisions and test flakiness.

CC, FYI @shollyman 

PR coming up.

shollyman:
Datasets are already created within a project-specific namespace, so if you're getting collisions on dataset names it means they already use the same project or its a more exotic test that uses bigquery DatasetInProject() rather than Dataset().

Have a failure I should look at for collision issues?

PR #503, opened 2018-05-21
Cleanup

shollyman:
* reduces some duplicate code by adding another runAndRead use
* removes the unused listRows implementation
* reduces whitespace in a multiline sql string
* adds a uniqueName func to the BQ snippets test rather than having string formatters littered throughout the tests.

Semi-Related: https://github.com/GoogleCloudPlatform/golang-samples/issues/501 is derived from a backend issue rather than a real collision in naming resources, but centralizing the logic should make it easier to deal with this should client issues arise in the future.

### DIFF
--- a/bigquery/snippets/snippet_test.go
+++ b/bigquery/snippets/snippet_test.go
@@ -33,6 +33,14 @@ func init() {
 	}
 }
 
+// uniqueName returns a more unique name for a resource id given a prefix string.
+// Does not leverage dashes or underscores, because datasets/tables can't have dashes
+// and storage buckets can't have underscores in the identifier.
+func uniqueName(prefix string) string {
+	t := time.Now()
+	return fmt.Sprintf("%s%d%d", prefix, t.Unix(), t.Nanosecond())
+}
+
 func TestAll(t *testing.T) {
 	tc := testutil.SystemTest(t)
 	ctx := context.Background()
@@ -42,7 +50,7 @@ func TestAll(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	datasetID := fmt.Sprintf("golang_example_dataset_%d", time.Now().Unix())
+	datasetID := uniqueName("golang_example_dataset")
 	if err := createDataset(client, datasetID); err != nil {
 		t.Errorf("createDataset(%q): %v", datasetID, err)
 	}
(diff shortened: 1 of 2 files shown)


bigquery/snippets/snippet_test.go
```diff
@@ -217,21 +222,21 @@ func TestImportExport(t *testing.T) {
...
 		t.Fatalf("importJSONExplicitSchema(dataset:%q table:%q): %v", datasetID, explicitJSON, err)
 	}
 
-	autodetectJSON := fmt.Sprintf("golang_example_dataset_importjson_autodetect_%d", time.Now().Unix())
+	autodetectJSON := uniqueName("golang_example_dataset_importjson_autodetect")
 	if err := importJSONAutodetectSchema(client, datasetID, autodetectJSON); err != nil {
 		t.Fatalf("importJSONAutodetectSchema(dataset:%q table:%q): %v", datasetID, autodetectJSON, err)
 	}
-	bucket := fmt.Sprintf("golang-example-bigquery-importexport-bucket-%d", time.Now().Unix())
+	bucket := uniqueName("golang-example-bigquery-importexport-bucket")
```
tbpg: Bucket names need to be globally unique. Maybe this should use `tc.ProjectID`?

bigquery/snippets/snippet_test.go
```diff
@@ -33,6 +34,26 @@ func init() {
 	}
 }
 
+// uniqueBQName returns a more unique name for a BigQuery resource.
+func uniqueBQName(prefix string) string {
+	t := time.Now()
+	return fmt.Sprintf("%s_%d_%d", sanitize(prefix, '_'), t.Unix(), t.Nanosecond())
```
tbpg: I think we can drop the nanosecond. I know bucket name lengths can be limited. Not sure about BQ.

bigquery/snippets/snippet_test.go
```diff
@@ -33,6 +34,26 @@ func init() {
...
+}
+
+// uniqueBucketName returns a more unique name cloud storage bucket.
+func uniqueBucketName(prefix, projectID string) string {
+	t := time.Now()
+	return fmt.Sprintf("%s-%s-%d%d", sanitize(prefix, '-'), sanitize(projectID, '-'), t.Unix(), t.Nanosecond())
+}
+func sanitize(s string, allowedSeparator rune) string {
```
tbpg: Nit: newline before this.

bigquery/snippets/snippet_test.go
shollyman: bucket names are bounded at 63 chars except for the dns-style name cases, so I added a check that truncates.

shollyman: BQ references can be up to 1k chars, so not worrying about that.

PR #511, opened 2018-06-14
Add DLP code samples for custom info types

mwdaub:
The DLP team is in the process of expanding code samples to include features that we have launched since the initial samples were written.

In this pull request, I updated inspect.go to include dictionary and regex based custom info types, and added some tests in inspect_test.go. Also updated deid.go and deid_test.go to include info type parameter required in API V2.


dlp/dlp_snippets/inspect.go
```diff
@@ -417,3 +433,37 @@ func inspectBigquery(w io.Writer, client *dlp.Client, project string, minLikelih
 }
 
 // [END dlp_inspect_bigquery]
+
+// Convert the custom dictionary word lists and custom regexes to a list of CustomInfoTypes.
+func buildCustomInfoTypes(customDictionaries []string, customRegexes []string) []*dlppb.CustomInfoType {
```
tbpg: How is this sample going to be displayed? I'm afraid it makes the samples above a bit harder to grok, but could be OK as a single function/sample demonstrating how to do this.

Also, the samples don't include `buildCustomInfoTypes`, so it would be difficult to understand when looking at the docs.

dlp/dlp_snippets/inspect_test.go
```diff
@@ -44,6 +44,33 @@ func TestInspectString(t *testing.T) {
 	}
 }
 
+func TestInspectStringWithCustomInfoTypes(t *testing.T) {
```
tbpg: If this stays as a single function, I'd prefer if there were a single test (can add fields to the tests struct).

dlp/dlp_snippets/inspect_test.go
```diff
@@ -33,7 +33,7 @@ func TestInspectString(t *testing.T) {
 	}
 	for _, test := range tests {
 		buf := new(bytes.Buffer)
-		inspectString(buf, client, projectID, dlppb.Likelihood_POSSIBLE, 0, true, []string{"US_SOCIAL_SECURITY_NUMBER"}, test.s)
+		inspectString(buf, client, projectID, dlppb.Likelihood_POSSIBLE, 0, true, []string{"US_SOCIAL_SECURITY_NUMBER"}, []string{}, []string{}, test.s)
```
tbpg: Can do nil, which acts like an empty slice most of the time (for things like `range` and `len`).

dlp/dlp_snippets/inspect.go
```diff
@@ -315,12 +328,14 @@ func inspectDatastore(w io.Writer, client *dlp.Client, project string, minLikeli
 // [START dlp_inspect_bigquery]
 
 // inspectBigquery searches for the given info types in the given Bigquery dataset table.
-func inspectBigquery(w io.Writer, client *dlp.Client, project string, minLikelihood dlppb.Likelihood, maxFindings int32, includeQuote bool, infoTypes []string, pubSubTopic, pubSubSub, dataProject, datasetID, tableID string) {
+func inspectBigquery(w io.Writer, client *dlp.Client, project string, minLikelihood dlppb.Likelihood, maxFindings int32, includeQuote bool, infoTypes []string, customDictionaries []string, customRegexes []string, pubSubTopic, pubSubSub, dataProject, datasetID, tableID string) {
```
tbpg: Do all of these need to use custom? There are already so parameters here, I'm hesitant to add more.

dlp/dlp_snippets/inspect.go
mwdaub: Ah, I wasn't thinking about how these would be displayed. I deleted the function and just put the logic inline.

dlp/dlp_snippets/inspect_test.go
mwdaub: Done.

dlp/dlp_snippets/inspect_test.go
mwdaub: Done.

dlp/dlp_snippets/inspect.go
mwdaub: Custom info types can be used in all types of inspection, so I wanted to include them in all the samples. You could certainly argue that just putting them in inspectString and inspectFile is sufficient, and if you feel strongly I can remove them from these methods.

tbpg: Whichever is consistent with the other languages. Seems most natural to me to have a single `inspectCustom` sample with corresponding docs which explains how to do custom types.

dlp/dlp_snippets/main.go
```diff
@@ -57,6 +59,14 @@ func main() {
 	flag.Parse()
 
 	infoTypesList := strings.Split(*infoTypesString, ",")
+	var customDictionariesList []string
+	if *customDictionaryString != "" {
+		customDictionariesList = []string{*customDictionaryString}
+	}
+	var customRegexesList []string
+	if *customRegexesString != "" {
+		customRegexesList = strings.Split(*customRegexesString, ",")
```
tbpg: This might cause problems if someone has a comma in their regex. So, maybe only allow a single regex?

tbpg:
Hey @mwdaub, any updates on consistency with other languages? Also, looks like there is a merge conflict that needs to be resolved.

dlp/dlp_snippets/main.go
mwdaub: Done.

dlp/dlp_snippets/inspect.go
mwdaub: The other code samples use custom info types in all versions of inspect*, so if you want it to be consistent then the current structure is right.

tbpg:
I rebased and fixed the conflict. LGTM, waiting for tests.

PR #518, opened 2018-06-25
DLP: Add autoPopulateTimespan option for creating job triggers.

micheldavid:


dlp/dlp_snippets/main.go
```diff
@@ -3,6 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // dlp is an example of using the DLP API.
+
```
tbpg: Please leave this newline in -- it serves as the package documentation. See https://github.com/golang/go/wiki/CodeReviewComments#package-comments.

dlp/dlp_snippets/triggers.go
```diff
@@ -69,6 +69,9 @@ func createTrigger(w io.Writer, client *dlp.Client, project string, minLikelihoo
 								},
 							},
 						},
+						TimespanConfig: &dlppb.StorageConfig_TimespanConfig{
```
tbpg: Would you mind adding a comment above this line with a short explanation of `TimespanConfig`? It's not immediately clear from the name.

micheldavid: I do not think the idea here is to provide documentation on each field, but on how to perform an operation.
How about a link to the proto doc?
https://cloud.google.com/dlp/docs/reference/rest/v2/InspectJobConfig#TimespanConfig

dlp/dlp_snippets/main.go
micheldavid: Oops, this was unintended. I will revert it.

dlp/dlp_snippets/triggers.go
tbpg: The goal is to provide a sample of the right way to call the API. Since it's not clear from the name/context what this field does, I think it would be helpful to have a short comment.

This will end up on https://cloud.google.com/dlp/docs/creating-job-triggers, which explains the field at the bottom of the page, but it would be confusing if you're reading from top to bottom.

micheldavid: I am adding code samples for the enable_auto_population_of_timespan_config field.
The code snippet at the bottom will be replaced with pointers to this code once all languages are covered.
I added a comment to the field.

tbpg: Unfortunately, this also shows up in the "Create a job trigger" section. So, we probably need a separate sample function or a comment (something like `Optionally limit scans to only new content. See [link].`)

PR #527, opened 2018-07-12
profiler: Add flags to configure version and function skew

aalexand:


profiler/hotapp/main.go
```diff
@@ -17,11 +18,16 @@ import (
...
-// There are several goroutines continuously fighting for this mutex.
-var mu sync.Mutex
-
-// Some allocated memory. Held in a global variable to protect it from GC.
-var mem [][]byte
+var (
+	// Service version to configure.
+	version = flag.String("version", "1.0.0", "service version")
+	// Skew of foo1 function over foo2, in the CPU busyloop, to simulate diff.
+	skew = flag.Int("skew", 100, "skew of foo2 over foo1 (100=none)")
```
tbpg: I'm not exactly sure what skew means here. Would you mind expanding/updating the flag message?

aalexand: Done.

PR #532, opened 2018-07-16
testing: only get deps for modified directories

tbpg:
Downloading dependencies is often the slowest part of a testing a PR. This PR changes the test script to only get dependencies needed to run tests in modified directories. `go vet` is also only run on modified directories since it requires the dependencies of what's being vetted.

This PR also simplifies the `git diff` command. From `man git-diff`:
```
"git diff A...B" is equivalent to "git diff $(git-merge-base A B) B".
```

### DIFF
--- a/testing/kokoro/system_tests.sh
+++ b/testing/kokoro/system_tests.sh
@@ -47,8 +47,29 @@ if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"system-tests"* && -n $GOLANG_SAMPLES_GO
   ./testing/kokoro/configure_gcloud.bash;
 fi
 
+RUN_ALL_TESTS="0"
+# If this is a nightly test (not a PR), run all tests.
+if [ -z ${KOKORO_GITHUB_PULL_REQUEST_NUMBER:-} ]; then
+  RUN_ALL_TESTS="1"
+fi
+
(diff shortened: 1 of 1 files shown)


testing/kokoro/system_tests.sh
```diff
@@ -47,8 +47,29 @@ if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"system-tests"* && -n $GOLANG_SAMPLES_GO
...
+RUN_ALL_TESTS="0"
+# If this is a nightly test (not a PR), run all tests.
+if [ -z ${KOKORO_GITHUB_PULL_REQUEST_NUMBER:-} ]; then
+  RUN_ALL_TESTS="1"
+fi
+
+# CHANGED_DIRS is the list of top-level directories that changed. CHANGED_DIRS will be empty when run on master.
+CHANGED_DIRS=$(git --no-pager diff --name-only HEAD..master | grep "/" | cut -d/ -f1 | sort | uniq || true)
+# If test configuration is changed, run all tests.
+if [[ $CHANGED_DIRS =~ "testing" ]]; then
```
broady: `internal`, too, I think.

tbpg: Done. This is a bit simpler since we don't have complex dependencies. But, good to know!

PR #536, opened 2018-07-20
speech: fix calls to Reader.Read

jba:
The livecaption and livecaption_from_file samples were calling Read
incorrectly: it is possible to get n > 0 bytes along with an error, so
it is important to process the bytes before checking the error.

### DIFF
--- a/speech/livecaption/livecaption.go
+++ b/speech/livecaption/livecaption.go
@@ -53,6 +53,15 @@ func main() {
 		buf := make([]byte, 1024)
 		for {
 			n, err := os.Stdin.Read(buf)
+			if n > 0 {
+				if err := stream.Send(&speechpb.StreamingRecognizeRequest{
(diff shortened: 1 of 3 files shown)


jba:
Tyler, feel free to re-route to whoever is appropriate.

jba:
Thanks, Tyler. Who maintains the kokoro CI? I can't figure out how to dig down to the actual error.

tbpg:
I do. I just made some updates in #528 to test with Go 1.11. I updated this branch and will wait for tests again.

jba:
Still broken. Also note that go 1.6 is failing too. 

Are there logs of the child processes somewhere?

tbpg:
I added another commit to always `go get ./internal/...`. Hopefully that fixes the problem. For some reason, we're missing a dependency (`preprocess`) on Go 1.6 and Go 1.11.

PR #543, opened 2018-07-28
testing: add test for spurious region tags in go docs

### LINKED ISSUE #438: testing: add check for spurious region tags in godoc
Some check like this. First step is to clean up the existing offenders.
```
$ for f in $(go list ./...); do go doc $f | grep START; done
[START bigquery_quickstart] Sample bigquery-quickstart creates a Google
[START datastore_quickstart] Sample datastore-quickstart fetches an entity
doc: multiple packages in directory /usr/local/google/home/cbro/src/github.com/GoogleCloudPlatform/golang-samples/datastore/snippets
exit status 1
[START complete_sample_application]
[START using_transactions]
[START sample_code] This sample gets the app displays 5 log Records at a
[START requests_and_HTTP]
[START logging]
[START intro]
[START package_example_1]
[START example_sending]
[START example_handler]
doc: multiple packages in directory /usr/local/google/home/cbro/src/github.com/GoogleCloudPlatform/golang-samples/kms/snippets
exit status 1
[START language_quickstart] Sample language-quickstart uses the Google Cloud
[START logging_quickstart] Sample logging-quickstart writes a log entry to
[START monitoring_quickstart] Sample monitoring-quickstart writes a data
[START pubsub_quickstart] Sample pubsub-quickstart creates a Google Cloud
[START speech_quickstart] Sample speech-quickstart uses the Google Cloud
[START storage_quickstart] Sample storage-quickstart creates a Google Cloud
[START translate_quickstart] Sample translate-quickstart translates "Hello,
[START vision_quickstart] Sample vision-quickstart uses the Google Cloud
```

danicat:
Fix #438.

### DIFF
--- a/regiontag_test.go
+++ b/regiontag_test.go
@@ -1,0 +1,58 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package samples
+
+import (
+	"bufio"
+	"bytes"
+	"log"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+func TestRegionTags(t *testing.T) {
+	type RegionTag struct {
+		tag  string
+		path string
+	}
+	errs := []RegionTag{}
+
+	cmd := exec.Command("go", "list", "./...")
+	out, err := cmd.Output()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	scanner := bufio.NewScanner(strings.NewReader(string(out)))
+	for scanner.Scan() {
+		path := scanner.Text()
+
+		goDoc := exec.Command("go", "doc", path)
+		grep := exec.Command("grep", "START")
+		grep.Stdin, _ = goDoc.StdoutPipe()
+
+		var buf bytes.Buffer
+		writer := bufio.NewWriter(&buf)
+		grep.Stdout = writer
+
+		grep.Start()
+		goDoc.Run()
+		grep.Wait()
+
+		writer.Flush()
+		s := buf.String()
+		if len(s) > 0 {
+			errs = append(errs, RegionTag{s, path})
+		}
+	}
+
+	if len(errs) > 0 {
+		for _, e := range errs {
+			t.Errorf("%#v", e.path)
+		}
+	}
+
+}
--- a/bigquery/bigquery_quickstart/main.go
+++ b/bigquery/bigquery_quickstart/main.go
@@ -3,6 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // [START bigquery_quickstart]
+
 // Sample bigquery-quickstart creates a Google BigQuery dataset.
 package main
 
--- a/datastore/datastore_quickstart/main.go
+++ b/datastore/datastore_quickstart/main.go
@@ -3,6 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // [START datastore_quickstart]
+
 // Sample datastore-quickstart fetches an entity from Google Cloud Datastore.
 package main
 
--- a/docs/appengine/blobstore/complete.go
+++ b/docs/appengine/blobstore/complete.go
@@ -3,6 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // [START complete_sample_application]
+
 package blobstore_example
 
 import (
--- a/docs/appengine/datastore/transactions/transactions.go
+++ b/docs/appengine/datastore/transactions/transactions.go
@@ -3,6 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // [START using_transactions]
+
 package counter
 
 import (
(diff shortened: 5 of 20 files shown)


regiontag_test.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	errs := []RegionTag{}
+
+	cmd := exec.Command("go", "list", "./...")
+	out, err := cmd.Output()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	scanner := bufio.NewScanner(strings.NewReader(string(out)))
```
tbpg: Use `bytes.NewReader` or hook directly into `cmd.StdoutPipe()`?

Something like this (untested):
```go
goDoc := exec.Command("go", "list", "./...")
out, err := goDoc.StdoutPipe()
if err != nil {
  t.Fatalf(err)
}
scanner := bufio.NewScanner(out)
```

regiontag_test.go
```diff
@@ -0,0 +1,58 @@
...
+		var buf bytes.Buffer
+		writer := bufio.NewWriter(&buf)
+		grep.Stdout = writer
+
+		grep.Start()
+		goDoc.Run()
+		grep.Wait()
+
+		writer.Flush()
+		s := buf.String()
```
tbpg: In general, we use the short form:

```go
if s := buf.String(); len(s) > 0 { ... }
```

regiontag_test.go
```diff
@@ -0,0 +1,58 @@
...
+		if len(s) > 0 {
+			errs = append(errs, RegionTag{s, path})
+		}
+	}
+
+	if len(errs) > 0 {
```
tbpg: I wonder if the errors could be printed as they're found, rather than accumulating in the `errs` slice? Might be able to get rid of the `RegionTag` type, too.


```go
t.Errorf("Region tag %q found in GoDoc of %q", s, path)
```

regiontag_test.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	scanner := bufio.NewScanner(strings.NewReader(string(out)))
+	for scanner.Scan() {
+		path := scanner.Text()
+
+		goDoc := exec.Command("go", "doc", path)
+		grep := exec.Command("grep", "START")
```
tbpg: I'm concerned about Windows users (or anyone without `grep` installed). Maybe this could "grep" in pure Go? Could maybe do something similar to above (untested):
```go
goDoc := exec.Command("go", "doc", path)
docOut, err := goDoc.StdoutPipe()
if err != nil {
  t.Fatalf(err)
}
docScanner := bufio.NewScanner(docOut)
for docScanner.Scan() {
  text := scanner.Text()
  if !strings.Contains(text, "[START") {
    continue
  }
  ...
}
```

broady: I'd lean towards using `go/doc`. @danicat, did you know about this package?

https://golang.org/pkg/go/doc/

broady:
This is really great.

@danicat, would you like to finish up this change? Otherwise, I'll address Tyler's comments and get this merged in.

regiontag_test.go
danicat: I keep forgetting that people still use windows... lol... will fix that :)

danicat: I was not familiar with that package @broady. I'll have a look at it.

danicat:
@broady I was actually doing some changes when you posted. If you can wait a few hours I'll push my updates. I'm going to make the test process a few files at once to speed it up a little bit.

regiontag_test.go
danicat: Done!

regiontag_test.go
danicat: Yes, I should have done that. Thanks!

regiontag_test.go
danicat: That was a leftover of a previous async implementation that failed. I've made a few changes and got the async version working so I think now it justify doing the things the way it's done... but if you think we should remove the type and handle things differently please let me know and I'll fix it.

regiontag_test.go
danicat: In my newest commit I have removed the grep dependency as suggested, but I haven't used go/doc yet. `doc.New` requires a package in ast format and I'm not sure of the right steps to load it in that form.

danicat:
@tbpg PTAL

tbpg:
Hmm. I liked the relative simplicity of the previous version. Was it too slow?

regiontag_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"os/exec"
+	"strings"
+	"sync"
+	"testing"
+)
+
+func listPackages() <-chan string {
+	c := make(chan string)
+	go func() {
+		cmd := exec.Command("go", "list", "./...")
```
broady: No need for this, I think.

Just find all of the Go files:

```
filepath.Glob("**/*.go")
```

tbpg:
LGTM. Definitely makes sense to have it be in parallel. Thank you for the background info!

tbpg:
Test failures caused by an unrelated network error.

PR #545, opened 2018-08-01
monitoring/uptime: initial samples commit

tbpg:


monitoring/uptime/uptime.go
```diff
@@ -0,0 +1,157 @@
...
+		log.Fatal(err)
+	}
+	fmt.Fprintf(w, "Successfully created uptime check %q\n", uc.GetDisplayName())
+}
+
+// [END monitoring_uptime_check_create]
+
+// [START monitoring_uptime_check_list_configs]
+
+// list is an example of listing the uptime checks in projectID.
```
grayside: This comment doesn't add anything to the documentation line "To list your current uptime checks, do the following:"

monitoring/uptime/uptime.go
```diff
@@ -0,0 +1,157 @@
...
+	defer client.Close()
+	req := &monitoringpb.CreateUptimeCheckConfigRequest{
+		Parent: "projects/" + projectID,
+		UptimeCheckConfig: &monitoringpb.UptimeCheckConfig{
+			DisplayName: "new uptime check",
+			Resource: &monitoringpb.UptimeCheckConfig_MonitoredResource{
+				MonitoredResource: &monitoredres.MonitoredResource{
+					Type: "uptime_url",
+					Labels: map[string]string{
+						"host": "example.com",
```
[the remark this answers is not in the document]
tbpg: I think this is most clear - example.com is OK for examples and it's very clear what it would be replaced with.

monitoring/uptime/uptime.go
tbpg: Agreed. I like to have valid GoDoc, though. So, I think it's OK.

tbpg:
Thanks!

PR #553, opened 2018-08-08
BigQuery: snippets for partitioning and clustering.

shollyman:
Partitioning samples deviate somewhat from the canonical python sample, as they use param binding for DATE type, which doesn't have a native type representation.  Addressed by inlining the param values as literals within the predicate.

Clustering examples are based around a sample transaction table, which has been placed in cloud-samples-data already.  We can expand by adding more transaction rows if desired.

Will add go lang reviewer once sample content is LGTM.


bigquery/snippets/snippet.go
```diff
@@ -16,6 +16,113 @@ import (
 	"google.golang.org/api/iterator"
 )
 
+// Use a common block to inline comments related to importing the library
```
tswast: Wheee!

tswast: I can see this getting pretty annoying to match up start and end tags, but I get why we might want to put this in a central place.

shollyman: I use the same pattern for the shared runAndRead function that inserts a query job and reads result rows.  If it grows too hairy we can consider doing something clever with templates in the docsite markup.

tbpg:
1. The rubric says it's optional for non-dynamic languages. Since the `go get` is at the top of the page, `goimports` should automatically find the right import and add it. This is standard for Go. Feel free to disagree.
2. It's idiomatic to have `file.go` and `file_test.go` in Go.

bigquery/snippets/snippet.go
```diff
@@ -297,6 +404,56 @@ func createTableExplicitSchema(client *bigquery.Client, datasetID, tableID strin
...
 
+func createTablePartitioned(client *bigquery.Client, datasetID, tableID string) error {
+	ctx := context.Background()
+	// [START bigquery_create_table_partitioned]
+	sampleSchema := bigquery.Schema{
+		{Name: "name", Type: bigquery.StringFieldType},
+		{Name: "post_abbr", Type: bigquery.IntegerFieldType},
+		{Name: "date", Type: bigquery.DateFieldType},
+	}
+	metaData := &bigquery.TableMetadata{
```
tbpg: s/metaData/metadata/g

I think it's one word?

bigquery/snippets/snippet.go
```diff
@@ -297,6 +404,56 @@ func createTableExplicitSchema(client *bigquery.Client, datasetID, tableID strin
 	return nil
 }
 
+func createTablePartitioned(client *bigquery.Client, datasetID, tableID string) error {
+	ctx := context.Background()
+	// [START bigquery_create_table_partitioned]
```
tbpg: This is the norm for the other snippets, but any particular reason to leave the `ctx` line out of the snippet?

shollyman: The decision predates me, but context construction feels very similar to the client creation. I'll add a comment to the shared include block to demonstrate background context creation as well as client creation.

PR #555, opened 2018-08-13
Use "http.MethodGet" instead of "GET".

yekurtal:
I think it would be better if the default Go package const is used instead of a string.

Cleaner code and line 57 does this too.

### DIFF
--- a/appengine/gophers/gophers-3/main.go
+++ b/appengine/gophers/gophers-3/main.go
@@ -40,7 +40,7 @@ func indexHandler(w http.ResponseWriter, r *http.Request) {
 	// [START handling]
 	params := templateParams{}
 
-	if r.Method == "GET" {
+	if r.Method == http.MethodGet {
 		indexTemplate.Execute(w, params)
 		return
 	}


yekurtal:
I signed it!

tbpg:
Thank you!

tbpg:
Would you mind updating `gophers-{4,5,6}`?

broady:
> I think it would be better if the default Go package const is used instead of a string.
> 
> Cleaner code and line 57 does this too.

Eh, this constant isn't used in the Go standard library at all:

```
~/go master $ git grep '\.Method.*"GET"' | wc -l
27
~/go master $ git grep '\.Method.*MethodGet' | wc -l
0
```

Comparing it to `http.StatusBadRequest` is not the same. `MethodGet` is not more descriptive than `"GET"`.

I vote to leave as-is, but will leave it for @tbpg to decide.

yekurtal:
@broady For beginners words are easier than numbers IMO, just because it is not used till now should not mean it should not be used onwards.

But the decision is yours and @tbpg.

broady:
Totally agree. "GET" is a word already.

yekurtal:
Should I close the PR?

tbpg:
The stdlib example is compelling. So, I think we'll stick with how it is now. Thank you for the PR! :heart:  Let us know if there is anything else that can be improved!

(By `gophers-{4,5,6}`, I meant the other directories in `appengine/gophers`. Due to the nature of the tutorial, code from one step has to be copied to the later steps.)

PR #556, opened 2018-08-13
[KMS] Remove global location from kms example to generalize for GCS

frankyn:
Hi @broady,

I'm removing global location explicitly from sample to generalize for GCS CMEK documentation. The sample is used to show developers how to set an IAM policy on a KMS key, but global location aren't supported and I want to generalize this part of the sample.

Also, after running `go gen` I found a few additional changes to `snippets_test.go`, from #294.

PTAL, thank you!

### DIFF
--- a/kms/snippets/generated/sample-template.go
+++ b/kms/snippets/generated/sample-template.go
@@ -147,9 +147,9 @@ func getRingPolicy(project, keyRing string) error {
 	return nil
 }
 
-func addMemberRingPolicy(project, keyRing, role, member string) error {
+func addMemberRingPolicy(project, location, keyRing, role, member string) error {
 	var client *cloudkms.Service // Boilerplate is inserted by gen.go
-	location := "global"
+
 	parent := fmt.Sprintf("projects/%s/locations/%s/keyRings/%s",
 		project, location, keyRing)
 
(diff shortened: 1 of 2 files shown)


kms/snippets/generated/sample-template.go
```diff
@@ -147,9 +147,9 @@ func getRingPolicy(project, keyRing string) error {
 	return nil
 }
 
-func addMemberRingPolicy(project, keyRing, role, member string) error {
+func addMemberRingPolicy(project, location, keyRing, role, member string) error {
```
broady: Just gotta run "go generate" in this directory.

kms/snippets/snippet_test.go
```diff
@@ -25,7 +25,7 @@ func init() {
 	_ = google.DefaultClient
 }
 
-func createKeyring(project, name string) error {
+func createKeyring(project, keyRing string) error {
```
broady: Gotta make this change in `sample-template.go` too, then run `go generate` from the `snippets directory.

frankyn: Thanks, I tried running `go generate` from the `snippets/` and nothing change.

PR #562, opened 2018-08-14
BigQuery: many new samples

shollyman:
Introduces a variety of sample snippets across the product.  ORC format, Parquet format, CMEK, schema manipulations.

Also, wraps common include block used by all samples in a no-op function, to ensure that samples have a consistent indent depth.

### DIFF
--- a/bigquery/snippets/snippet.go
+++ b/bigquery/snippets/snippet.go
@@ -23,6 +23,7 @@ func noOpCommentFunc() {
 	// [START bigquery_add_empty_column]
 	// [START bigquery_add_column_query_append]
 	// [START bigquery_browse_table]
+	// [START bigquery_cancel_job]
 	// [START bigquery_copy_table]
 	// [START bigquery_copy_table_cmek]
 	// [START bigquery_copy_table_multiple_source]
@@ -40,6 +41,7 @@ func noOpCommentFunc() {
 	// [START bigquery_extract_table_json]
 	// [START bigquery_get_dataset]
 	// [START bigquery_get_dataset_labels]
+	// [START bigquery_get_job]
 	// [START bigquery_get_table]
 	// [START bigquery_get_table_labels]
 	// [START bigquery_label_dataset]
@@ -93,6 +95,7 @@ func noOpCommentFunc() {
 	// [END bigquery_add_empty_column]
 	// [END bigquery_add_column_query_append]
 	// [END bigquery_browse_table]
+	// [END bigquery_cancel_job]
 	// [END bigquery_copy_table]
 	// [END bigquery_copy_table_cmek]
 	// [END bigquery_copy_table_multiple_source]
@@ -110,6 +113,7 @@ func noOpCommentFunc() {
 	// [END bigquery_extract_table_json]
 	// [END bigquery_get_dataset]
 	// [END bigquery_get_dataset_labels]
+	// [END bigquery_get_job]
 	// [END bigquery_get_table]
 	// [END bigquery_get_table_labels]
 	// [END bigquery_label_dataset]
@@ -155,6 +159,17 @@ func noOpCommentFunc() {
 	// [END bigquery_update_dataset_expiration]
 	// [END bigquery_update_table_description]
 	// [END bigquery_update_table_expiration]
+}
+
+func cancelJob(client *bigquery.Client, jobID string) error {
+	ctx := context.Background()
+	// [START bigquery_cancel_job]
+	job, err := client.JobFromID(ctx, jobID)
+	if err != nil {
+		return nil
+	}
+	return job.Cancel(ctx)
+	// [END bigquery_cancel_job]
 }
 
 func createDataset(client *bigquery.Client, datasetID string) error {
@@ -1324,6 +1339,30 @@ func deleteAndUndeleteTable(client *bigquery.Client, datasetID, tableID string) 
 	ds.Table(recoverTableID).Delete(ctx)
 	return nil
 
+}
+
+func getJobInfo(client *bigquery.Client, jobID string) error {
+	ctx := context.Background()
+	// [START bigquery_get_job]
+	job, err := client.JobFromID(ctx, jobID)
+	if err != nil {
+		return err
+	}
+
+	status := job.LastStatus()
+	state := "Unknown"
+	switch status.State {
+	case bigquery.Pending:
+		state = "Pending"
+	case bigquery.Running:
+		state = "Running"
+	case bigquery.Done:
+		state = "Done"
+	}
+	fmt.Printf("Job %s was created %v and is in state %s\n",
+		jobID, status.Statistics.CreationTime, state)
+	// [END bigquery_get_job]
+	return nil
 }
 
 func importCSVFromFile(client *bigquery.Client, datasetID, tableID, filename string) error {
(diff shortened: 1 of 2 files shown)


shollyman:
I forgot this was dependent on some changes to the client library that I haven't submitted yet.  Will ping this once those changes are in.

shollyman:
Dependent changes went in with https://code-review.googlesource.com/c/gocloud/+/31670

bigquery/snippets/snippet.go
```diff
@@ -16,115 +16,161 @@ import (
...
-// [END bigquery_update_dataset_access]
-// [END bigquery_update_dataset_description]
-// [END bigquery_update_dataset_expiration]
-// [END bigquery_update_table_description]
-// [END bigquery_update_table_expiration]
+func noOpCommentFunc() {
+	// Use a common block to inline comments related to importing the library
+	// and constructing a client.  Inside a func to ensure the indentation is
+	// consistent between multiple includes.
+	// [START bigquery_add_empty_column]
```
tswast: 👍 on alphabetizing this.

bigquery/snippets/snippet.go
```diff
@@ -407,6 +453,194 @@ func createTableExplicitSchema(client *bigquery.Client, datasetID, tableID strin
...
+				{Name: "city", Type: bigquery.StringFieldType},
+				{Name: "state", Type: bigquery.StringFieldType},
+				{Name: "zip", Type: bigquery.StringFieldType},
+				{Name: "numberOfYears", Type: bigquery.StringFieldType},
+			}},
+	}
+
+	metaData := &bigquery.TableMetadata{
+		Schema:         sampleSchema,
+		ExpirationTime: time.Now().AddDate(1, 0, 0), // Table will be automatically deleted in 1 year.
```
tswast: Is the expiration necessary for this sample? It seems unrelated to the actual schema demonstration.

bigquery/snippets/snippet.go
```diff
@@ -407,6 +453,194 @@ func createTableExplicitSchema(client *bigquery.Client, datasetID, tableID strin
...
+	// Iterate through the schema to set all Required fields to false (nullable).
+	var relaxed bigquery.Schema
+	for _, v := range original.Schema {
+		v.Required = false
+		relaxed = append(relaxed, v)
+	}
+	newMeta := bigquery.TableMetadataToUpdate{
+		Schema: relaxed,
+	}
+	if _, err := tableRef.Update(ctx, newMeta, ""); err != nil {
```
tswast: This seems like a prime example were we would want to use the etag, right? We're doing a get-modify-update.

bigquery/snippets/snippet.go
```diff
@@ -407,6 +453,194 @@ func createTableExplicitSchema(client *bigquery.Client, datasetID, tableID strin
...
+	// [START bigquery_relax_column]
+	sampleSchema := bigquery.Schema{
+		{Name: "full_name", Type: bigquery.StringFieldType, Required: true},
+		{Name: "age", Type: bigquery.IntegerFieldType, Required: true},
+	}
+	original := &bigquery.TableMetadata{
+		Schema: sampleSchema,
+	}
+	tableRef := client.Dataset(datasetID).Table(tableID)
+	if err := tableRef.Create(ctx, original); err != nil {
```
tswast: A table Get might be more natural for this sample (with the create outside the sample).

bigquery/snippets/snippet.go
```diff
@@ -1136,6 +1474,134 @@ func importJSONAutodetectSchema(client *bigquery.Client, datasetID, tableID stri
...
+	if status.Err() != nil {
+		return fmt.Errorf("Job completed with error: %v", status.Err())
+	}
+	// [END bigquery_load_table_gcs_orc]
+	return nil
+}
+
+func importORCTruncate(client *bigquery.Client, datasetID, tableID string) error {
+	ctx := context.Background()
+	// [START bigquery_load_table_gcs_orc_truncate]
```
tswast: Do we want to combo all the `_append` and `_truncate` samples like we did in Python to save on system test time?

https://github.com/GoogleCloudPlatform/google-cloud-python/blob/b176b888d56651689bea8768d39e57e518660ce7/docs/bigquery/snippets.py#L1543-L1559

bigquery/snippets/snippet.go
shollyman: Potentially, but getting an Etag would require another roundtrip of metadata fetch, since Go doesn't return the metadata on create (only any potential errors).

Schema evolution is more restrictive in that it only allows "compatible" evolutions of the schema, so the modify would fail regardless if the new schema is incompatible (e.g. an intermediate modify added a column).  Perhaps a comment rather than another fetch?

bigquery/snippets/snippet.go
shollyman: Will yank it.

bigquery/snippets/snippet.go
shollyman: It would simplify the other point about schema evolution, will do that instead.

shollyman: s/that/this/

bigquery/snippets/snippet.go
shollyman: My primary question is how are people consuming the samples?  If they're consuming from the docs, then the sample is straightforward.  If they're using the "view on github", then tag interleaving within a method like that makes for somewhat confusing samples.

bigquery/snippets/snippet.go
tswast: > Perhaps a comment rather than another fetch?

SGTM

bigquery/snippets/snippet.go
tswast: Our hypothesis is copy-paste from cloud.google.com, but it's a good question.

bigquery/snippets/snippet.go
```diff
@@ -457,6 +694,47 @@ func createTableClustered(client *bigquery.Client, datasetID, tableID string) er
...
+	ctx := context.Background()
+	// [START bigquery_update_table_cmek]
+	tableRef := client.Dataset(datasetID).Table(tableID)
+	meta, err := tableRef.Metadata(ctx)
+	if err != nil {
+		return err
+	}
+	update := bigquery.TableMetadataToUpdate{
+		EncryptionConfig: &bigquery.EncryptionConfig{
+			KMSKeyName: "projects/cloud-samples-tests/locations/us-central1/keyRings/test/cryptoKeys/otherkey",
```
tbpg: Add TODO line like the other samples?

bigquery/snippets/snippet.go
```diff
@@ -1136,6 +1477,134 @@ func importJSONAutodetectSchema(client *bigquery.Client, datasetID, tableID stri
...
+	return nil
+}
+
+func importORCTruncate(client *bigquery.Client, datasetID, tableID string) error {
+	ctx := context.Background()
+	// [START bigquery_load_table_gcs_orc_truncate]
+	gcsRef := bigquery.NewGCSReference("gs://cloud-samples-data/bigquery/us-states/us-states.orc")
+	gcsRef.SourceFormat = bigquery.ORC
+	loader := client.Dataset(datasetID).Table(tableID).LoaderFrom(gcsRef)
+	// default for import jobs is to append data to a table.  WriteTruncate
```
tbpg: Nit: Capitalize.

PR #563, opened 2018-08-14
kms: add Asymmetric Keys samples

DanSanche:
Added samples for the KMS Asymmetric Keys launch

### DIFF
--- a/kms/asymmetric/samples.go
+++ b/kms/asymmetric/samples.go
@@ -1,0 +1,189 @@
+package main
+
+import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/asn1"
+	"encoding/base64"
+	"encoding/pem"
+	"fmt"
+	"golang.org/x/net/context"
+	"google.golang.org/api/cloudkms/v1"
+	"log"
+	"math/big"
+
+	"golang.org/x/oauth2/google"
+)
+
+// [START kms_get_asymmetric_public]
+// Retrieve a public key from a saved asymmetric key pair on KMS
+func getAsymmetricPublicKey(client *cloudkms.Service, ctx context.Context, keyPath string) (interface{}, error) {
+	response, err := client.Projects.Locations.KeyRings.CryptoKeys.CryptoKeyVersions.
+		GetPublicKey(keyPath).Context(ctx).Do()
+	if err != nil {
+		return nil, err
+	}
+	if response == nil {
+		return nil, fmt.Errorf("no response from GetPublicKey")
+	}
+	keyBytes := []byte(response.Pem)
+	block, _ := pem.Decode(keyBytes)
+	publicKey, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return nil, err
+	}
+	return publicKey, nil
+}
+
+// [END kms_get_asymmetric_public]
+
+// [START kms_decrypt_rsa]
+// Attempt to decrypt ciphertext with saved RSA key
+func decryptRSA(client *cloudkms.Service, ctx context.Context, ciphertext, keyPath string) (string, error) {
+	decryptRequest := &cloudkms.AsymmetricDecryptRequest{
+		Ciphertext: ciphertext,
+	}
+	response, err := client.Projects.Locations.KeyRings.CryptoKeys.CryptoKeyVersions.
+		AsymmetricDecrypt(keyPath, decryptRequest).Context(ctx).Do()
+	if err != nil {
+		return "", err
+	}
+	messageArr, err := base64.StdEncoding.DecodeString(response.Plaintext)
+	if err != nil {
+		return "", err
+	}
+	message := fmt.Sprintf("%s", messageArr)
+	return message, nil
+}
+
+// [END kms_decrypt_rsa]
+
+// [START kms_encrypt_rsa]
+// Encrypt plaintext message using saved RSA public key
+func encryptRSA(client *cloudkms.Service, ctx context.Context, message, keyPath string) (string, error) {
+	abstractKey, err := getAsymmetricPublicKey(client, ctx, keyPath)
+	if err != nil {
+		return "", err
+	}
+
+	//perform type assertion to get rsa key
+	rsaKey := abstractKey.(*rsa.PublicKey)
+
+	ciphertextBytes, err := rsa.EncryptOAEP(sha256.New(), rand.Reader, rsaKey, []byte(message), nil)
+	if err != nil {
+		return "", err
+	}
+	ciphertextStr := base64.StdEncoding.EncodeToString(ciphertextBytes)
+	return ciphertextStr, nil
+}
+
+// [END kms_encrypt_rsa]
+
+// [START kms_sign_asymmetric]
+// Sign a message using saved asymmetric private key
+func signAsymmetric(client *cloudkms.Service, ctx context.Context, message, keyPath string) (string, error) {
+	//find hash of message
+	digest := sha256.New()
+	digest.Write([]byte(message))
+	digestStr := base64.StdEncoding.EncodeToString(digest.Sum(nil))
+
+	asymmetricSignRequest := &cloudkms.AsymmetricSignRequest{
+		Digest: &cloudkms.Digest{
+			Sha256: digestStr,
+		},
+	}
+
+	response, err := client.Projects.Locations.KeyRings.CryptoKeys.CryptoKeyVersions.
+		AsymmetricSign(keyPath, asymmetricSignRequest).Context(ctx).Do()
+	if err != nil {
+		return "", err
+	}
+
+	return response.Signature, nil
+}
+
+// [END kms_sign_asymmetric]
+
+// [START kms_verify_signature_rsa]
+// Verify the cryptographic signature for a message signed with an RSA private key
+func verifySignatureRSA(client *cloudkms.Service, ctx context.Context, signature, message, keyPath string) error {
+	abstractKey, err := getAsymmetricPublicKey(client, ctx, keyPath)
+	if err != nil {
+		return err
+	}
+	//perform type assertion to get rsa key
+	rsaKey := abstractKey.(*rsa.PublicKey)
+	decodedSignature, _ := base64.StdEncoding.DecodeString(signature)
+
+	digest := sha256.New()
+	digest.Write([]byte(message))
+	hash := digest.Sum(nil)
+
+	pssOptions := rsa.PSSOptions{SaltLength: len(hash), Hash: crypto.SHA256}
+	err = rsa.VerifyPSS(rsaKey, crypto.SHA256, hash, decodedSignature, &pssOptions)
+	if err != nil {
+		return fmt.Errorf("verification failed")
+	}
+	return nil
+}
+
+// [END kms_verify_signature_rsa]
+
+// [START kms_verify_signature_ec]
+// Verify the cryptographic signature for a message signed with an Elliptic Curve private key
+func verifySignatureEC(client *cloudkms.Service, ctx context.Context, signature, message, keyPath string) error {
+	abstractKey, err := getAsymmetricPublicKey(client, ctx, keyPath)
+	if err != nil {
+		return err
+	}
+	//perform type assertion to get elliptic curve key
+	ecKey := abstractKey.(*ecdsa.PublicKey)
+	decodedSignature, err := base64.StdEncoding.DecodeString(signature)
+
+	var parsedSig struct{ R, S *big.Int }
+	rest, err := asn1.Unmarshal(decodedSignature, &parsedSig)
+	if err != nil || len(rest) != 0 {
+		return err
+	}
+
+	digest := sha256.New()
+	digest.Write([]byte(message))
+	hash := digest.Sum(nil)
+
+	if !ecdsa.Verify(ecKey, hash, parsedSig.R, parsedSig.S) {
+		return fmt.Errorf("verification failed")
+	}
+	return nil
+}
+
+// [END kms_verify_signature_ec]
+
+func main() {
+	project_id := "sanche-testing-project"
+	location := "global"
+	keyName := "ec-sign"
+	keyring := "test-ring"
+	keyVersion := "1"
+
+	keyPath := fmt.Sprintf("projects/%s/locations/%s/keyRings/%s/cryptoKeys/%s/cryptoKeyVersions/%s", project_id, location, keyring, keyName, keyVersion)
+
+	ctx := context.Background()
+	client, err := google.DefaultClient(ctx, cloudkms.CloudPlatformScope)
+	if err != nil {
+		log.Fatal(err)
+	}
+	kmsClient, err := cloudkms.New(client)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	message := "test msg"
+	sig, _ := signAsymmetric(kmsClient, ctx, message, keyPath)
+	fmt.Println(sig)
+	success := verifySignatureEC(kmsClient, ctx, sig, message, keyPath)
+	fmt.Println(success)
+}
--- a/kms/asymmetric/samples.go
+++ b/kms/asymmetric/samples.go
@@ -1,3 +1,8 @@
+// Copyright 2018 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Samples for asymmetric keys feature of Cloud Key Management Service: https://cloud.google.com/kms/
 package main
 
 import (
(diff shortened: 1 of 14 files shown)


kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/asn1"
+	"encoding/base64"
+	"encoding/pem"
+	"fmt"
+	"golang.org/x/net/context"
+	"google.golang.org/api/cloudkms/v1"
```
ahmetb: separate vendored pkgs vs stdlib pkgs, see other samples for examples of this

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+	"log"
+	"math/big"
+
+	"golang.org/x/oauth2/google"
+)
+
+// [START kms_get_asymmetric_public]
+// Retrieve a public key from a saved asymmetric key pair on KMS
```
ahmetb: godoc format begins with the method name.

    // getAsymmetricPublicKey does x y z.

add a trailing comma.

Also see if other samples export these helper functions or not.

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+func getAsymmetricPublicKey(client *cloudkms.Service, ctx context.Context, keyPath string) (interface{}, error) {
```
ahmetb: I think often ctx would be the first param by convention. See other methods in this repo if there's precedence.

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+	response, err := client.Projects.Locations.KeyRings.CryptoKeys.CryptoKeyVersions.
+		GetPublicKey(keyPath).Context(ctx).Do()
+	if err != nil {
+		return nil, err
+	}
+	if response == nil {
```
ahmetb: remove this since expected to return

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, fmt.Errorf("no response from GetPublicKey")
+	}
+	keyBytes := []byte(response.Pem)
+	block, _ := pem.Decode(keyBytes)
+	publicKey, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return nil, err
```
ahmetb: `return nil, fmt.Errorf("failed to parse public key: %+v", err)`

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+		GetPublicKey(keyPath).Context(ctx).Do()
+	if err != nil {
+		return nil, err
+	}
+	if response == nil {
+		return nil, fmt.Errorf("no response from GetPublicKey")
+	}
+	keyBytes := []byte(response.Pem)
+	block, _ := pem.Decode(keyBytes)
+	publicKey, err := x509.ParsePKIXPublicKey(block.Bytes)
```
ahmetb: instead of returning interface{} you can perhaps do

```
type pubKey struct {
   rsa rsa.PublicKey
   ecdsa ecdsa.PublicKey
}
```

```
switch v:= publicKey.(type) {
case *rsa.PublicKey: return &pubKey{rsa: v}, nil
case *ecdsa.PublicKey: return &pubKey{ecdsa: v}, nil
default: return nil, fmt.Errorf("unrecognized key format %T", v)
}
```

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+		AsymmetricDecrypt(keyPath, decryptRequest).Context(ctx).Do()
+	if err != nil {
+		return "", err
+	}
+	messageArr, err := base64.StdEncoding.DecodeString(response.Plaintext)
+	if err != nil {
+		return "", err
+	}
+	message := fmt.Sprintf("%s", messageArr)
+	return message, nil
```
ahmetb: just fmt here

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	//perform type assertion to get rsa key
+	rsaKey := abstractKey.(*rsa.PublicKey)
+
+	ciphertextBytes, err := rsa.EncryptOAEP(sha256.New(), rand.Reader, rsaKey, []byte(message), nil)
+	if err != nil {
+		return "", err
+	}
+	ciphertextStr := base64.StdEncoding.EncodeToString(ciphertextBytes)
```
ahmetb: use `ciphertext` 
just return it too

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+		return "", err
+	}
+	message := fmt.Sprintf("%s", messageArr)
+	return message, nil
+}
+
+// [END kms_decrypt_rsa]
+
+// [START kms_encrypt_rsa]
+// Encrypt plaintext message using saved RSA public key
```
ahmetb: `encryptRSA creates a ciphertext from a plain message using a RSA public key saved at the specified keyPath.`

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+	}
+
+	response, err := client.Projects.Locations.KeyRings.CryptoKeys.CryptoKeyVersions.
+		AsymmetricSign(keyPath, asymmetricSignRequest).Context(ctx).Do()
+	if err != nil {
+		return "", err
+	}
+
+	return response.Signature, nil
```
ahmetb: `return response.GetSignature(), err`

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+// [START kms_verify_signature_rsa]
+// Verify the cryptographic signature for a message signed with an RSA private key
+func verifySignatureRSA(client *cloudkms.Service, ctx context.Context, signature, message, keyPath string) error {
+	abstractKey, err := getAsymmetricPublicKey(client, ctx, keyPath)
+	if err != nil {
+		return err
+	}
+	//perform type assertion to get rsa key
+	rsaKey := abstractKey.(*rsa.PublicKey)
+	decodedSignature, _ := base64.StdEncoding.DecodeString(signature)
```
ahmetb: dont ignore this error

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+
+	digest := sha256.New()
+	digest.Write([]byte(message))
+	hash := digest.Sum(nil)
+
+	pssOptions := rsa.PSSOptions{SaltLength: len(hash), Hash: crypto.SHA256}
+	err = rsa.VerifyPSS(rsaKey, crypto.SHA256, hash, decodedSignature, &pssOptions)
+	if err != nil {
+		return fmt.Errorf("verification failed")
```
ahmetb: include the err in the msg

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return err
+	}
+	//perform type assertion to get elliptic curve key
+	ecKey := abstractKey.(*ecdsa.PublicKey)
+	decodedSignature, err := base64.StdEncoding.DecodeString(signature)
+
+	var parsedSig struct{ R, S *big.Int }
+	rest, err := asn1.Unmarshal(decodedSignature, &parsedSig)
+	if err != nil || len(rest) != 0 {
```
ahmetb: ignore rest OR return a different error when len(rest)!=0

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+// [START kms_verify_signature_ec]
+// Verify the cryptographic signature for a message signed with an Elliptic Curve private key
+func verifySignatureEC(client *cloudkms.Service, ctx context.Context, signature, message, keyPath string) error {
+	abstractKey, err := getAsymmetricPublicKey(client, ctx, keyPath)
+	if err != nil {
+		return err
+	}
+	//perform type assertion to get elliptic curve key
+	ecKey := abstractKey.(*ecdsa.PublicKey)
+	decodedSignature, err := base64.StdEncoding.DecodeString(signature)
```
ahmetb: use the err

kms/asymmetric/samples.go
broady: yes, generally, use [goimports](https://golang.org/x/tools/cmd/goimports) :)

kms/asymmetric/samples.go
broady: and you'll want to put a newline between the doc string and the region tag.

kms/asymmetric/samples.go
broady: also... why aren't we using the gRPC client?

https://godoc.org/cloud.google.com/go/kms/apiv1

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
+package main
```
broady: missing header and package-level comment

kms/asymmetric/samples_test.go
```diff
@@ -0,0 +1,160 @@
...
+	ctx            context.Context
+	message        string
+	rsaDecryptPath string
+	rsaSignPath    string
+	ecSignPath     string
+}
+
+func setup(t *testing.T) TestVariables {
+	tc := testutil.SystemTest(t)
+	projectId := tc.ProjectID
```
broady: s/projectId/projectID

broady: but really, just inline this. no need to create another variable.

kms/asymmetric/samples_test.go
```diff
@@ -0,0 +1,160 @@
...
+	ctx := context.Background()
+	client, err := google.DefaultClient(ctx, cloudkms.CloudPlatformScope)
+	if err != nil {
+		t.Fatal(err)
+	}
+	kmsClient, err := cloudkms.New(client)
+	if err != nil {
+		t.Fatal(err)
+	}
+	//create cryptokeys
```
broady: comments should be full sentences with punctuation and capitalization.

kms/asymmetric/samples_test.go
```diff
@@ -0,0 +1,160 @@
...
+	kmsClient.Projects.Locations.KeyRings.Create(parent, &cloudkms.KeyRing{}).KeyRingId(keyRing).Do()
+	kmsClient.Projects.Locations.KeyRings.CryptoKeys.Create(
```
broady: Why create the key for every test method? Shouldn't this happen just once?

kms/asymmetric/samples_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+func TestGetPublicKey(t *testing.T) {
+	v := setup(t)
+
+	rsaDecryptPub, err := getAsymmetricPublicKey(v.client, v.ctx, v.rsaDecryptPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, ok := rsaDecryptPub.(*rsa.PublicKey)
+	assertEqual(t, ok, true)
```
broady: please, no helper functions. Just write the code.

This `assertEqual` function loses all its context. I'd expect a test failure message like:

```
if _, ok := rsaDecryptPub.(*rsa.PublicKey); !ok {
   t.Errorf("type of rsaDecryptPub: got %T, want *rsa.PublicKey", rsaDecryptPub)
}
```

The current message would say: "false != true", which is not very helpful.

kms/asymmetric/samples_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+func TestECSignVerify(t *testing.T) {
+	v := setup(t)
+
+	sig, err := signAsymmetric(v.client, v.ctx, v.message, v.ecSignPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(sig) < 50 || len(sig) > 300 {
+		t.Errorf("signature length outside expected range. Length: %d", len(sig))
```
broady: Go style is "got; want"

Read more: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

kms/asymmetric/samples_test.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	err = verifySignatureEC(v.client, v.ctx, sig, v.message, v.ecSignPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = verifySignatureEC(v.client, v.ctx, sig, v.message+".", v.ecSignPath)
```
broady:
```
if err := ...; err != nil {
```

kms/asymmetric/samples_test.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(sig) < 50 || len(sig) > 300 {
+		t.Errorf("signature length outside expected range. Length: %d", len(sig))
+	}
+
+	err = verifySignatureEC(v.client, v.ctx, sig, v.message, v.ecSignPath)
+	if err != nil {
+		t.Fatal(err)
```
broady: Error message please.

kms/asymmetric/samples_test.go
```diff
@@ -0,0 +1,160 @@
...
+			},
+		}).CryptoKeyId("rsa-decrypt").Do()
+	kmsClient.Projects.Locations.KeyRings.CryptoKeys.Create(
+		parent+"/keyRings/"+keyRing, &cloudkms.CryptoKey{
+			Purpose: "ASYMMETRIC_SIGN",
+			VersionTemplate: &cloudkms.CryptoKeyVersionTemplate{
+				Algorithm: "RSA_SIGN_PSS_2048_SHA256",
+			},
+		}).CryptoKeyId("rsa-sign").Do()
+	kmsClient.Projects.Locations.KeyRings.CryptoKeys.Create(
```
broady: No need for the line wraps, Go has no line length limit.

ahmetb: IMO the wrapped version is way more readable.

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+// [START kms_get_asymmetric_public]
+// Retrieve a public key from a saved asymmetric key pair on KMS
+func getAsymmetricPublicKey(client *cloudkms.Service, ctx context.Context, keyPath string) (interface{}, error) {
+	response, err := client.Projects.Locations.KeyRings.CryptoKeys.CryptoKeyVersions.
+		GetPublicKey(keyPath).Context(ctx).Do()
+	if err != nil {
+		return nil, err
+	}
+	if response == nil {
+		return nil, fmt.Errorf("no response from GetPublicKey")
```
white0ut: nit: when there's no formatting required, use errors.New("...") instead of fmt.Errorf

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+	if !ecdsa.Verify(ecKey, hash, parsedSig.R, parsedSig.S) {
+		return fmt.Errorf("verification failed")
+	}
+	return nil
+}
+
+// [END kms_verify_signature_ec]
+
+func main() {
+	project_id := "sanche-testing-project"
```
white0ut: An idea to float around is to make this a command-line flag or env var so it can be more plug & playable

kms/asymmetric/samples.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	kmsClient, err := cloudkms.New(client)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	message := "test msg"
+	sig, _ := signAsymmetric(kmsClient, ctx, message, keyPath)
```
white0ut: don't ignore

white0ut:
Thanks for adding these, Dan - looks great!

kms/asymmetric/samples.go
DanSanche: I didn't realize "google.golang.org/api/cloudkms/v1" was deprecated, but "cloud.google.com/go/kms/apiv1" doesn't look like it has the new asymmetric keys functions. Do you know what's going on with this @white0ut ?

broady: Looks like the protobuf hasn't been published:

https://github.com/googleapis/googleapis/tree/master/google/cloud/kms/v1

kms/asymmetric/samples.go
DanSanche: good catch. The main function won't be part of the samples, but I'll make sure to take out my hardcoded projectID

kms/asymmetric/samples.go
```diff
@@ -10,83 +15,82 @@ import (
...
 
 // [START kms_sign_asymmetric]
-// Sign a message using saved asymmetric private key
-func signAsymmetric(client *cloudkms.Service, ctx context.Context, message, keyPath string) (string, error) {
-	//find hash of message
+
+// signAsymmetric will sign a plaintext message using a saved asymmetric private key.
+func signAsymmetric(ctx context.Context, client *cloudkms.Service, message, keyPath string) (string, error) {
+	// Find the hash of the plaintext message.
 	digest := sha256.New()
```
bdhess: sha384 for EC P-384

kms/asymmetric/samples_test.go
```diff
@@ -17,144 +25,180 @@ type TestVariables struct {
...
+
+func TestMain(m *testing.M) {
+	tc, ok := testutil.ContextMain(m)
+	v, err := getTestVariables(tc.ProjectID)
+	parent := "projects/" + tc.ProjectID + "/locations/global"
+	if ok && err == nil {
+		//Create cryptokeys in the test project if needed.
+		s1 := createKeyHelper(v, v.rsaDecryptId, v.rsaDecryptPath, "ASYMMETRIC_DECRYPT", "RSA_DECRYPT_OAEP_2048_SHA256", parent)
+		s2 := createKeyHelper(v, v.rsaSignId, v.rsaSignPath, "ASYMMETRIC_SIGN", "RSA_SIGN_PSS_2048_SHA256", parent)
+		s3 := createKeyHelper(v, v.ecSignId, v.ecSignPath, "ASYMMETRIC_SIGN", "EC_SIGN_P224_SHA256", parent)
```
bdhess: P224 was dropped from the API

PR #570, opened 2018-08-16
Video Intelligence region tag update

alixhami:
Updates region tags to be consistent with those in other languages. Region tag update on cloud.google.com is also in progress and will be submitted after all language PRs are merged.

Note that I was not able to run `go generate` successfully. Please let me know if this region tag change will work with the template.


tbpg:
Thanks! There is a test to confirm `go generate` is up to date. I also checked this out locally to be sure. I merged master and will merge after tests.

PR #571, opened 2018-08-16
Cloud Talent Solution v3 API Golang code samples

tanyanino:
Add Golang code samples for CTS v3 API, including

/XXX_sample.go   API samples with XXXEntry() as test entry
/quickstart             a quick start example
/sampe_test.go     function test for all sample entries


tanyanino:
Close this for fixing CLA check fails

PR #578, opened 2018-08-17
add accessTokenSample

salrashid123:
Add sample for `JWTAccessTokenSourceFromJSON`

- [https://godoc.org/golang.org/x/oauth2/google#JWTAccessTokenSourceFromJSON](https://godoc.org/golang.org/x/oauth2/google#JWTAccessTokenSourceFromJSON)

The token type can be used directly with GCP apis. 
If this repo isn't the correct place for this sample, LMK where.

### DIFF
--- a/auth/accesstoken.go
+++ b/auth/accesstoken.go
@@ -1,5 +1,3 @@
-package authsnippets
-
 // Copyright 2018 Google Inc. All rights reserved.
 // Use of this source code is governed by the Apache 2.0
 // license that can be found in the LICENSE file.
@@ -8,8 +6,9 @@ package authsnippets
 // This sample demonstrates AccessTokenCredentials:
 // https://godoc.org/golang.org/x/oauth2/google#JWTAccessTokenSourceFromJSON
 
-// To use, create a service accountJSON file and allow it PubSubAdmin IAM
+// To use, create a service accountJSON file and allow atleast Pub/Sub Viewer IAM
 // permissions on allow listing Topics on a project.
+package authsnippets
 
 import (
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


tbpg:
How will this example be used? Will it be referenced in any documentation? Seems related to https://cloud.google.com/docs/authentication/production - maybe one of those is enough?

(I like the sample, but want to make sure there isn't a better spot for it and it's not covered by something else.)

salrashid123:
@tbpg 

Its really a standalone sample demonstrating this:

- [https://developers.google.com/identity/protocols/OAuth2ServiceAccount#jwt-auth](https://developers.google.com/identity/protocols/OAuth2ServiceAccount#jwt-auth)

The docs above describes `FireStore` which threw me off a bit...a sample w/ `Pubsub` above and a link to find out the `aud:` field for an abitrary GCP service took me sometime to connect the dots.

Unlinke the samples in the link above, golang oauth provides a 'first class' support for this via the direct api call `JWTAccessTokenSourceFromJSON` is pretty useful.

So, i think the best place for this is another 'tab' in the docs cited above.  If i need to file an internal docbug too, LMK.  To make it consistent with the above examples, i can rework mine to use Firestore (or change the existing ones to `Pubsub` in the same docbug...but thats just personal preference.

i've just added some annotation to the code snippt (eg
`[START|END jwtaccesstoken_sample]`

Finally, i do think there should be a mention of this mechanism from the docs here
[https://cloud.google.com/docs/authentication/production](https://cloud.google.com/docs/authentication/production) to the one cited above

tbpg:
Closing this for now, but we can always reopen at a later date when there is a clear picture of where this should land in the docs.

For anyone in the future, you can see a walkthrough of this feature in a blog post by Salmaan here: https://medium.com/google-cloud/faster-serviceaccount-authentication-for-google-cloud-platform-apis-f1355abc14b2

Issue #580, opened 2018-08-20
kms: asymmetric keys feedback from bdhess


broady:
https://github.com/GoogleCloudPlatform/golang-samples/pull/563#discussion_r211395914

> sha384 for EC P-384

> P224 was dropped from the API

/cc @bdhess

DanSanche:
added a PR with the fixes (@bdhess)

tbpg:
Is this fixed?

bdhess:
Yes (#584)

PR #584, opened 2018-08-21
KMS Asymmetric Text Fixes

DanSanche:
Addresses the issues in https://github.com/GoogleCloudPlatform/golang-samples/issues/580:
- clarifies which RSA key algorithm is used for encrypt/decrypt
- makes a note that signAsymmetric will need changes for certain key algorithms
- changed EC_SIGN_P224_SHA256 to EC_SIGN_P256_SHA256 in test file

### DIFF
--- a/kms/asymmetric/samples.go
+++ b/kms/asymmetric/samples.go
@@ -45,7 +45,7 @@ func getAsymmetricPublicKey(ctx context.Context, client *cloudkms.Service, keyPa
 
 // [START kms_decrypt_rsa]
 
-// decryptRSA will attempt to decrypt a given ciphertext with saved a RSA key.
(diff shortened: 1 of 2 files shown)


kms/asymmetric/samples.go
```diff
@@ -67,7 +67,7 @@ func decryptRSA(ctx context.Context, client *cloudkms.Service, ciphertext, keyPa
 
 // [START kms_encrypt_rsa]
 
-// encryptRSA creates a ciphertext from a plain message using a RSA public key saved at the specified keyPath.
+// encryptRSA will encrypt a message locally using an 'RSA_DECRYPT_OAEP_2048_SHA256' public key retrieved from Cloud KMS
```
broady: no need for the single quotes here, above, and below.

PR #587, opened 2018-08-22
Update Region Tags - GAE

averikitsch:
Ongoing GAE region tag updates
* Docs are frozen at last commit *

### DIFF
--- a/appengine/gophers/gophers-6/index.html
+++ b/appengine/gophers/gophers-6/index.html
@@ -36,18 +36,21 @@
 
 {{with .Notice}}<div id="notice">{{.}}</div>{{end}}
 
+<!-- [START form] -->
 <form id="post-form" enctype="multipart/form-data" action="/" method="post" hidden=true>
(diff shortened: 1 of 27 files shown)


appengine/gophers/gophers-6/index.html
```diff
@@ -36,21 +36,18 @@ <h1>The Gopher Network</h1>
 
 {{with .Notice}}<div id="notice">{{.}}</div>{{end}}
 
-<!-- [START form] -->
```
tbpg: gophers-6 is about to be published.

averikitsch: Currently, region tags for gophers-6 are not used in any of the docs. @angelafunk please let me know if you are planning on using these region tags so we can get them on the samples tracker. Thanks!

angelafunk: Yes! will be using region tags for gophers-6

PR #588, opened 2018-08-23
Publish that scales with asynchronous error handling

anguillanneuf:
The original publish example shows publishing a single message. 

The added function shows publishing multiple messages with error handling.

### DIFF
--- a/pubsub/topics/main.go
+++ b/pubsub/topics/main.go
@@ -156,36 +156,39 @@ func publish(client *pubsub.Client, topic, msg string) error {
 
 func publishThatScales(client *pubsub.Client, topic string, n uint64) error {
 	ctx := context.Background()
-	var wg sync.WaitGroup
-	var n_err uint64
 	// [START pubsub_publish_with_error_handling_that_scales]
+	var wg sync.WaitGroup
+	var totalErrors uint64
 	t := client.Topic(topic)
 
 	for i := 0; i < int(n); i++ {
 		result := t.Publish(ctx, &pubsub.Message{
-			Data: []byte("Message No. " + strconv.Itoa(i)),
+			// data must be a ByteString
+			Data: []byte("Message " + strconv.Itoa(i)),
 		})
 
 		wg.Add(1)
-		go func(i int) {
+		go func(i int, res *pubsub.PublishResult) {
 			defer wg.Done()
-			// The Get method blocks until the result is returned and a
-			// server-generated ID is returned for the published message.
-			id, err := result.Get(ctx)
+			// The Get method blocks until a server-generated ID or
+			// an error is returned for the published message.
+			id, err := res.Get(ctx)
 			if err != nil {
-				// Error handling code can be here.
+				// Error handling code can be added here.
 				log.Output(1, fmt.Sprintf("Failed to publish: %v", err))
-				atomic.AddUint64(&n_err, 1)
+				atomic.AddUint64(&totalErrors, 1)
 			} else {
-				fmt.Printf("Published message No. %d; msg ID: %v\n", i, id)
+				fmt.Printf("Published message %d; msg ID: %v\n", i, id)
 			}
-		}(i)
-
-		wg.Wait()
+		}(i, result)
 	}
 
-	if n_err > 0 {
-		return errors.New("some messages did not publish successfully")
+	wg.Wait()
+
+	if totalErrors > 0 {
+		return errors.New(
+			fmt.Sprintf("%d of %d messages did not publish successfully",
+				totalErrors, n))
 	}
 	return nil
 	// [END pubsub_publish_with_error_handling_that_scales]
--- a/pubsub/topics/main.go
+++ b/pubsub/topics/main.go
@@ -157,7 +157,7 @@ func publish(client *pubsub.Client, topic, msg string) error {
 func publishThatScales(client *pubsub.Client, topic string, n uint64) error {
 	ctx := context.Background()
 	var wg sync.WaitGroup
-	var ops uint64
+	var n_err uint64
 	// [START pubsub_publish_with_error_handling_that_scales]
 	t := client.Topic(topic)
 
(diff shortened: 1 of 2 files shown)


pubsub/topics/main.go
```diff
@@ -145,6 +149,37 @@ func publish(client *pubsub.Client, topic, msg string) error {
...
+				log.Output(1, fmt.Sprintf("Failed to publish: %v", err))
+			} else {
+				fmt.Printf("Published message No. %d; msg ID: %v\n", i, id)
+			}
+		}(i)
+
+	}
+
+	fmt.Scanln()
+	fmt.Println("You pressed ENTER to stop publishing.\n")
```
broady: I think using a sync.WaitGroup is the right thing to do here.
`Scanln` makes testing very difficult.

pubsub/topics/main.go
```diff
@@ -145,6 +149,37 @@ func publish(client *pubsub.Client, topic, msg string) error {
 	return nil
 }
 
+func publishThatScales(client *pubsub.Client, topic string, n int) {
```
tbpg: Would you mind adding a test?

pubsub/topics/main.go
tbpg: You could also use a `context.Context` and `switch` on `ctx.Done()`.

pubsub/topics/main.go
anguillanneuf: That's a good idea. This is my first piece of Go code. Hungry for comments/feedback. :)

anguillanneuf: I have added a return value for PublishThatScales and wrote a test for it.

pubsub/topics/main.go
```diff
@@ -145,6 +154,43 @@ func publish(client *pubsub.Client, topic, msg string) error {
 	return nil
 }
 
+func publishThatScales(client *pubsub.Client, topic string, n uint64) error {
+	ctx := context.Background()
+	var wg sync.WaitGroup
+	var n_err uint64
+	// [START pubsub_publish_with_error_handling_that_scales]
+	t := client.Topic(topic)
```
davidcavazos: Nit: I would rename `t` to `topic` to help on readability, and the current `topic` (string) to `topicName`

pubsub/topics/main.go
```diff
@@ -145,6 +154,43 @@ func publish(client *pubsub.Client, topic, msg string) error {
 	return nil
 }
 
+func publishThatScales(client *pubsub.Client, topic string, n uint64) error {
+	ctx := context.Background()
+	var wg sync.WaitGroup
+	var n_err uint64
+	// [START pubsub_publish_with_error_handling_that_scales]
```
davidcavazos: Let's have this just below the `ctx` definition, the `wg` object might not be obvious and it plays an important role in the synchronization.

pubsub/topics/main.go
```diff
@@ -145,6 +154,43 @@ func publish(client *pubsub.Client, topic, msg string) error {
 	return nil
 }
 
+func publishThatScales(client *pubsub.Client, topic string, n uint64) error {
+	ctx := context.Background()
+	var wg sync.WaitGroup
+	var n_err uint64
```
davidcavazos: Nit: How about calling this `totalErrors` for readability?

pubsub/topics/main.go
```diff
@@ -145,6 +154,43 @@ func publish(client *pubsub.Client, topic, msg string) error {
...
+			if err != nil {
+				// Error handling code can be here.
+				log.Output(1, fmt.Sprintf("Failed to publish: %v", err))
+				atomic.AddUint64(&n_err, 1)
+			} else {
+				fmt.Printf("Published message No. %d; msg ID: %v\n", i, id)
+			}
+		}(i)
+
+		wg.Wait()
```
davidcavazos: Doesn't this have to be outside the loop?

pubsub/topics/main.go
```diff
@@ -145,6 +154,43 @@ func publish(client *pubsub.Client, topic, msg string) error {
...
+	}
+
+	if n_err > 0 {
+		return errors.New("some messages did not publish successfully")
```
davidcavazos: Since we're already counting the messages with errors, it might be nice to say something like `%d of %d messages did not publish successfully`

pubsub/topics/main.go
```diff
@@ -145,6 +154,43 @@ func publish(client *pubsub.Client, topic, msg string) error {
...
+		result := t.Publish(ctx, &pubsub.Message{
+			Data: []byte("Message No. " + strconv.Itoa(i)),
+		})
+
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			// The Get method blocks until the result is returned and a
+			// server-generated ID is returned for the published message.
+			id, err := result.Get(ctx)
```
davidcavazos: I think you also want to pass the `result` as a function argument to give it a copy of the result object, otherwise all the goroutines will be reading the latest value of `result` rather than their own copy.

anguillanneuf: Thanks David for pointing that out!

pubsub/topics/main.go
```diff
@@ -145,6 +154,46 @@ func publish(client *pubsub.Client, topic, msg string) error {
...
+		wg.Add(1)
+		go func(i int, res *pubsub.PublishResult) {
+			defer wg.Done()
+			// The Get method blocks until a server-generated ID or
+			// an error is returned for the published message.
+			id, err := res.Get(ctx)
+			if err != nil {
+				// Error handling code can be added here.
+				log.Output(1, fmt.Sprintf("Failed to publish: %v", err))
+				atomic.AddUint64(&totalErrors, 1)
```
tbpg: Add a return here then remove the else? That way, only the error flow is indented. See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow.

anguillanneuf: Booked marked the Code Review link. Thanks a lot!

tbpg:
cc @jadekler for Pub/Sub usage.

PR #594, opened 2018-08-31
Memorystore GCE : Replace GCS_APP_LOCATION with GCS_BUCKET_NAME

jabubake:
Replacing GCS_APP_LOCATION with GCS_BUCKET_NAME to minimize user error


jabubake:
@kurtisvg FYI

PR #596, opened 2018-09-04
appengine/helloworld: go111 quickstart

grayside:
* Move <= go 1.9 code to subfolder
* Add back helloworld sample for quickstart & basic how to build an app guide
* Added some basic HTTP handler test

PR #598, opened 2018-09-04
appengine: move existing samples to go19 folder

grayside:
Fast-tracking this change so it's not blocked by the release of #596


grayside:
Yes, I learned about the docs/appengine yesterday, if it's appropriate I will add that to this PR.

Coordination is in progress, Angela is ahead of me.

If we want to exempt directories because of potential links, I'm not sure we can move any of them. Could key blog posts be updated?

Issue #599, opened 2018-09-06
all: remove support for Go 1.6 and 1.8


tbpg:
They're being deprecated. So, we can drop sample support for them.

tbpg:
I'll handle updating Kokoro and context usage (also see #575 by @anguillanneuf). @broady will rework some tests (setup, teardown, etc.).

There is a good amount to cover here. So, this issue may be closed before everything has dropped support.

tbpg:
Closing this in favor of more specific issues. We've disabled the old builds and moved to the standard library `context` package, which are the biggest changes.

PR #601, opened 2018-09-06
all: use stdlib context package

tbpg:
I replaced all of the old `"golang.org/x/net/context"` imports with `"context"`.

There are also some changes to the `testing` directory.

Part of #599.

### DIFF
--- a/appengine/bigquery/app.go
+++ b/appengine/bigquery/app.go
@@ -7,12 +7,12 @@
 package sample
 
 import (
+	"context"
 	"fmt"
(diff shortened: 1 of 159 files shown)


tbpg:
I think it's OK for samples to drop support sooner. Any new code should be using new versions. If the user is on an old version, they'll know changes to make. Libraries are a bit different since they're depended on by user code (whereas this is copied & pasted from docs).

If you disagree, I could be convinced to wait until November.

PR #612, opened 2018-09-17
profiler: do not collect allocation profiles in profiler quickstart.

nolanmar511:
This guarantees that the first profile collected is a CPU profile, since only CPU profiles will be collected. This is necessary for the profiler quickstart.


tbpg:
LGTM.

PR #613, opened 2018-09-17
appengine/go11x/helloworld: Initial port of hello.

grayside:
Rebuild this on a clean branch.


appengine/go11x/helloworld/app.yaml
```diff
@@ -0,0 +1 @@
+runtime: go111
```
tbpg: Add a newline at the end?

grayside: done

PR #617, opened 2018-09-19
docs/appengine: move defer sample to another file

tbpg:
This also deletes the URL_endpoints sample since it's not used.

### DIFF
--- a/docs/appengine/taskqueue/push/defer.go
+++ b/docs/appengine/taskqueue/push/defer.go
@@ -11,7 +11,7 @@ import (
 )
 
 func example() {
-	var ctx context.Context
+	ctx := context.Background()
 
 	// [START deferred_tasks]
 	var expensiveFunc = delay.Func("some-arbitrary-key", func(ctx context.Context, a string, b int) {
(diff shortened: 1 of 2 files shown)


docs/appengine/taskqueue/push/defer.go
```diff
@@ -0,0 +1,24 @@
...
+package counter
+
+import (
+	"context"
+
+	"google.golang.org/appengine/delay"
+)
+
+func example() {
+	var ctx context.Context
```
enocom: This should be `ctx := context.Background` or something similar. Using `ctx` as is might mislead readers, since it's a nil interface and needs a backing type.

docs/appengine/taskqueue/push/defer.go
```diff
@@ -0,0 +1,24 @@
...
+
+	// [START deferred_tasks]
+	var expensiveFunc = delay.Func("some-arbitrary-key", func(ctx context.Context, a string, b int) {
```
enocom: Consider swapping `ctx` for `_` to avoid shadowing the top level `ctx` variable.

tbpg: I think this is OK to shadow? This defines a new function, then `ctx` is passed in when it's called. I left it outside the region tag because users probably don't want want to use `context.Background()` directly in prod.

docs/appengine/taskqueue/push/defer.go
tbpg: Done.

PR #621, opened 2018-09-21
profiler: add function to hotapp which repeatedly allocates memory

nolanmar511:
This will allow non-empty Heap Alloc profiles to be collected on this app.

### DIFF
--- a/profiler/hotapp/main.go
+++ b/profiler/hotapp/main.go
@@ -83,7 +83,7 @@ func allocMany() {
 		for i := 0; i < 16; i++ {
 			_ = make([]byte, 64*1024)
 		}
-		time.Sleep(time.Second)
+		time.Sleep(100 * time.Millisecond)
 	}
 }
 
--- a/profiler/hotapp/main.go
+++ b/profiler/hotapp/main.go
@@ -75,6 +75,18 @@ func allocImpl() {
 	}
 }
 
+// Simulates a function which allocates a lot of memory, but does not hold on
+// to that memory.
+func allocMany() {
+	// Allocate 1 MiB of 64 KiB chunks repeatedly.
+	for {
+		for i := 0; i < 16; i++ {
+			_ = make([]byte, 64*1024)
+		}
+		time.Sleep(time.Second)
+	}
+}
+
 // Simulates a CPU-intensive computation.
 func busyloop() {
 	for {
@@ -134,6 +146,9 @@ func main() {
 
 	// Simulate some memory allocation.
 	alloc()
+
+	// Simulate repeated memory allocations.
+	go allocMany()
 
 	// Simulate CPU load.
 	busyloop()
--- a/profiler/hotapp/main.go
+++ b/profiler/hotapp/main.go
@@ -75,8 +75,8 @@ func allocImpl() {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


nolanmar511:
CC/ @aalexand, @jqll

profiler/hotapp/main.go
```diff
@@ -135,6 +147,9 @@ func main() {
 	// Simulate some memory allocation.
 	alloc()
```
aalexand: Would it make sense to rename alloc to allocOnce so that it's possible to use a "show stacks" filter with "allocOnce" to filter in just function if that's ever needed?

nolanmar511: Done!

profiler/hotapp/main.go
```diff
@@ -75,6 +75,18 @@ func allocImpl() {
...
 
+// Simulates a function which allocates a lot of memory, but does not hold on
+// to that memory.
+func allocMany() {
+	// Allocate 1 MiB of 64 KiB chunks repeatedly.
+	for {
+		for i := 0; i < 16; i++ {
+			_ = make([]byte, 64*1024)
+		}
+		time.Sleep(time.Second)
```
aalexand: This loop allocates 1 MiB per second which isn't a lot and that's 10 MiBs per the profile duration of 10 seconds. The variation between profiles may be significant since the sampling threshold is 512 KiB by default. And I wouldn't be surprised if some profiles will miss samples altogether. I would increase the allocation rate at least 10x, maybe 100x. Basically, I would increase it sufficiently to make the allocation metric consistent enough between individual profiles (you can look at it using the weight filter which will show the distribution), but also not too much to avoid too much CPU time spent in GC.

nolanmar511: This now allocates 100MiB per second.
Over 71 profiles, allocMany has an average of 104.7MiB. Weight filters show profiles have between 50MiB and 140MiB.

aalexand: I don't think it's 100 MiB per second, I think it's per 10 seconds? Should be fine if at least average works out to be ~100 MiB per 10 second profile, just confirming the math.

aalexand: Looks good, I'll let the repo owners to approve the PR now.

profiler/hotapp/main.go
```diff
@@ -75,6 +75,18 @@ func allocImpl() {
 	}
 }
 
+// Simulates a function which allocates a lot of memory, but does not hold on
```
tbpg: Comments should start with the thing being described. For example,

```
// allocMany simulates a function...
```

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

profiler/hotapp/main.go
```diff
@@ -133,7 +145,10 @@ func main() {
 	}
 
 	// Simulate some memory allocation.
-	alloc()
+	allocOnce()
+
+	// Simulate repeated memory allocations.
+	go allocMany()
```
tbpg: I like how this leaves concurrency up to the caller of the function, rather than creating the goroutine inside of `allocMany`. :+1:

profiler/hotapp/main.go
nolanmar511: Done!

PR #624, opened 2018-09-24
firestore: update error handling

kurtisvg:
Update Firestore snippets error handling to logging the error and passing it back outside of the snippet context. This allows the snippet to be more easily copy/pasted from the docs site.


firestore/firestore_snippets/save.go
```diff
@@ -22,10 +23,11 @@ func addDocAsMap(ctx context.Context, client *firestore.Client) error {
 		"country": "USA",
 	})
 	if err != nil {
-		return err
+		// Handle the error correctly here.
+		log.Printf("An error has occurred: %s", err)
```
tbpg: It's more common to return the error here. It's OK to log and return then return `nil` below. The semantics are the same, but it's easier to read.

firestore/firestore_snippets/save.go
```diff
@@ -22,10 +23,11 @@ func addDocAsMap(ctx context.Context, client *firestore.Client) error {
 		"country": "USA",
 	})
 	if err != nil {
-		return err
+		// Handle the error correctly here.
```
tbpg: Say `TODO: Handle the error correctly.`?

firestore/firestore_snippets/save.go
```diff
@@ -164,23 +174,24 @@ func updateDocNested(ctx context.Context, client *firestore.Client) error {
 	}
 
 	// [START_EXCLUDE]
-	_, err := client.Collection("users").Doc("frank").Set(ctx, initialData)
-	if err != nil {
-		return err
+	_, preErr := client.Collection("users").Doc("frank").Set(ctx, initialData)
```
tbpg: Any particular reason to change this?

kurtisvg: Yes - this section is excluded in the snippets. Declaring `err :=` down below requires the part in scope of the snippets to reassign rather than declare, which is a compilation error if the snippet is copy and pasted.

firestore/firestore_snippets/save.go
kurtisvg: If the return is inside the region tags but the function declaration isn't, the return doesn't make sense  in the scope of the snippet.

firestore/firestore_snippets/save.go
kurtisvg: It's not really a TODO, it's a comment for the user instructing them that they should handle the error inside that block.

firestore/firestore_snippets/save.go
tbpg: How about:

```
if _, err := client.Collection("users").Doc("frank").Set(ctx, initialData); err != nil {
  return err
}
```

PR #634, opened 2018-09-26
monitoring/custommetric: use testutil.Retry

tbpg:
The current logic is flaky.


tbpg:
This is breaking nightly tests. So, merging.

PR #635, opened 2018-09-27
vision/detect: add localize objects samples

dizcology:


vision/detect/generated/sample-template.go
```diff
@@ -388,3 +388,32 @@ func detectAsyncDocument(w io.Writer, gcsSourceURI, gcsDestinationURI string) er
...
+// localizeObjects gets objects and bounding boxes from the Vision API for an image at the given file path.
+func localizeObjects(w io.Writer, file string) error {
+	var client *vision.ImageAnnotatorClient // Boilerplate is inserted by gen.go
+	annotations, err := client.LocalizeObjects(ctx, image, nil)
+	if err != nil {
+		return err
+	}
+
+	if len(annotations) == 0 {
+		fmt.Fprintln(w, "No objects found.")
```
tbpg: Maybe add a return nil here, then everything inside the else can be un-indented. See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow for inspiration (not an error condition necessarily).

PR #636, opened 2018-09-27
speech: add model selection sample

davidcavazos:
Region tag: `speech_transcribe_model_selection`

Hi @tbpg , can you have a look at this? Thanks!


davidcavazos:
Updated, should be ready to go :)

PR #637, opened 2018-10-02
pubsub: standardize sample topic & sub names

kenfesta:
Using "my-topic" and "my-sub" in all samples and gcloud examples


kenfesta:
Hi Tyler, 

that's actually the bug that I'm working on, which is to make all of these consistent across gcloud, dotnet, go, java, nodejs, php, python, and ruby (not all of these suggest specific example topic/sub names)...I'll send you that link.

tbpg:
Thank you!

PR #638, opened 2018-10-05
appengine/go11x: add example of serving static files

enocom:
The static example demonstrates how to serve static assets from a
directory and how to render templates with multiple variables.

/cc @grayside @tbpg


appengine/go11x/static/main.go
```diff
@@ -0,0 +1,83 @@
...
+var (
+	indexTmpl = template.Must(
+		template.ParseFiles(filepath.Join("templates", "index.html")),
+	)
+)
+
+func main() {
+	http.HandleFunc("/", indexHandler)
+
+	// Serve static files out of the public directory.
```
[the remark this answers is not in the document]
enocom: Done.

appengine/go11x/static/main.go
```diff
@@ -0,0 +1,83 @@
...
+	if r.URL.Path != "/" {
+		http.NotFound(w, r)
+		return
+	}
+	type indexData struct {
+		Logo        string
+		Style       string
+		RequestTime string
+	}
+	data := indexData{
```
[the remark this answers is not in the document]
enocom: Either seems fine to me.

appengine/go11x/static/main.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+var (
+	indexTmpl = template.Must(
+		template.ParseFiles(filepath.Join("templates", "index.html")),
+	)
+)
+
+func main() {
+	http.HandleFunc("/", indexHandler)
```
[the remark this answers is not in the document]
enocom: I've included the withLogging middleware for consistency for now.

appengine/go11x/static/main.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2018 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package main
```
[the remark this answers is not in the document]
enocom: Good point.

enocom:
@grayside PTAL

appengine/go11x/static/app.yaml
```diff
@@ -0,0 +1,12 @@
+runtime: go111
```
tbpg: Do we need a license header here and the other files (html, css)?

appengine/go11x/static/main.go
```diff
@@ -0,0 +1,86 @@
...
+		RequestTime string
+	}
+	data := indexData{
+		Logo:        "/public/gcp-gopher.svg",
+		Style:       "/public/style.css",
+		RequestTime: time.Now().Format(time.RFC822),
+	}
+	err := indexTmpl.Execute(w, data)
+	if err != nil {
+		log.Fatal(err)
```
tbpg: Log and do an `http.Error`?

appengine/go11x/static/main.go
```diff
@@ -0,0 +1,86 @@
...
+		Style       string
+		RequestTime string
+	}
+	data := indexData{
+		Logo:        "/public/gcp-gopher.svg",
+		Style:       "/public/style.css",
+		RequestTime: time.Now().Format(time.RFC822),
+	}
+	err := indexTmpl.Execute(w, data)
+	if err != nil {
```
tbpg:
```
if err := indexTmpl.Execute(w, data); err != nil { ... }
```

appengine/go11x/static/main.go
```diff
@@ -0,0 +1,86 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		resp := &statusResponse{ResponseWriter: w, statusCode: http.StatusOK}
+		next.ServeHTTP(resp, r)
+		log.Printf("[%d] %s %s", resp.statusCode, r.Method, r.URL.Path)
+	})
+}
+
+// statusResponse wraps http.ResponseWriter to record the response's status
+// code.
+type statusResponse struct {
```
tbpg: Not sure I understand this. Why doesn't the normal response code setup work?

enocom: In the context of middleware, there's no way of telling what a subsequent handlers has done. The approach saves the status code for use after the subsequent handlers are complete. Happy to consider alternative approaches.

enocom:
@tbpg PTAL.

appengine/go11x/static/main.go
tbpg: Gotcha. It looks like `statusCode` isn't used, though? If the example is for serving static files, everything else should be minimal? I might be missing the context of the sample.

appengine/go11x/static/public/style.css
```diff
@@ -0,0 +1,4 @@
+body {
```
tbpg: I think the CSS and HTML files should have licenses. Maybe app.yaml doesn't support it? If other files don't have the header, we should add it.

appengine/go11x/static/main.go
enocom: The `statusCode` field is used in the call to `log.Printf` above.

@grayside proposed adding logging to the sample to help users see when the static assets were being served. But it might be too much of a distraction from the purpose of this sample. I'm happy to remove it.

tbpg: Ahh. I understand now. `next` can call `WriteHeader`. SGTM.

appengine/go11x/static/public/style.css
enocom: Done.

enocom:
@tbpg I've added license headers to all files (except the svg) and removed the logging middleware to keep this sample to the point. PTAL.

PR #647, opened 2018-10-12
license_test: update to allow Google LLC.

tbpg:
This switches the test to use `regexp` and check the year is 4 digits. There does not seem to be an impact on test speed.


broady:
So odd that mailgun only failed on 1.11...

tbpg:
`go vet` is only run with Go 1.11. So, I guess if there are no tests, it doesn't try to compile the package?

PR #649, opened 2018-10-16
stackdriver: update region tags

tbpg:
This also adds a new snippet that was previously inlined in docs.

~For posterity, the Stackdriver Exporter looks in application default credentials
for the project, so we don't need to specify the project manually.~ Uses `os.Getenv("GOOGLE_CLOUD_PROJECT")` to avoid auto-detection (makes it easier to test on multiple projects with a single service account, too).


broady:
from memory, project id is only available from creds if obtained via service account JSON.

Issue #551, opened 2018-08-06
spanner: snippet tests are flaky


tbpg:
We should add the project name to the database name. Relying on timestamps to be unique can cause data races. See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/spanner/spanner_snippets/snippet_test.go#L29

```
snippet_test.go:44: run("addnewcolumn", "projects/golang-samples-tests/instances/golang-samples-tests/databases/test-1533489027"): rpc error: code = FailedPrecondition desc = Duplicate column name Albums.MarketingBudget.
snippet_test.go:44: run("addstoringindex", "projects/golang-samples-tests/instances/golang-samples-tests/databases/test-1533489027"): rpc error: code = FailedPrecondition desc = Duplicate name in schema: AlbumsByAlbumTitle2.
snippet_test.go:44: run("createtabledocswithhistorytable", "projects/golang-samples-tests/instances/golang-samples-tests/databases/test-1533489027"): rpc error: code = FailedPrecondition desc = Duplicate name in schema: Documents.
```

cpanato:
i will take this one as first contrib

tbpg:
Thank you! It might be difficult to test without a Spanner instance, but tests will run when you create the PR. Let me know if you have any questions. :)

[Here](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/storage/buckets/main_test.go#L44) is an example of using `testutil.SystemTest` to get the `ProjectID` and use it in a bucket name.

PR #651, opened 2018-10-17
spanner: replace time with the projectID when creating a DB for tests

cpanato:
fixes: https://github.com/GoogleCloudPlatform/golang-samples/issues/551

### DIFF
--- a/spanner/spanner_snippets/snippet_test.go
+++ b/spanner/spanner_snippets/snippet_test.go
@@ -19,6 +19,8 @@ import (
 )
 
 func TestSample(t *testing.T) {
+	tc := testutil.SystemTest(t)
+
 	instance := os.Getenv("GOLANG_SAMPLES_SPANNER")
 	if instance == "" {
 		t.Skip("Skipping spanner integration test. Set GOLANG_SAMPLES_SPANNER.")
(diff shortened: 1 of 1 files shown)


internal/testutil/testutil.go
```diff
@@ -13,6 +13,8 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
```
tbpg: Please delete line.

tbpg:
Thank you!

internal/testutil/testutil.go
cpanato: The blank line or the import? Thanks

cpanato: I cant see the error :(

tbpg: The import (and the blank line, but `gofmt`/`goimports` should take care of that).

The import is for the package it's in (kind of like `package foo` has `import foo`, if that makes sense), so it causes a cycle right now:

```
# github.com/GoogleCloudPlatform/golang-samples/spanner/spanner_snippets
import cycle not allowed
package github.com/GoogleCloudPlatform/golang-samples/spanner/spanner_snippets (test)
	imports github.com/GoogleCloudPlatform/golang-samples/internal/testutil
	imports github.com/GoogleCloudPlatform/golang-samples/internal/testutil
FAIL	github.com/GoogleCloudPlatform/golang-samples/spanner/spanner_snippets [setup failed]
```

tbpg: Here is how to run the tests (they will skip right now without Spanner setup, but it should still show you the compilation failure):

```
cd $GOPATH/src/github.com/GoogleCloudPlatform/golang-samples/spanner/spanner_snippets
go test -v ./...
```

(Note: the default `GOPATH` is `~/go`. You can see the current value with `go env GOPATH`. [CONTRIBUTING.md](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md) has some guidance for how to setup your local environment.)

broady:
Nice, thanks @cpanato!

PR #654, opened 2018-10-18
added import comments

DanSanche:
added imports to snippet comments (b/113025156)

### DIFF
--- a/kms/asymmetric/samples.go
+++ b/kms/asymmetric/samples.go
@@ -5,8 +5,6 @@
 // Samples for asymmetric keys feature of Cloud Key Management Service: https://cloud.google.com/kms/
 package samples
 
-// [START kms_get_asymmetric_public]
-
 import (
 	"context"
 	"crypto"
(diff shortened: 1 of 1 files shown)


kms/asymmetric/samples.go
```diff
@@ -25,11 +23,13 @@ import (
 	"google.golang.org/api/cloudkms/v1"
 )
 
-// [END kms_get_asymmetric_public]
-
 // [START kms_get_asymmetric_public]
 
 // getAsymmetricPublicKey retrieves the public key from a saved asymmetric key pair on KMS.
+//
+// Requires:
```
tbpg: These lines are a little bit weird since they don't list every import, only some of them. Would you mind splitting all of these functions into separate files and include the `import()` block in the region tag? I think that's the pattern the canonical snippet will follow, so it would be a step in the right direction.

DanSanche: I've had conversations with Kurtis (https://github.com/GoogleCloudPlatform/java-docs-samples/pull/1218) and the product team (b/113025156) on how best to handle this. We decided that until we have an agreed upon standard way of handling imports, it would be best to leave the imports in the comments for now. This gives the users the information they need, and it's easy to be consistent with how we present the information across all supported languages without significant refactors.

As for not including all imports, I assumed standard imports like _fmt_, _errors_, and _context_ would be considered a given for anyone familiar with go, but I can add them if you think it would be useful

tbpg: I'm not sure. These are all standard library imports. So, `goimports` should find them. So, I'm not sure this is worth adding.

If we decide on the right way to include imports in the canonical snippet, we can update this.

Does that make sense?

DanSanche: I suppose. I'd prefer to include these for consistency sake since these comments are needed for python and java. Even in go it might not be clear whether we're using standard crypto libraries or 3rd party ones with similar names, so I'd prefer to be explicit. But I guess these aren't as necessary for go

I am going to want to remove the region tags from the import block though, so this PR should stay open either way

PR #658, opened 2018-10-22
appengine/warmup: initial port of warmup to go 1.11

grayside:
This change needs to be tested to confirm warmup works as expected.


sbuss:
sample lgtm

appengine/go11x/warmup/app.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func warmupHandler(w http.ResponseWriter, r *http.Request) {
+	// Perform warmup tasks, including ones that require a context,
+	// such as retrieving data from Datastore.
+
+	log.Println("warmup done")
+}
+
+// [END gae_go111_warmup]
```
tbpg: Newline.

appengine/go11x/warmup/app.go
```diff
@@ -0,0 +1,37 @@
...
+		port = "8080"
+		log.Printf("Defaulting to port %s", port)
+	}
+
+	log.Printf("Listening on port %s", port)
+	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
+}
+
+func warmupHandler(w http.ResponseWriter, r *http.Request) {
+	// Perform warmup tasks, including ones that require a context,
```
tbpg: context.Context

appengine/go11x/warmup/app.yaml
```diff
@@ -0,0 +1,12 @@
+runtime: go111
+
+inbound_services:
+  - warmup
+
+automatic_scaling:
+  min_idle_instances: 1
```
tbpg: Why is this needed?

broady: I think it is needed, since warmup requests are for priming idle instances.

It'd be good to add a comment or link to the docs.

grayside: I've incorporated some feedback, but am holding off on linking to docs while I work out what docs should be linked.

Also working to confirm this is how we want to use warmup in go.

grayside:
Updates:
* Removed unnecessary min_instances configuration
* Renamed to warmup.go
* Added "real" example of setup tasks and put it in main for consistency on instance startup.
* Tested on app engine, though I did not happen to observe any warmup requests

sbuss:
> Tested on app engine, though I did not happen to observe any warmup requests

Hm. Try again with the thing that may-or-may-not be necessary?

```yaml
automatic_scaling:
  min_idle_instances: 1
```

grayside:
It's less that I expect there's a problem and more than I don't have a way to trigger a warmup request on demand.

appengine/go11x/warmup/warmup.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Printf("Defaulting to port %s", port)
+	}
+
+	log.Printf("Listening on port %s", port)
+	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
+}
+
+// setup executes per-instance one-time setup actions.
+func setup(ctx context.Context) error {
+	var err error
```
tbpg: Please move this declaration to right before it's used.

appengine/go11x/warmup/warmup.go
```diff
@@ -0,0 +1,70 @@
...
+
+var startupTime time.Time
+var client *storage.Client
+
+func main() {
+	if err := setup(context.Background()); err != nil {
+		log.Fatalf("setup: %v", err)
+	}
+
+	http.HandleFunc("/_ah/warmup", func(w http.ResponseWriter, r *http.Request) {
```
tbpg: I wonder if `setup` should be inline and warmup should be a different function, with a comment for what you might do in a warmup function?

It's not clear that `setup()` is key to understanding warmup. But, feel free to disagree if there was a goal with this setup.

grayside: My intention was to perform all warmup tasks as part of main and leave the warmup handler dedicated to logging that the official warmup request had successful triggered.

I wanted to demonstrate it this way because it seemed more unit testable to have a separate setup that could receive any context.

Would the concern be addressed if I rename setup to warmup and add comments explaining this?

grayside: @tbpg could you look at this again? I've made some comment and function name changes which hopefully help to clarify.

tbpg: I think my confusion comes my gut saying, "we should do warmup stuff during the warmup request." But, maybe the warmup request is a signal the instance is started and ready to go? In which case, I think naming the function `setup` is better than `warmup()` (otherwise, why is `warmup()` not connected to `/_ah/warmup`?). Whatever you think is most clear.

grayside: Switching to setup but keeping the comments.

The reason is that a warmup request is determined by scheduler magic, and if I understand correctly, doesn't preclude cold starts. It also ensures that local or other hosting of the application that uses a "normal" startup has the correct initialization.

I can't think of any initialization steps that I would **only** want on warmup but not on a cold start.

PR #666, opened 2018-10-24
[WIP] Update GCE bookshelf

ace-n:
**Do not merge** until that I confirm that everything works.

Related to [this Node PR](https://github.com/GoogleCloudPlatform/nodejs-getting-started/pull/198/files).


broady:
Not sure what you mean. In the context of Bookshelf, I think the only secret is the OAuth client secret.

ace-n:
What about things like SQL DB passwords?

ace-n:
Per our offline chat - I'll refactor this to use a config file.

ace-n:
Right now, it seems like we're fetching a copy of the `bookshelf` package from Github to run alongside the local `main` one. In other languages (perhaps not Go), this might come off as odd behavior - people expect their local changes to be integrated into the code.

Fixing this may require either a) vendoring the `bookshelf` package into the `main` one, or b) moving everything under one `main` package. Is such a change worth making?

tbpg:
Sounds like a `GOPATH` issue - it should work with local changes.

getting-started/bookshelf/config.go
```diff
@@ -37,84 +41,99 @@ var (
 	_ mgo.Session
 )
 
-const PubsubTopicID = "fill-book-details"
-
 func init() {
 	var err error
 
+	// Read config.json
```
tbpg: I don't think we need the read/parse comments.

getting-started/bookshelf/config.go
```diff
@@ -37,84 +41,99 @@ var (
...
+	configFile, err := ioutil.ReadFile("config.json")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Parse config.json
+	var configJson map[string]interface{}
```
tbpg: All of the values are strings. So, this can be `map[string]string`.

Alternatively, make it a `struct` with all of the fields defined, so it's a bit easier to use.

getting-started/bookshelf/config.go
```diff
@@ -37,84 +41,99 @@ var (
...
+	err = json.Unmarshal(configFile, &configJson)
+	if err != nil {
```
tbpg:
```suggestion
	if err := json.Unmarshal(configFile, &configJson); err != nil {
```

getting-started/bookshelf/config.go
tbpg: s/configJson/config/g

Also, initialisms in Go should have consistent case. So, this would be `configJSON`. https://github.com/golang/go/wiki/CodeReviewComments#initialisms

ace-n:
This PR is outdated, and is being abandoned.

PR #670, opened 2018-10-26
profiler: add a sample simulating multiple calls to an expensive library function.

aalexand:
See the code comments for details.

### DIFF
--- a/profiler/hotmid/main.go
+++ b/profiler/hotmid/main.go
@@ -1,0 +1,118 @@
+// Copyright 2018 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// [START hotmid]
+
+// Sample hotmid is an application that simulates multiple calls to a library
+// function made via different call paths. Each of these calls is not
+// particularly expensive (and so does not stand out on the flame graph). But
+// in the aggregate these calls add up to a significant time which can be
+// identified via looking at the flat list of functions' self and total time.
+package main
+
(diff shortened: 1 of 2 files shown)


profiler/hotmid/main.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"flag"
+	"log"
+	"time"
+
+	"cloud.google.com/go/profiler"
+)
+
+var (
+	// Service version to configure.
```
tbpg: Love these comments. Would you mind starting them with the thing being described? https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

profiler/hotmid/main.go
```diff
@@ -0,0 +1,118 @@
...
+func foo5() {
+	processItem(15)
+}
+
+func foo6() {
+	processItem(15)
+}
+
+func run() {
+	for start := time.Now(); *seconds == 0 || time.Since(start) < time.Duration(*seconds)*time.Second; {
```
tbpg: I wonder if this would be more clear as (untested):
```suggestion
	duration := time.Duration(*seconds)*time.Second
	for start := time.Now(); *seconds == 0 || time.Since(start) < duration; {
```

profiler/hotmid/main.go
aalexand: Done.

profiler/hotmid/main.go
aalexand: Done.

PR #677, opened 2018-10-31
gettingstarted/bookshelf/app: don't rely on port 8080 being open in tests

shollyman:
It appears there's test failures due to connection refused for
TestMainFunc, which invokes the application rather than the other
tests which render handlers.

On the suspicion something else is leaving the static port 8080 in
a temporarily unusuable state, this change picks a port randomly
and then passes it via the PORT environment variable which is used
by appengine.Main to override the serving port.

Also fixes a minor typo in internal/webtest.

### DIFF
--- a/getting-started/bookshelf/app/app_test.go
+++ b/getting-started/bookshelf/app/app_test.go
@@ -39,7 +39,7 @@ func TestMainFunc(t *testing.T) {
 	m := testutil.BuildMain(t)
 	defer m.Cleanup()
 	m.Run(env, func() {
-		appWt := webtest.New(nil, fmt.Sprintf(":%s", port))
+		appWt := webtest.New(nil, fmt.Sprintf("localhost:%s", port))
 		appWt.WaitForNet()
 		bodyContains(t, appWt, "/books", "No books found")
 	})
--- a/getting-started/bookshelf/app/app_test.go
+++ b/getting-started/bookshelf/app/app_test.go
@@ -14,6 +14,7 @@ import (
 	"testing"
 
 	"github.com/GoogleCloudPlatform/golang-samples/getting-started/bookshelf"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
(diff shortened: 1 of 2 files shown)


shollyman:
Example from http://sponge/a35b6719-a1c6-44da-a258-cb326a250da1

=== RUN   TestMainFunc
--- FAIL: TestMainFunc (1.97s)
    app_test.go:135: Get http://localhost:8080/books: dial tcp [::1]:8080: connect: connection refused

shollyman:
So, was able to get more insight here.  It seems the issue is related to changes in the appengine packages, likely https://github.com/golang/appengine/pull/170.  Because this is a flex app, it attempts to resolve information by interrogating a GCE metadata server.  This causes panics in everywhere that's not a GCE vm, and its racy between goroutines which is yielding the different errors.

The strategy of running this app locally is likely to continue to cause problems.  Likely the quickest option is to remove this build+run test for now, and leave the tests that exercise handlers.  I'll go ahead and remove the test with a followup change, but feel free to let me know if you'd prefer to see a different resolution here.

getting-started/bookshelf/app/app_test.go
```diff
@@ -28,16 +27,6 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-func TestMainFunc(t *testing.T) {
-	wt := webtest.New(t, "localhost:8080")
-	m := testutil.BuildMain(t)
-	defer m.Cleanup()
-	m.Run(nil, func() {
```
tbpg: I wonder if we should still build the main package, but not necessarily run it (to make sure it compiles). What do you think?

shollyman: It all depends on how well exercised the app is.  It doesn't hurt, but I don't think it provides a great deal of additional utility.  I'll pull it back in.

shollyman:
Remaining test failure is related to DLP.  Tracking that internally as 118816377.

PR #694, opened 2018-11-22
iot: add sample for iot core commands 

hongalex:
will add tests in separate pull request

### DIFF
--- a/iot/manager/manager.go
+++ b/iot/manager/manager.go
@@ -107,8 +107,9 @@ func deleteRegistry(w io.Writer, projectID string, region string, registryID str
 
 // [END iot_delete_registry]
 
+// [START iot_get_registry]
+
 func getRegistry(w io.Writer, projectID string, region string, registryID string) (*cloudiot.DeviceRegistry, error) {
-	// [START iot_get_registry]
 	// Authorize the client using Application Default Credentials.
 	// See https://g.co/dv/identity/protocols/application-default-credentials
 	ctx := context.Background()
--- a/iot/manager/manager.go
+++ b/iot/manager/manager.go
@@ -1,6 +1,16 @@
-// Copyright 2018 Google Inc. All rights reserved.
-// Use of this source code is governed by the Apache 2.0
-// license that can be found in the LICENSE file.
+// Copyright 2018 Google LLC
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     https://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 // Command manager lets you manage Cloud IoT Core devices and registries.
 package main
@@ -9,6 +19,7 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -25,9 +36,10 @@ import (
 
 // Registry Management
 
+// [START iot_create_registry]
+
 // createRegistry creates a device registry.
-func createRegistry(projectID string, region string, registryID string, topicName string) (*cloudiot.DeviceRegistry, error) {
-	// [START iot_create_registry]
+func createRegistry(w io.Writer, projectID string, region string, registryID string, topicName string) (*cloudiot.DeviceRegistry, error) {
 	// Authorize the client using Application Default Credentials.
 	// See https://g.co/dv/identity/protocols/application-default-credentials
 	ctx := context.Background()
@@ -55,19 +67,21 @@ func createRegistry(projectID string, region string, registryID string, topicNam
 		return nil, err
 	}
 
-	fmt.Println("Created registry:")
-	fmt.Println("\tID: ", response.Id)
-	fmt.Println("\tHTTP: ", response.HttpConfig.HttpEnabledState)
-	fmt.Println("\tMQTT: ", response.MqttConfig.MqttEnabledState)
-	fmt.Println("\tName: ", response.Name)
-	// [END iot_create_registry]
+	fmt.Fprintf(w, "Created registry:\n")
+	fmt.Fprintf(w, "\tID: %s\n", response.Id)
+	fmt.Fprintf(w, "\tHTTP: %s\n", response.HttpConfig.HttpEnabledState)
+	fmt.Fprintf(w, "\tMQTT: %s\n", response.MqttConfig.MqttEnabledState)
+	fmt.Fprintf(w, "\tName: %s\n", response.Name)
 
 	return response, err
 }
 
+// [END iot_create_registry]
+
+// [START iot_delete_registry]
+
 // deleteRegistry deletes a device registry
-func deleteRegistry(projectID string, region string, registryID string) (*cloudiot.Empty, error) {
-	// [START iot_delete_registry]
+func deleteRegistry(w io.Writer, projectID string, region string, registryID string) (*cloudiot.Empty, error) {
 	// Authorize the client using Application Default Credentials.
 	// See https://g.co/dv/identity/protocols/application-default-credentials
 	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


iot/manager/manager.go
```diff
@@ -613,6 +614,38 @@ func setConfig(projectID string, region string, registry string, deviceID string
 	return response, err
 }
 
+// sendCommand sends a command to a device listening for commands
+func sendCommand(projectID string, region string, registry string, deviceID string, sendData string) (*cloudiot.SendCommandToDeviceResponse, error) {
```
tbpg: Test?

iot/manager/manager.go
```diff
@@ -613,6 +614,38 @@ func setConfig(projectID string, region string, registry string, deviceID string
...
+	// [START iot_send_command]
```
tbpg: We're generally putting region tags outside the function. But, it's better to be consistent within a given product.

iot/manager/manager.go
tbpg: It's common to pass an `io.Writer` to write output to. Then, the line below would be `fmt.Fprintf(w, ...)`. That is much easier to test (can ensure the output is correct) and matches HTTP handlers.

iot/manager/manager.go
hongalex: Yeah for our other languages we have the region tags inside the functions. @gguuss any thoughts on this?

gguuss: I don't have a strong opinion on this -- AFAIK we're exposing these code snippets only within the context of static examples and are not composing snippets from multiple functions so either way they should appear in the documentation in a way that will work for our users.

tbpg: For Go, we've decided on region tags outside the function declaration. Some languages are inside, some are outside. But, if the other Go samples on this page have it a particular way, it's probably good to keep it consistent.

hongalex: I'll move the region tags outside since I need to refactor the other samples anyway to use `io.Writer` anyway

iot/manager/manager.go
```diff
@@ -85,14 +90,16 @@ func deleteRegistry(projectID string, region string, registryID string) (*cloudi
...
 	return response, err
 }
 
-func getRegistry(projectID string, region string, registryID string) (*cloudiot.DeviceRegistry, error) {
-	// [START iot_get_registry]
+// [END iot_delete_registry]
+
+// [START iot_get_registry]
+
+func getRegistry(w io.Writer, projectID string, region string, registryID string) (*cloudiot.DeviceRegistry, error) {
```
tbpg: Please add a function comment.

iot/manager/manager.go
```diff
@@ -110,14 +117,21 @@ func getRegistry(projectID string, region string, registryID string) (*cloudiot.
...
 
 	return response, err
 }
 
+// [START iot_get_iam_policy]
+
 // getRegistryIam gets the IAM policy for a device registry.
-func getRegistryIam(projectID string, region string, registryID string) (*cloudiot.Policy, error) {
-	// [START iot_get_iam_policy]
+func getRegistryIam(w io.Writer, projectID string, region string, registryID string) (*cloudiot.Policy, error) {
```
tbpg: s/Iam/IAM/

iot/manager/manager.go
```diff
@@ -138,21 +152,23 @@ func getRegistryIam(projectID string, region string, registryID string) (*cloudi
 		return nil, err
 	}
 
-	fmt.Println("Policy:")
+	fmt.Fprintf(w, "Policy:\n")
```
tbpg: Small comment, `Fprintln` is OK if you aren't printing any variables. Whatever you prefer.

iot/manager/manager.go
```diff
@@ -171,18 +187,20 @@ func listRegistries(projectID string, region string) ([]*cloudiot.DeviceRegistry
...
 }
 
+// [END iot_list_registries]
+
+// [START iot_set_iam_policy]
+
 // setRegistryIam sets the IAM policy for a device registry
-func setRegistryIam(projectID string, region string, registryID string, member string, role string) (*cloudiot.Policy, error) {
-	// [START iot_set_iam_policy]
+func setRegistryIam(w io.Writer, projectID string, region string, registryID string, member string, role string) (*cloudiot.Policy, error) {
```
tbpg: IAM

iot/manager/manager.go
```diff
@@ -211,17 +229,19 @@ func setRegistryIam(projectID string, region string, registryID string, member s
...
+// [END iot_set_iam_policy]
+
 // Device Management
 
+// [START iot_create_es_device]
+
 // createEs creates a device in a registry with ES credentials
-func createEs(projectID string, region string, registry string, deviceID string, keyPath string) (*cloudiot.Device, error) {
-	// [START iot_create_es_device]
+func createEs(w io.Writer, projectID string, region string, registry string, deviceID string, keyPath string) (*cloudiot.Device, error) {
```
tbpg: Should this be `createES`?

iot/manager/manager.go
```diff
@@ -257,15 +277,17 @@ func createEs(projectID string, region string, registry string, deviceID string,
...
 }
 
+// [END iot_create_es_device]
+
+// [START iot_create_rsa_device]
+
 // createRsa creates a device in a registry with RS credentials
-func createRsa(projectID string, region string, registry string, deviceID string, keyPath string) (*cloudiot.Device, error) {
-	// [START iot_create_rsa_device]
+func createRsa(w io.Writer, projectID string, region string, registry string, deviceID string, keyPath string) (*cloudiot.Device, error) {
```
tbpg: RSA

iot/manager/manager.go
```diff
@@ -301,15 +323,17 @@ func createRsa(projectID string, region string, registry string, deviceID string
...
+	fmt.Fprintf(w, "Successfully created RSA device\n")
 
 	return response, err
 }
 
+// [END iot_create_rsa_device]
+
+// [START iot_create_unauth_device]
+
 // createUnauth creates a device in a registry without credentials
```
tbpg: Nit: Periods at the end of sentences.

iot/manager/manager.go
```diff
@@ -531,15 +567,17 @@ func patchDeviceEs(projectID string, region string, registry string, deviceID st
...
 }
 
+// [END iot_patch_es]
+
+// [START iot_patch_rsa]
+
 // patchDeviceRsa patches a device to use RSA credentials
-func patchDeviceRsa(projectID string, region string, registry string, deviceID string, keyPath string) (*cloudiot.Device, error) {
-	// [START iot_patch_rsa]
+func patchDeviceRsa(w io.Writer, projectID string, region string, registry string, deviceID string, keyPath string) (*cloudiot.Device, error) {
```
tbpg: RSA

iot/manager/manager_test.go
```diff
@@ -0,0 +1,113 @@
...
+	s := m.Run()
+	log.SetOutput(os.Stderr)
+	shutdown()
+	os.Exit(s)
+}
+
+func setup() {
+	ctx := context.Background()
+	// Retrieve project ID from console
+	projectID = os.Getenv("GOLANG_SAMPLES_PROJECT_ID")
```
tbpg: Please use `testutil.SystemTest` to get the projectID.

Example: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/kms/asymmetric/samples_test.go#L101

iot/manager/manager.go
```diff
@@ -677,7 +753,10 @@ func main() {
 	flag.Parse()
 
 	// Retrieve project ID from console
-	projectID := os.Getenv("GCLOUD_PROJECT")
+	projectID := os.Getenv("GOLANG_SAMPLES_PROJECT_ID")
```
tbpg: I think it's OK to leave this one out and only do the normal ones. The test setup sets the `GOOGLE_CLOUD_PROJECT` variable.

iot/manager/manager.go
hongalex: yup, fixed

iot/manager/manager.go
hongalex: Done

iot/manager/manager.go
hongalex: Got it. Some responses don't use same-case initialisms, like `response.Id` but I'll change it for everything else

iot/manager/manager.go
hongalex: Good catch. Changed to using `Fprintln` for no variables prints

tbpg:
One Go version failed due to network flakiness. Merging.

PR #698, opened 2018-12-04
testing/gimmeproj: add option to output list

### LINKED ISSUE #669: testing: enhance gimmeproj to support scripting changes across projects
When making a change to the setup of projects, it's fairly common to implement something like this:

```
gcloud config set project golang-samples-tests
gcloud <my project change>

for i in {2..10}; do
  gcloud config set project golang-samples-tests-$i
  gcloud <my project change>
done

gcloud config set project $GOOGLE_CLOUD_PROJECT
```

That change seems like it could easily be facilitated, and avoid the need to know the naming conventions of the underlying project.

## Solution Options

These options are not mutually exclusive.

### Option A: Listing Projects

`gimmeproj list` provides an iterable list of the projects. This would include the meta project, and allow for scripting like so:

```
for $proj in $(gimmeproj list); do foo $proj; done
```

**Pros:** Versatile, Low maintenance
**Cons:** Not Terse, More Scripting Required, Usage Samples Needed

### Option B: Execution Wrapper

```
gimmeproj exec <foo>
gimmeproj exec --file=<foo>
```

In this case, gimmeproj wraps your one-liner or script in it's own for loop.
Project wrangling could be done by injecting $GOOGLE_PROJECT_NAME as an environment variable, and/or by explicitly calling `gcloud config set project $GOOGLE_PROJECT_NAME` before running the command.

Future enhancements could facilitate output aggregation, impose confirmation steps, and so on.

**Pros:** Smooth Experience, Minimal Memorization
**Cons:** Higher Maintenance Overhead

grayside:
Fixes #669

I exempted list from extraneous output to facilitate it's scripting focus. Happy to change the approach as needed.


testing/gimmeproj/main.go
```diff
@@ -125,12 +126,19 @@ Administrative commands:
 		return fmt.Errorf("datastore.NewClient: %v", err)
 	}
 
+	// List is intended explicitly for scripting.
+	if flag.Arg(0) != "list" {
+		fmt.Println(flag.Arg(0), flag.Arg(1))
+	}
+
 	switch flag.Arg(0) {
 	case "help":
```
broady: add list here.

if facilitating scripting, specifically, add a `-json` or `-output=json` flag.

would adding that to "status" work fine?

testing/gimmeproj/main.go
```diff
@@ -125,12 +126,19 @@ Administrative commands:
 		return fmt.Errorf("datastore.NewClient: %v", err)
 	}
 
+	// List is intended explicitly for scripting.
+	if flag.Arg(0) != "list" {
+		fmt.Println(flag.Arg(0), flag.Arg(1))
+	}
```
broady: remove? looks like this is for debugging?

testing/gimmeproj/main.go
```diff
@@ -199,6 +207,16 @@ func lease(ctx context.Context, duration string) error {
 	return nil
 }
 
+func list(ctx context.Context) error {
+	return withPool(ctx, func(pool *Pool) error {
+		for _, proj := range pool.Projects {
+			fmt.Println(proj.ID)
+		}
```
broady: if facilitating scripting, specifically, add a `-json` or `-output=json` flag.

would adding that to "status" work fine?

broady:
list seems pretty similar to status, consider other ways to enable scripting while keeping the command list small(ish)?

testing/gimmeproj/main.go
grayside: I have no opinions either way. The println statement itself already existed and I assumed it's intended to confirm the action being taken, since the various commands are not always transparent in what they do.

tbpg:
Maybe including the expiration should be from a `--verbose` flag, and just the project list should be the default, plus rename `status` to `list`? I don't have strong opinions either way.

Environment variables seem like overkill.

testing/gimmeproj/main.go
broady: I don't see the existing print statement.

broady:
LGTM, @tbpg for approval

tbpg:
Remove the one print statement (looks new?) then LGTM. No need to do the idea above unless you particularly like it.

testing/gimmeproj/main.go
grayside: Done.

testing/gimmeproj/main.go
```diff
@@ -92,16 +93,18 @@ func submain() error {
 	usage := errors.New(`
 Usage:
 	gimmeproj -project=[meta project ID] command
+	gimmeproj -project=[meta project ID] -output=list status
 
 Commands:
 	lease [duration]    Leases a project for a given duration. Prints the project ID to stdout.
+	list [project ID]   List all projects in the pool.
```
broady: not used

broady:
LGTM

tbpg:
Failures are unrelated (3/4 failed to download a dependency. I can't wait for the proxy.)

PR #703, opened 2018-12-05
KMS: Updated for new client library

DanSanche:
- Updated all samples to use the new libraries
- Added missing tests in the process

tracking bug: b/120218275


kms/asymmetric.go
```diff
@@ -0,0 +1,260 @@
...
+		Name: keyName,
+	}
+	// Call the API.
+	response, err := client.GetPublicKey(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch public key: %+v", err)
+	}
+	// Parse the key.
+	keyBytes := []byte(response.Pem)
+	block, _ := pem.Decode(keyBytes)
```
tbpg: Should we check this error?

kms/asymmetric.go
```diff
@@ -0,0 +1,260 @@
...
+	return nil
+}
+
+// [END kms_create_asymmetric_key]
+
+// [START kms_get_asymmetric_public]
+
+// getAsymmetricPublicKey retrieves the public key from a saved asymmetric key pair on KMS.
+// example keyName: "projects/PROJECT_ID/locations/global/keyRings/RING_ID/cryptoKeys/KEY_ID/cryptoKeyVersions/1""
+func getAsymmetricPublicKey(keyName string) (interface{}, error) {
```
tbpg: The `get` prefix is usually discouraged in Go. But, I think this is OK if you think it's the best name. https://golang.org/doc/effective_go.html#Getters

kms/asymmetric.go
```diff
@@ -0,0 +1,260 @@
...
+// example keyName: "projects/PROJECT_ID/locations/global/keyRings/RING_ID/cryptoKeys/KEY_ID/cryptoKeyVersions/1""
+func signAsymmetric(keyName string, message []byte) ([]byte, error) {
+	// Note: some key algorithms will require a different hash function.
+	// For example, EC_SIGN_P384_SHA384 requires SHA-384.
+	ctx := context.Background()
+	client, err := cloudkms.NewKeyManagementClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+	// Find the digest of the message
```
tbpg: Period at the end.

kms/asymmetric.go
```diff
@@ -0,0 +1,260 @@
...
+	digest := sha256.New()
+	digest.Write(message)
+	// Build the signing request
```
tbpg: Period at the end.

kms/asymmetric.go
```diff
@@ -0,0 +1,260 @@
...
+		return err
+	}
+
+	// Retrieve the public key from KMS.
+	response, err := client.GetPublicKey(ctx, &kmspb.GetPublicKeyRequest{Name: keyName})
+	if err != nil {
+		return fmt.Errorf("failed to fetch public key: %+v", err)
+	}
+	// Parse the key.
+	block, _ := pem.Decode([]byte(response.Pem))
```
tbpg: Check this error?

kms/asymmetric.go
```diff
@@ -0,0 +1,260 @@
...
+	abstractKey, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return fmt.Errorf("failed to parse public key: %+v", err)
+	}
+	rsaKey := abstractKey.(*rsa.PublicKey)
```
tbpg: Use the "comma ok" pattern to check this assertion?
```suggestion
	rsaKey, ok := abstractKey.(*rsa.PublicKey)
	if !ok {
		return fmt.Errorf(...)
	}
```

kms/asymmetric.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	return nil
+}
+
+// [END kms_verify_signature_rsa]
+
+// [START kms_verify_signature_ec]
+
+// verifySignatureEC will verify that an 'EC_SIGN_P256_SHA256' signature is valid for a given message.
+// example keyName: "projects/PROJECT_ID/locations/global/keyRings/RING_ID/cryptoKeys/KEY_ID/cryptoKeyVersions/1""
```
tbpg:
```suggestion
// example keyName: "projects/PROJECT_ID/locations/global/keyRings/RING_ID/cryptoKeys/KEY_ID/cryptoKeyVersions/1"
```

kms/quickstart.go
```diff
@@ -0,0 +1,54 @@
+// Copyright 2017 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Sample quickstart is a basic program that uses Cloud KMS.
+package kms
+
+// [START kms_quickstart]
```
tbpg: I think we should include the package declaration in the region tag -- looks a little weird without it.

kms/quickstart.go
```diff
@@ -0,0 +1,54 @@
+// Copyright 2017 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Sample quickstart is a basic program that uses Cloud KMS.
+package kms
```
tbpg: Should this be a main package? Weird to have `func main` outside of `package main`.
```suggestion
package main
```

kms/quickstart.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"log"
+
+	cloudkms "cloud.google.com/go/kms/apiv1"
+	"google.golang.org/api/iterator"
+	kmspb "google.golang.org/genproto/googleapis/cloud/kms/v1"
+)
+
+func main() {
+	projectID := "your-project-id"
```
tbpg: Use an env variable for the project?

```suggestion
	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
	if projectID == "" {
		log.Fatal("Must set the GOOGLE_CLOUD_PROJECT environment variable."
	}
```

kms/asymmetric.go
DanSanche: I think it would be best to keep get. We want the name to be consistent with other languages, and this isn't exactly a getter anyway

kms/quickstart.go
DanSanche: Using a string like this seems to be a standard used by many quickstarts. I think we should keep it as is to show the minimum viable code, and to be consistent with the other languages

tbpg:
Changes LGTM. But, we need to figure out how to avoid quota issues (retry with sleep? different project config? increase quota?)

DanSanche:
I added some testutil.Retry blocks in there and it seems to be working now. If there are any more quota issues, let me know and I can figure out how to extend it for the test projects

tbpg:
Another quota failure :(
```
snippets_test.go:167: disableCryptoKeyVersion(projects/golang-samples-tests/locations/global/keyRings/kms-samples/cryptoKeys/symmetric/cryptoKeyVersions/1): rpc error: code = ResourceExhausted desc = Quota exceeded for quota metric 'cloudkms.googleapis.com/write_requests' and limit 'WriteRequestsPerMinutePerProject' of service 'cloudkms.googleapis.com'
```

DanSanche:
I didn't realize those tests used the quota too. I added more retry blocks. I'll make a note to look into increasing the quota after the holidays too

bdhess:
CC @SurferJeffAtGoogle -- seeing quota failures in .NET docs samples as well at dotnet-docs-samples#720

We should probably coordinate our approach to the quota issues across the samples for all languages.  Getting quota increases for the projects we use is straightforward.  I wonder, though, if it isn't time for the Cloud KMS team to consider raising the default quota given that even a fairly trivial set of interactions in a test environment is causing to hit limits.

DanSanche:
@tbpg are you still seeing quota issues after my latest commit? Do you think it's worth looking into increasing the quota before merging?

Otherwise, this should be ready to merge (I don't have permissions to do it myself)

PR #712, opened 2018-12-28
appengine/tasks: create_tasks executable

grayside:
### Goal

Make create_tasks into an executable in line with other languages.

### Approach

* Make creation and handling of tasks "equal" apps in separate directories
* Add a main.go which handles the CLI wrapping for the snippet.
* Use the nodejs approach of CLI arguments to pass in project, location, and queue. 

Since the Go snippet supports a parameterized message, this is also available as a fourth optional argument.


appengine/go11x/tasks/create_task/create_task.go
```diff
@@ -4,8 +4,8 @@
 
 // [START cloud_tasks_appengine_create_task]
 
-// Package tasks contains snippets for using Cloud Tasks with App Engine.
-package tasks
+// Sample Create Task contains snippets for using Cloud Tasks with App Engine.
```
tbpg:
```suggestion
// Sample create_task contains snippets for using Cloud Tasks with App Engine.
```

Maybe "Command create_task"?

appengine/go11x/tasks/create_task/main.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2018 Google LLC. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Sample Create Task contains snippets for using Cloud Tasks with App Engine.
```
tbpg: Update this comment? This doesn't contain snippets and is a command.

averikitsch:
This is for a quickstart. The executable was added to match other languages.

tbpg:
I don't see the quickstart tag? Maybe I'm missing it. We should probably have the `func main()` as part of the region if we'd like to make the sample executable. Having `package main` without `func main` in the snippet is a little odd. Otherwise, I don't think we should have the `func main()`.

PR #719, opened 2019-01-04
testutil:  revisit BuildMain's runner

shollyman:
This provides a non-interactive runner for the testutil binary builder that
exposes stdout/stderr after the fact.

Checking the repo, there are apparently no active callers using
the existing Run(), so it could potentially be replaced with this
non-interactive runner.

There's some oddity around naming collisions with the Context
declarations from testutil.go for defining system vs e2e tests.
I could potentially remove the use of the context package since it's
only for leveraging the exec.CommandContext.

Here's the sample test from the sample app that would consume
something like this.

```
func TestApp(t *testing.T) {
	m := testutil.BuildMain(t)
	defer m.Cleanup()

	if !m.Built() {
		t.Errorf("failed to build app")
	}

	env := make(map[string]string)
	if p, ok := os.LookupEnv("GOLANG_SAMPLES_PROJECT_ID"); ok {
		env["GOOGLE_CLOUD_PROJECT"] = p
	}
	if ok := m.RunNonInteractive(env, 30*time.Second); !ok {
		t.Errorf("execution failed: %v", string(m.Stderr))
	}
}

```


internal/testutil/runmain.go
```diff
@@ -5,6 +5,8 @@
 package testutil
 
 import (
+	realContext "context"
```
tbpg: We should rename the `testutil.context()` function and not rename this.

internal/testutil/runmain.go
```diff
@@ -102,3 +106,35 @@ func (r *Runner) Run(env map[string]string, f func()) {
 	case <-done:
 	}
 }
+
+// RunNonInteractive runs the build binary until terminated or timeout has
+// been reached, and indicates successful execution on return.
+func (r *Runner) RunNonInteractive(env map[string]string, timeout time.Duration) bool {
```
tbpg: I'm good to rewrite `Run` if it's not used anywhere.

internal/testutil/runmain.go
```diff
@@ -102,3 +106,35 @@ func (r *Runner) Run(env map[string]string, f func()) {
...
+	cmd.Env = environ
+
+	out, err := cmd.Output()
+	r.Stdout = out
+	if err != nil {
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			r.Stderr = exitErr.Stderr
+		}
+
+		// propagate this, or let callers decide?
```
tbpg: What if we return the error (and exit code)? Then, the caller has a bit more control than a bool.

shollyman:
There's a variety of test failures unrelated to this change, sending other cleanups to address those.

internal/testutil/runmain.go
```diff
@@ -59,46 +63,55 @@ func (r *Runner) Cleanup() {
...
+// Run executes runs the built binary until terminated or timeout has
+// been reached, and indicates successful execution on return.
+func (r *Runner) Run(env map[string]string, timeout time.Duration) error {
 	if !r.Built() {
-		r.t.Error("Tried to run when binary not built.")
-		return
+		return fmt.Errorf("tried to run when binary not built")
 	}
+	r.stderr.Reset()
+	r.stdout.Reset()
```
broady: I don't particularly like holding so much state on the "Runner" struct.

A separate struct for each run seems to make sense to me, that way you can build once and run several times.

If we only ever need to run once, I'd make a Runner only valid once. This will prevent unexpected concurrency issues.

internal/testutil/runmain.go
```diff
@@ -59,46 +63,55 @@ func (r *Runner) Cleanup() {
...
 	cmd.Env = environ
 
-	if err := cmd.Start(); err != nil {
-		r.t.Error(err)
-		return
+	stdErr, err := cmd.StderrPipe()
+	if err != nil {
+		return fmt.Errorf("could not get stderr pipe")
+	}
+	stdOut, err := cmd.StdoutPipe()
```
broady: cmd.Stdout = r.stdout
cmd.Stderr = r.stderr

No need to mess around with pipes.

internal/testutil/runmain.go
shollyman: Good point.  I'll cull the state from Runner and return everything directly from Run().

internal/testutil/runmain.go
shollyman: also done

internal/testutil/runmain.go
shollyman: done

internal/testutil/runmain.go
shollyman: per cbro's suggestion, returning stdout/stderr and error value.

internal/testutil/runmain.go
```diff
@@ -59,46 +61,33 @@ func (r *Runner) Cleanup() {
...
 
-	select {
-	case <-time.After(5 * time.Second):
-		r.t.Error("Timed out with SIGINT, trying SIGKILL.")
-		if err := cmd.Process.Kill(); err != nil {
-			r.t.Error(err)
-		}
-	case <-done:
+	if err := cmd.Wait(); err != nil {
+		return bufOut.Bytes(), bufErr.Bytes(), fmt.Errorf("error waiting for termination: %v", err)
```
broady: this isn't an error waiting for termination, rather it might just be a non-zero exit code.
I don't think you really need to wrap this error.

internal/testutil/runmain.go
```diff
@@ -59,46 +61,33 @@ func (r *Runner) Cleanup() {
 	}
 }
 
-// Run runs the built binary with the given environment.
-// After f returns, the running process is shut down.
-func (r *Runner) Run(env map[string]string, f func()) {
```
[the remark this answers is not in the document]
tbpg: Quickstarts are still runnable.

PR #723, opened 2019-01-08
Speech multichannel

blechdom:
Created new Speech Multichannel Transcription snippet. Added stereo audio file to testdata.

PR #725, opened 2019-01-08
speech: add multichannel sample

blechdom:
Refactored Mulitchannel Speech Snippet according to feedback I received from previous pull request #723 - added test, function comment, gofmt, and direct function instead of run


speech/snippets/multichannel.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Package snippets contains speech examples.
+package snippets
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	// [START imports]
```
tbpg: Is this region tag going to be used? If so, the region tag should start with "speech" (required by the samples tracker).

speech/snippets/multichannel.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+
+	speech "cloud.google.com/go/speech/apiv1p1beta1"
+	speechpb "google.golang.org/genproto/googleapis/cloud/speech/v1p1beta1"
+	// [END imports]
+)
+
+// TranscribeMultichannel generates a transcript of a multichannel speech file and tags the speech from each channel.
```
tbpg: The comment should be on the line directly above the function declaration and have the same case.

```
// transcribeMultichannel ...
func transcribeMultichannel(...)...
```

speech/snippets/multichannel.go
```diff
@@ -0,0 +1,61 @@
...
+
+func transcribeMultichannel(w io.Writer, path string) error {
+	ctx := context.Background()
+
+	client, err := speech.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewClient: %v", err)
+	}
+  	// path = "../testdata/commercial_stereo.wav"
```
tbpg: Can delete this comment - might be confusing when viewing embedded in the docs.

speech/snippets/multichannel_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestMultichannel(t *testing.T) {
+	testutil.SystemTest(t)
+
+	var buf bytes.Buffer
+	err := transcribeMultichannel(&buf, "../testdata/commercial_stereo.wav")
+	if err != nil {
```
tbpg: Prefer inline declarations if none of the variables are needed outside the if statement. Helps to avoid forgetting to check the error or accidentally shadowing a variable.

```suggestion
	if err := transcribeMultichannel(&buf, "../testdata/commercial_stereo.wav"); err != nil {
```

speech/snippets/multichannel_test.go
```diff
@@ -0,0 +1,27 @@
...
+		t.Fatal(err)
+	}
+
+	if got := buf.String(); !strings.Contains(got, "Channel 1: hi I'd like to buy a Chromecast I'm always wondering whether you could help me with that\nChannel 2: certainly which color would you like we have blue black and red\nChannel 1:  let's go with the black one\nChannel 2:  would you like the new Chromecast Ultra model or the regular Chromecast\nChannel 1:  regular Chromecast is fine thank you\nChannel 2:  okay sure would you like to ship it regular or Express\nChannel 1:  express please\nChannel 2:  terrific it's on the way thank you\nChannel 1:  thank you very much bye\n") {
```
tbpg: It might be better to only check a small portion of the result. It's common for the transcription to change over time, which can cause the test to fail.

speech/snippets/multichannel_test.go
```diff
@@ -0,0 +1,27 @@
...
+		t.Fatalf(`transcribeMultichannel(../testdata/commercial_stereo.wav) = %q; want "Channel 1: hi I'd like to buy a Chromecast I'm always wondering whether you could help me with that\nChannel 2: certainly which color would you like we have blue black and red\nChannel 1:  let's go with the black one\nChannel 2:  would you like the new Chromecast Ultra model or the regular Chromecast\nChannel 1:  regular Chromecast is fine thank you\nChannel 2:  okay sure would you like to ship it regular or Express\nChannel 1:  express please\nChannel 2:  terrific it's on the way thank you\nChannel 1:  thank you very much bye\n"`, got)
```
tbpg: Create a `want` variable to avoid duplicating the string?

Also, "want to contain".

speech/snippets/multichannel.go
blechdom: reordered the imports

speech/snippets/multichannel.go
blechdom: fixed

speech/snippets/multichannel.go
blechdom: done

speech/snippets/multichannel_test.go
blechdom: got it

speech/snippets/multichannel_test.go
blechdom: yep

speech/snippets/multichannel_test.go
blechdom: fixed

speech/snippets/multichannel.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2016 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+// Package snippets contains speech examples.
+package snippets
+
```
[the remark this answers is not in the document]
blechdom: added region tag

PR #727, opened 2019-01-08
kms: add 2 region tags

WalterHub:
Add tags for kms_remove_member_from_keyring_policy and kms_get_cryptokey_policy.

### DIFF
--- a/kms/snippets.go
+++ b/kms/snippets.go
@@ -15,6 +15,7 @@ import (
 )
 
 // [START kms_create_keyring]
+
 // createKeyRing creates a new ring to store keys on KMS.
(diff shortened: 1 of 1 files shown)


kms/snippets.go
```diff
@@ -209,6 +209,7 @@ func getRingPolicy(keyRingName string) (*iam.Policy, error) {
 
 // [END kms_get_keyring_policy]
 
+// [START kms_get_cryptokey_policy]
```
tbpg: Please add newlines between the region tag and the start of the function comments. Otherwise, they can show up in the GoDoc for the function.

WalterHub: Done

PR #738, opened 2019-01-24
storage: add Bucket Policy Only samples

frankyn:
Hi Eno, could you review these samples?

### General information
PR to add GCS Bucket Policy Only samples for Go. Library change are not yet released and will be required before merging samples.


frankyn:
Thanks @enocom, I have updated the variable name. PTAL.

frankyn:
Thanks @enocom! I will merge when library is released.

PR #740, opened 2019-01-26
iot: add mqtt gateway snippets + test

hongalex:
Added samples for Gateway features including
- Having a gateway send data on behalf of a bound device
- Having a gateway listen for messages on a bound device's topic (state, config, errors)


hongalex:
I have excluded tests for `create_jwt` mostly because it doesn't test functionality for IoT Core and because we don't have it for other samples, but open to discuss this.

iot/mqttsnippets/attach_detach_device.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2018 Google LLC. All rights reserved.
```
gguuss: Nit, if it's a new file, 2019

iot/mqttsnippets/attach_detach_device.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+
+	"github.com/eclipse/paho.mqtt.golang"
+)
+
+// [START iot_attach_device]
+
+// attaches a devices to a gateway.
```
gguuss: attaches a device to a gateway

iot/mqttsnippets/attach_detach_device.go
```diff
@@ -0,0 +1,51 @@
...
+		return token.Error()
+	}
+	return nil
+}
+
+// [END iot_attach_device]
+
+// [START iot_detach_device]
+
+// detaches a devices to a gateway.
```
gguuss: detaches a device from a gateway

gguuss: Does this (and attach) return nil on success?

iot/mqttsnippets/gateway_send_data_test.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	client, err := cloudiot.New(httpClient)
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
+
+// [END iot_get_client]
```
gguuss: Any reason for the include code blocks in the test?

iot/mqttsnippets/attach_detach_device.go
```diff
@@ -0,0 +1,51 @@
...
+		return token.Error()
+	}
+	return nil
+}
+
+// [END iot_attach_device]
+
+// [START iot_detach_device]
+
+// detaches a device from a gateway.
```
tbpg:
```suggestion
// detachDevice detaches a device from a gateway.
```

iot/mqttsnippets/create_jwt.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"errors"
+	"io/ioutil"
+	"time"
+
+	jwt "github.com/dgrijalva/jwt-go"
+)
+
+// createJWT creates a Cloud IoT Core JWT for the given project id.
+// algorithm can be one of ["RSA256", "ES256"]
```
tbpg: Period at the end.

iot/mqttsnippets/create_jwt.go
```diff
@@ -0,0 +1,45 @@
...
+func createJWT(projectID string, privateKeyPath string, algorithm string, expiresMin int) (string, error) {
```
tbpg: Make expiresMin a `time.Duration` and use `time.Now().Add(expiration).Unix()`? Alternatively, make the expiration the `time.Time` it should expire?

iot/mqttsnippets/create_jwt.go
```diff
@@ -0,0 +1,45 @@
...
+
+	switch algorithm {
+	case "RS256":
+		privKey, _ := jwt.ParseRSAPrivateKeyFromPEM(keyBytes)
+		return token.SignedString(privKey)
+	case "ES256":
+		privKey, _ := jwt.ParseECPrivateKeyFromPEM(keyBytes)
+		return token.SignedString(privKey)
+	default:
+		return "", errors.New("Cannot find JWT algorithm. Specify 'ES256' or 'RS256'")
```
tbpg: This works, but I might move the final return out of the `switch` and drop the `default` branch to make it more clear that's what happens if we haven't returned yet.

iot/mqttsnippets/gateway_send_data.go
```diff
@@ -0,0 +1,116 @@
...
+	"io"
+	"math/rand"
+	"time"
+
+	"github.com/eclipse/paho.mqtt.golang"
+)
+
+// [START iot_send_data_from_bound_device]
+
+// Sends data on behalf of a bound device
```
tbpg:
```suggestion
// sendDataFromBoundDevice sends data on behalf of a bound device.
```

iot/mqttsnippets/gateway_send_data.go
```diff
@@ -0,0 +1,116 @@
...
+		fmt.Fprintln(w, "Failed to connect client")
+		return
+	}
+
+	if err := attachDevice(deviceID, client, ""); err != nil {
+		fmt.Fprintf(w, "Failed to attach device %s\n", err)
+		return
+	}
+
+	time.Sleep(5 * time.Second)
```
tbpg: Add a comment explaining the sleep?

iot/mqttsnippets/gateway_send_data.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Create and connect a client using the above options.
+	client := mqtt.NewClient(opts)
+	if token := client.Connect(); token.Wait() && token.Error() != nil {
+		fmt.Fprintln(w, "Failed to connect client")
+		return
+	}
+
+	if err := attachDevice(deviceID, client, ""); err != nil {
+		fmt.Fprintf(w, "Failed to attach device %s\n", err)
```
tbpg: Return an error?

iot/mqttsnippets/gateway_send_data.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	time.Sleep(5 * time.Second)
+
+	gatewayStateTopic := fmt.Sprintf("/devices/%s/state", gatewayID)
+	deviceStateTopic := fmt.Sprintf("/devices/%s/state", deviceID)
+
+	gatewayInitPayload := fmt.Sprintf("Starting gateway at time: %d", time.Now().Unix())
+	if token := client.Publish(gatewayStateTopic, 1, false, gatewayInitPayload); token.Wait() && token.Error() != nil {
+		fmt.Fprintln(w, "Failed to publish initial gateway payload")
```
tbpg: Return the error?

iot/mqttsnippets/gateway_send_data.go
```diff
@@ -0,0 +1,116 @@
...
+		return
+	}
+
+	for i := 1; i <= numMessages; i++ {
+		if shouldBackoff {
+			if backoffTime > maximumBackoffTime {
+				fmt.Fprintln(w, "Exceeded max backoff time.")
```
tbpg: Return error.

iot/mqttsnippets/gateway_send_data.go
```diff
@@ -0,0 +1,116 @@
...
+				return
+			}
+
+			waitTime := backoffTime + rand.Intn(1000)/1000.0
```
tbpg: I'm not familiar with backoff strategies - why is this added?

iot/mqttsnippets/gateway_send_data.go
```diff
@@ -0,0 +1,116 @@
...
+	deviceStateTopic := fmt.Sprintf("/devices/%s/state", deviceID)
+
+	gatewayInitPayload := fmt.Sprintf("Starting gateway at time: %d", time.Now().Unix())
+	if token := client.Publish(gatewayStateTopic, 1, false, gatewayInitPayload); token.Wait() && token.Error() != nil {
+		fmt.Fprintln(w, "Failed to publish initial gateway payload")
+		return
+	}
+
+	for i := 1; i <= numMessages; i++ {
+		if shouldBackoff {
```
tbpg: `shouldBackoff` feels weird to me. Would it be enough to check if `backoffTime > 0`?

iot/mqttsnippets/gateway_send_data.go
```diff
@@ -0,0 +1,116 @@
...
+			backoffTime *= 2
+			client = mqtt.NewClient(opts)
+		}
+
+		deviceStatePayload := fmt.Sprintf("%s, #%d", payload, i)
+
+		fmt.Fprintf(w, "Publishing message: %s to %s\n", deviceStatePayload, deviceStateTopic)
+
+		if token := client.Publish(deviceStateTopic, 1, false, payload); token.Wait() && token.Error() != nil {
+			fmt.Fprintln(w, "Failed to publish payload to device state topic")
```
tbpg: Return error.

iot/mqttsnippets/gateway_send_data_test.go
```diff
@@ -0,0 +1,284 @@
...
+	"golang.org/x/oauth2/google"
+	cloudiot "google.golang.org/api/cloudiot/v1"
+)
+
+var privateKeyRSA = os.Getenv("GOLANG_SAMPLES_IOT_PRIV")
+var pubKeyRSA = os.Getenv("GOLANG_SAMPLES_IOT_PUB")
+var region = "us-central1"
+
+// returns a v1 UUID for a resource: e.g. topic, registry, gateway, device
+func createIDForTest(resource string) string {
```
tbpg: This is in a test file. So, I don't think we need the `ForTest`.

iot/mqttsnippets/gateway_send_data_test.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/google/uuid"
+	"golang.org/x/oauth2/google"
+	cloudiot "google.golang.org/api/cloudiot/v1"
+)
+
+var privateKeyRSA = os.Getenv("GOLANG_SAMPLES_IOT_PRIV")
+var pubKeyRSA = os.Getenv("GOLANG_SAMPLES_IOT_PUB")
+var region = "us-central1"
+
+// returns a v1 UUID for a resource: e.g. topic, registry, gateway, device
```
tbpg:
```suggestion
// createID returns a v1 UUID for a resource: e.g. topic, registry, gateway, device.
```

iot/mqttsnippets/gateway_send_data_test.go
```diff
@@ -0,0 +1,284 @@
...
+func createIDForTest(resource string) string {
+	uuid, err := uuid.NewRandom()
+	if err != nil {
+		log.Fatalf("Could not generate uuid: %v", err)
+	}
+	id := fmt.Sprintf("golang-test-%s-%s", resource, uuid.String())
```
tbpg:
```suggestion
	return fmt.Sprintf("golang-test-%s-%s", resource, uuid.String())
```

iot/mqttsnippets/gateway_send_data_test.go
```diff
@@ -0,0 +1,284 @@
...
+	client, err := cloudiot.New(httpClient)
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
+
+// createRegistry creates a IoT Core device registry associated with a PubSub topic
+func createRegistry(w io.Writer, projectID string, region string, registryID string, topicName string) (*cloudiot.DeviceRegistry, error) {
```
tbpg: Are there samples for all of these helpers? Or they're in another package?

iot/mqttsnippets/gateway_send_data_test.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+func TestSendDataFromBoundDevice(t *testing.T) {
+	projectID := testutil.SystemTest(t).ProjectID
+
+	registryID := "golang-iot-test-registry"
+	gatewayID := createIDForTest("gateway")
+	deviceID := createIDForTest("device")
+
+	testutil.Retry(t, 1, 10*time.Second, func(r *testutil.R) {
```
tbpg: Might be better to have a retry for each part that could fail, rather than the end-to-end test. As is, if the last step fails, you have to redo all of the others.

iot/mqttsnippets/gateway_send_data_test.go
```diff
@@ -0,0 +1,284 @@
...
+		want := fmt.Sprintf("Publishing message: %s", "test")
+		if got := buf.String(); !strings.Contains(got, want) {
+			r.Errorf("SendDataFromBoundDevice got %s, want substring %q", got, want)
+		}
+	})
+
+	// cleanup
+	unbindDeviceFromGateway(ioutil.Discard, projectID, region, registryID, gatewayID, deviceID)
+	deleteDevice(ioutil.Discard, projectID, region, registryID, deviceID)
+	deleteDevice(ioutil.Discard, projectID, region, registryID, gatewayID)
```
tbpg: deleteDevice on a gateway? Confirming this is expected.

iot/mqttsnippets/gateway_send_data_test.go
```diff
@@ -0,0 +1,284 @@
...
+		sendDataFromBoundDevice(buf, projectID, region, registryID, gatewayID, deviceID, privateKeyRSA, "RS256", 2, "test")
+
+		want := fmt.Sprintf("Publishing message: %s", "test")
+		if got := buf.String(); !strings.Contains(got, want) {
+			r.Errorf("SendDataFromBoundDevice got %s, want substring %q", got, want)
+		}
+	})
+
+	// cleanup
+	unbindDeviceFromGateway(ioutil.Discard, projectID, region, registryID, gatewayID, deviceID)
```
tbpg: Might be good to have a retry for these, too.

iot/mqttsnippets/subscribe_gateway_device_topic.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+	"time"
+
+	"github.com/eclipse/paho.mqtt.golang"
+)
+
+// [START iot_subuscribe_config_bound_device]
```
tbpg: If it's in a single file, the region should include the imports.

iot/mqttsnippets/subscribe_gateway_device_topic.go
```diff
@@ -0,0 +1,98 @@
...
+
+// subscribeGatewayToDeviceTopic creates a gateway client that subscribes to a topic of a bound device.
+// Currently supported topics include: "config", "state", "commands", "errors"
+func subscribeGatewayToDeviceTopic(w io.Writer, projectID string, region string, registryID string, gatewayID string, deviceID string, privateKeyPath string, algorithm string, clientDuration int, topic string) {
```
tbpg: Should return an error.

iot/mqttsnippets/subscribe_gateway_device_topic_test.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func TestSubscribeGatewayToDeviceTopic(t *testing.T) {
+	projectID := testutil.SystemTest(t).ProjectID
+
+	registryID := "golang-iot-test-registry"
+	gatewayID := createIDForTest("gateway")
+	deviceID := createIDForTest("device")
+
+	testutil.Retry(t, 1, 10*time.Second, func(r *testutil.R) {
```
tbpg: Same retry comment.

iot/mqttsnippets/attach_detach_device.go
tbpg: Yes, returning `nil` when there is no error is idiomatic and indicates success.

iot/mqttsnippets/gateway_send_data_test.go
hongalex: Done.

iot/mqttsnippets/subscribe_gateway_device_topic_test.go
```diff
@@ -44,10 +44,10 @@ func TestSubscribeGatewayToDeviceTopic(t *testing.T) {
 	projectID := testutil.SystemTest(t).ProjectID
 
 	registryID := "golang-iot-test-registry"
-	gatewayID := createIDForTest("gateway")
-	deviceID := createIDForTest("device")
+	gatewayID := createID("gateway")
+	deviceID := createID("device")
 
-	testutil.Retry(t, 1, 10*time.Second, func(r *testutil.R) {
+	testutil.Retry(t, 5, 5*time.Second, func(r *testutil.R) {
```
tbpg: Sorry, I think I was unclear. Should all of these `create` calls be in separate `testutil.Retry` blocks? Try to create the gateway, then try to create device, then try to bind them. And so on. That way, if everything succeeds except binding the device, the test doesn't have to create the gateway/device again. Or maybe `createGateway` and `createDevice` do nothing if they already exist?

hongalex: So as it stands right now, you're right, trying to create a device that already exists will throw an error. I could rewrite the existing code to account for this, but I think wrapping each call in separate `testutil.Retry` probably makes more sense.

iot/mqttsnippets/gateway_send_data_test.go
tbpg: Could we put them in the same directory/package? I think the cost of having samples in the same directory is lower than copying the helpers. Feel free to disagree. It's OK the manager is `package main` - these could be, too.

hongalex: Potentially worried about scaling long term. If we continue to add samples at the rate of 1 file per sample, this directory could get very big very fast. Willing to entertain the idea though since it decreases duplication, but I think @broady @gguuss might have more ideas.

gguuss: Anecdotally, we have put one snippet per file in PHP; there are a good number of files already but it's not unmaintainable. If we wanted to coalesce the snippets to each feature, that's one way we could group them and reduce the file count.  Longer term, I want to ship e2e demos showing the features, which could be put in the main package.

hongalex: I'd be willing to move to a one directory snippet, but as of right now, the refactor into one directory would cause a lot of issues. Basically it would involve updating the docs to point to a different directory (since I can't just move MQTT samples into the `manager` directory as well as fixing the tests as well. Also, it would make this specific PR very large since there will be a good amount of renaming files and such. I would prefer if we pushed this to a later time and keep it the way it is for now.

hongalex: Also, having a single directory for device management and MQTT samples would be hard for pages like this: https://cloud.google.com/iot/docs/samples/mqtt-samples which have links that point to a specific directory.

tbpg: Got it. Yeah, if the directory is linked from a doc, it's probably good to keep it clean.

PR #744, opened 2019-01-30
appengine: add import region tag

llatif:
Added region tags for partial code imports into documentation

### DIFF
--- a/appengine/go11x/static/main.go
+++ b/appengine/go11x/static/main.go
@@ -64,4 +64,5 @@ func indexHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "Internal server error", http.StatusInternalServerError)
 	}
 }
+
 // [END handlers]
--- a/appengine/go11x/static/main.go
+++ b/appengine/go11x/static/main.go
@@ -20,6 +20,7 @@ var (
 	)
 )
 
+// [START main]
 func main() {
 	http.HandleFunc("/", indexHandler)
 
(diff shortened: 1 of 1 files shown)


appengine/go11x/static/main.go
```diff
@@ -39,7 +40,9 @@ func main() {
 	log.Printf("Listening on port %s", port)
 	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
 }
+//[END main]
```
grayside: Need an extra newline between the closing brace and the end comment.

appengine/go11x/static/main.go
```diff
@@ -61,3 +64,4 @@ func indexHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "Internal server error", http.StatusInternalServerError)
 	}
 }
+// [END handlers]
```
grayside: Need an extra newline between the closing brace and the end comment.

appengine/go11x/static/main.go
```diff
@@ -20,6 +20,7 @@ var (
 	)
 )
 
+// [START main]
```
tbpg: The region tag should start with the product name and be in the samples tracker, which requires that. Otherwise, the sample will be untracked. If you let me know the updated name, I can update it here, if you prefer.

tbpg:
Going to close this for now, but feel free to reopen or ping me if you need help getting the region tag set up.

PR #745, opened 2019-01-30
functions: correct project ID environment variable

stewblr:
GOOGLE_CLOUD_PROJECT is not set automatically for Cloud Functions. GCP_PROJECT is set automatically.


tbpg:
Would you mind updating functions/helloworld/hello_pubsub_system_test.go, too?

PR #746, opened 2019-01-30
vision: correct detection tests

shollyman:
Two issues:

Changes to detection may cause inconsistencies in the capitalization
for detected labels.  This relaxes comparisons to be case insensitive.

Changes in the backend of logo detection are causing failures with the
sample logo (internal issue 123358697).  This change corrects the local
file, but tests will still fail until the test data in the GCS test
bucket is updated in similar fashion.

gs://python-docs-samples-tests/vision/logos.png is the artifact needing
update out of band.


tbpg:
CLA reason: both authors are Googlers.

PR #754, opened 2019-02-07
iam: add quickstart and doc snippets

awkoren:
Add code snippets for IAM docs: https://cloud.google.com/iam/docs/


iam/quickstart/quickstart.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func main() {
+	// Get credentials
+	client, _ := google.DefaultClient(
+		context.Background(),
+		iam.CloudPlatformScope)
+
+	// Create the Cloud IAM service object
+	iamService, _ := iam.New(client)
```
tbpg:
```suggestion
	iamService, err := iam.New(client)
	if err != nil {
		log.Fatalf("iam.New: %v", err)
	}
```

iam/quickstart/quickstart.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// Call the Cloud IAM Roles API
+	response, _ := iamService.Roles.List().Do()
```
tbpg:
```suggestion
	resp, err := iamService.Roles.List().Do()
	if err != nil {
		log.Fatalf("List.Do: %v", err)
	}
```

iam/quickstart/quickstart.go
```diff
@@ -0,0 +1,39 @@
...
+	roles := response.Roles
+
+	// Process the response
+	for _, role := range roles {
```
tbpg:
```suggestion
	for _, role := range resp.Roles {
```

Can delete the `roles` variable.

iam/quickstart/quickstart.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"fmt"
+
+	"golang.org/x/net/context"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/iam/v1"
+)
+
+func main() {
+	// Get credentials
```
tbpg: Please add periods at the end of comments.

iam/snippets/snippets.go
```diff
@@ -0,0 +1,232 @@
+// Copyright 2018 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package snippets
+
+import (
+	"fmt"
+
+	"golang.org/x/net/context"
```
tbpg:
```suggestion
	"context"
```

iam/snippets/snippets.go
```diff
@@ -0,0 +1,232 @@
...
+	"golang.org/x/oauth2/google"
+	iam "google.golang.org/api/iam/v1"
+)
+
+var client, _ = google.DefaultClient(
```
tbpg: Same comment (line length and error).

iam/snippets/snippets.go
```diff
@@ -0,0 +1,232 @@
...
+	context.Background(),
+	iam.CloudPlatformScope)
+var iamService, _ = iam.New(client)
+
+// [START iam_view_grantable_roles]
+func viewGrantableRoles(fullResourceName string) {
+	request := iam.QueryGrantableRolesRequest{FullResourceName: fullResourceName}
+	response, _ := iamService.Roles.QueryGrantableRoles(&request).Do()
```
tbpg: Each snippet should create the service.

iam/snippets/snippets.go
```diff
@@ -0,0 +1,232 @@
...
+	iam "google.golang.org/api/iam/v1"
+)
+
+var client, _ = google.DefaultClient(
+	context.Background(),
+	iam.CloudPlatformScope)
+var iamService, _ = iam.New(client)
+
+// [START iam_view_grantable_roles]
+func viewGrantableRoles(fullResourceName string) {
```
tbpg: We usually pass the project ID as an argument. It's a bit easier to test and it shows more clearly how to construct the full resource name.

tbpg: Add an error return value then check errors below, adding context with `fmt.Errorf` as needed.

tbpg: Please add a comment with what this function does, starting with the function name. There should be a newline between the region tag and the comment so the tag doesn't show up in GoDoc.

iam/snippets/snippets.go
```diff
@@ -0,0 +1,232 @@
...
+		fmt.Println("Key: ", key.Name)
+	}
+	return response.Keys
+}
+
+// [END iam_list_keys]
+
+// [START iam_delete_key]
+func deleteKey(fullKeyName string) {
+	iamService.Projects.ServiceAccounts.Keys.Delete(fullKeyName).Do()
```
tbpg: Check for errors.

iam/snippets/snippets.go
```diff
@@ -0,0 +1,232 @@
...
+	fmt.Println("Created key: ", key.Name)
+	return *key
+}
+
+// [END iam_create_key]
+
+// [START iam_list_keys]
+func listKeys(serviceAccountEmail string) []*iam.ServiceAccountKey {
+	resource := "projects/-/serviceAccounts/" + serviceAccountEmail
+	response, _ := iamService.Projects.ServiceAccounts.Keys.List(
```
tbpg: Single line + errors.

iam/snippets/snippets_test.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"math/rand"
+	"os"
+	"strconv"
+	"testing"
+	"time"
+)
+
+func TestViewGrantableRoles(t *testing.T) {
+	projectID := os.Getenv("GOLANG_SAMPLES_PROJECT_ID")
```
tbpg: Please use `testutil` to get the project ID. Example: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/monitoring/snippets/channel_delete_test.go#L22

iam/snippets/snippets.go
```diff
@@ -0,0 +1,232 @@
+// Copyright 2018 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package snippets
```
tbpg: We've started writing each sample in separate files so the region tag can include the import statements. Might want to proactively do it here so we don't have to update later. For example, see https://github.com/GoogleCloudPlatform/golang-samples/pull/706.

iam/snippets/snippets_test.go
```diff
@@ -0,0 +1,53 @@
...
+	resource := "//cloudresourcemanager.googleapis.com/projects/" + projectID
+	viewGrantableRoles(resource)
```
tbpg: Pass in a buffer then check the output includes what you expect. Example: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/monitoring/snippets/channel_delete_test.go#L27

iam/snippets/snippets.go
tbpg: Please have `w io.Writer` as the first argument then write to it with `fmt.Fprintln(w, ...)`. This makes it easier to test, doesn't lead to printing things out when running tests, and is the same as writing to an `http.ResponseWriter` in an HTTP handler. Example: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/monitoring/snippets/channel_delete.go#L21

iam/snippets/snippets_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func TestQueryTestablePermissions(t *testing.T) {
+	projectID := os.Getenv("GOLANG_SAMPLES_PROJECT_ID")
+	resource := "//cloudresourcemanager.googleapis.com/projects/" + projectID
+	queryTestablePermissions(resource)
+}
+
+func TestRoles(t *testing.T) {
+	rand.Seed(time.Now().UnixNano())
+	roleName := "gotest" + strconv.Itoa(rand.Intn(100000))
```
tbpg: Does the role name need to be globally unique? Can this reuse the same role name? Trying to insert randomness has unfortunately lead to flaky tests in the past. So, we try to make it non-random. If we do need a random name, please use the `gofrs/uuid` package.

iam/snippets/snippets_test.go
tbpg: Check no error is returned.

iam/snippets/snippets.go
awkoren: This pattern deviates from the rest of snippet in the docs. For example, see: https://cloud.google.com/iam/docs/creating-managing-service-accounts#iam-service-accounts-create-csharp
Is there a reason why you're moving towards this approach?

iam/snippets/snippets.go
awkoren: What if the resource isn't a project?

iam/snippets/snippets.go
awkoren: This makes the snippets long and repetitive, especially in docs with lots of snippets demonstrating very similar tasks. And as I said earlier, it deviates from the approach used by every other language.

iam/snippets/snippets.go
awkoren: It might be easier to test, but it's harder for users to read and copy/paste from the docs. User friendliness is more important than ease of testing.

iam/snippets/snippets_test.go
awkoren: It needs to be unique because roles persist 30 days in the system even after being deleted, so you can't re-use the same role name. Changed to use a GUID instead.

iam/snippets/snippets_test.go
awkoren: This reduces the usability of the snippet for the customer.

awkoren: Added error checking to all calls w/ log.Fatalf, which will cause the test to fail if the call fails.

iam/quickstart/quickstart.go
```diff
@@ -8,38 +8,39 @@
 package main
 
 import (
-	"fmt"
+	"log"
```
tbpg: Please remove this newline.

iam/snippets/snippets.go
tbpg: We want samples to be idiomatic for every language. That differs a bit for each language when it comes to regions inside/outside the function declaration, imports inside or outside (or can they be directly in the function), etc. Including the imports in region seems the most clear for Go.

Not a requirement for now.

iam/snippets/snippets.go
tbpg: Looks like this still needs to be updated.

iam/snippets/snippets.go
tbpg: If that's part of the documentation, sounds good.

tbpg: There are lots of options for how to write samples, and this is the best all-around way we've come up with for writing them. Using `fmt.Fprintf` is common and makes this very easy to convert to an HTTP handler.

iam/snippets/snippets_test.go
tbpg: `log.Fatal` does not work well with tests. It stops test execution completely and will not run anymore tests.

Returning errors is idiomatic Go. Calling `log.Fatal` in a library function is not.

iam/snippets/snippets.go
awkoren: Yes, the docs explain what fullResourceName should be.

iam/snippets/snippets_test.go
awkoren: Passed in buffer. For testing, I just wrote tests against the response directly.

iam/snippets/snippets_test.go
awkoren: This gives an error because I'm working in a fork and can't reference the internal main package in the main repo. I don't want to reference my fork because that'll be wrong once the PR is merged. What's a good workaround?

awkoren:
Refactored to follow the new guidance. Think I addressed all issues, except the testutil one -- not sure how to resolve that.
Let me know if you have more feedback.

iam/snippets/roles_undelete.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	"golang.org/x/oauth2/google"
+	iam "google.golang.org/api/iam/v1"
+)
+
+// undeleteRole restores a deleted custom role.
+func undeleteRole(w io.Writer, name string, projectID string) (*iam.Role, error) {
```
tbpg: Might want to put `projectID` first and document what `name` is. Should it be `fullResourceName` (and the code updated)?

iam/snippets/roles_view_grantable.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	"golang.org/x/oauth2/google"
+	iam "google.golang.org/api/iam/v1"
+)
+
+// viewGrantableRoles lists roles grantable on a resource.
+func viewGrantableRoles(w io.Writer, fullResourceName string) ([]*iam.Role, error) {
```
tbpg: Might be helpful to document what `fullResourceName` looks like.

iam/snippets/service_accounts_create.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"io"
+
+	"golang.org/x/oauth2/google"
+	iam "google.golang.org/api/iam/v1"
+)
+
+// createServiceAccount creates a service account.
+func createServiceAccount(w io.Writer, projectID string, name string,
+	displayName string) (*iam.ServiceAccount, error) {
```
tbpg: One line.

iam/snippets/snippets_test.go
tbpg: You should have a single repo locally and a different remote for your fork/the origin. See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#contributing. Then, you'll be able to import the `testutil` package.

iam/snippets/roles_view_grantable.go
awkoren: This is in the documentation where the snippet will live: https://cloud.google.com/iam/docs/viewing-grantable-roles

iam/snippets/roles_undelete.go
awkoren: Swapped vars. Throughout the docs, we use "name" when we mean local name, and "fullResourceName" when the whole thing is required. I don't think this will confuse customers.

awkoren:
Review feedback addressed.

PR #755, opened 2019-02-07
testutil:  add a multiple string tester

shollyman:
For your consideration.  I think it would be helpful to have some kind of standard substring testing for the non-deterministic APIs, based on recent failures in vision and speech testing.

### DIFF
--- a/internal/testutil/multistringtest.go
+++ b/internal/testutil/multistringtest.go
@@ -1,0 +1,44 @@
+// Copyright 2019 Google Inc. All rights reserved.
+// Use of this source code is governed by the Apache 2.0
+// license that can be found in the LICENSE file.
+
+package testutil
+
+import (
+	"fmt"
+	"strings"
+)
+
+// MultiStringTest provides a way for testing many substrings in a response.
+// By setting a set of expected substrings, and then indicating a nonzero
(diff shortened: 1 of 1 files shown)


internal/testutil/multistringtest.go
```diff
@@ -0,0 +1,44 @@
...
+// Intended use is for checking the responses from non-deterministic sources,
+// such as ML-based APIs where exact string matching is flaky/problematic.
+type MultiStringTest struct {
+	Expected []string
+	MinPass  int
+	MinFail  int
+}
+
+// Evaluate returns an error if either the pass or failure rate is nonzero and unmet.
+func (mst *MultiStringTest) Evaluate(got string) (passed, failed []string, err error) {
```
tbpg: I think I'd prefer a simpler function that only checks a single string and doesn't do any minimum logic.

The speech test would have failed regardless of if this function were used. I think the solution there was a shorter test string. What was the vision failure, again? If I recall correctly, I'm not sure this would have prevented that, either.

This has definitely come up, though. So, it could be nice to have some sort of standard function to do string contains. Would make it easier for new contributors to get it right. To make it easy to use in a test, it should probably only return a bool.

PR #762, opened 2019-02-20
bigtable: add region tags

billyjacobson:
Adding region tags for Cloud Bigtable samples

### DIFF
--- a/bigtable/helloworld/main.go
+++ b/bigtable/helloworld/main.go
@@ -16,6 +16,7 @@
 // library to perform basic CRUD operations
 package main
 
+// [START bigtable_hw_imports]
 import (
 	"context"
 	"flag"
@@ -24,6 +25,7 @@ import (
 
 	"cloud.google.com/go/bigtable"
 )
+// [END bigtable_hw_imports]
 
 // User-provided constants.
 const (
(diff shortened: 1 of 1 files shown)


bigtable/helloworld/main.go
```diff
@@ -86,14 +91,18 @@ func main() {
 			log.Fatalf("Could not create column family %s: %v", columnFamilyName, err)
 		}
 	}
+	// [END bigtable_hw_create_table]
 
 	// Set up Bigtable data operations client.
 	// NewClient uses Application Default Credentials to authenticate.
+	// [START bigtable_hw_connect_data]
```
broady: Do you not want the preceding comments? Seems useful!

bigtable/helloworld/main.go
```diff
@@ -59,11 +61,14 @@ func main() {
 
 	// Set up admin client, tables, and column families.
 	// NewAdminClient uses Application Default Credentials to authenticate.
+	// [START bigtable_hw_connect]
```
broady: Same here.

bigtable/helloworld/main.go
billyjacobson: The docs already have comments about it, so it seems easier for them to be maintained there. Same for below. Thoughts? 

https://cloud.google.com/bigtable/docs/samples-go-hello

bigtable/helloworld/main.go
```diff
@@ -24,6 +25,7 @@ import (
 
 	"cloud.google.com/go/bigtable"
 )
+// [END bigtable_hw_imports]
```
broady: Need to add a newline before this to make `gofmt` happy.

bigtable/helloworld/main.go
broady: SGTM!

PR #763, opened 2019-02-21
storage/gcsupload: public flag does not work

junichi-tanaka:
The public flag in storage/gcsupload.go does not work due to using the fixed value.
So, the uploaded object can always be accessed from public.


junichi-tanaka:
I signed it!

PR #764, opened 2019-02-24
functions/helloworld: update logging messages

tbpg:
Stderr is usually used for normal logs, not just errors. This PR updates the logged messages to say where they go, rather than sample content.

We could also change `logger` to `errLogger`, but "normal" logging goes to stderr in Go, so I'm not sure we want to change it.

See b/125910326.


functions/helloworld/hello_logging.go
```diff
@@ -31,8 +31,8 @@ var (
 
 // HelloLogging logs messages.
 func HelloLogging(w http.ResponseWriter, r *http.Request) {
-	stdLogger.Println("I am a log entry!")
-	logger.Println("I am an error!")
+	stdLogger.Println("This is sent to os.Stdout")
+	logger.Println("This is sent to os.Stderr")
```
broady: maybe outLogger and errLogger?

broady: or ...

log.Println("This is stderr")
fmt.Println("This is stdout")

log format with timestamps doesn't really matter for logging because the logging service attaches its own timestamp

tbpg: Done.

PR #767, opened 2019-02-26
endpoints: fix sample when both api-key and token are specified

basharal:
Existing code is broken where it would attempt to create a context
inside a context as opposed to appending both keys to the context.
Therefore, when both api-key and token are specified, only token
would take effect and not api-key.

### DIFF
--- a/endpoints/getting-started-grpc/client/main.go
+++ b/endpoints/getting-started-grpc/client/main.go
@@ -95,12 +95,12 @@ func main() {
 	}
 	if *token != "" {
 		log.Printf("Using authentication token: %s", *token)
-                _, ok := metadata.FromOutgoingContext(ctx)
+		_, ok := metadata.FromOutgoingContext(ctx)
 		if ok {
-                  ctx = metadata.AppendToOutgoingContext(ctx, "Authorization", fmt.Sprintf("Bearer %s", *token))
-                } else {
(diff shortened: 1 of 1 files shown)


endpoints/getting-started-grpc/client/main.go
```diff
@@ -95,7 +95,12 @@ func main() {
 	}
 	if *token != "" {
 		log.Printf("Using authentication token: %s", *token)
-		ctx = metadata.NewOutgoingContext(ctx, metadata.Pairs("Authorization", fmt.Sprintf("Bearer %s", *token)))
+		_, ok := metadata.FromOutgoingContext(ctx)
+		if ok {
+			ctx = metadata.AppendToOutgoingContext(ctx, "Authorization", fmt.Sprintf("Bearer %s", *token))
```
tbpg: From reading https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#sending-metadata, we should use `AppendToOutgoingContext` every time, even when there is no metadata yet.

> This can be used with or without existing metadata on the context. When there is no prior metadata, metadata is added; when metadata already exists on the context, kv pairs are merged in.

basharal: Verified that 
AppendToOutgoingContext works as you mentioned.

PR #771, opened 2019-03-04
clarification on serving static files via appengine standard and moving to runtime go111

superwhykz:
It took a bit effort to find out it,  just added a bit of description.


superwhykz:
I signed it!

tbpg:
Thank you for the PR! Unfortunately, I don't think we're ready to make this change yet. It will need to be done for all steps of the tutorial and the documentation surrounding it will need to be updated at the same time. We'll also probably wait for `go111` to be GA.

superwhykz:
@tbpg I see... make sense.

PR #784, opened 2019-03-07
all: clean up unused and bad files, add test

broady:
I noticed a binary was accidentally checked in, so thought a test to
catch these types of files might be useful.

I started with mime type checking but went to primarily a whitelist of
extensions and paths. MIME type might still be useful to catch really
bad stuff (like Go binaries, from running "go build").


broady:
I don't think key.json.enc is used anymore, it was from days of using Travis.

tbpg:
Merged, but realizing now the test wasn't included - need to add?

broady:
Oops.

PR #790, opened 2019-03-08
all: add test for bad files

broady:
Forgot to add the actual test in PR #784.

I noticed a binary was accidentally checked in, so thought a test to
catch these types of files might be useful.

I started with mime type checking but went to primarily a whitelist of
extensions and paths. MIME type might still be useful to catch really
bad stuff (like Go binaries, from running "go build").


tbpg:
Unrelated test failures. Merging.

PR #796, opened 2019-03-13
docs(samples): Update datastore_transactional_get_or_create to check that err != nil

BenWhitehead:


broady:
Sample looks fine before.

"get *or* create"

after this change, the entity would be re-written even if it already exists.

PR #799, opened 2019-03-15
tasks: add HTTP push queue sample

averikitsch:
Test projects will have to be granted permission for HTTP targets or wait until whitelisting is off.


appengine/go11x/tasks/create_http_task/create_http_task.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Command create_task constructs and adds a task to an App Engine Queue.
+package main
+
+import (
+	"context"
+	"fmt"
+
+	cloudtasks "cloud.google.com/go/cloudtasks/apiv2beta3"
+	taskspb "google.golang.org/genproto/googleapis/cloud/tasks/v2beta3"
```
grayside: Still using beta3 API.

appengine/go11x/tasks/create_http_task/create_http_task.go
```diff
@@ -0,0 +1,71 @@
...
+	// Build the Task payload.
+	// https://godoc.org/google.golang.org/genproto/googleapis/cloud/tasks/v2beta3#CreateTaskRequest
+	req := &taskspb.CreateTaskRequest{
+		Parent: queuePath,
+		Task: &taskspb.Task{
+			// https://godoc.org/google.golang.org/genproto/googleapis/cloud/tasks/v2beta3#HttpRequest
+			PayloadType: &taskspb.Task_HttpRequest{
+				HttpRequest: &taskspb.HttpRequest{
+					HttpMethod:  taskspb.HttpMethod_POST,
+					Url: url,
```
grayside: Curiosity: Can you set a default URL for the queue or is a URL per task required?

appengine/go11x/tasks/create_http_task/create_http_task_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+	"testing"
+)
+
+func TestCreateHttpTask(t *testing.T) {
+	tc := testutil.SystemTest(t)
+	locationID := "us-central1"
+	queueID := "my-appengine-queue"
```
grayside: Given this is not a queue for appengine, should it still target `my-appengine-queue`? Maybe we should have a second queue to test HTTP Push. Not sure if test concurrency could be a concern, maybe tbp@ has insight there?

appengine/go11x/tasks/create_http_task/create_http_task.go
averikitsch: Per task is needed.

appengine/go11x/tasks/create_http_task/create_http_task_test.go
averikitsch: Queues can handle both app engine and http routing. We can create new queue with a better name if this is confusing.

tbpg: Not sure I understand - does this need to be App Engine? I noticed the `.appspot.com/task_handler` URL below - we probably shouldn't depend on a certain App Engine app being deployed when this test runs. We could deploy it as part of the test, but if it's possible to run the test without that would be even better.

If it's not related to App Engine, could this be in a completely separate directory, say `golang-samples/tasks`?

appengine/go11x/tasks/create_http_task/create_http_task.go
```diff
@@ -0,0 +1,69 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START cloud_tasks_create_http_task]
+
+// Command create_task constructs and adds a task to an App Engine Queue.
+package main
```
tbpg: Generally, snippets should not be in a package main. From the sample rubric, snippets should not be runnable unless they're a quickstart because we've had issues maintaining them. Tests should ensure the snippet works.

appengine/go11x/tasks/create_http_task/create_http_task.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"context"
+	"fmt"
+
+	cloudtasks "cloud.google.com/go/cloudtasks/apiv2beta3"
+	taskspb "google.golang.org/genproto/googleapis/cloud/tasks/v2beta3"
+)
+
+// createTask creates a new task in your App Engine queue.
+func createHttpTask(projectID, locationID, queueID, url, message string) (*taskspb.Task, error) {
```
tbpg: s/Http/HTTP/g

appengine/go11x/tasks/create_http_task/create_http_task.go
```diff
@@ -0,0 +1,69 @@
...
+					HttpMethod: taskspb.HttpMethod_POST,
+					Url:        url,
+				},
+			},
+		},
+	}
+
+	// Add a payload message if one is present.
+	if message != "" {
+		req.Task.GetHttpRequest().Body = []byte(message)
```
tbpg: What happens if `Body` is set to an empty `[]byte`? Curious if it would still work.

appengine/go11x/tasks/create_http_task/create_http_task_test.go
```diff
@@ -0,0 +1,49 @@
...
+			message: "task details for handler processing",
+		},
+		{
+			name:    "No Message",
+			message: "",
+		},
+	}
+
+	for _, test := range tests {
+		_, err := createHttpTask(tc.ProjectID, locationID, queueID, url, test.message)
```
tbpg: Any sort of cleanup we should do for the task?

appengine/go11x/tasks/create_http_task/create_http_task.go
averikitsch: That's the correct version for this feature.

appengine/go11x/tasks/create_http_task/create_http_task.go
averikitsch: This is a quickstart, but if it wouldn't be confusing to run it as a test I'd be open to working with someone to change it over.

appengine/go11x/tasks/create_http_task/create_http_task_test.go
averikitsch: This sample first needed an app engine task handler and the old task queues were tied to app engine hence why it is in the appengine folder. It can be moved if necessary. The url is a placeholder and it doesn't matter for the test to pass. The URL currently set so it mimics the sample url.

appengine/go11x/tasks/create_http_task/create_http_task_test.go
averikitsch: We can implement delete task to clean up the created task.

appengine/go11x/tasks/create_http_task/create_http_task.go
averikitsch: Empty payload is fine. Updated code to reflect this.

appengine/go11x/tasks/create_http_task/create_http_task.go
averikitsch: done.

appengine/go11x/tasks/create_http_task/create_http_task_test.go
averikitsch: This would be good as an additional sample (not just a test). I currently don't have time to implement.

grayside:
The existing sample is focused on Cloud Tasks with individual queues tied to an AppEngine service. In preparation for the launch of HTTP Targets Beta, we need samples where the task processing hits an arbitrary (presumably verified) URL.

My request for changes is tangled up with points 1 & 2 below. Point 3 is a useful clarification to be made.

1. Will Cloud Tasks need to keep the existing App Engine integration samples, even if only to facilitate migration for apps currently on first-gen runtimes? If so, we should keep existing samples under appengine/go11x.
2. A sample demonstrating HTTP Targets should be under /tasks, as that feature does not have an appengine constraint.
3. If we are keeping the existing App Engine handler sample, we should keep my-appengine-queue specifically for it. This queue resources is pre-created because when the sample was first created it could take 1-2 minutes to create the queue. Not sure if that is still the case. Pre-created queue for HTTP Target system tests should not reference appengine to avoid ambiguity, as the feature only integrates via URL, not via hosting provider.

averikitsch:
Note: I am about to push another sample that has an authentication token for HTTP requests. 

1. Can you explain what you mean here? We will need to keep the existing quickstart with the handler and create_task sample for App Engine Task queues. 
At first, this seemed like this needed to be a "quickstart", but the App Engine quickstart may be enough.
2. I agree that HTTP tasks should go under /tasks, but should the App Engine quickstart move as well?
3. I agree that this queue may be poorly named, but the queue needs to be whitelisted for the tests to pass and the queue can handle all types of targets. I can add a bug for myself to correct this when whitelisting is off.

averikitsch:
I moved the samples. The tasks samples are passing.

averikitsch:
ping @tbpg (Sorry to bug you again, but trying to get this done by the EOW)

tasks/create_http_task.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	cloudtasks "cloud.google.com/go/cloudtasks/apiv2beta3"
+	taskspb "google.golang.org/genproto/googleapis/cloud/tasks/v2beta3"
+)
+
+// createTask creates a new task in your App Engine queue.
```
tbpg: Creates a new Cloud Task?

tasks/create_http_task.go
```diff
@@ -0,0 +1,67 @@
...
+func createHTTPTask(projectID, locationID, queueID, url, message string) (*taskspb.Task, error) {
```
tbpg: Can we move any of these arguments to be variables in the function? With the new canonical sample format, we're trying to keep args to a minimum. For example,

```
	locationID := "us-central1"
	queueID := "my-queue"
	url := "https://example.com/task_handler"
```

tasks/create_http_task.go
```diff
@@ -0,0 +1,67 @@
...
+				HttpRequest: &taskspb.HttpRequest{
+					HttpMethod: taskspb.HttpMethod_POST,
+					Url:        url,
+				},
+			},
+		},
+	}
+
+	// Add a payload message if one is present.
+	req.Task.GetHttpRequest().Body = []byte(message)
```
tbpg: Can do this directly when creating `req`. Same with the other sample.

```go
	req := &taskspb.CreateTaskRequest{
		Parent: queuePath,
		Task: &taskspb.Task{
			// https://godoc.org/google.golang.org/genproto/googleapis/cloud/tasks/v2beta3#HttpRequest
			PayloadType: &taskspb.Task_HttpRequest{
				HttpRequest: &taskspb.HttpRequest{
					Body: []byte(message),
					HttpMethod: taskspb.HttpMethod_POST,
					Url:        url,
				},
			},
		},
	}
```

tasks/create_http_task.go
grayside: This is following the pattern of the existing snippet. Can refactoring the basic pattern be done in a follow-up?

tbpg: SGTM.

tasks/create_http_task.go
grayside: This comes from the original snippet for appengine queues. While it **can** be done directly, identifying how to modify message is tricky, so I kept this separate as an alternative to adding documentation or explicit samples for that use case.

I'd rather not merge this into one statement without a plan forward on improving discoverability of `req.Task.GetHttpRequest().Body = []byte(message)`.

averikitsch:
Can this get merged?

PR #804, opened 2019-03-22
trace: call Body.Close to end the client HTTP span in OpenCensus

bogdandrutu:
The  client span.End() will be invoked after a read from resp.Body returns io.EOF or when resp.Body.Close() is invoked.


broady:
oh, wow, the span doesn't end if the response body isn't closed?

bogdandrutu:
@broady done.

PR #811, opened 2019-03-29
storage: add v4 signed URL get and put samples

frankyn:
Hi @kurtisvg,

Could you help review these samples?

They're blocked until the Go library is released.


storage/objects/main.go
```diff
@@ -424,6 +426,75 @@ func downloadUsingRequesterPays(client *storage.Client, object, bucketName, loca
 	return nil
 }
 
+func generateV4GetObjectSignedUrl(client *storage.Client, bucketName, objectName, serviceAccount string) (string, error) {
+	// [START storage_generate_signed_url_v4]
```
tbpg: Might be better to make new samples follow the new canonical sample format, putting these in their own file and having the region tag include imports. See #650.

storage/objects/main.go
```diff
@@ -424,6 +426,75 @@ func downloadUsingRequesterPays(client *storage.Client, object, bucketName, loca
 	return nil
 }
 
+func generateV4GetObjectSignedUrl(client *storage.Client, bucketName, objectName, serviceAccount string) (string, error) {
```
tbpg: Each function should create its own client, rather than have it passed as an argument. It might be better to stay consistent within the same set of samples, but I'm not sure.

storage/objects/main.go
```diff
@@ -424,6 +426,75 @@ func downloadUsingRequesterPays(client *storage.Client, object, bucketName, loca
...
+		PrivateKey:     conf.PrivateKey,
+		Expires:        time.Now().Add(15*time.Minute),
+	}
+
+	u, err := storage.SignedURL(bucketName, objectName, opts)
+	if err != nil {
+		return "", fmt.Errorf("Unable to generate a signed URL: %v", err)
+	}
+
+	fmt.Printf("Generated GET signed URL:\n");
```
tbpg: Please pass an `io.Writer` as an argument and print to it like this:
```
fmt.Fprintln(w, "Generated GET signed URL:")
```

tbpg: No semicolons needed at the ends of the lines. :)

storage/objects/main_test.go
```diff
@@ -224,6 +225,60 @@ func TestKMSObjects(t *testing.T) {
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var (
+		bucketName = tc.ProjectID + "-signed-url-bucket-name"
+
+		objectName = "foo.txt"
+
+		serviceAccount = os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
```
tbpg: Can use the shorter variable declaration syntax and get rid of the `var()` wrapper and newlines.

```
serviceAccount := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
```

storage/objects/main_test.go
```diff
@@ -224,6 +225,60 @@ func TestKMSObjects(t *testing.T) {
 	}
 }
 
+func TestV4SignedUrl(t *testing.T) {
+	tc := testutil.SystemTest(t)
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatal(err)
```
tbpg:
```suggestion
		r.Fatalf("storage.NewClient: %v", err)
```

`log.Fatal` will stop the test suite. `t.Fatal` fails the test (immediately).

storage/objects/main_test.go
```diff
@@ -224,6 +225,60 @@ func TestKMSObjects(t *testing.T) {
...
+		t.Errorf("can't generate a put signed url: %v", err)
+	}
+
+	httpClient := &http.Client{}
+	request, err := http.NewRequest("PUT", putUrl, strings.NewReader("hello world"))
+	request.ContentLength = 11
+	request.Header.Set("Content-Type", "application/octet-stream")
+	response, err := httpClient.Do(request)
+	if err != nil {
+		t.Errorf("Err: %v", err)
```
tbpg: Usually want to identify the function/test that failed in the error message.

storage/objects/main_test.go
```diff
@@ -224,6 +225,60 @@ func TestKMSObjects(t *testing.T) {
...
+
+	cleanBucket(t, ctx, client, tc.ProjectID, bucketName)
+
+	putUrl, err := generateV4PutObjectSignedUrl(client, bucketName, objectName, serviceAccount)
+	if err != nil {
+		t.Errorf("can't generate a put signed url: %v", err)
+	}
+
+	httpClient := &http.Client{}
+	request, err := http.NewRequest("PUT", putUrl, strings.NewReader("hello world"))
```
tbpg: Maybe create a variable for the input string so it can be used for `ContentLength` and `want` below?

storage/objects/main.go
frankyn: I'd rather keep the existing format without changing the format unless it's required.

storage/objects/main_test.go
frankyn: I'm following the existing code, should this be updated elsewhere as well?

storage/objects/main.go
tbpg: See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#initialize-clients-and-services-in-every-sample (new canonical sample style guide).

Eventually, all samples should match this style. But, if you want to be internally consistent for these samples, that works too.

storage/objects/main_test.go
tbpg: Yes, please, if you don't mind!

storage/objects/main_test.go
frankyn: It's the way samples are designed right now. I like the idea of showing at least one concrete example instead of making it generic.

storage/objects/main.go
frankyn: Thanks for clarifying! 

I want to follow the format that exists in these samples to be consistent.

storage/objects/main.go
frankyn: lol... I can Go, I promise! Thanks for catching those.

frankyn:
Thanks @tbpg, I have updated the PR based on your comments. PTAL.

storage/objects/main.go
tbpg: Mind updating this? Causes a lot of printing for tests that makes it more difficult to debug.

frankyn: done. PTAL, thank you!

Issue #812, opened 2019-03-29
container_registry/container_analysis: re-enable TestPubSub


tbpg:
It's currently broken. Also see #785.

Example, line 2257 of https://source.cloud.google.com/results/invocations/78aedd16-b211-4a6d-9613-f4d50f8066a9/targets/cloud-devrel%2Fgo%2Fgolang-samples%2Fsystem-tests%2Fgo19/log.

DanSanche:
pub/sub has been re-enabled, and the test is now passing

PR #813, opened 2019-03-29
container_registry/container_analysis: disable broken PubSub test

tbpg:
See #812.


tbpg:

PR #814, opened 2019-03-29
iot: fix send command test

tbpg:


tbpg:

PR #815, opened 2019-03-29
monitoring/uptime: fix update test

tbpg:
The path now has a leading slash.


tbpg:

PR #816, opened 2019-03-29
monitoring: check bounds of time series points

tbpg:


tbpg:
Submitting without review because this is causing flaky tests (and flaky sample reliability).

PR #825, opened 2019-04-11
opencensus: use metric exporter interface

rghetia:
Modify quick_start  example to use new metric interface from opencensus. The new interface supports receiving gauges and derived gauges.

### DIFF
--- a/opencensus/metrics_quickstart/main.go
+++ b/opencensus/metrics_quickstart/main.go
@@ -66,7 +66,7 @@ func main() {
 	defer exporter.Flush()
 
 	if err := exporter.StartMetricsExporter(); err != nil {
-		log.Fatalf("error starting metric exporter. error=%v", err)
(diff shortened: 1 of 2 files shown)


opencensus/metrics_quickstart/main.go
```diff
@@ -65,10 +65,8 @@ func main() {
 	// Flush must be called before main() exits to ensure metrics are recorded.
 	defer exporter.Flush()
 
-	view.RegisterExporter(exporter)
-
-	// The minimum reporting period for Stackdriver is 1 minute.
-	view.SetReportingPeriod(60 * time.Second)
+	exporter.StartMetricsExporter()
```
tbpg: Should we check if this returns an error?

rghetia: added the check.

PR #826, opened 2019-04-12
jobs/v4: fix TestHistogramSearch

### LINKED ISSUE #822: jobs/v4: TestHistogramSearch is failing
Line 4180 of https://source.cloud.google.com/results/invocations/f860af24-65c1-42b2-9391-4c34b8fa3f99/targets/cloud-devrel%2Fgo%2Fgolang-samples%2Fsystem-tests%2Fgo112/log:
```
--- FAIL: TestHistogramSearch (13.59s)
    retry.go:44: FAILED after 10 attempts:
        histogram_search_sample_test.go:36: histogramSearch got "Job: \"projects/golang-samples-tests-4/tenants/null/jobs/132306366103462598\"\n", want to contain "projects/golang-samples-tests-4/tenants/547c9536-cd00-0000-0000-00b36ca93e2a/jobs/132306366103462598"
```

I'll disable the test and this issue will be used to track re-enabling it.

anguillanneuf:
Fix #822 

```
FAIL: TestJobTitleAutocomplete (0.42s)
        auto_complete_sample_test.go:34: jobTitleAutoComplete got "Auto complete results:\tSoftware Engineer (m/f)\n", want "Software Developer"

FAIL: TestHistogramSearch (14.45s)
	retry.go:44: FAILED after 10 attempts:
		histogram_search_sample_test.go:36: histogramSearch got "Job: \"projects/tz-playground-bigdata/tenants/null/jobs/80807335254991558\"\n", want to contain "projects/tz-playground-bigdata/tenants/b85ba847-2e00-0000-0000-0074e215da1d/jobs/80807335254991558"
```
@beccasaurus The job name field has a new optional `tenants/{tenant_id}` component. This was the cause of the test failure. Shall we add any samples showing what this tenant field is for?

Adding @chenlei1216 for comments.


jobs/v4/howto/auto_complete_sample_test.go
```diff
@@ -29,7 +29,7 @@ func TestJobTitleAutocomplete(t *testing.T) {
 	if _, err := jobTitleAutocomplete(buf, tc.ProjectID, companyID, "Software"); err != nil {
 		t.Fatalf("jobTitleAutoComplete: %v", err)
 	}
-	want := "Software Developer"
+	want := "Software Engineer"
```
beccasaurus: Was this working before and then broke because 'Developer' is no longer being returned in the top results?

To make this less brittle, I recommend updating the PageSize to 10. Seems reasonable that 'Software Developer' would show up in the top 10 and, if not, it might be a sign that the API is broken.

Else simply verify that the output includes a name with 'Software' in its title

chenlei1216: @anguillanneuf Yes the autocompletion ranking could change over time based on the job counts in live dataset, so I think updating the PageSize to 10 or verifying the output includes 'Software' would be better.

Sorry for a few back and forth in the past few weeks. We've already finalized our implementation so we should not expect more code sample test failures going forwards, otherwise it'll be considered as a bug on our end.

anguillanneuf: I updated the PageSize to 10. I hope this fix is enough.

PR #829, opened 2019-04-17
run: service-to-service auth snippet

ahmetb:
Service to service authentication method sample for Cloud Run.

### DIFF
--- a/run/authentication/auth.go
+++ b/run/authentication/auth.go
@@ -15,7 +15,7 @@
 // Package authentication contains the authentication samples for Cloud Run.
 package authentication
 
-// [START service_to_service_auth]
+// [START run_service_to_service_auth]
 import (
 	"fmt"
 	"net/http"
@@ -41,4 +41,4 @@ func makeGetRequest(serviceURL string) (*http.Response, error) {
 	return http.DefaultClient.Do(req)
 }
 
-// [END service_to_service_auth]
+// [END run_service_to_service_auth]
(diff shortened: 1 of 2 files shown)


ahmetb:
cc: @mcdonamp

broady:
LGTM, will leave to @tbpg to merge

run/authentication/auth.go
```diff
@@ -0,0 +1,44 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package authentication contains the authentication samples for Cloud Run.
+package authentication
+
+// [START service_to_service_auth]
```
tbpg: The region tag should be in the samples tracker, which requires the region tags to start with the product name. So, this should be something like `run_service_to_service_auth`.

ahmetb: Thanks! just created sample number 1692 in the system with that `tag`. Any ideas what's next?

broady:
maybe we need a TokenSource for this?

similar to this, which also accepts an audience:
https://godoc.org/golang.org/x/oauth2/google#JWTAccessTokenSourceFromJSON

ahmetb:
Yeah not a bad idea except

1. other languages won't have it so it'll create asymmetry (e.g. I can't look at Go example and code my own PHP)
2. introduces an extra dependency in user code.

PR #830, opened 2019-04-17
testing/kokoro: always go get top-level deps

tbpg:
cc @ahmetb

### DIFF
--- a/testing/kokoro/system_tests.sh
+++ b/testing/kokoro/system_tests.sh
@@ -93,7 +93,7 @@ popd
 
 # Always download top-level and internal dependencies.
 go get -t ./internal/...
-go get -t .
+go get -t -d .
 
 go get github.com/jstemmer/go-junit-report
 go install -v $GO_IMPORTS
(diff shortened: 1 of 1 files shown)


testing/kokoro/system_tests.sh
```diff
@@ -91,8 +91,9 @@ git checkout v2.0.0
 go get -v ./...
 popd
 
-# Always download internal dependencies.
-go get ./internal/...
+# Always download top-level and internal dependencies.
+go get -t ./internal/...
```
ahmetb: (I'm not familiar with the repo structure but) looks like `./internal/...` packages are used via subpackage tests (like `./bigquery/snippets`) and not in the root `.`

So, when subpackages are updated, `./internal/...`'s dependencies are going to be captured indirectly into `GO_IMPORTS`.

Even if `./internal/...` was used in root `.`, the `go get -t .` would be sufficient.

tl;dr not sure if this line is necessary.

tbpg: Added this after internal dependencies weren't picked up and it wouldn't compile. So, I think we need it (for one reason or another).

PR #836, opened 2019-04-26
spanner: check for database presence at test start

shollyman:
Spanner tests use stable names for testing, rather than generating
random names on each test invocation.  This change issues a single
pessimistic delete at the start of the snippet test and logs the
output, to reduce flakiness of this approach.

### DIFF
--- a/spanner/spanner_snippets/snippet_test.go
+++ b/spanner/spanner_snippets/snippet_test.go
@@ -45,9 +45,12 @@ func TestSample(t *testing.T) {
 	defer adminClient.Close()
 	defer dataClient.Close()
 
-	// Issue a pessimistic delete of the database before starting, as it's possible that resources have
-	// not been fully cleaned up from prior runs.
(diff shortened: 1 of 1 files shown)


spanner/spanner_snippets/snippet_test.go
```diff
@@ -45,6 +45,10 @@ func TestSample(t *testing.T) {
 	defer adminClient.Close()
 	defer dataClient.Close()
 
+	// Issue a pessimistic delete of the database before starting, as it's possible that resources have
+	// not been fully cleaned up from prior runs.
+	t.Logf("pessimistic spanner delete result: %v", adminClient.DropDatabase(ctx, &adminpb.DropDatabaseRequest{Database: dbName}))
```
tbpg: Maybe we should check if the database exists first? No need to log an error if delete doesn't work because the database doesn't exist (and shouldn't).

shollyman: Good point.  Will augment to check beforehand.

PR #840, opened 2019-05-03
webrisk: remove unnecessary fields from proto

JJLionHeart:


webrisk/internal/webrisk_proto/webrisk.pb.go
```diff
@@ -1,21 +1,26 @@
-// Copyright 2019 Google LLC
```
tbpg: Include copyright header?

JJLionHeart: Done. (I think)

Issue #559, opened 2018-08-13
testing: add firestore testing to nightly E2E


broady:
Firestore can't run in our regular test projects (incompatible with Datastore).

Currently we don't run our Firestore tests because `GOLANG_SAMPLES_FIRESTORE_PROJECT` is unset. We should set this in our nightly E2E tests.

tbpg:
Created a project. Setting up settings and confirming everything works before sending a PR.

PR #842, opened 2019-05-04
firebase: add collection group samples

tbpg:
See #559 for enabling tests.


BenWhitehead:
Thanks so much for getting this written up!

PR #843, opened 2019-05-06
healthcare: more samples

tbpg:


healthcare/dataset_iam_set.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+
+	healthcare "google.golang.org/api/healthcare/v1beta1"
+)
+
+// setDatasetIAMPolicy sets an IAM policy for the dataset.
+func setDatasetIAMPolicy(w io.Writer, projectID, location, datasetID string) error {
```
broady: we should probably talk to Benedict about this sample. I don't think it's very useful - it seems pretty rare you'd set a policy outright, rather than augmenting an existing one.

healthcare/dicom_iam_get.go
```diff
@@ -0,0 +1,48 @@
...
+	dicomService := healthcareService.Projects.Locations.Datasets.DicomStores
+
+	name := fmt.Sprintf("projects/%s/locations/%s/datasets/%s/dicomStores/%s", projectID, location, datasetID, dicomStoreID)
+
+	policy, err := dicomService.GetIamPolicy(name).Do()
+	if err != nil {
+		return fmt.Errorf("GetIamPolicy: %v", err)
+	}
+
+	fmt.Fprintf(w, "IAM Policy version: %v\n", policy.Version)
```
broady: I believe Version is deprecated - print something else? Etag?

healthcare/fhir_resource_export.go
```diff
@@ -0,0 +1,73 @@
...
+		return fmt.Errorf("Export: %v", err)
+	}
+
+	operationsService := healthcareService.Projects.Locations.Datasets.Operations
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.New("timeout waiting for operation to finish")
```
broady: `return ctx.Err()` i think is more appropriate

healthcare/fhir_resource_get.go
```diff
@@ -0,0 +1,62 @@
...
+
+	defer resp.Body.Close()
+
+	respBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("could not read response: %v", err)
+	}
+
+	if resp.StatusCode > 299 {
+		return fmt.Errorf("Create: status %d %s: %s", resp.StatusCode, resp.Status, respBytes)
```
broady: s/Create/Read

healthcare/fhir_resource_history_get.go
```diff
@@ -0,0 +1,60 @@
...
+
+	defer resp.Body.Close()
+
+	respBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("could not read response: %v", err)
+	}
+
+	if resp.StatusCode > 299 {
+		return fmt.Errorf("Create: status %d %s: %s", resp.StatusCode, resp.Status, respBytes)
```
broady: s/Create/History/

healthcare/dataset_iam_set.go
tbpg: Done. Changed to append (because it doesn't add much complexity, but matches the shell sample Benedict pointed out).

PR #846, opened 2019-05-08
Update the README to create a network bridge for GCE deployment

qiwzhang:
Signed-off-by: Wayne Zhang <qiwzhang@google.com>

The current README is outdated.  Updated it to use container network bridge


tbpg:
I just went through the GCE deployment (with the existing instructions) and it worked. Is the old way deprecated or expected not to work?

qiwzhang:
It really depends on which version of GCE your were using.  Mine is Ubunto 16.4.  It doesn't work.

I guess the --link flag is deprecated.  You have to create a network bridge.

PR #850, opened 2019-05-13
container_registry: container analysis filter

DanSanche:
Two small fixes:
- I changed the filter string that is exposed in the docs to be more precise, but used region tags to invisibly fall back to the original filter since the new one can't be easily tested
- I renamed PubSub.java to Subscriptions.java. It seems PubSub was causing warnings in the docs CLs, since the proper name is Pub/Sub


container_registry/container_analysis/poll_discovery_finished.go
```diff
@@ -52,8 +52,19 @@ func pollDiscoveryOccurrenceFinished(resourceURL, projectID string, timeout time
 		case <-ticker.C:
 			req := &grafeaspb.ListOccurrencesRequest{
 				Parent: fmt.Sprintf("projects/%s", projectID),
-				Filter: fmt.Sprintf(`kind="DISCOVERY" AND resourceUrl=%q`, resourceURL),
+				// Vulbnerability discovery occurrences are always published under the
+				// PACKAGE_VULNERABILITY note in the "goog-analysis" GCP project.
```
tbpg: Not sure I understand this - if the project is locked down, should it be in this example?

DanSanche: Whenever a user pushes an image to GCR with vulnerability scanning enabled, Google will scan the image and provide details about any vulnerabilities found in a publicly readable GCP project.  This sample is showing how to query that project to find the results of the vulnerability scanning, which is a common use case. 

The issue comes when we want to write to the project for testing, which is why I added the fall back filter

container_registry/container_analysis/poll_discovery_finished.go
```diff
@@ -52,8 +52,19 @@ func pollDiscoveryOccurrenceFinished(resourceURL, projectID string, timeout time
 		case <-ticker.C:
 			req := &grafeaspb.ListOccurrencesRequest{
 				Parent: fmt.Sprintf("projects/%s", projectID),
-				Filter: fmt.Sprintf(`kind="DISCOVERY" AND resourceUrl=%q`, resourceURL),
+				// Vulbnerability discovery occurrences are always associated with the
+				// PACKAGE_VULNERABILITY note in the "goog-analysis" GCP project.
+				Filter: fmt.Sprintf("resourceUrl=\"%s\" AND noteProjectId=\"%s\" AND noteId=\"%s\"",
+					resourceUrl, "goog-analysis", "PACKAGE_VULNERABILITY"),
```
tbpg: No need to use placeholders for static strings? Also, can use `%q` to automatically put quotes around the value. Can also use backticks (`) to avoid having to escape ", if you need.

DanSanche: Yeah, I thought bringing the static strings out made it easier to read, both by displaying the template format more clearly and making the line shorter. I can change it if you think it's better to include them in the string though, I don't feel too strongly about it

i didn't know about %q, I'll add that

container_registry/container_analysis/poll_discovery_finished.go
```diff
@@ -51,9 +51,19 @@ func pollDiscoveryOccurrenceFinished(resourceURL, projectID string, timeout time
 			return nil, fmt.Errorf("timeout while retrieving discovery occurrence")
 		case <-ticker.C:
 			req := &grafeaspb.ListOccurrencesRequest{
+				Parent: fmt.Sprintf("projects/%s", projectID),
+				// Vulbnerability discovery occurrences are always associated with the
```
tbpg: Sorry I missed this, typo in vulnerability.

DanSanche: good catch

PR #853, opened 2019-05-14
spanner: add queryWithParameter sample

jsimonweb:


spanner/spanner_snippets/snippet.go
```diff
@@ -1097,25 +1098,57 @@ func updateUsingDMLStruct(ctx context.Context, w io.Writer, client *spanner.Clie
 // [START spanner_dml_getting_started_insert]
 
 func writeUsingDML(ctx context.Context, w io.Writer, client *spanner.Client) error {
-	_, err := client.ReadWriteTransaction(ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
-		stmt := spanner.Statement{
-			SQL: `INSERT Singers (SingerId, FirstName, LastName) VALUES
+	_, err := client.ReadWriteTransaction(
+		ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
```
tbpg: I think I preferred it being on the same line. Generally try to avoid breaking in the middle of function calls (not including struct literals).

tbpg:
Also, heads up our new canonical sample format has all samples in their own file so imports can be included in the region tag. Don't have to do it now, but heads up for the future.

See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#one-file-per-sample.

PR #856, opened 2019-05-15
spanner: use correct/consistent logic to transfer budget

hegemonic:
According to the [Cloud Spanner docs](https://cloud.google.com/spanner/docs/getting-started/go/), the `dmlwritetxn` snippet moves 100000 from Album1 to Album2. However, it was actually moving 200000, and that only worked if you ran the tests in a specific order.

The `writetransaction` snippet was also moving 200000, but in the opposite direction, and without confirming that the budget was actually available before making the transfer.

I cleaned up both snippets, made them consistent with one another, fixed the logic error in `writetransaction`, and updated the tests to make the results more predictable.

### DIFF
--- a/spanner/spanner_snippets/snippet.go
+++ b/spanner/spanner_snippets/snippet.go
@@ -552,14 +552,17 @@ func writeWithTransaction(ctx context.Context, w io.Writer, client *spanner.Clie
 			if err != nil {
 				return err
 			}
-			const transfer = 200000
-			album1Budget += transfer
-			album2Budget -= transfer
-			cols := []string{"SingerId", "AlbumId", "MarketingBudget"}
-			txn.BufferWrite([]*spanner.Mutation{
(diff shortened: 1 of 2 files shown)


hegemonic:
It's not clear to me why the `go110` build (and only the `go110` build) is failing. [Here's the build log](https://source.cloud.google.com/results/invocations/4e6c417c-e409-4d1a-993a-32cff0cbb48e/targets/cloud-devrel%2Fgo%2Fgolang-samples%2Fpresubmit%2Fgo110/log). This appears to be the relevant error:

```
# cd /tmpfs/src/gopath/src/cloud.google.com/go; git pull --ff-only
fatal: unable to access 'https://code.googlesource.com/gocloud/': The requested URL returned error: 502
package cloud.google.com/go/spanner: exit status 1
```

Any suggestions on how to get a clean CI build?

PR #868, opened 2019-05-29
functions: add sample for Cloud Functions + Memorystore for Redis

ericschmidtatwork:
I adapted the existing sample from https://cloud.google.com/memorystore/docs/redis/connect-redis-instance-flex.


functions/memorystore/redis/function.go
```diff
@@ -0,0 +1,53 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START functions_memorystore_redis]
+
+package visitcount
```
tbpg: Conventionally, the package name is the same as the directory name. Not a requirement, but maybe we should update either the directory or package name (is there more than one `memorystore` sample where we need that subdirectory?)

Also, please add a package comment (https://github.com/golang/go/wiki/CodeReviewComments#package-comments).

functions/memorystore/redis/function.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+  "fmt"
```
tbpg: Looks like you need to run `gofmt` and `goimports`. There's usually a way to setup your editor to run these automatically every time you save.

functions/memorystore/redis/function.go
```diff
@@ -0,0 +1,53 @@
...
+  redisPort := os.Getenv("REDISPORT")
+  redisAddr := fmt.Sprintf("%s:%s", redisHost, redisPort)
+
+  const maxConnections = 10
+  redisPool = redis.NewPool(func() (redis.Conn, error) {
+    return redis.Dial("tcp", redisAddr)
+  }, maxConnections)
+}
+
+func VisitCount(w http.ResponseWriter, r *http.Request) {
```
tbpg: Please add a function comment. https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

functions/memorystore/redis/function.go
```diff
@@ -0,0 +1,53 @@
...
+  "net/http"
+  "os"
+
+  "github.com/gomodule/redigo/redis"
+)
+
+var redisPool *redis.Pool
+
+// Initialize the connection pool on instance startup
+func init() {
```
tbpg: Is there a way to test this function? That might change how we initialize the `*redis.Pool`.

functions/memorystore/redis/function.go
```diff
@@ -0,0 +1,53 @@
...
+  redisHost := os.Getenv("REDISHOST")
+  redisPort := os.Getenv("REDISPORT")
```
tbpg: Please check if either of these variables is empty and `log.Fatal` if so.

```go
if redisHost == "" {
  log.Fatal("REDISHOST must be set")
}
redisPort := os.Getenv("REDISPORT")
if redisPort == "" {
  log.Fatal("REDISPORT must be set")
}
```

Also, are these standard names for these variables? Wondering if they should be `REDIS_HOST` and `REDIS_PORT`.

functions/memorystore/redis/go.mod
```diff
@@ -0,0 +1,5 @@
+module visitcount
```
tbpg: Should probably have the full `github.com/GoogleCloudPlatform/golang-samples/...` path.

grant:
Comments need to be addressed before continuing.

functions/memorystore/redis/function.go
ericschmidtatwork: The subdirectory is necessary (see https://github.com/GoogleCloudPlatform/golang-samples/tree/master/memorystore/redis for precedent).

I've updated the package name and added a package comment.

functions/memorystore/redis/function.go
ericschmidtatwork: Done :+1:

functions/memorystore/redis/function.go
ericschmidtatwork: Done :+1:

functions/memorystore/redis/function.go
ericschmidtatwork: I've sent you a message about this :slightly_smiling_face:

functions/memorystore/redis/function.go
ericschmidtatwork: Done, and I'm not sure if they're standard but our other Memorystore samples use these names. Figured I'd be consistent.

functions/memorystore/redis/go.mod
ericschmidtatwork: Done.

functions/memorystore/redis/redis.go
```diff
@@ -0,0 +1,64 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START functions_memorystore_redis]
+
+// Package redis provides a basic Cloud Function that connects
```
tbpg: Delete "basic." https://developers.google.com/style/tone#some-things-to-avoid-where-possible

functions/memorystore/redis/redis.go
```diff
@@ -0,0 +1,64 @@
...
+// to a managed Redis instance.
+package redis
```
tbpg: Sorry for the package name churn. Better to have it not match the directory name and be something other than `redis` since you're importing a package named `redis` below. Things get confusing with collisions like that.

functions/memorystore/redis/redis.go
```diff
@@ -0,0 +1,64 @@
...
+
+// VisitCount increments the visit count on the Redis instance
+// and prints the current count in the HTTP response.
+func VisitCount(w http.ResponseWriter, r *http.Request) {
+	conn := redisPool.Get()
+	defer conn.Close()
+
+	counter, err := redis.Int(conn.Do("INCR", "visits"))
+	if err != nil {
+		http.Error(w, "Error incrementing visit count", http.StatusInternalServerError)
```
tbpg: Please also log the error above the call to `http.Error`.

```
log.Printf("redis.Int: %v", err)
```

This way you can check the function logs and see more detail about what went wrong.

functions/memorystore/redis/redis.go
```diff
@@ -0,0 +1,64 @@
...
+	redisPort := os.Getenv("REDISPORT")
+	if redisPort == "" {
+		log.Fatal("REDISPORT must be set")
+	}
+	redisAddr := fmt.Sprintf("%s:%s", redisHost, redisPort)
+
+	const maxConnections = 10
+	redisPool = redis.NewPool(func() (redis.Conn, error) {
+		return redis.Dial("tcp", redisAddr)
+	}, maxConnections)
```
tbpg: Is there some sort of check to make sure this successfully connected?

functions/memorystore/redis/redis.go
ericschmidtatwork: Done :+1:

functions/memorystore/redis/redis.go
ericschmidtatwork: No worries, done :+1:

functions/memorystore/redis/redis.go
ericschmidtatwork: Done :+1:

functions/memorystore/redis/redis.go
ericschmidtatwork: I rewrote this since `redis.NewPool` is apparently deprecated, and included a check :slightly_smiling_face:

functions/memorystore/redis/visitcount.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	redisAddr := fmt.Sprintf("%s:%s", redisHost, redisPort)
+
+	const maxConnections = 10
+	redisPool = &redis.Pool{
+		MaxIdle: maxConnections,
+		Dial: func() (redis.Conn, error) {
+			c, err := redis.Dial("tcp", redisAddr)
+			if err != nil {
+				log.Fatal("Error connecting to Redis instance")
```
tbpg:
```suggestion
				log.Fatal("redis.Dial: %v", err)
```

functions/memorystore/redis/visitcount.go
```diff
@@ -0,0 +1,72 @@
...
+	"net/http"
+	"os"
+
+	"github.com/gomodule/redigo/redis"
+)
+
+var redisPool *redis.Pool
+
+// Initialize the connection pool on instance startup
+func init() {
```
tbpg: I updated #877 to include testing this. For now, please change this to `func initializeRedis() (*redis.Pool, error)` and have it return errors (`return nil, fmt.Errorf("...", err)`).

Then, in `VisitCount`, you can do:

```go
if redisPool == nil {
	// Pre-declare err to avoid shadowing redisPool.
	var err error
	redisPool, err = initializeRedis()
	if err != nil {
		log.Fatalf("initializeRedis: %v", err)
	}
	// ...
}
```

This makes writing to the global `redisPool` variable explicit - it happens at the beginning of `VisitCount`.

It also makes it easier to test because you don't initialize the pool until you run the function. That means a test can check the environment variables are set before testing the function -- ideally when someone clones this repo and runs `go test`, everything would pass.

Hopefully we won't have to modify the sample code at all when we write the test.

functions/memorystore/redis/visitcount.go
ericschmidtatwork: Done

functions/memorystore/redis/visitcount.go
ericschmidtatwork: Done - let me know if the rewrite looks okay!

PR #871, opened 2019-06-03
spanner: more updates to samples that transfer money

hegemonic:
These changes iterate on #856 to address some requests from @jsimonweb (see GoogleCloudPlatform/java-docs-samples#1432):

- Move $200,000 from Album2 to Album1, rather than $100,000 from Album1 to Album2.
- Just check the transfer amount against the Album2 budget, not another arbitrary number.

These changes reduce the scope of updates to other code samples and docs.


spanner/spanner_snippets/snippet.go
```diff
@@ -1187,26 +1185,26 @@ func writeWithTransactionUsingDML(ctx context.Context, w io.Writer, client *span
 
 		// Transfer the marketing budget from one album to another. By keeping the actions
 		// in a single transaction, it ensures the movement is atomic.
-		const transferAmt = 100000
-		var album1budget, album2budget int64
+		const transferAmt = 200000
+		var album1Budget, album2Budget int64
 		var err error
-		if album1budget, err = getBudget(1, 1); err != nil {
+		if album2Budget, err = getBudget(2, 2); err != nil {
```
tbpg: This was already here, but it's a little weird to pre-declare `err`, `album1Budget`, and `album2Budget` so we can do an inline error check. Would probably more clear to initialize the variables outside the `if` and not pre-declare:

```go
const transferAmt = 200000
album2Budget, err := getBudget(2, 2)
if err != nil {
  return err
}
if album2Budget >= transferAmt {
  album1Budget, err := getBudget(1, 1)
  if err != nil {
    return err
  }
  // ...
}
```

hegemonic: Done (and thanks for the extremely clear instructions!).

hegemonic:
Looks like the merge commit caused CLA issues. If you need me to do anything in response, please let me know!

tbpg:
I squashed the commits and force-pushed to clean up the history (the merge added recent commits to this diff).

PR #875, opened 2019-06-05
getting-started/background_events: initial commit

tbpg:
This still needs region tags, but we can start the review without (and that might change the region tags we want).

The README has minimal documentation for now as it should point to the public tutorial once it's written.


getting-started/background/index/main.go
```diff
@@ -0,0 +1,165 @@
...
+// requests.
+type app struct {
+	pubsubClient    *pubsub.Client
+	pubsubTopic     *pubsub.Topic
+	firestoreClient *firestore.Client
+	tmpl            *template.Template
+}
+
+// newApp creates a new app.
+func newApp(projectID, templateDir string) (*app, error) {
```
jadekler: nit: maybe below main? I personally prefer main at the top ideally, at the bottom less ideally, but dislike in the middle. :)

getting-started/background/index/main.go
```diff
@@ -0,0 +1,165 @@
...
+	}, nil
+}
+
+func main() {
+	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
+	if projectID == "" {
+		log.Fatalf("GOOGLE_CLOUD_PROJECT must be set")
+	}
+
+	a, err := newApp(projectID, "index")
```
jadekler: If it's just going to be "index" in actual code, do you need the second argument?

getting-started/background/index/main.go
```diff
@@ -0,0 +1,165 @@
...
+		http.Error(w, fmt.Sprintf("Error getting translations: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	translations := []background.Translation{}
+	for _, d := range docs {
+		t := background.Translation{}
+		if err := d.DataTo(&t); err != nil {
+			http.Error(w, "Error reading translations", http.StatusInternalServerError)
+			log.Printf("DataTo: %v", err)
```
jadekler: Should this return? And, should the log be before the http.Error like below?

getting-started/background/index/main.go
```diff
@@ -0,0 +1,165 @@
...
+
+// index lists the current translations.
+func (a *app) index(w http.ResponseWriter, r *http.Request) {
+	docs, err := a.firestoreClient.Collection("translations").Documents(r.Context()).GetAll()
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Error getting translations: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	translations := []background.Translation{}
```
jadekler: var translations []background.Translation

getting-started/background/index/main.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		translations = append(translations, t)
+	}
+
+	a.tmpl.Execute(w, translations)
+}
+
+// requestTranslation parses the request, validates it, and sends it to Pub/Sub.
+func (a *app) requestTranslation(w http.ResponseWriter, r *http.Request) {
+	r.ParseForm()
```
jadekler: check err

getting-started/background/index/main.go
```diff
@@ -0,0 +1,165 @@
...
+		"en": true,
+		"es": true,
+		"fr": true,
+		"ja": true,
+		"sw": true,
+	}
+	lang := r.PostFormValue("lang")
+	if !acceptableLanguages[lang] {
+		log.Printf("Unsupported language: %v", lang)
+		http.Error(w, fmt.Sprintf("Unsupported language: %v", lang), http.StatusBadRequest)
```
jadekler: should this return?

getting-started/background/index/main.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	msg, err := json.Marshal(t)
+	if err != nil {
+		log.Printf("json.Marshal: %v", err)
+		http.Error(w, "Error requesting translation", http.StatusInternalServerError)
+		return
+	}
+
+	res := a.pubsubTopic.Publish(r.Context(), &pubsub.Message{Data: msg})
+	if _, err := res.Get(r.Context()); err != nil {
```
jadekler: This races. Wait for Ready before calling Get https://godoc.org/cloud.google.com/go/pubsub#PublishResult.Ready

getting-started/background/translate.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Translate translates the given message to French.
+func Translate(ctx context.Context, m PubSubMessage) error {
+	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
+	if projectID == "" {
+		return fmt.Errorf("GOOGLE_CLOUD_PROJECT must be set")
+	}
+
+	if translateClient == nil {
+		var err error
```
jadekler: Why?

getting-started/background/translate.go
```diff
@@ -0,0 +1,96 @@
...
+		translateClient, err = translate.NewClient(context.Background())
+		if err != nil {
+			return fmt.Errorf("translate.NewClient: %v", err)
+		}
+	}
+	if firestoreClient == nil {
+		var err error
```
jadekler: Why?

getting-started/background/translate.go
```diff
@@ -0,0 +1,96 @@
...
+// Translate translates the given message to French.
+func Translate(ctx context.Context, m PubSubMessage) error {
+	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
+	if projectID == "" {
+		return fmt.Errorf("GOOGLE_CLOUD_PROJECT must be set")
+	}
+
+	if translateClient == nil {
+		var err error
+		translateClient, err = translate.NewClient(context.Background())
```
jadekler: Is there a reason to use context.Background instead of propagating context? (here and elsewhere)

getting-started/background/translate.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return fmt.Errorf("language.Parse: %v", err)
+	}
+
+	outs, err := translateClient.Translate(ctx, []string{t.Original}, lang, nil)
+	if err != nil {
+		return fmt.Errorf("Translate: %v", err)
+	}
+
+	t.Translated = outs[0].Text
```
jadekler: Maybe perform a len(outs) check first? Returning an error or something would be a better situation than a panic.

getting-started/background/translate_test.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"testing"
+
+	"cloud.google.com/go/firestore"
+)
+
+func deleteAll(projectID string) error {
```
jadekler: nit: I think it looks nicer to have tests at the top and helpers elsewhere/at the bottom. When I open a test file, I expect to see tests.

Feel free to ignore this if you feel differently.

getting-started/background/translate_test.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, err
+	}
+	docs, err := client.Collection("translations").Documents(context.Background()).GetAll()
+	if err != nil {
+		return nil, err
+	}
+	translations := []Translation{}
+	for _, doc := range docs {
+		t := Translation{}
+		doc.DataTo(&t)
```
jadekler: check err

getting-started/background/translate_test.go
```diff
@@ -0,0 +1,98 @@
...
+func getAll(projectID string) ([]Translation, error) {
+	client, err := firestore.NewClient(context.Background(), projectID)
+	if err != nil {
+		return nil, err
+	}
+	docs, err := client.Collection("translations").Documents(context.Background()).GetAll()
+	if err != nil {
+		return nil, err
+	}
+	translations := []Translation{}
```
jadekler: var translations []Translation

getting-started/background/translate_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	return translations, nil
+}
+
+func TestTranslate(t *testing.T) {
+	projectID := os.Getenv("GOLANG_SAMPLES_FIRESTORE_PROJECT")
+	if projectID == "" {
+		t.Skip("Skipping Firestore test. Set GOLANG_SAMPLES_FIRESTORE_PROJECT.")
+	}
+	os.Setenv("GOOGLE_CLOUD_PROJECT", projectID)
```
jadekler: Why not just require this env var is set?

And, does this change the running user's shell? If so, two thumbs down heh.

getting-started/background/translate_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// Remove any old translations.
+	if err := deleteAll(projectID); err != nil {
+		t.Fatalf("deleteAll: %v", err)
+	}
+
+	ctx := context.Background()
+	msg, _ := json.Marshal(Translation{
```
jadekler: check err?

getting-started/background/translate_test.go
```diff
@@ -0,0 +1,98 @@
...
+		t.Skip("Skipping Firestore test. Set GOLANG_SAMPLES_FIRESTORE_PROJECT.")
+	}
+	os.Setenv("GOOGLE_CLOUD_PROJECT", projectID)
+
+	// Remove any old translations.
+	if err := deleteAll(projectID); err != nil {
+		t.Fatalf("deleteAll: %v", err)
+	}
+
+	ctx := context.Background()
```
jadekler: nit: here you're defining a ctx, but you're not using it in getAll / deleteAll. I'd say either use context everywhere (instead of inline context.Background) or inline context.Background everywhere (instead of defining a ctx).

getting-started/background/translate_test.go
```diff
@@ -0,0 +1,98 @@
...
+	})
+	m := PubSubMessage{Data: msg}
+	if err := Translate(ctx, m); err != nil {
+		t.Fatalf("Translate: %v", err)
+	}
+	translations, err := getAll(projectID)
+	if err != nil {
+		t.Fatalf("getAll: %v", err)
+	}
+	if len(translations) != 1 {
```
jadekler: nit: you could simplify this next block by using cmp.Diff (which does len and value check) if you want.

getting-started/background/translate_test.go
```diff
@@ -96,3 +66,40 @@ func TestTranslate(t *testing.T) {
 		t.Fatalf("Translate got:\n%+v\nWant:\n%+v", translations[0], want)
 	}
 }
+
+func deleteAll(ctx context.Context, projectID string) error {
+	client, err := firestore.NewClient(ctx, projectID)
```
broady: Why create a new client on each request?

tbpg: Done (only used for tests, but might as well be more efficient).

broady: Oops, missed that this was in _test.go

getting-started/background/index/main.go
jadekler: WHAT. Holy cow how did I miss that. Nice :+1:

getting-started/background/translate.go
jadekler: Ohhhh - gotcha, thanks for the explanation! :+1:

getting-started/background/translate_test.go
jadekler: Gotcha! Thanks!

getting-started/background/translate_test.go
jadekler: :+1:

PR #879, opened 2019-06-10
container_registry: Container Analysis v1

DanSanche:
Updated Container analysis samples to use new v1 libraries

### DIFF
--- a/container_registry/container_analysis/create_note.go
+++ b/container_registry/container_analysis/create_note.go
@@ -32,7 +32,7 @@ func createNote(noteID, projectID string) (*grafeaspb.Note, error) {
 		return nil, fmt.Errorf("NewGrafeasV1Beta1Client: %v", err)
 	}
 	defer client.Close()
-	
+
 	projectName := fmt.Sprintf("projects/%s", projectID)
 
 	req := &grafeaspb.CreateNoteRequest{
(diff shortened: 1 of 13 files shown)


container_registry/container_analysis/create_occurrence.go
```diff
@@ -41,16 +40,27 @@ func createOccurrence(resourceURL, noteID, occProjectID, noteProjectID string) (
...
+							AffectedPackage: "your-package-here",
+							MinAffectedVersion: &grafeaspb.Version{
+								Kind: grafeaspb.Version_MINIMUM,
+							},
+							FixedVersion: &grafeaspb.Version{
+								Kind: grafeaspb.Version_MAXIMUM,
+							},
+						},
+					},
+				},
```
DanSanche: FYI @Katee : It looks like the minimal Note and Occurrence has got a lot more complicated in v1 compared to beta1. Just wanted to make sure that was intentional (and let me know if this looks like a good sample, or if you have another idea in mind)

container_registry/container_analysis/create_note.go
```diff
@@ -20,20 +20,19 @@ import (
...
 func createNote(noteID, projectID string) (*grafeaspb.Note, error) {
 	ctx := context.Background()
-	client, err := containeranalysis.NewGrafeasV1Beta1Client(ctx)
+	client, err := containeranalysis.NewClient(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("NewGrafeasV1Beta1Client: %v", err)
 	}
 	defer client.Close()
-
+
```
tbpg: Looks like you need to run `gofmt -s` (there's usually a way to have your editor run it for you every time you save). Kokoro failed the `gofmt` check.

If you're running it from command line, you'll want to run `gofmt -w -s .` (write files directly, do extra simplifications, all files in this directory (recursive)).

container_registry/container_analysis/poll_discovery_finished.go
```diff
@@ -20,10 +20,10 @@ import (
 	"context"
 	"fmt"
 	"time"
+	"google.golang.org/api/iterator"
```
tbpg: Nit: would you mind moving this down one line with the other non-standard library imports?

container_registry/container_analysis/create_note.go
DanSanche: sorry, should have caught that! fixed

PR #884, opened 2019-06-12
speech: add/move region tags

davidcavazos:
Adding region tags to the Text-to-Speech streaming from file code snippet


davidcavazos:
Done, I also went ahead and included the imports on `livecaption` for consistency since they're kind of twins :)

Issue #569, opened 2018-08-16
vision: sample generation template needs a GCS-only and local-only options


alixhami:
The sample here: https://github.com/GoogleCloudPlatform/golang-samples/blob/057d899bf97d4ea75c883da4b410fc1699090188/vision/detect/generated/sample-template.go#L340

is only for GCS, but a local sample was also generated (with the same code). Please add the ability to specify GCS or Local only samples to the template.

@davidcavazos mentioned that it was not possible to fix this by creating another non-generated file within the same directory.

broady:
This API method only works with GCS, not on a local file?

alixhami:
Yes, it's GCS only.

tbpg:
Generation was removed in #782.

alixhami:
In this case, the `vision_text_detection_pdf` sample should be deleted.

PR #885, opened 2019-06-13
vision/detect: delete unused sample and add subtests

tbpg:
Using parallel subtests for TestDetect saves about ~20 seconds per run
(~22 seconds -> ~2 seconds).

This PR also makes it so we don't need `time.Now().Unix()` to make a unique bucket name. Instead, it cleans and uses the same bucket every time.

Noticed by @alixhami in #569.

### DIFF
--- a/vision/detect/detect_test.go
+++ b/vision/detect/detect_test.go
@@ -58,12 +58,11 @@ func TestDetect(t *testing.T) {
 		t.Run(tt.name+"/local", func(t *testing.T) {
 			t.Parallel()
 			var buf bytes.Buffer
-			err := tt.local(&buf, "../testdata/"+tt.path)
-			if err != nil {
+			if err := tt.local(&buf, "../testdata/"+tt.path); err != nil {
 				t.Fatalf("Local %s(%q): got %v, want nil err", tt.name, tt.path, err)
 			}
-			if got := buf.String(); !strings.Contains(strings.ToLower(got), strings.ToLower(tt.wantContain)) {
-				t.Errorf("Local %s(%q): got %q, want to contain %q", tt.name, tt.path, got, tt.wantContain)
+			if got, wantContain := strings.ToLower(buf.String()), strings.ToLower(tt.wantContain); !strings.Contains(got, wantContain) {
+				t.Errorf("Local %s(%q): got %q, want to contain %q", tt.name, tt.path, got, wantContain)
 			}
 		})
 	}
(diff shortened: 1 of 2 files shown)


vision/detect/detect_test.go
```diff
@@ -57,30 +55,34 @@ func TestDetect(t *testing.T) {
...
-		if err != nil {
-			t.Fatalf("GCS %s(%q): got %v, want nil err", tt.name, tt.path, err)
-		}
-		if got := buf.String(); !strings.Contains(strings.ToLower(got), strings.ToLower(tt.wantContain)) {
-			t.Errorf("GCS %s(%q): got %q, want to contain %q", tt.name, tt.path, got, tt.wantContain)
-		}
+		t.Run(tt.name+"/gcs", func(t *testing.T) {
+			t.Parallel()
+			var buf bytes.Buffer
+			err := tt.gcs(&buf, "gs://python-docs-samples-tests/vision/"+tt.path)
```
jadekler: nit: `if err := [...]; err != nil { ... }`

(here and elsewhere)

vision/detect/detect_test.go
```diff
@@ -57,30 +55,34 @@ func TestDetect(t *testing.T) {
...
-			t.Errorf("Local %s(%q): got %q, want to contain %q", tt.name, tt.path, got, tt.wantContain)
-		}
+		t.Run(tt.name+"/local", func(t *testing.T) {
+			t.Parallel()
+			var buf bytes.Buffer
+			err := tt.local(&buf, "../testdata/"+tt.path)
+			if err != nil {
+				t.Fatalf("Local %s(%q): got %v, want nil err", tt.name, tt.path, err)
+			}
+			if got := buf.String(); !strings.Contains(strings.ToLower(got), strings.ToLower(tt.wantContain)) {
```
jadekler: nit: may be nicer to do:

```
if got, wantContains := strings.ToLower(buf.String), strings.ToLower(tt.wantContains); !strings.Contains(got, wantContains) {
    ...
}
```

(making the assertion just about asserting, and putting all the "munge up the data into the way I want it" into the declaration part)

vision/detect/detect_test.go
tbpg: Done.

vision/detect/detect_test.go
tbpg: Done.

PR #890, opened 2019-06-21
testing: remove special case for mailgun

tbpg:
Go 1.9.7 and 1.10.3 have minimal support for semantic import versioning.


tbpg:
Mailgun updated to modules and started using SIV. So, older versions of Go stopped compiling because they couldn't find the /v2 dependencies. Later versions of Go got minimal support for SIV. I upgraded the Go versions in the base test images, so we don't need any more special pinning for Mailgun.

Hope that helps.

PR #893, opened 2019-06-25
appengine: change runtime to go112 and add comment for go111

supersandra:
Changing go111->go112 and putting the go111 value in a comment.

### DIFF
--- a/appengine/go11x/helloworld/app.yaml
+++ b/appengine/go11x/helloworld/app.yaml
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-runtime: go111
+runtime: go112 # replace with go111 for Go 1.11


appengine/go11x/helloworld/app.yaml
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-runtime: go111
+runtime: go112 # replace with go111 for Go 1.11
```
tbpg: I don't think we need the comment for saying how to do it with Go 1.11?

supersandra: Both runtimes are supported and this sample is used in both docs, so I do think we need it.
e.g.
https://cloud.google.com/appengine/docs/standard/go112/building-app/#creating_the_appyaml_file
https://cloud.google.com/appengine/docs/standard/go111/building-app/#creating_the_appyaml_file

PR #660, opened 2018-10-22
all: add go.mod file

broady:
DO NOT SUBMIT

Created with:
$ go mod init
$ go get -u github.com/mailgun/mailgun-go@master

(mailgun has made some backwards incompatible changes, and they're
likely tagging this repo with v2 soon)

Everything builds with go build ./... from outside a GOPATH.

We probably want separate go.mod files for deployed apps (App Engine samples,
Bookshelf app).

PR #895, opened 2019-06-26
run/pubsub: add sample app

grayside:
Previously reviewed sample.


grayside:
Cloud Run samples don't necessarily try to be backwards compatible, so while this does not have special runtime requirements like #896, maybe we'd rather run this inside the container as well?

```sh
docker build --target builder -t pubsub:builder .
docker run --rm pubsub:builder go test -v
```

run/pubsub/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# Use the offical Golang image to create a build artifact.
+# This is based on Debian and sets the GOPATH to /go.
+# https://hub.docker.com/_/golang
+FROM golang:1.12 as builder
+
+# Copy local code to the container image.
+WORKDIR /app
+COPY . .
+
+# Build the outyet command inside the container.
```
tbpg: s/outyet/hellopubsub/

run/pubsub/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# (You may fetch or manage dependencies here,
+# either manually or with a tool like "godep".)
```
tbpg: I'd leave out references to godep.

run/pubsub/go.mod
```diff
@@ -0,0 +1,3 @@
+module run_pubsub
```
tbpg: Probably use the default module name (with the full `golang-samples` path)? That's what we've done for others. Not clear what the best way to handle nested modules in golang-samples is.

run/pubsub/main.go
```diff
@@ -0,0 +1,72 @@
...
+		log.Printf("Defaulting to port %s", port)
+	}
+	// Start HTTP server.
+	log.Printf("Listening on port %s", port)
+	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
+}
+
+// [END run_pubsub_server]
+
+// [START run_pubsub_handler]
```
tbpg: Newline after the region tag.

run/pubsub/main.go
```diff
@@ -0,0 +1,72 @@
...
+// PubSubMessage is the payload of a Pub/Sub event. Please refer to the docs for
+// additional information regarding Pub/Sub events.
```
tbpg: Link to https://cloud.google.com/functions/docs/tutorials/pubsub, even though it's about Cloud Functions?

run/pubsub/main_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+func TestHelloPubSubErrors(t *testing.T) {
+	tests := []struct {
+		name string
+	}{
+		{name: "no_payload"},
+		{name: "not_base64"},
+	}
+	for _, test := range tests {
+		var payload *strings.Reader
```
tbpg: `payload` should be a part of the test struct, right? Also `not_encoded` and `fmt.Sprintf` are not needed below, so it would be fine to inline it all.

Or maybe if you want to keep symmetry with the normal test below, add an `encode` value to the test struct? Relying on the test name feels weird.

run/pubsub/main_test.go
```diff
@@ -0,0 +1,97 @@
...
+			jsonStr := fmt.Sprintf(`{"message":{"data":"%s","id":"test-123"}}`, not_encoded)
+			payload = strings.NewReader(jsonStr)
+		}
+		req := httptest.NewRequest("GET", "/", payload)
+		rr := httptest.NewRecorder()
+
+		HelloPubSub(rr, req)
+
+		if rr.Result().StatusCode != http.StatusBadRequest {
+			t.Errorf("HelloPubSub(%q) should get BadRequest response", test.name)
```
tbpg: Print the response it got. https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

run/pubsub/main_test.go
```diff
@@ -0,0 +1,97 @@
...
+		rr := httptest.NewRecorder()
+
+		HelloPubSub(rr, req)
+
+		w.Close()
+		log.SetOutput(os.Stderr)
+		log.SetFlags(originalFlags)
+
+		if rr.Result().StatusCode == http.StatusBadRequest {
+			t.Errorf("HelloPubSub received invalid input (%q)", test.data)
```
tbpg: Print the status code.

go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/GoogleCloudPlatform/golang-samples
```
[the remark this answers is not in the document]
grayside: I'm not sure how #660 applies here. Note this module is probably namespaced incorrectly, this should be a standalone module.

run/pubsub/main.go
grayside: Why not link to cloud.google.com/run/docs/tutorials/pubsub? The reason this is vague is that the specific docs pages to look at for a definition of event structure are unclear and in flux, but there's probably something closer in the Pub/Sub docs. Do you have a preference?

grayside:
Addressed code feedback.

go.mod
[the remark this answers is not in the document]
grayside: That won't work for building a sample inside a container. Minimally this would block end-to-end tests.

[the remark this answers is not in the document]
grayside: In that light, should this be 
```
module github.com/GoogleCloudPlatform/golang-samples/run/pubsub
```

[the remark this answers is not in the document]
tbpg: Sorry, I think I was unclear. This is a base-level repository `go.mod`. This PR is adding two `go.mod` files right now:
* `golang-samples/go.mod`
* `golang-samples/run/pubsub/go.mod`.

We should delete this one -- we only need the `pubsub` one (which is `module github.com/GoogleCloudPlatform/golang-samples/run/pubsub` already).

Shouldn't block any tests or anything.

grayside:
Removed the root-level go.mod. Not sure how it got there.

grayside:
> "Cloud Run samples don't necessarily try to be backwards compatible."

My take on code that is meant to run in a container is that it matters that they run inside the container, because the container is effectively production. Running in any other environment is a planning mechanism to know whether planned environment changes might break.

Will file issues to discuss some more targeted parts of this as follow-up.

PR #898, opened 2019-06-28
container_registry/container_analysis: error message fix

DanSanche:
The v1 Container Analysis samples had error messages that still references the beta library. Fixed the string


tbpg:
@jadekler Are the breaking grafeas proto changes being rolled back? Wondering if we should update these samples or if those changes are being rolled back.

PR #906, opened 2019-07-11
bigquery/bigquery_storage_quickstart: update example code.

shollyman:
Makes more explicit use of sharding strategy and data format
now that additional options exist in the API.  Also, migrates
row progress to the parent message as it's no longer format-dependent.


shollyman:
Failure was 1.12 and due to backend issue in an unrelated BQ query snippet: 
snippet_test.go:148: queryBasic: googleapi: Error 500: An internal error occurred and the request could not be completed., internalError

PR #909, opened 2019-07-12
getting-started/background: split Translate into smaller functions

tbpg:
I put them in the order they'll be shown in the doc. I put the cloud function (`Translate`) last so there won't be compile errors before the other helper functions are defined.

I'm open to splitting it up more or less, if you think that will be more clear.

### DIFF
--- a/getting-started/background/translate.go
+++ b/getting-started/background/translate.go
@@ -92,8 +92,10 @@ func initializeClients() error {
 
 // [START getting_started_background_translate_string]
 
-// translateString translates text to lang, returning the translated text,
-// the automatically detected source language, and an error.
+// translateString translates text to lang, returning:
+// * the translated text,
+// * the automatically detected source language, and
+// * an error.
 func translateString(ctx context.Context, text string, lang string) (translated string, originalLang string, err error) {
 	l, err := language.Parse(lang)
 	if err != nil {
--- a/getting-started/background/translate.go
+++ b/getting-started/background/translate.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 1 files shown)


getting-started/background/translate.go
```diff
@@ -80,6 +85,40 @@ func Translate(ctx context.Context, m PubSubMessage) error {
...
 		}
 	}
+	return nil
+}
+
+// [END getting_started_background_translate_init]
+
+// [START getting_started_background_translate_string]
+
+// translateString translates text to lang, returning the translated text,
```
fhinkel: nit: I'd find this easier to read with a list of return objects.

getting-started/background/translate.go
```diff
@@ -80,6 +85,40 @@ func Translate(ctx context.Context, m PubSubMessage) error {
...
+	}
+
+	return outs[0].Text, outs[0].Source.String(), nil
+}
+
+// [END getting_started_background_translate_string]
+
+// [START getting_started_background_translate]
+
+// Translate translates the given message and stores the result in Firestore.
```
fhinkel: Is this a convention to include the function name in the comment? I'd leave it out:
* You never risk that comment and function name diverge
* Comment is one word shorter without losing any information

(I might just be used to seeing comments like `Translate the given message` and my brain is trained to associate that with the function below)

getting-started/background/translate.go
tbpg: Done (changed to a list).

getting-started/background/translate.go
tbpg: Yes, it's standard for comments to start with the thing being described. https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

PR #928, opened 2019-08-01
functions: Adds sample functions_helloworld_error_test

grant:
Adds GCF Error function sample
DevRel Sample: 1563
Bug: 129762816, 129762354
Doc: https://cloud.google.com/functions/docs/monitoring/error-reporting


functions/helloworld/hello_error.go
```diff
@@ -0,0 +1,36 @@
...
+package helloworld
+
+import (
+	"log"
+	"net/http"
+)
+
+// HelloError is an HTTP Cloud Function with a request parameter.
+func HelloError(w http.ResponseWriter, r *http.Request) {
+	// [START functions_helloworld_error]
```
tbpg: Please start the region tag above the package declaration and end it after the end of the function (the pattern we've come up with for functions). Also, there should not be a newline after the package comment. So, the final result will be:

```
// [START functions_helloworld_error]

// Package helloworld provides a set of Cloud Functions samples.
package helloworld
```

functions/helloworld/hello_error.go
```diff
@@ -0,0 +1,36 @@
...
+// Package helloworld provides a set of Cloud Functions samples.
+
+package helloworld
+
+import (
+	"log"
+	"net/http"
+)
+
+// HelloError is an HTTP Cloud Function with a request parameter.
```
tbpg: Update comment.

functions/helloworld/hello_error.go
```diff
@@ -0,0 +1,36 @@
...
+func HelloError(w http.ResponseWriter, r *http.Request) {
+	// [START functions_helloworld_error]
+	// These WILL be reported to Stackdriver Error Reporting
+	log.Fatal("I failed you")
+	
+	// These will NOT be reported to Stackdriver Error Reporting
+	log.Print("I failed you")
+
+	http.Error(w, "I failed you", 500)
```
tbpg: Comment with what this one does?

functions/helloworld/hello_error.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"log"
+	"net/http"
+)
+
+// HelloError is an HTTP Cloud Function with a request parameter.
+func HelloError(w http.ResponseWriter, r *http.Request) {
+	// [START functions_helloworld_error]
+	// These WILL be reported to Stackdriver Error Reporting
```
tbpg: Minor: please add periods at the end of comments. https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

functions/helloworld/hello_error_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"io/ioutil"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestHelloError(t *testing.T) {
+	payload := strings.NewReader("")
+	req := httptest.NewRequest("GET", "/", payload)
```
tbpg: The payload is not used - maybe it can just be nil?

functions/helloworld/hello_error_test.go
```diff
@@ -0,0 +1,39 @@
...
+package helloworld
+
+import (
+	"io/ioutil"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestHelloError(t *testing.T) {
```
tbpg: I'm not sure I understand this test. As soon as it calls `HelloError`, the function will `log.Fatal` and the test will fail (see Kokoro logs).

grant:
@tbpg PTAL. It looks like some samples are not region tagged...

functions/tips/error.go
```diff
@@ -12,11 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// [START functions_helloworld_error]
+
 package tips
 
 import (
+	// [END functions_helloworld_error]
```
tbpg: Delete the GCS function below to avoid having to do this? Doesn't look like it's included anywhere.

grant: Done.

PR #929, opened 2019-08-01
[Storage] Bucket metadata samples

frankyn:
cc: (Go sample) @JesseLovelace


storage/buckets/main.go
```diff
@@ -137,6 +137,40 @@ func list(client *storage.Client, projectID string) ([]string, error) {
...
+	fmt.Fprintln(w, "RequesterPays: %v", attrs.RequesterPays)
+	fmt.Fprintln(w, "VersioningEnabled: %v", attrs.VersioningEnabled)
+	fmt.Fprintln(w, "LogBucket: %v", attrs.BucketLogging.LogBucket)
+	fmt.Fprintln(w, "LogObjectPrefix: %v", attrs.BucketLogging.LogObjectPrefix)
+	fmt.Fprintln(w, "\n\n\nLabels:")
+	for key, value := range attrs.Labels {
+		fmt.Fprintln(w, "\t%v = %v", key, value)
+	}
+	// [END storage_get_bucket_metadata]
+	return nil
```
kurtisvg: nit: The other `get*` functions return the object as well - should you do that for this one?

frankyn: I instead verify the output of the sample in this case. Is that fair?

kurtisvg: Yeah, it's more consistent with the new format but let consistent with the samples around it to print instead. I'm just pointing it out, I'm don't think it makes a significant different either way

frankyn: I split it up. PTAL

PR #935, opened 2019-08-08
bigquery: alter retry strategy for storage sample, improve output.

shollyman:
The demonstrated strategy was more indicative of a unary RPC.  This changes the retry logic to be better aligned with a long running stream that may continue to consume results for minutes or hours.

This change also improves the printed output by replacing rather than appending elements into a slice.

### DIFF
--- a/bigquery/bigquery_storage_quickstart/main.go
+++ b/bigquery/bigquery_storage_quickstart/main.go
@@ -205,7 +205,7 @@ func processStream(ctx context.Context, client *bqStorage.BigQueryStorageClient,
 	retryLimit := 3
 
 	for {
-		retries := retryLimit
+		retries := 0
 		// Send the initiating request to start streaming row blocks.
 		rowStream, err := client.ReadRows(ctx, &bqStoragepb.ReadRowsRequest{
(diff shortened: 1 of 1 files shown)


bigquery/bigquery_storage_quickstart/main.go
```diff
@@ -197,9 +199,13 @@ func valueFromTypeMap(field interface{}) interface{} {
...
 func processStream(ctx context.Context, client *bqStorage.BigQueryStorageClient, st *bqStoragepb.Stream, ch chan<- *bqStoragepb.AvroRows) error {
 	var offset int64
-	streamRetry := 3
+
+	// Streams may be long-running.  Rather than using a global retry for the
+	// stream, implement a retry that resets once progress is made.
+	retryLimit := 3
 
 	for {
+		retries := retryLimit
```
tbpg: Optional: Intuitively, `retries` seems like it should start at 0 to me. Makes sense once you look below and check it's being decremented, but while you're changing this, consider making it start a 0.

shollyman: Swapped.

PR #938, opened 2019-08-09
pubsub: add missing samples for concurrency control and sync pull

hongalex:
Add two missing samples:
* concurrency control
* async pull

Another issue is that we currently do not have tests for topics/subscriptions since they're currently in separate packages (both as main). Will need to refactor at some point, per https://github.com/GoogleCloudPlatform/golang-samples/blob/6cd4636d36a0189d74a2f9444b551c498e679d6e/pubsub/topics/main_test.go#L99


pubsub/subscriptions/main.go
```diff
@@ -167,6 +170,37 @@ func pullMsgsSettings(client *pubsub.Client, subName string) error {
 	return nil
 }
 
+func pullMsgsConcurrencyControl(client *pubsub.Client, subName string, numGoroutines int) error {
+	ctx := context.Background()
+	// [START pubsub_subscriber_concurrency_control]
+	sub := client.Subscription(subName)
+	sub.ReceiveSettings.NumGoroutines = numGoroutines
```
tbpg: Does `numGoroutines` need to be an argument? Might be easier to understand if this uses a number directly.

(If that conflicts with samples in other languages, it's fine to leave it as-is.)

pubsub/subscriptions/main.go
```diff
@@ -167,6 +170,37 @@ func pullMsgsSettings(client *pubsub.Client, subName string) error {
...
+	err := sub.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
+		fmt.Printf("Got message %q\n", string(msg.Data))
+		msg.Ack()
+	})
+	if err != nil {
+		return err
```
tbpg:
```suggestion
		return fmt.Errorf("Receive: %v", err)
```

pubsub/subscriptions/main.go
anguillanneuf: I would consider adding a comment here just to clarify that `numGoroutines` threads are used for  executing callbacks.

pubsub/subscriptions/main.go
```diff
@@ -167,6 +170,37 @@ func pullMsgsSettings(client *pubsub.Client, subName string) error {
...
+	}
+	// [END pubsub_subscriber_sync_pull]
+	return nil
+}
```
anguillanneuf: We need a more elaborate sample here that does synchronous pull with lease management. The sample needs to show how to modify message ack deadline for long-running tasks. Here are the existing [Python](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/7d01d749fb616eb5332512413c47634b199b28e6/pubsub/cloud-client/subscriber.py#L293) and [Node](https://github.com/googleapis/nodejs-pubsub/blob/b9408118dbc5e4d8c580e23a586c0e19caeae0db/samples/subscriptions.js#L273) samples.

hongalex: Done. I noticed that the samples used in the docs are actually inconsistent. We want 2 samples in this case: one to show how to do simple Synchronous Pull, and another sample to show lease management. The Python/Node samples are missing the simple case, since [the docs](https://cloud.google.com/pubsub/docs/pull#synchronous_pull) actually say:

> Here is some sample code to pull and acknowledge a fixed number of messages:

This is hard to read for users who want to know _how_ to do sync pull. We should add another section in the docs referencing doing synchronous pull with lease management as you said.

anguillanneuf: Let's come back to this after discussing with Ken/Kir.

pubsub/subscriptions/sync_pull.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"cloud.google.com/go/pubsub"
+	pubsubV1 "cloud.google.com/go/pubsub/apiv1"
+	pb "google.golang.org/genproto/googleapis/pubsub/v1"
+)
+
+func pullMsgsSync(w io.Writer, projectID, subName string, topic *pubsub.Topic, maxMessages int32) ([]string, error) {
+	// projectID := "my-project-id"
+	// subName := projectID + "-example-sub"
+	// topic of type https://godoc.org/cloud.google.com/go/pubsub#Topic
```
tbpg: Does `maxMessages` need to be an argument? If so, please document it.

pubsub/subscriptions/sync_pull.go
```diff
@@ -0,0 +1,84 @@
...
+	ctx := context.Background()
+
+	// Publish 10 messages on the topic.
+	var results []*pubsub.PublishResult
+	for i := 0; i < 15; i++ {
```
tbpg: 10 messages?
```suggestion
	for i := 0; i < 10; i++ {
```

pubsub/subscriptions/sync_pull.go
```diff
@@ -0,0 +1,84 @@
...
+		res := topic.Publish(ctx, &pubsub.Message{
+			Data: []byte(fmt.Sprintf("hello world #%d", i)),
+		})
+		results = append(results, res)
+	}
+
+	// Check that all messages were published.
+	for _, r := range results {
+		_, err := r.Get(ctx)
+		if err != nil {
```
tbpg:
```suggestion
		if _, err := r.Get(ctx); err != nil {
```

pubsub/subscriptions/sync_pull.go
```diff
@@ -0,0 +1,84 @@
...
+			return nil, fmt.Errorf("Get: %v", err)
+		}
+	}
+
+	// Instantiate a client
+	subClient, err := pubsubV1.NewSubscriberClient(ctx)
```
tbpg: Should be using `cloud.google.com/go/pubsub` rather than one of the versioned packages.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,51 @@
...
+// [START pubsub_subscriber_concurrency_control]
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/pubsub"
+)
+
+func pullMsgsConcurrent(w io.Writer, projectID, subName string) error {
```
tbpg: Is this function tested?

hongalex: We currently don't have a good way to test this sample. 

https://github.com/GoogleCloudPlatform/golang-samples/blob/121fb3eec469dc207cc7ac3af7fdd7c394d862e5/pubsub/subscriptions/pull_concurrency.go#L41, `sub.Receive` is long running until the context is cancelled. I can refactor the sample to take in an external context, but I would probably need to add another comment like `ctx := context.Background` which might be confusing. What do you think?

tbpg: Ah. That's problematic...

Tests seem more important than the `context.Context` pattern. What do you think of passing a `context.Context` in or doing `ctx := context.WithTimeout(ctx, 5 * time.Second)` in the sample? I'm good with either, so whatever you prefer.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,51 @@
...
+	ctx := context.Background()
+	client, err := pubsub.NewClient(ctx, projectID)
+	if err != nil {
+		return fmt.Errorf("pubsub.NewClient: %v", err)
+	}
+
+	sub := client.Subscription(subName)
+	// NumGoroutines is the number of goroutines Receive will spawn to pull messages concurrently.
+	sub.ReceiveSettings.NumGoroutines = 4
+	// If ReceiveSettings.Synchronous is set to true, NumGoroutines is overriden to 1. To enable
```
anguillanneuf: I would prefer a little more emphasis on concurrency settings here by saying something like `Must set Synchronous to false to enable concurrency settings. Otherwise NumGoroutines will default to 1` and reverse the order of line 40 & line 37.

pubsub/subscriptions/pull_settings.go
```diff
@@ -34,7 +34,11 @@ func pullMsgsSettings(w io.Writer, projectID, subName string) error {
 
 	sub := client.Subscription(subName)
 	sub.ReceiveSettings.Synchronous = true
+	// MaxOutstandingMessages is the maximum number of unprocessed messages (unacknowledged but not yet expired).
+	// This is only guaranteed with ReceiveSettings.Synchronous is set to true.
```
anguillanneuf: `with` -> `when`?

pubsub/subscriptions/sync_pull.go
tbpg: Not sure I fully understand. Sounds good to me, but adding @jadekler for additional review.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,64 @@
...
+	// Must set ReceiveSettings.Synchronous to false to enable concurrency settings.
+	// Otherwise, NumGoroutines will be set to 1.
+	sub.ReceiveSettings.Synchronous = false
+	// NumGoroutines is the number of goroutines sub.Receive will spawn to pull messages concurrently.
+	sub.ReceiveSettings.NumGoroutines = numGoroutines
+
+	// Receive messages for 10 seconds.
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+	var msgs []string
```
jadekler: How about a channel? Otherwise sub.Receive becomes serialized.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,64 @@
...
+	var lock sync.Mutex
+	err = sub.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
+		lock.Lock()
+		defer lock.Unlock()
+		msgs = append(msgs, string(msg.Data))
+		fmt.Fprintf(w, "Got message: %s\n", string(msg.Data))
+		msg.Ack()
```
jadekler: If the program crashes before msgs gets handled, all these messages are lost. Is that OK? Should we document that this should only be used in programs where loss is OK?

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,64 @@
...
+	client, err := pubsub.NewClient(ctx, projectID)
+	if err != nil {
+		return nil, fmt.Errorf("pubsub.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	sub := client.Subscription(subName)
+	// Must set ReceiveSettings.Synchronous to false to enable concurrency settings.
+	// Otherwise, NumGoroutines will be set to 1.
+	sub.ReceiveSettings.Synchronous = false
```
jadekler: This is the default, right? If so, do we need to explicitly set it?

pubsub/subscriptions/subscription_test.go
```diff
@@ -155,6 +156,84 @@ func TestIAM(t *testing.T) {
...
+			AckDeadline: 20 * time.Second,
+		})
+		if err != nil {
+			t.Fatalf("CreateSubscription: %v", err)
+		}
+	}
+
+	maxMessages := 5
+	buf := new(bytes.Buffer)
+	msgs, err := pullMsgsSync(buf, tc.ProjectID, subName, topic, int32(maxMessages))
```
jadekler: What's the advantage of this over Synchronous=true + Receive?

pubsub/subscriptions/sync_pull.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	var msgs, ackIDs []string
+	for _, msg := range resp.GetReceivedMessages() {
+		ackIDs = append(ackIDs, msg.GetAckId())
+		message := string(msg.GetMessage().Data)
+		fmt.Fprintf(w, "Got message %q\n", message)
+		msgs = append(msgs, message)
+	}
+
+	subClient.Acknowledge(ctx, &pb.AcknowledgeRequest{
```
jadekler: As before, doing an ack before the msg actually gets handled is dangerous.

pubsub/subscriptions/sync_pull.go
hongalex: @jadekler Is the comment clear enough? If you don't think so, we can remove all uses of the gapic client altogether and just focus on the functionality exposed in the handwritten client.

pubsub/subscriptions/pull_concurrency.go
jadekler: In google-cloud-go we tend to do,

```
_ = msg // TODO: handle message.
```

I think this is a bit more obvious, but will leave it to you to make the call.

hongalex: Done.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,80 @@
...
+	// NumGoroutines is the number of goroutines sub.Receive will spawn to pull messages concurrently.
+	sub.ReceiveSettings.NumGoroutines = numGoroutines
+
+	// Receive messages for 10 seconds.
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	var numMsgs uint64
+	// Create a channel to send messages to as they come in.
```
anguillanneuf: Nit: Instead of saying "send", can we change it to "handle"?

pubsub/subscriptions/subscription_test.go
```diff
@@ -155,6 +158,103 @@ func TestIAM(t *testing.T) {
...
+	buf := new(bytes.Buffer)
+	err = pullMsgsSync(buf, tc.ProjectID, subName, int32(maxMessages))
+	if err != nil {
+		t.Fatalf("failed to pull messages: %v", err)
+	}
+	got := buf.String()
+	want := "Got message \"hello world #4\""
+	if !strings.Contains(got, want) {
+		t.Fatalf("got output: %v\nwant %v", got, want)
+	}
```
anguillanneuf: There's some redundancy in the test. Ten messages got published, but the script pulls for 5 of them, then checks if message #4 is there. Can we pull all ten messages and make sure message #0-9 are all there?

pubsub/subscriptions/sync_pull.go
anguillanneuf: This method is useful when only a specific number of messages are wanted by an end application that takes a long time to process each incoming message. Because Pub/Sub can flood the end application with tons of messages, many messages would make wasted trips across the network if we don't avoid asking for them. 

@hongalex [The docs site](https://cloud.google.com/pubsub/docs/pull#synchronous_pull) mentions that this is a REST method, not gRPC, is that correct? If not, we may still want to call out that this is a [`PullRequest`](https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#google.pubsub.v1.PullRequest) is not [`StreamingPullRequest`](https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#streamingpullrequest).

pubsub/subscriptions/subscription_test.go
hongalex: I want the sync pull test to only pull max 5 messages. The publish methods sends 10 so that we can test that sync pull works properly. I can change the test to not be 0 indexed to be less confusing if that helps?

anguillanneuf: Okay, publishing more than the number you want to pull makes sense. 

I think Indexed messages are fine. I guess I don't understand how you could make sure `#4` is in among the pulled messages.

hongalex: Oh you're totally right. Maybe I'll check for number of lines instead then. Good catch!

hongalex:
@anguillanneuf, after further consideration, I think we should remove the example detailing the use of the gapic client. We should only have 1 synchronous pull sample (as you mentioned) that handles lease management. I'll be updating this PR shortly.

pubsub/subscriptions/pull_settings.go
```diff
@@ -34,7 +34,16 @@ func pullMsgsSettings(w io.Writer, projectID, subName string) error {
 
 	sub := client.Subscription(subName)
 	sub.ReceiveSettings.Synchronous = true
+	// MaxOutstandingMessages is the maximum number of unprocessed messages the
+	// client will pull from the server before pausing.
+	// This is only guaranteed when ReceiveSettings.Synchronous is set to true.
+	// When Synchronous is set to false, the StreamingPull RPC is used which
+	// can pull a single large batch of messages at once that is greater than
+	// MaxOustandingMessages before pausing.
```
anguillanneuf: Thank you for adding this!

pubsub/subscriptions/sync_pull.go
```diff
@@ -0,0 +1,64 @@
...
+	// Turn on synchronous mode. This makes the subscriber use the Pull RPC rather
+	// than the StreamingPull RPC, which is useful for guaranteeing MaxOutstandingMessages,
+	// the max number of messages the client will hold in memory.
+	sub.ReceiveSettings.Synchronous = true
+	sub.ReceiveSettings.MaxOutstandingMessages = 10
+
+	cctx, cancel := context.WithCancel(ctx)
+	err = sub.Receive(cctx, func(ctx context.Context, msg *pubsub.Message) {
+		mu.Lock()
+		defer mu.Unlock()
```
anguillanneuf: Just want to understand it better. How does this protect `received`?

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,80 @@
...
+	// Create a channel to handle messages to as they come in.
+	cm := make(chan *pubsub.Message)
+
+	// Handle individual messages in a goroutine.
+	go func() {
+		for {
+			select {
+			case msg := <-cm:
+				_ = msg // TODO: handle message
+				atomic.AddUint64(&numMsgs, 1)
```
tbpg: I don't think this needs to be atomic? There is only one goroutine that ever adds to `numMsgs`, so there won't be a race condition.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,80 @@
...
+				return
+			}
+		}
+	}()
+
+	err = sub.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
+		cm <- msg
+	})
+	if err != nil {
+		return fmt.Errorf("Error in Receive: %v", err)
```
tbpg:
```suggestion
		return fmt.Errorf("Receive: %v", err)
```

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,80 @@
...
+		return fmt.Errorf("pubsub.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	sub := client.Subscription(subName)
+	// Must set ReceiveSettings.Synchronous to false to enable concurrency settings.
+	// Otherwise, NumGoroutines will be set to 1.
+	sub.ReceiveSettings.Synchronous = false
+	// NumGoroutines is the number of goroutines sub.Receive will spawn to pull messages concurrently.
+	sub.ReceiveSettings.NumGoroutines = numGoroutines
```
tbpg: Does `numGoroutines` need to be an argument? If it's consistent with samples in other languages, it's probably OK to keep. But, in general, prefer fewer arguments if they aren't needed for tests.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Receive messages for 10 seconds.
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
tbpg: Would 5 seconds work? This means tests will take > 10 seconds and I've been trying to speed overall tests up.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -0,0 +1,80 @@
...
+				atomic.AddUint64(&numMsgs, 1)
+				msg.Ack()
+			case <-ctx.Done():
+				close(cm)
+				return
+			}
+		}
+	}()
+
+	err = sub.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
```
tbpg: Does `Receive` block until `ctx` is `Done`? Might be worth adding a comment. Right now, it kind of reads like this will receive one message then be done.

pubsub/subscriptions/pull_settings.go
```diff
@@ -34,7 +34,18 @@ func pullMsgsSettings(w io.Writer, projectID, subName string) error {
...
+	// client will pull from the server before pausing.
+	//
+	// This is only guaranteed when ReceiveSettings.Synchronous is set to true.
+	// When Synchronous is set to false, the StreamingPull RPC is used which
+	// can pull a single large batch of messages at once that is greater than
+	// MaxOustandingMessages before pausing. For more info, see
+	// https://cloud.google.com/pubsub/docs/pull#streamingpull_dealing_with_large_backlogs_of_small_messages.
 	sub.ReceiveSettings.MaxOutstandingMessages = 10
+	// MaxOutstandingBytes is the maximum size of unprocessed messages,
+	// that the client will pull from the server before pausing.
```
tbpg: Does this also have the caveat above with `Synchronous`? Maybe say to `See MaxOutstandingMessages`?

pubsub/subscriptions/subscription_test.go
```diff
@@ -155,6 +158,69 @@ func TestIAM(t *testing.T) {
...
+
+	buf := new(bytes.Buffer)
+	err := pullMsgsSync(buf, tc.ProjectID, subName, topic)
+	if err != nil {
+		t.Fatalf("failed to pull messages: %v", err)
+	}
+	// Check for number of newlines, which should correspond with number of messages.
+	got := strings.Count(buf.String(), "\n")
+	if got != numMessages {
```
tbpg:
```suggestion
	if got := strings.Count(buf.String(), "\n"); got != numMessages {
```

pubsub/subscriptions/subscription_test.go
```diff
@@ -155,6 +158,69 @@ func TestIAM(t *testing.T) {
...
+		t.Fatalf("got %d messages, want %d", got, numMessages)
```
tbpg:
```suggestion
		t.Fatalf("pullMsgsSync got %d messages, want %d", got, numMessages)
```

pubsub/subscriptions/subscription_test.go
```diff
@@ -155,6 +158,69 @@ func TestIAM(t *testing.T) {
...
+		result.Get(ctx)
+	}
+
+	buf := new(bytes.Buffer)
+	if err := pullMsgsConcurrenyControl(buf, tc.ProjectID, subName, 4); err != nil {
+		t.Fatalf("failed to pull messages: %v", err)
+	}
+	got := buf.String()
+	want := fmt.Sprintf("Received %d messages", numMsgs)
+	if !strings.Contains(got, want) {
```
tbpg:
```suggestion
	if got := buf.String(); !strings.Contains(got, want) {
```

pubsub/subscriptions/subscription_test.go
```diff
@@ -155,6 +158,69 @@ func TestIAM(t *testing.T) {
...
+		t.Fatalf("got output: %v, want %v", got, want)
```
tbpg:
```suggestion
		t.Fatalf("pullMsgsConcurrenyControl got:\n----\n%v\n----\nWant to contain:\n----%v\n----", got, want)
```

pubsub/subscriptions/sync_pull.go
```diff
@@ -0,0 +1,64 @@
...
+		return fmt.Errorf("pubsub.NewClient: %v", err)
+	}
+
+	var mu sync.Mutex
+	received := 0
+	sub := client.Subscription(subName)
+
+	// Turn on synchronous mode. This makes the subscriber use the Pull RPC rather
+	// than the StreamingPull RPC, which is useful for guaranteeing MaxOutstandingMessages,
+	// the max number of messages the client will hold in memory.
```
tbpg: Does `Synchronous = true` mean only one message will be processed by `Receive` at a time? If so, do we need the mutex? Why a mutex instead of atomic? Or send to a channel processed by a single goroutine like the other sample?

pubsub/subscriptions/sync_pull.go
```diff
@@ -0,0 +1,64 @@
...
+	sub.ReceiveSettings.MaxOutstandingMessages = 10
+
+	cctx, cancel := context.WithCancel(ctx)
+	err = sub.Receive(cctx, func(ctx context.Context, msg *pubsub.Message) {
+		mu.Lock()
+		defer mu.Unlock()
+		fmt.Fprintf(w, "Got message :%q\n", string(msg.Data))
+		_ = msg // TODO: handle message.
+		msg.Ack()
+		if received++; received == 10 {
```
tbpg: Optional: I might put the `received++` on its own line. That's the important part and `++` doesn't return a value. Inline `if` declarations are normally used for declarations of stuff used in the `if` statement (like errors and `got`).

pubsub/subscriptions/subscription_test.go
```diff
@@ -155,6 +158,69 @@ func TestIAM(t *testing.T) {
...
+
+func TestPullMsgsConcurrencyControl(t *testing.T) {
+	ctx := context.Background()
+	tc := testutil.SystemTest(t)
+	client := setup(t)
+
+	topic := client.Topic(topicName)
+
+	// Publish 100 message to test with.
+	numMsgs := 100
```
tbpg: Why 100? Looks like only 10 are pulled? (OK if we need 100, but I was confused.)

pubsub/subscriptions/sync_pull.go
hongalex: I was having a weird data race error, but have since resolved it. Removed.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -37,42 +36,40 @@ func pullMsgsConcurrenyControl(w io.Writer, projectID, subName string, numGorout
...
 	// Create a channel to handle messages to as they come in.
 	cm := make(chan *pubsub.Message)
-
 	// Handle individual messages in a goroutine.
 	go func() {
 		for {
 			select {
 			case msg := <-cm:
+				fmt.Fprintf(w, "Got message :%q\n", string(msg.Data))
 				_ = msg // TODO: handle message
```
tbpg: Delete now that we're printing `msg.Data`?

pubsub/subscriptions/sync_pull.go
```diff
@@ -44,18 +43,17 @@ func pullMsgsSync(w io.Writer, projectID, subName string, topic *pubsub.Topic) e
...
-		mu.Lock()
-		defer mu.Unlock()
+	// Receive messages for 5 seconds.
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	// Receive blocks until the passed in context is done.
+	err = sub.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
 		fmt.Fprintf(w, "Got message :%q\n", string(msg.Data))
 		_ = msg // TODO: handle message.
```
tbpg: Delete since `msg.Data` is printed?

pubsub/subscriptions/sync_pull.go
```diff
@@ -44,18 +43,17 @@ func pullMsgsSync(w io.Writer, projectID, subName string, topic *pubsub.Topic) e
...
 		msg.Ack()
-		if received++; received == 10 {
-			cancel()
-		}
 	})
-	if err != nil && err != context.Canceled {
+	if err != nil && err == context.Canceled {
```
tbpg: It's an error if the context gets cancelled? Why check `!= nil` and `== context.Canceled`?

pubsub/subscriptions/async_pull.go
```diff
@@ -57,8 +57,9 @@ func pullMsgs(w io.Writer, projectID, subName string, topic *pubsub.Topic) error
 	sub := client.Subscription(subName)
 	cctx, cancel := context.WithCancel(ctx)
 	err = sub.Receive(cctx, func(ctx context.Context, msg *pubsub.Message) {
-		msg.Ack()
 		fmt.Fprintf(w, "Got message: %q\n", string(msg.Data))
+		_ = msg // TODO: handle message.
```
tbpg: Delete since `msg.Data` is printed?

hongalex:
@tbpg Seems like the recent builds are failing with this error:

```
--- FAIL: TestBadFiles (0.03s)
    badfiles_test.go:134: Likely bad file checked in: .github/renovate.json. MIME type: {  }
```

Is this possibly related to https://github.com/Pandinosaurus/golang-samples/pull/131?

go.mod
```diff
@@ -43,9 +43,9 @@ require (
 	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45
 	golang.org/x/text v0.3.2
 	google.golang.org/api v0.11.0
-	google.golang.org/appengine v1.6.5
+	google.golang.org/appengine v1.6.1
```
tbpg: I don't think we need the `go.mod` and `go.sum` updates?

pubsub/subscriptions/async_pull.go
tbpg: Ping.

pubsub/subscriptions/sync_pull.go
hongalex: Fixed. This was actually entirely wrong, since instead of checking for `context.Cancelled`, I needed to check for the grpc `CANCELLED` code.

hongalex:
I don't feel good about the test flakiness issue being resolved. Marking this as WIP until I figure out what's causing the tests to fail occasionally.

pubsub/subscriptions/sync_pull.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"cloud.google.com/go/pubsub"
+)
+
+func pullMsgsSync(w io.Writer, projectID, subName string, topic *pubsub.Topic) error {
+	// projectID := "my-project-id"
+	// subName := projectID + "-example-sub"
```
tbpg: Same `subID` comment.

PR #942, opened 2019-08-13
add gimme-acc tool for SA HMAC testing

jkwlui:


testing/gimme-acc/Makefile
```diff
@@ -0,0 +1,17 @@
+.PHONY: bin cross upload
```
frankyn: Add copyright to Makefile.

testing/gimme-acc/README.md
```diff
@@ -0,0 +1,36 @@
...
+### Example use in integration tests
+
+```
+set -e -x
+
+curl https://storage.googleapis.com/gimme-acc/linux_amd64/gimme-acc > gimme-acc
+chmod +x gimme-acc
+./gimme-acc version
+
+export HMAC_SERVICE_ACCOUNT=$(./gimme-acc -project gimme-acc lease 15m)
```
frankyn: please switch up the project of `gimme-acc` to maybe `your-project-id`. It confused me a little bit at first seeing double.

testing/gimme-acc/main.go
```diff
@@ -0,0 +1,288 @@
+// Copyright 2019 Google LLC
```
frankyn: Is there a test?

jkwlui:
Thanks @tbpg. 

A bit of context here - we needed a service account pool to test a new feature in GCS where users are able to create keys against service accounts. Because the limitations on the number of keys you can have on a service account, languages that run parallels tests in the CI would run into problems if they used the same service accounts.

I think you're right - this is essentially the same thing as `gimmeproj`, perhaps we should just close this and reuse the same binaries but plug in a different project to `-project`? @frankyn

frankyn:
+1, thanks @tbpg! I missed that support in this tool.

PR #944, opened 2019-08-14
functions/ocr: Go samples for OCR function

LogicalShark:
Using similar flow as existing Python and NodeJS examples.


functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ocr
+
+// [START functions_ocr_detect]
```
tbpg: Generally include the package for Cloud Functions samples so people can copy and paste the whole thing.

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	pubsub "cloud.google.com/go/pubsub/apiv1"
+	"cloud.google.com/go/translate"
+	vision "cloud.google.com/go/vision/apiv1"
+	"golang.org/x/text/language"
+	pb "google.golang.org/genproto/googleapis/cloud/vision/v1"
```
tbpg: Normally imported as `visionpb`.

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+	// fileName := "menu.jpg"
+	fmt.Fprintf(w, "Looking for text in image %v", fileName)
+	ctx := context.Background()
+	visionClient, err := vision.NewImageAnnotatorClient(ctx)
+	if err != nil {
+		return fmt.Errorf("vision.NewImageAnnotatorClient: %v", err)
+	}
+	maxResults := 1
+	annotations, err := visionClient.DetectTexts(ctx,
+		&pb.Image{
```
tbpg: Create a separate variable for the image and change this function call to a single line.

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+				GcsImageUri: fmt.Sprintf("gs://%s/%s", bucketName, fileName),
+			},
+		},
+		&pb.ImageContext{}, maxResults,
+	)
+	if err != nil {
+		return fmt.Errorf("DetectTexts: %v", err)
+	}
+	text := ""
+	if len(annotations) > 0 {
```
tbpg: Return an error if `len(annotations) == 0`?

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+		text = annotations[0].Description
+	}
+	fmt.Fprintf(w, "Extracted text %q from image (%d chars).", text, len(text))
```
tbpg: Return an error if `len(text) == 0`? I assume `DetectLanguage` would not work in that case.

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return fmt.Errorf("DetectLanguage: %v", err)
+	}
+	if len(detectResponse) == 0 || len(detectResponse[0]) == 0 {
+		return fmt.Errorf("DetectLanguage gave empty response")
+	}
+	srcLang := detectResponse[0][0].Language.String()
+	fmt.Fprintf(w, "Detected language %q for text %q.", srcLang, text)
+
+	data, err := ioutil.ReadFile("config.json")
```
tbpg: Use `os.Open` and `json.Decoder` to avoid having to manage `data`.

tbpg: Should probably read `config.json` in an `init()` function to avoid reading+parsing every time the function is invoked. Same with initializing all of the clients.

The `setup` function should do all of this and save it in global variables. Probably name it `init`, though, so it runs during package initialization.

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+
+// [START functions_ocr_detect]
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	pubsub "cloud.google.com/go/pubsub/apiv1"
```
tbpg: Just `cloud.google.com/go/pubsub` -- a manually written wrapper around the specific versions. Will need to update calls below.

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// Submit a message to the bus for each target language
+	publisher, err := pubsub.NewPublisherClient(ctx)
+	if err != nil {
+		return fmt.Errorf("translate.NewClient: %v", err)
+	}
+	for _, targetLang := range config.ToLang {
+		topicName := config.TranslateTopic
+		if srcLang == targetLang || srcLang == "und" {
```
tbpg: What is `und`?

functions/ocr/app/setup.go
```diff
@@ -0,0 +1,85 @@
...
+
+type config struct {
+	ResultTopic    string   `json:"RESULT_TOPIC"`
+	ResultBucket   string   `json:"RESULT_BUCKET"`
+	TranslateTopic string   `json:"TRANSLATE_TOPIC"`
+	Translate      bool     `json:"TRANSLATE"`
+	ToLang         []string `json:"TO_LANG"`
+}
+
+type ocrmessage struct {
```
tbpg: ocrMessage

https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+	config := &config{}
+	err = json.Unmarshal(data, config)
+	if err != nil {
+		return fmt.Errorf("json.Unmarshal: %v", err)
+	}
+
+	// Submit a message to the bus for each target language
+	publisher, err := pubsub.NewPublisherClient(ctx)
+	if err != nil {
+		return fmt.Errorf("translate.NewClient: %v", err)
```
tbpg: `pubsub.NewClient`

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,129 @@
...
+		message, err := json.Marshal(ocrmessage{
+			Text:     text,
+			FileName: fileName,
+			Lang:     targetTag,
+			SrcLang:  srcTag,
+		})
+		if err != nil {
+			return fmt.Errorf("json.Marshal: %v", err)
+		}
+		messageData := []byte(message)
```
tbpg: Probably don't need this variable. Only used once.

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+	"golang.org/x/text/language"
+	pubsubpb "google.golang.org/genproto/googleapis/pubsub/v1"
+)
+
+const (
+	menuName        = "menu.jpg"
+	signName        = "sign.png"
+	bucketName      = "result-bucket-test"
+	imageBucketName = "ocr-image-bucket123"
```
tbpg: Bucket names should be based on the project ID. That way, there is less likely to be an issue with being able to use whatever bucket for running tests. Also, tests run in parallel in multiple projects for golang-samples, so we have to be careful about race conditions. For example, see https://github.com/GoogleCloudPlatform/golang-samples/blob/master/asset/quickstart/export-assets/main.go#L37 (not a test, but the same idea).

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"cloud.google.com/go/storage"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+	"golang.org/x/text/language"
+	pubsubpb "google.golang.org/genproto/googleapis/pubsub/v1"
+)
+
+const (
+	menuName        = "menu.jpg"
+	signName        = "sign.png"
```
tbpg: Do these files need to be checked in? Do Python and Java use a common bucket we can reuse here? In general, it's good to avoid checking binary files into version control.

functions/ocr/app/detect.go
LogicalShark: Language tag string for undefined, should I Errorf instead?

functions/ocr/app/detect.go
LogicalShark: Done

functions/ocr/app/ocr_test.go
LogicalShark: The other version have an images folder, but since the files are also in cloud storage they don't need to be in the local directory. Should I remove them?

functions/ocr/app/ocr_test.go
LogicalShark: Done

functions/ocr/app/setup.go
LogicalShark: Done

functions/ocr/app/ocr_test.go
tbpg: Yes, please remove them.

LogicalShark: Done

functions/ocr/app/save.go
```diff
@@ -0,0 +1,58 @@
...
+		err := json.Unmarshal(messageData, &message)
+		if err != nil {
+			return fmt.Errorf("json.Unmarshal: %v", err)
+		}
+	} else {
+		return fmt.Errorf("Empty data")
+	}
+	text := message.Text
+	fileName := message.FileName
+	lang := message.Lang
```
tbpg: Probably don't need these variables - can reference directly. Even if the later lines are longer, you don't have to keep these new variables in your head as you're reading.

functions/ocr/app/save.go
```diff
@@ -0,0 +1,58 @@
...
+
+	log.Printf("Received request to save file %q.", fileName)
+
+	bucketName := config.ResultBucket
+	resultFilename := fmt.Sprintf("%s_%s.txt", fileName, lang)
+	bucket := storageClient.Bucket(bucketName)
+
+	log.Printf("Saving result to %q in bucket %q.", resultFilename, bucketName)
+
+	file := bucket.Object(resultFilename).NewWriter(ctx)
```
tbpg: `w` is a common name for writers.

functions/ocr/app/setup.go
```diff
@@ -0,0 +1,109 @@
...
+	"time"
+
+	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/storage"
+	"cloud.google.com/go/translate"
+	vision "cloud.google.com/go/vision/apiv1"
+	"golang.org/x/text/language"
+)
+
+type configType struct {
```
tbpg: Maybe `configuration`, with variables named `config`? Weird to have `Type` in the name of a type.

functions/ocr/app/setup.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	d := json.NewDecoder(cfgFile)
+	config = &configType{}
+	err = d.Decode(config)
+	if err != nil {
+		log.Fatalf("Decode: %v", err)
+	}
+
+	projectID := config.ProjectID
```
tbpg: Delete.

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
+// // Copyright 2019 Google LLC
```
tbpg: Slashes are doubled up -- `// //`.

functions/ocr/app/process.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"context"
+	"fmt"
+	"log"
+)
+
+// ProcessImage is executed when a file is uploaded to the Cloud Storage bucket you created for uploading images.
+// It runs detectText, which processes the image for text.
+func ProcessImage(ctx context.Context, event GCSEvent) error {
```
tbpg: All of the Cloud Functions need to call `setup` to make sure the config is parsed and clients are initialized, right?

functions/ocr/app/setup.go
```diff
@@ -0,0 +1,109 @@
...
+
+var (
+	visionClient    *vision.ImageAnnotatorClient
+	translateClient *translate.Client
+	publisher       *pubsub.Client
+	storageClient   *storage.Client
+	config          *configType
+)
+
+func setup(ctx context.Context) {
```
tbpg: Should probably return an error.

functions/ocr/app/setup.go
```diff
@@ -0,0 +1,109 @@
...
+	cfgFile, err := os.Open("config.json")
+	if err != nil {
+		log.Fatalf("os.Open: %v", err)
+	}
+
+	d := json.NewDecoder(cfgFile)
+	config = &configType{}
+	err = d.Decode(config)
```
tbpg: Inline.

functions/ocr/app/setup.go
```diff
@@ -0,0 +1,109 @@
...
+var (
+	visionClient    *vision.ImageAnnotatorClient
+	translateClient *translate.Client
+	publisher       *pubsub.Client
+	storageClient   *storage.Client
+	config          *configType
+)
+
+func setup(ctx context.Context) {
+	cfgFile, err := os.Open("config.json")
```
tbpg: Should be able to run this over and over without consequence. So, it should check `if cconfig == nil`and initialize it if so. Same with the other clients. This also makes it so tests will pass (since you can initialize `config` and the clients before calling the functions).

functions/ocr/app/translate.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	var message ocrMessage
+	if event.Data != nil {
+		messageData := event.Data
+		err := json.Unmarshal(messageData, &message)
+		if err != nil {
+			return fmt.Errorf("json.Unmarshal: %v", err)
+		}
+	} else {
+		return fmt.Errorf("Empty data")
```
tbpg: Do this check first (same comment as before).

functions/ocr/app/translate.go
```diff
@@ -0,0 +1,100 @@
...
+
+// TranslateText is executed when a message is published to the Cloud Pub/Sub topic specified
+// by TRANSLATE_TOPIC in config.json, and translates the text using the Google Translate API.
+func TranslateText(ctx context.Context, event PubSubMessage) error {
+	if event.Data == nil {
+		return fmt.Errorf("Empty data")
+	}
+	var message ocrMessage
+	if event.Data != nil {
+		messageData := event.Data
```
tbpg: Delete.

functions/ocr/app/translate.go
```diff
@@ -0,0 +1,100 @@
...
+		err := json.Unmarshal(messageData, &message)
```
tbpg: Inline.

functions/ocr/app/translate.go
```diff
@@ -0,0 +1,100 @@
...
+			return fmt.Errorf("json.Unmarshal: %v", err)
+		}
+	} else {
+		return fmt.Errorf("Empty data")
+	}
+
+	text := message.Text
+	fileName := message.FileName
+	targetTag := message.Lang
+	srcTag := message.SrcLang
```
tbpg: Delete these variables.

functions/ocr/app/translate.go
```diff
@@ -0,0 +1,100 @@
...
+	translateResponse, err := translateClient.Translate(ctx, []string{text}, targetTag, &opts)
+	if err != nil {
+		return fmt.Errorf("Translate: %v", err)
+	}
+	if len(translateResponse) == 0 {
+		return fmt.Errorf("Empty Translate response")
+	}
+	translatedText := translateResponse[0]
+
+	topicName := config.ResultTopic
```
tbpg: Delete.

functions/ocr/app/translate.go
```diff
@@ -0,0 +1,100 @@
...
+		return fmt.Errorf("Exists: %v", err)
+	}
+	if !ok {
+		topic, err = publisher.CreateTopic(ctx, topicName)
+		if err != nil {
+			return fmt.Errorf("CreateTopic: %v", err)
+		}
+	}
+	r := topic.Publish(ctx,
+		&pubsub.Message{
```
tbpg: Separate `msg` variable and make this a single line `err` declaration in the `if` statement.

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,108 @@
...
+			return fmt.Errorf("Exists: %v", err)
+		}
+		if !ok {
+			topic, err = publisher.CreateTopic(ctx, topicName)
+			if err != nil {
+				return fmt.Errorf("CreateTopic: %v", err)
+			}
+		}
+		r := topic.Publish(ctx,
+			&pubsub.Message{
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/GoogleCloudPlatform/golang-samples/functions/ocr/app
+
+go 1.12
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/ocr_test.go
LogicalShark: Fixed

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
...
+	menuName = "menu.jpg"
+	signName = "sign.png"
+)
+
+var (
+	bucketName      string
+	imageBucketName string
+)
+
+func TestMain(t *testing.T) {
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
...
+	ctx := context.Background()
+	tc := testutil.SystemTest(t)
+	var err error
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
...
+	imageBucketName = fmt.Sprintf("%s-image", tc.ProjectID)
+	config = &configType{
+		ProjectID:      tc.ProjectID,
+		ResultTopic:    "test-result-topic",
+		ResultBucket:   bucketName,
+		TranslateTopic: "test-translate-topic",
+		Translate:      true,
+		ToLang:         []string{"en", "fr", "es", "ja", "ru"},
+	}
+	projectID := config.ProjectID
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
...
+		log.Fatalf("translate.NewClient: %v", err)
+	}
+
+	storageClient, err = storage.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("storage.NewClient: %v", err)
+	}
+}
+
+func TestSaveResult(t *testing.T) {
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Errorf("language.Parse: %v", err)
+	}
+	data, err := json.Marshal(ocrMessage{
+		Text:     "Hello",
+		FileName: menuName,
+		Lang:     en,
+		SrcLang:  fr,
+	})
+	log.SetOutput(buf)
+	err = SaveResult(ctx, PubSubMessage{
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
...
+		Data: data,
+	})
+	if err != nil {
+		t.Errorf("TestSaveResult: %v", err)
```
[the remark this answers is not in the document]
LogicalShark: Fixed

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	r, err := bkt.Object(fmt.Sprintf("%s_%s.txt", menuName, en)).NewReader(ctx)
+	if err != nil {
+		t.Errorf("NewReader: %v", err)
+	}
+	fbuf := make([]byte, 100, 100)
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
...
+	if want := "Bonjour"; !strings.Contains(got, want) {
+		t.Errorf("got %q, want %q", got, want)
+	}
+}
+
+func TestDetectText(t *testing.T) {
+	ctx := context.Background()
+	testutil.SystemTest(t)
+	buf := new(bytes.Buffer)
+	storageClient.Bucket(imageBucketName)
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,170 @@
...
+	err := detectText(ctx, imageBucketName, menuName)
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/process.go
```diff
@@ -0,0 +1,41 @@
...
+
+package ocr
+
+import (
+	"context"
+	"fmt"
+	"log"
+)
+
+// ProcessImage is executed when a file is uploaded to the Cloud Storage bucket you created for uploading images.
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/process.go
```diff
@@ -0,0 +1,41 @@
...
+// It runs detectText, which processes the image for text.
+func ProcessImage(ctx context.Context, event GCSEvent) error {
+	if event.Bucket == "" {
+		return fmt.Errorf("empty file.Bucket")
+	}
+	if event.Name == "" {
+		return fmt.Errorf("empty file.Name")
+	}
+	err := detectText(ctx, event.Bucket, event.Name)
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/save.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"log"
+)
+
+// SaveResult is executed when a message is published to the Cloud Pub/Sub topic specified by
+// RESULT_TOPIC in config.json file, and saves the data packet to a file in GCS.
+func SaveResult(ctx context.Context, event PubSubMessage) error {
+	var message ocrMessage
+	if event.Data != nil {
+		messageData := event.Data
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/save.go
```diff
@@ -0,0 +1,58 @@
...
+		err := json.Unmarshal(messageData, &message)
+		if err != nil {
+			return fmt.Errorf("json.Unmarshal: %v", err)
+		}
+	} else {
```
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/save.go
LogicalShark: Done

functions/ocr/app/save.go
LogicalShark: Done

functions/ocr/app/setup.go
LogicalShark: Changed

functions/ocr/app/setup.go
LogicalShark: Done, was leftover from init

functions/ocr/app/setup.go
LogicalShark: Done

functions/ocr/app/setup.go
LogicalShark: Done

functions/ocr/app/translate.go
LogicalShark: Done

functions/ocr/app/translate.go
LogicalShark: Done

functions/ocr/app/translate.go
LogicalShark: Done

functions/ocr/app/translate.go
LogicalShark: Done

functions/ocr/app/translate.go
LogicalShark: Done

functions/ocr/app/translate.go
LogicalShark: Done

functions/ocr/app/setup.go
LogicalShark: Done

functions/ocr/app/process.go
LogicalShark: Yes

functions/ocr/app/ocr_test.go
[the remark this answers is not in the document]
LogicalShark: Done

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,186 @@
...
+	imageBucketName string
+)
+
+// TestMain sets up the config rather than using the config file
+// which contains placeholder values.
+func TestMain(m *testing.M) {
+	ctx := context.Background()
+	tc, ok := testutil.ContextMain(m)
+	if !ok {
+		log.Fatalf("testutil.ContextMain failed")
```
tbpg: Should not use `log.Fatal` in tests - it stops the whole test suite. If `ContextMain` failed, it's probably because `GOLANG_SAMPLES_PROJECT_ID` is not set, in which case we shouldn't fail.

```
log.Print("GOLANG_SAMPLES_PROJECT_ID is unset. Skipping.")
return
```

functions/ocr/app/process.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"fmt"
+	"log"
+)
+
+// ProcessImage is executed when a file is uploaded to the Cloud Storage bucket you
+// created for uploading images. It runs detectText, which processes the image for text.
+func ProcessImage(ctx context.Context, event GCSEvent) error {
+	err := setup(ctx)
```
tbpg: Inline and update error message to be `"setup: %v"`.

Same with the other calls.

functions/ocr/app/setup.go
```diff
@@ -0,0 +1,119 @@
...
+
+// PubSubMessage is the payload of a Pub/Sub event.
+type PubSubMessage struct {
+	Data []byte `json:"data"`
+}
+
+var (
+	visionClient    *vision.ImageAnnotatorClient
+	translateClient *translate.Client
+	publisher       *pubsub.Client
```
tbpg: Name this `pubsubClient` for consistency with the other clients?

functions/ocr/app/translate.go
```diff
@@ -0,0 +1,91 @@
...
+	if !ok {
+		topic, err = publisher.CreateTopic(ctx, config.ResultTopic)
+		if err != nil {
+			return fmt.Errorf("CreateTopic: %v", err)
+		}
+	}
+	msg := &pubsub.Message{
+		Data: messageData,
+	}
+	_, err = topic.Publish(ctx, msg).Get(ctx)
```
tbpg: Inline.

functions/ocr/app/ocr_test.go
LogicalShark: Done

functions/ocr/app/process.go
LogicalShark: Done

functions/ocr/app/setup.go
LogicalShark: It's publisher in the other languages, changing anyway for consistency

functions/ocr/app/translate.go
LogicalShark: Done

functions/ocr/app/ocr_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"log"
+	"os"
+	"strings"
+	"testing"
+
+	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/storage"
+	"cloud.google.com/go/translate"
+	vision "cloud.google.com/go/vision/apiv1"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
```
tbpg: Sorry, we should get rid of this dependency and `os.Getenv("GOLANG_SAMPLES_PROJECT_ID")` manually. Can define a function to do the check to make it easy to call. Having this dependency means `github.com/GoogleCloudPlatform/golang-samples` shows up in the `go.mod` and gets downloaded whenever the function is deployed, which will slow deployments down.

functions/ocr/app/detect.go
```diff
@@ -0,0 +1,106 @@
...
+	if len(annotations) > 0 {
+		text = annotations[0].Description
+	}
+	if len(annotations) == 0 || len(text) == 0 {
+		log.Printf("No text detected in image %q. Returning early.", fileName)
+		return nil
+	}
+	log.Printf("Extracted text %q from image (%d chars).", text, len(text))
+
+	detectResponse, err := translateClient.DetectLanguage(ctx, []string{text})
```
tbpg: Why do we need to detect the source language? The Translate API can auto-detect the source language when translating, so this API call is not needed.

LogicalShark: It's in the other samples, I think if it detects if the text is already in the target language it uploads it to the result bucket without calling the translate cloud function

functions/ocr/app/ocr_test.go
LogicalShark: Done

functions/ocr/app/detect.go
LogicalShark: Also I noticed the TRANSLATE bool in config isn't checked anywhere in the other samples, should I leave it out here as well?

tbpg: Ahh. That makes sense. Yeah, I would delete it from the config if it's not referenced anywhere.

cc @grant

PR #946, opened 2019-08-15
healthcare: add DICOMweb methods and tests

noerog:
Also minor changes in tests for the UIDs, and update dicomweb_study_retrieve.go to download a study file.


healthcare/dicom_test.go
```diff
@@ -86,14 +99,92 @@ func TestDICOMStore(t *testing.T) {
...
 	testutil.Retry(t, 10, 2*time.Second, func(r *testutil.R) {
-		if err := dicomWebRetrieveStudy(ioutil.Discard, tc.ProjectID, location, datasetID, dicomStoreID, "studies/1.3.6.1.4.1.11129.5.5.111396399361969898205364400549799252857604"); err != nil {
-			r.Errorf("dicomWebSearchInstances got err: %v", err)
+		// Remove the output file if it already exists.
+		os.Remove(studyOutputFile)
+		buf.Reset()
+
+		err := dicomWebRetrieveStudy(buf, tc.ProjectID, location, datasetID, dicomStoreID, studyUID, studyOutputFile)
+		if err != nil {
+			t.Errorf("Testing dicomWebRetrieveStudy: %v", err)
```
tbpg: No need for "Testing".

https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures (OK to not print args here since it's not a table test).

tbpg: Use `r` instead of `t` for all of the failures to trigger the retry logic.

healthcare/dicom_test.go
```diff
@@ -86,14 +99,92 @@ func TestDICOMStore(t *testing.T) {
...
+
+		got := buf.String()
+
+		if want := "Study retrieved and downloaded to file: study.multipart\n"; !strings.Contains(got, want) {
+			t.Errorf("got %q, want %q", got, want)
+		}
+
+		stat, err := os.Stat(studyOutputFile)
+		if err != nil {
+			t.Error("Could not stat output file")
```
tbpg: Include error. Probably:

```
r.Errorf("os.Stat: %v", err)
```

Same below.

healthcare/dicomweb_study_retrieve.go
```diff
@@ -50,10 +50,15 @@ func dicomWebRetrieveStudy(w io.Writer, projectID, location, datasetID, dicomSto
 		return fmt.Errorf("could not read response: %v", err)
 	}
 
-	if resp.StatusCode > 299 {
```
tbpg: Probably good to check the status code?

healthcare/dicomweb_study_retrieve.go
```diff
@@ -50,10 +50,15 @@ func dicomWebRetrieveStudy(w io.Writer, projectID, location, datasetID, dicomSto
...
-		return fmt.Errorf("SearchForInstances: status %d %s: %s", resp.StatusCode, resp.Status, respBytes)
+	if err = ioutil.WriteFile(outputFile, respBytes, 0644); err != nil {
```
tbpg: I'm generally hesitant to do IO -- do we need to write the file out? Was there confusion about what to do with the bytes?

healthcare/dicom_test.go
noerog: Done. You still need to print err, correct?

healthcare/dicomweb_study_retrieve.go
noerog: Simply writing out the bytes is not particularly useful and doesn't demonstrate what users would actually do. Our other samples all write the bytes out to a file so it would also be good to be consistent with those.

healthcare/dicomweb_instance_retrieve.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	healthcare "google.golang.org/api/healthcare/v1beta1"
+)
+
+// dicomWebRetrieveInstance retrieves a specific instance.
+func dicomWebRetrieveInstance(w io.Writer, projectID, location, datasetID, dicomStoreID, dicomWebPath string, outputFile string) error {
```
tbpg: Sorry I missed this in the first review. We added a new style guide for samples and now want to include sample values for all of the arguments to functions.

See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#function-arguments for the format.

healthcare/dicomweb_instance_retrieve.go
```diff
@@ -0,0 +1,67 @@
...
+	call := storesService.RetrieveInstance(parent, dicomWebPath)
+	call.Header().Set("Accept", "application/dicom; transfer-syntax=*")
+	resp, err := call.Do()
+	if err != nil {
+		return fmt.Errorf("RetrieveInstance: %v", err)
+	}
+
+	defer resp.Body.Close()
+
+	respBytes, err := ioutil.ReadAll(resp.Body)
```
tbpg: We can avoid reading the file into memory using `io.Copy` (need to move the `StatusCode` check above this):

```
file, err := os.Create(outputFile)
if err != nil {
	return fmt.Errorf("os.Create: %v", err)
}
if _, err := io.Copy(file, resp.Body); err != nil {
	return fmt.Errorf("io.Copy: %v", err)
}
```

healthcare/dicomweb_study_retrieve.go
tbpg: SGTM. Thanks.

healthcare/dicomweb_instance_retrieve.go
noerog: Done for all of the files in this PR.

healthcare/dicomweb_instance_retrieve.go
noerog: Done for this and the other files.

healthcare/dicomweb_instance_retrieve.go
```diff
@@ -19,13 +19,19 @@ import (
...
 )
 
 // dicomWebRetrieveInstance retrieves a specific instance.
 func dicomWebRetrieveInstance(w io.Writer, projectID, location, datasetID, dicomStoreID, dicomWebPath string, outputFile string) error {
+        // projectID := fmt.Sprintf("my-project")
+        // location := fmt.Sprintf("us-central1")
+        // datasetID := fmt.Sprintf("my-dataset")
+        // dicomStoreID := fmt.Sprintf("my-dicom-store")
+        // dicomWebPath := fmt.Sprintf("studies/1.3.6.1.4.1.11129.5.5.1113639985/series/1.3.6.1.4.1.11129.5.5.1953511724/instances/1.3.6.1.4.1.11129.5.5.9562821369")
+        // outputFile := fmt.Sprintf("instance.dcm")
```
tbpg: No need to `fmt.Sprintf` normal strings.

healthcare/dicomweb_instance_retrieve.go
```diff
@@ -46,18 +52,18 @@ func dicomWebRetrieveInstance(w io.Writer, projectID, location, datasetID, dicom
...
-                return fmt.Errorf("RetrieveInstance: status %d %s: %s", resp.StatusCode, resp.Status, respBytes)
+	if resp.StatusCode > 299 {
+                return fmt.Errorf("RetrieveInstance: status %d %s: %s", resp.StatusCode, resp.Status, resp.Body)
         }
 
-	if err = ioutil.WriteFile(outputFile, respBytes, 0644); err != nil {
-		return err
-	}
+	file, err := os.Create(outputFile)
+        if err != nil {
```
tbpg: Need to run `gofmt`. Your editor should be able to do it on save.

healthcare/dicomweb_instance_retrieve.go
noerog: Done for all files.

healthcare/dicomweb_instance_retrieve.go
noerog: Done for all files.

PR #951, opened 2019-08-22
getting-started/bookshelf: initial rewrite

tbpg:
Areas to review:
* Should this be more than one package? I moved everything to `package main`.
* Is the design easy to understand? Should we move files/packages around?
* `func NewBookshelf` is a little weird to me. Is there a better way to write it?
* Should we use `Bookshelf.logWriter`, is there a better way to do it, or should we just print logs when testing?

Summary of changes:
* Switch to GAE Standard (go112).
* Close() returns an error.
* Booskshelf methods take a context.
* Remove Datastore and replace it with Firestore.
* Remove auth/login.
* Remove databases other than Firestore and Memory.
* Remove GCE and GKE deployments.
* Add a go.mod and go.sum.
* Move webtest into this directory to avoid depending on golang-samples.
* Move all Go code into a root directory package main.
* Make handlers methods on Bookshelf, rather than standalone functions.
* Always initialize storage (see config.go) and give runtime error on
  image upload.
* Firestore doc IDs stored as part of the Book in Firestore, rather than
  assigned during editing and listing.


tbpg:
cc @dmitshur for review (can't add as a PR reviewer).

getting-started/bookshelf/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/GoogleCloudPlatform/golang-samples/getting-started/bookshelf
+
+go 1.13
+
+require (
+	cloud.google.com/go v0.44.3
+	github.com/gofrs/uuid v3.2.0+incompatible
```
dmitshur: Perhaps consider switching to another version that isn't incompatible.

For example, `github.com/gofrs/uuid/v3 v3.1.2` or `github.com/google/uuid v1.1.1`.

getting-started/bookshelf/db_firestore.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+// firestoreDB persists books to Cloud Firestore.
+// See https://cloud.google.com/firestore/docs.
+type firestoreDB struct {
+	client *firestore.Client
+}
+
+// Ensure firestoreDB conforms to the BookDatabase interface.
+var _ BookDatabase = &firestoreDB{}
```
dmitshur: This is completely okay as is, but consider moving it to a _test.go file since it's not a piece of code that is needed for functionality of the app.

Doing so causes a failure for `*firestoreDB` to implement the `BookDatabase` interface to be reported when you do `go test` rather than `go build`, which feels more fitting IMO.

getting-started/bookshelf/main.go
```diff
@@ -50,43 +46,44 @@ func main() {
...
+
+	log.Printf("Listening on localhost:%s", port)
 	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
 }
 
-func registerHandlers() {
+func (b *Bookshelf) registerHandlers() {
 	// Use gorilla/mux for rich routing.
-	// See http://www.gorillatoolkit.org/pkg/mux
+	// See http://www.gorillatoolkit.org/pkg/mux.
```
dmitshur: s,http://,https://,

getting-started/bookshelf/main.go
```diff
@@ -243,78 +210,50 @@ func uploadFileFromForm(r *http.Request) (url string, err error) {
...
-	b, err := json.Marshal(bookID)
-	if err != nil {
-		return
-	}
-	topic := bookshelf.PubsubClient.Topic(bookshelf.PubsubTopicID)
-	_, err = topic.Publish(ctx, &pubsub.Message{Data: b}).Get(ctx)
-	log.Printf("Published update to Pub/Sub for Book ID %d: %v", bookID, err)
-}
-
 // http://blog.golang.org/error-handling-and-go
```
dmitshur: s,http://,https://,

getting-started/bookshelf/db_memory.go
```diff
@@ -27,43 +29,45 @@ var _ BookDatabase = &memoryDB{}
...
 	return &memoryDB{
-		books:  make(map[int64]*Book),
+		books:  make(map[string]*Book),
 		nextID: 1,
 	}
 }
 
 // Close closes the database.
-func (db *memoryDB) Close() {
+func (db *memoryDB) Close(ctx context.Context) error {
```
dmitshur: The parameter `ctx` is unused, so it'd be slightly better to leave it out, as you have already done in the `firestoreDB.Close` method.

getting-started/bookshelf/db_firestore.go
tbpg: I think I like it in here - when `firestoreDB` stops implementing `BookDatabase`, you know right away in your editor, unlike when it's in the test file (for VS Code, at least).

getting-started/bookshelf/db_memory.go
tbpg: Done.

getting-started/bookshelf/go.mod
tbpg: Unfortunately, that version also doesn't work. So, sticking with incompatible.

getting-started/bookshelf/main.go
tbpg: Done.

getting-started/bookshelf/main.go
tbpg: Done.

getting-started/bookshelf/db_memory.go
```diff
@@ -12,58 +12,61 @@
...
 // memoryDB is a simple in-memory persistence layer for books.
 type memoryDB struct {
 	mu     sync.Mutex
-	nextID int64           // next ID to assign to a book.
-	books  map[int64]*Book // maps from Book ID to Book.
+	nextID int64            // next ID to assign to a book.
+	books  map[string]*Book // maps from Book ID to Book.
 }
 
 func newMemoryDB() *memoryDB {
```
broady: memoryDB is not used anymore either

getting-started/bookshelf/main.go
```diff
@@ -0,0 +1,279 @@
...
+		return appErrorf(err, "could not list books: %v", err)
+	}
+
+	return listTmpl.Execute(w, r, books)
+}
+
+// bookFromRequest retrieves a book from the database given a book ID in the
+// URL's path.
+func (b *Bookshelf) bookFromRequest(r *http.Request) (*Book, error) {
+	id := mux.Vars(r)["id"]
```
broady: valid if empty string?

getting-started/bookshelf/main.go
```diff
@@ -0,0 +1,279 @@
...
+func (b *Bookshelf) updateHandler(w http.ResponseWriter, r *http.Request) *appError {
+	id := mux.Vars(r)["id"]
+
+	book, err := b.bookFromForm(r)
+	if err != nil {
+		return appErrorf(err, "could not parse book from form: %v", err)
+	}
+	book.ID = id
+
+	if err := b.DB.UpdateBook(r.Context(), book); err != nil {
```
broady: `ctx := r.Context()` at top of each function

getting-started/bookshelf/main_test.go
```diff
@@ -16,47 +16,57 @@ package main
...
 
 func TestMain(m *testing.M) {
+	projectID := os.Getenv("GOLANG_SAMPLES_FIRESTORE_PROJECT")
+	if projectID == "" {
+		fmt.Fprintln(os.Stderr, "GOLANG_SAMPLES_FIRESTORE_PROJECT not set. Skipping.")
+		return
+	}
+
+	// Don't log anything during testing.
+	log.SetOutput(ioutil.Discard)
```
broady: maybe capture output in a buffer and print only if tests failed? probably doesn't matter

getting-started/bookshelf/main_test.go
```diff
@@ -16,47 +16,57 @@ package main
...
-var wt *webtest.W
+var (
+	wt *webtest.W
+	b  *Bookshelf
+)
 
 func TestMain(m *testing.M) {
+	projectID := os.Getenv("GOLANG_SAMPLES_FIRESTORE_PROJECT")
+	if projectID == "" {
+		fmt.Fprintln(os.Stderr, "GOLANG_SAMPLES_FIRESTORE_PROJECT not set. Skipping.")
```
broady: use in-memory DB for tests, firestore for e2e tests?

getting-started/bookshelf/db_memory.go
broady: but maybe it should be (for tests? see comment below)

getting-started/bookshelf/db_firestore.go
broady: Won't it fail for `go test`, too? Package must compile for the tests to run.

getting-started/bookshelf/config.go
```diff
@@ -12,133 +12,92 @@
...
-	// [END auth]
-
-	// [START sessions]
-	// Configure storage method for session-wide information.
-	// Update "something-very-secret" with a hard to guess string or byte sequence.
-	cookieStore := sessions.NewCookieStore([]byte("something-very-secret"))
-	cookieStore.Options = &sessions.Options{
-		HttpOnly: true,
+	var err error
+	b.DB, err = configureFirestoreDB(projectID)
```
broady: I think configureFirestoreDB can be inlined now.

getting-started/bookshelf/config.go
```diff
@@ -12,133 +12,92 @@
...
-	// PubsubClient, err = configurePubsub("<your-project-id>")
-	// [END pubsub]
 
+	// This Cloud Storage bucket must exist to be able to upload book pictures.
+	// You can create it and make it public by running:
+	//     gsutil mb my-project_bucket
+	//     gsutil defacl set public-read gs://my-project_bucket
+	// replacing my-project with your project ID.
+	b.StorageBucketName = projectID + "_bucket"
+	b.StorageBucket, err = configureStorage(b.StorageBucketName)
```
broady: and configureStorage (inline)

getting-started/bookshelf/config.go
```diff
@@ -12,133 +12,92 @@
...
-	//
-	// PubsubClient, err = configurePubsub("<your-project-id>")
-	// [END pubsub]
 
+	// This Cloud Storage bucket must exist to be able to upload book pictures.
+	// You can create it and make it public by running:
+	//     gsutil mb my-project_bucket
+	//     gsutil defacl set public-read gs://my-project_bucket
+	// replacing my-project with your project ID.
+	b.StorageBucketName = projectID + "_bucket"
```
broady: Why the `_bucket` suffix?

Since this is going to GAE, is the `<project>.appspot.com` "default bucket" not suitable? I can't remember whether this is meant to be used or not.

getting-started/bookshelf/main.go
```diff
@@ -0,0 +1,279 @@
...
+		PublishedDate: r.FormValue("publishedDate"),
+		ImageURL:      imageURL,
+		Description:   r.FormValue("description"),
+	}
+
+	return book, nil
+}
+
+// uploadFileFromForm uploads a file if it's present in the "image" form field.
+func (b *Bookshelf) uploadFileFromForm(r *http.Request) (url string, err error) {
```
broady: pass in ctx explicitly rather than getting it from r.

getting-started/bookshelf/config.go
broady: If the bucket wasn't created, I think the first error the user/developer will see is when they upload a picture. That seems OK but maybe a friendly message to tell them the commands to make the bucket would be good.

getting-started/bookshelf/main_test.go
tbpg: Will leave as-is, but we can change if if/when this fails and we need debugging.

getting-started/bookshelf/main_test.go
tbpg: Done. It ended up changing the API quite a bit -- now pass a `BookDatabase` to `NewBookshelf`. It makes sense for testing, but is a slightly awkward API.

I updated this to always test the in memory database and only test the Firestore one if it's able to.

getting-started/bookshelf/config.go
tbpg: Matching the existing tutorial with the name.

The error when you try to upload an image is "could not parse book from form: could not upload file: bucket "my-project_bucket" does not exist: check config.go". The `gsutil mb` and `defacl` explanation is a bit long, so I left it as a reference to config.go.

getting-started/bookshelf/db_memory.go
tbpg: Added to tests - see other comment.

getting-started/bookshelf/main.go
```diff
@@ -229,6 +247,7 @@ func (b *Bookshelf) createHandler(w http.ResponseWriter, r *http.Request) *appEr
 
 // updateHandler updates the details of a given book.
 func (b *Bookshelf) updateHandler(w http.ResponseWriter, r *http.Request) *appError {
+	ctx := r.Context()
 	id := mux.Vars(r)["id"]
```
broady: check empty ID here too

tbpg: Done.

PR #955, opened 2019-08-26
getting-started/sessions: initial commit

tbpg:
~~firestoregorilla is not yet released, so this doesn't compile yet. This is the sample that will use it.~~

~~I'll add a go.mod once firestoregorilla is released.~~


tbpg:
`firestore-gorilla-sessions` was released. I updated this PR to use it.

Please take another look.

getting-started/sessions/main.go
```diff
@@ -0,0 +1,114 @@
...
+	name := "color-views"
+	session, err := a.store.Get(r, name)
+	if err != nil {
+		// Could not get the session. Log an error and continue, saving a new
+		// session.
+		log.Printf("store.Get: %v", err)
+	}
+
+	if session.IsNew {
+		// firestoregorilla uses JSON, which unmarshals numbers as float64s.
```
broady: you can use int:

https://play.golang.org/p/-YQT8hnnOWj

tbpg: I think it works for unmarshaling into a typed struct. But, the JSON is unmarshaled into `map[string]interface{}` and the default numeric type seems to be `float64`.

```
panic: interface conversion: interface {} is float64, not int [recovered]
	panic: interface conversion: interface {} is float64, not int
```

https://play.golang.org/p/crHAwrzDvOV

PR #958, opened 2019-08-27
vision: add product_search purge_products_in_product_set sample

davidcavazos:
Adding `product_search` sample for `purge_products_in_product_set`.


davidcavazos:
Looks like tests are failing because the service is being updated. Will comment back when tests are passing.

vision/product_search/purge_products_in_product_set.go
```diff
@@ -0,0 +1,64 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package productsearch contains samples for Google Cloud Vision API Product Search.
+package productsearch
+
+// [START vision_product_search_purge_products_in_product_set]
+
```
tbpg: Delete newline between, otherwise this shows up when embedded.

vision/product_search/purge_products_in_product_set.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	vision "cloud.google.com/go/vision/apiv1"
+	visionpb "google.golang.org/genproto/googleapis/cloud/vision/v1"
+)
+
+// purgeProductsInProductSet deletes all products in a product set.
+func purgeProductsInProductSet(w io.Writer, projectID string, location string, productSetID string, force bool) error {
```
tbpg: Please give example values for the arguments other than `w`.

tbpg: Does `force` need to be an argument? In general, try to have as few arguments as possible, keeping what's needed for testing.

vision/product_search/purge_products_in_product_set.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+		Force: force,
+	}
+
+	// The purge operation is async.
+	op, err := c.PurgeProducts(ctx, req)
+	if err != nil {
+		return fmt.Errorf("PurgeProducts: %v", err)
+	}
+	fmt.Fprintf(w, "Processing operation name: %s\n", op.Name())
```
tbpg: Consider using `%q` to get quotes around the string.

vision/product_search/purge_products_in_product_set_test.go
```diff
@@ -0,0 +1,93 @@
...
+	if err := listProductsInProductSet(&buf, tc.ProjectID, location, productSetID); err != nil {
+		t.Fatalf("listProductsInProductSet: %v", err)
+	}
+	if got := buf.String(); strings.Contains(got, productID) {
+		t.Errorf("Product ID %s was not removed", productID)
+	}
+
+	// Clean up.
+	deleteProduct(&buf, tc.ProjectID, location, productID)
+	deleteProductSet(&buf, tc.ProjectID, location, productSetID)
```
tbpg: Consider deferring these calls right after the resources get created.

vision/product_search/purge_products_in_product_set.go
davidcavazos: Sure, done

vision/product_search/purge_products_in_product_set.go
davidcavazos: Great point! Done

davidcavazos: Got it, I was just following the [canonical Python example](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2349/files), but it does make sense, so I'll hardcode the `force: true`.

vision/product_search/purge_products_in_product_set.go
davidcavazos: Thanks!

vision/product_search/purge_products_in_product_set_test.go
davidcavazos: Good catch, done

davidcavazos:
Library is ready and tests are passing locally, as well as in `go19` and `go110`. Tests in other versions might be failing due to Go modules.

PR #961, opened 2019-08-28
vision: add product_search purge_orphan_products sample

davidcavazos:
Adding `product_search` sample for `purge_orphan_products`.


davidcavazos:
Looks like tests are failing because the service is being updated. Will comment back when tests are passing.

davidcavazos:
Sure thing, moving this to #958

Issue #89, opened 2016-08-23
testing: add more e2e tests


broady:
Need more support for doing temporary string replacements before deployment. Or adding files to the deployment (e.g., to set up config.go for bookshelf)

broady:
Storage added.

tbpg:
We have a few and @grayside has been moving forward for Cloud Run samples. Closing this.

PR #963, opened 2019-08-30
internal/e2e: switch to only one GAE Standard test

tbpg:
Related to #89.

E2E tests are currently failing since GAE Flex doesn't support modules, which were enabled in #660.

This will also have the benefit (?) of not doing 3 flex deploys when running E2E tests, which take a while.

`flexible_test.go` shows up as deleted, but it was really renamed to `standard_test.go`. I changed `bodyShouldContain` to use `testutil.Retry`. I also removed comments/build tags about Go 1.7 (since we only support 1.9+).


tbpg:
Failure caused by an unrelated back-end issue. Submitting TBR to fix the build.

PR #964, opened 2019-08-30
run: add gRPC service on Cloud Run sample

mchmarny:
Simple example of creating and deploying gRPC service on Cloud Run.


steren:
CC @grayside

mchmarny:
@grayside PTAL

grayside:
I've started a review on this, but after thinking it over and discussing with Mark, we're probably not going to use this right now, and instead incorporate this into something in the future. Leaving for now while planning is in motion.

tbpg:
Marking as do not merge until that point.

tbpg:
Going to close this PR for now, but feel free to reopen later if we decide on needing it.

PR #966, opened 2019-09-04
mod: update

broady:
$ go get -u
$ go mod tidy


broady:
Possibly two real failures in pubsub and bigquery.

PR #983, opened 2019-09-15
profiler: remove the docdemo sample, merge it with hotapp

aalexand:
Add a flag to hotapp sample to run some local work to make the flame
graph more interesting, thus allowing to use it as a doc sample.

Tested with command:

go run main.go -project_id oval-time-515 -service aalexand-test -version 1.0.0 -local_work

Which gives

Z9aYQGS0XGE


tbpg:
cc @padelaney

profiler/hotapp/main.go
```diff
@@ -37,6 +37,8 @@ var (
 	version = flag.String("version", "1.0.0", "service version")
 	// Skew of foo1 function over foo2, in the CPU busyloop, to simulate diff.
 	skew = flag.Int("skew", 100, "skew of foo2 over foo1: foo2 will consume skew/100 CPU time compared to foo1 (default is no skew)")
+	// Whether to run some local CPU work to increase the self metric.
+	localWork = flag.Bool("local_work", false, "whether to run some local CPU work")
```
tbpg: Remind me again why we can't do the local work every time?

padelaney: The original version of this program is used for benchmarking. We want to preserve that capability.

PR #985, opened 2019-09-17
testing: include region tags in XML test output

tbpg:
This can be used to associate a region tag with the test that covers it.

Current coverage:
```
573/1126 (50.89%) of region tags are tested.
```

See the package comment for additional details.

This ends up with XML like:

```xml
<testsuites>
	<testsuite tests="12" failures="0" time="0.039" name="github.com/GoogleCloudPlatform/golang-samples/functions/http">
		<properties>
			<property name="go.version" value="go1.14"></property>
		</properties>
		<testcase classname="http" name="TestCORSEnabledFunctionAuth" time="0.000">
			<properties>
				<property name="region_tags" value="functions_http_cors_auth"></property>
			</properties>
		</testcase>
		<testcase classname="http" name="TestCORSEnabledFunctionAuthPOST" time="0.000">
			<properties>
				<property name="region_tags" value="functions_http_cors_auth"></property>
			</properties>
		</testcase>
```


testing/sampletests/main.go
```diff
@@ -0,0 +1,245 @@
...
+// like:
+//     TestCreate__spanner_create__bigtable_create
+// Again, this is an acceptable tradeoff. It only happens when both regions are
+// tested. If only one is tested, only one would show up. Failures are
+// attributed to both snippets, though.
+//
+// The test coverage over all regions is printed at the end.
+//
+// To get the number of unique region tags in the repo manually, run:
+//     grep -RoPh '\[START \K(.+)\]' | sort | uniq | wc -l
```
BenWhitehead: you could also use `sort -u` in place of `sort | uniq` so that it's all done when sorting. For large datasets this tends to be faster.

testing/sampletests/main.go
```diff
@@ -0,0 +1,245 @@
...
+				}
+			}
+			if file == nil {
+				return nil, fmt.Errorf("file for %q not found", id.Name)
+			}
+			path, exact := astutil.PathEnclosingInterval(file, id.Pos(), id.End())
+			if !exact {
+				return nil, fmt.Errorf("PathEnclosingInterval got not exact path for %q in %v", id.Name, file)
+			}
+			ast.Inspect(path[1], func(node ast.Node) bool {
```
ace-n: I'm assuming this `ast` utility won't translate to other languages?

testing/sampletests/main.go
```diff
@@ -0,0 +1,245 @@
...
+	"regexp"
+	"strings"
+
+	"golang.org/x/tools/go/ast/astutil"
+	"golang.org/x/tools/go/packages"
+	"golang.org/x/tools/go/types/typeutil"
+)
+
+var startRe = regexp.MustCompile("\\[START ([^\\]]+)\\]")
+var endRe = regexp.MustCompile("\\[END ([^\\]]+)\\]")
```
broady:
```suggestion
var (
	startRe = regexp.MustCompile("\\[START ([^\\]]+)\\]")
	endRe   = regexp.MustCompile("\\[END ([^\\]]+)\\]")
)
```

testing/sampletests/main.go
```diff
@@ -0,0 +1,245 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The sampletests command adds the region tags tested by each test to the test
+// name. It reads from stdin and writes to stdout.
+//
+// For example:
+//     TestFoo -> TestFoo__foo_hello_world__foo_hello_gopher
```
broady: in this case, are there two regions?

testing/sampletests/main.go
```diff
@@ -0,0 +1,245 @@
...
+			path, exact := astutil.PathEnclosingInterval(file, id.Pos(), id.End())
+			if !exact {
+				return nil, fmt.Errorf("PathEnclosingInterval got not exact path for %q in %v", id.Name, file)
+			}
+			ast.Inspect(path[1], func(node ast.Node) bool {
+				call, ok := node.(*ast.CallExpr)
+				if !ok {
+					return true
+				}
+				calleeObj := typeutil.Callee(pkg.TypesInfo, call)
```
broady: will this get callees of function references?

e.g.

```
func Foo() {}
func Bar() {}
func Baz() {}

func TestX() {
	var funcs []func() = []func(){Foo, Bar, Baz}
	...
}
```

testing/sampletests/main.go
tbpg: Yes. Updated the comment.

testing/sampletests/main.go
tbpg: Done.

testing/sampletests/main.go
tbpg: No, it's specific to Go. Other languages might have other AST parsers, but this one won't work.

testing/sampletests/main.go
tbpg: Done.

testing/sampletests/main.go
tbpg: No, I don't think so. The `CallExpr` where those functions are called will end up being in a `for` loop over `funcs`. Would need to look for transitive function calls. `TestX` uses/references `Foo`, which calls `DoFoo`, which is part of region `do_foo`.

broady:
An idea:
https://godoc.org/golang.org/x/tools/go/callgraph

Another idea:
Is there call graph information in the test coverage output? I don't think so but that would seem promising

ace-n:
@broady we've already considered coverage output, and found that it didn't really suit our needs. (I'll ping you a link internally to the design doc section that covers this.)

tbpg:
> https://godoc.org/golang.org/x/tools/go/callgraph
Looks promising... Looking.

> Is there call graph information in the test coverage output? I don't think so but that would seem promising
Unfortunately, no. Coverage only says if a line ran, not what test caused it to run (Java does). We could run coverage per-test, but that would be too slow.

broady:
> Unfortunately, no. Coverage only says if a line ran, not what test caused it to run (Java does). We could run coverage per-test, but that would be too slow.

+1, and some tests would need to be re-written. some packages depend on test ordering:

e.g.
```
func TestCreateBucket(...) {}
func TestCreateObject(...) {}
```

tbpg:
Pending approval from @ace-n, @broady, and Kokoro, I think this is ready to go.

I changed the behavior of this to add `properties` to the test cases rather than change the names of the tests. I also changed the logic to actually look within each package, rather than doing a global search and replace, so the results are more accurate.

@broady, I tried using `callgraph` (as you can see from the commit history), but I couldn't get it working the same way. I'm sure it's possible, but I think this works for today.

ace-n:
LGTM assuming you've confirmed our internal systems can access this.

(They'll have to be updated accordingly; so maybe that's something we should sync w/ @jmdobry at some point about.)

testing/sampletests/main.go
```diff
@@ -0,0 +1,260 @@
...
+			return err
+		}
+
+		lines := strings.Split(string(src), "\n")
+		for lineNum, line := range lines {
+			if regions := startRe.FindAllStringSubmatch(line, -1); len(regions) > 0 {
+				region := regions[0][1]
+				uniqueRegionTags[region] = struct{}{}
+				for _, r := range testFileRanges[path] {
+					if r.start >= lineNum && r.end >= lineNum {
```
codyoss, line 172: Should these both be `>=`?

testing/sampletests/main.go
```diff
@@ -0,0 +1,365 @@
...
+			ast.Inspect(path[1], func(node ast.Node) bool {
+				call, ok := node.(*ast.CallExpr)
+				if !ok {
+					return true
+				}
+				callee := typeutil.StaticCallee(pkg.TypesInfo, call)
+				if callee == nil {
+					return true
+				}
+				if callee.Pkg() != obj.Pkg() {
```
codyoss, line 220: Do we ever test from external packages in this repo? "hello" != "hello_test". Should this be a special case?

codyoss:
LGTM, just the one final questions. Having the tests and being able to play with them made this much easier to reason about. Thanks for adding those!

testing/sampletests/main.go
tbpg: Good idea! I checked and there are some `_test` packages, so I added support.

PR #988, opened 2019-09-18
spanner: drop leaderboard db before trying to create

tbpg:
There is some flake caused by the db not being dropped and the create
operation failing (happens if tests are cancelled in the middle, before
delete runs).


tbpg:
Updated to only drop the database if it already exists. Good to start fresh with the database because I've seen other errors about rows already existing in the table, so I'd rather just start from scratch.

tbpg:
Submitting TBR to fix the test. The Kokoro infra issue is being worked on in #989.

PR #991, opened 2019-09-19
getting-started/sessions: switch to greetings & add regions

tbpg:


getting-started/sessions/main.go
```diff
@@ -71,7 +83,7 @@ func newApp(projectID string) (*app, error) {
 		log.Fatalf("firestoregorilla.New: %v", err)
 	}
 
-	tmpl, err := template.New("Index").Parse("<body bgcolor={{.color}}>Views {{.views}}</body>")
+	tmpl, err := template.New("Index").Parse(`<body>{{.views}} {{if eq .views 1.0}}view{{else}}views{{end}} for "{{.greeting}}"</body>`)
```
fhinkel: Are you sure you want to work with a template here instead of just posting plain text? I know that the template is much nicer and idiomatic - but is it confusing for the devs looking at this tutorial for the firs time?

tbpg: Yes. See https://github.com/GoogleCloudPlatform/golang-samples/pull/955#discussion_r317703082.

getting-started/sessions/main.go
```diff
@@ -102,13 +119,17 @@ func (a *app) index(w http.ResponseWriter, r *http.Request) {
...
 	}
 	session.Values["views"] = session.Values["views"].(float64) + 1
 	if err := session.Save(r, w); err != nil {
 		log.Printf("Save: %v", err)
 		// Don't return early so the user still gets a response.
 	}
 
-	a.tmpl.Execute(w, session.Values)
+	if err := a.tmpl.Execute(w, session.Values); err != nil {
+		log.Printf("Execute: %v", err)
```
jadekler: Why are these log.Printf instead of log.Exitf?

jadekler: Oh it's an HTTP app. Nevermind.

PR #992, opened 2019-09-19
monitoring: add retries and unique names to metric get/delete tests

tbpg:


tbpg:
Still failing, so I added the do not merge label and reached out internally.

tbpg:
Switched to unique names (per advice I was given) and tests are passing now. There are some optimizations when the same metric/name is used.

PR #1003, opened 2019-09-27
profiler: add README.md

padelaney:
Add a readme file that contains usage instructions and links to the public documentation.


profiler/README.md
```diff
@@ -0,0 +1,107 @@
...
+### hotapp
+
+The sample `hotapp` is uses an infinite loop with two call stacks.
+
+The [Profiler documentation](https://cloud.google.com/profiler/docs)
+includes images generated from this sample. 
+If you wish to generate profile data consistent with that included in the Profiler documentation,
+run the `hotapp` service with the following command line options:
+```
+       go run main.go -service=docdemo-service -local_work -skew=75 -version=1.75.0
```
aalexand: The indent is redundant here.

profiler/README.md
```diff
@@ -0,0 +1,107 @@
...
+To execute a sample and collect profiling data in your GCP project, do the following:
+
+1.  If you have a new GCP, you need to enable the Profiler API for your project. Choose one of the following methods.
+
+    From the Cloud console, go to **APIs & Services** and then click **Enable APIS and Services**.
+    Search for **Profiler**.  If the API isn't enabled, click **Enable**.
+
+    From the Cloud Shell, run the following command:
+    
+```
```
aalexand: Indent the whole backtick block, there shouldn't be added indent for the code.

profiler/README.md
```diff
@@ -0,0 +1,107 @@
...
+    gcloud services enable cloudprofiler.googleapis.com
+``` 
+
+2.  If you aren't running on GCP, then you need to create a service account. For details on these steps, see 
+    [Profiling applications runing outside of Google Cloud Platform](https://cloud.google.com/profiler/docs/profiling-external).
+    
+3.  From your clone of the GitHub repository, change to the source directory of the program you want to execute.
+    For example, the following command changes the working directory to that for the sample `hotapp`:
+```
```
aalexand: Same here and for all code blocks.

profiler/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+A few seconds after you start the program, the message `profiler has started` is displayed.
+New messages are displayed each time a profile is uploaded to your GCP project.
+To stop the program, enter `Ctrl-C`.
+
+### Viewing your data
+
+To view your profile data, do the following:
+
+1. Go to the [Cloud console](https://console.cloud.google.com).
```
aalexand: Should this be "Cloud Console" rather than "Cloud console"? See https://cloud.google.com/cloud-console/.

I am also unsure about using "the" with both "Cloud Console" and "Cloud Shell". https://cloud.google.com/cloud-console/ doesn't use that and I think it's more common that proper product names go without "the". I may be wrong, I am an ESL person.

profiler/README.md
padelaney: Changes were made. Thanks.

profiler/README.md
padelaney: Changes made. Thanks.

profiler/README.md
padelaney: Changes made. Thanks.

profiler/README.md
padelaney: Changed  casing. Checked w/ editors, we normally don't use "the" with Cloud Shell but do use "the" with Cloud Console. Thanks.

padelaney:
> We generally avoid creating README files because they require extra maintenance and we prefer for people to use the cloud.google.com documentation directly. Did someone ask for this or are we not able to embed the necessary docs? Could we have a short README that essentially says, "See cloud.google.com/..."?

The request was made by TL aalexand to have user-facing documentation within the repo for ease of use. The README does reference the official documentation. The README is expected to be slowly changing. I support your merge decision.

Updated per feedback from aalexand.    Thanks!

PR #1013, opened 2019-10-09
appengine,appengine_flexible,docs/appengine: delete unused samples

### LINKED ISSUE #1005: appengine: Remove appengine examples prior to go11x
Go 1.9 on App Engine Standard is deprecated and new deploys are no longer allowed. All users must migrate to `runtime: go111` or later. We can delete the old appengine examples and anything that mentions `goapp`.

tbpg:
I left the websockets sample since we will probably want it in the
future.

Fixes #1005. There might be others that we can delete eventually, but here is a first pass.

I didn't find any internal references to any of this code.


tbpg:
cc @sbuss @juliehockett

PR #1017, opened 2019-10-12
Pubsub: Correct subscription reference

Elgarni:
The current code sample references a Subscription by concatenating the project id with the subscription identifier:
```go
func pullMsgs(w io.Writer, projectID, subName string, topic *pubsub.Topic) error {
	// subName := projectID + "-example-sub"
        ...
```  

However, only the subscription `id` is needed, so that it can be used to form the fully qualified name. The following is from the pubsub [source code](https://github.com/googleapis/google-cloud-go/blob/master/pubsub/subscription.go#L61):
```go
// SubscriptionInProject creates a reference to a subscription in a given project.
func (c *Client) SubscriptionInProject(id, projectID string) *Subscription {
	return &Subscription{
		c:    c,
		name: fmt.Sprintf("projects/%s/subscriptions/%s", projectID, id),
	}
}
```

This PR changes the subscription identifier to `my-sub`, to be consistent with the [PubSub docs](https://cloud.google.com/pubsub/docs/quickstart-client-libraries). All relevant files were updated too.

### DIFF
--- a/pubsub/subscriptions/add_users.go
+++ b/pubsub/subscriptions/add_users.go
@@ -24,16 +24,16 @@ import (
 )
 
 // addUsers adds all IAM users to a subscription.
-func addUsers(projectID, subName string) error {
+func addUsers(projectID, subID string) error {
 	// projectID := "my-project-id"
-	// subName := "my-sub" // subscription id
+	// subID := "my-sub"
 	ctx := context.Background()
 	client, err := pubsub.NewClient(ctx, projectID)
 	if err != nil {
 		return fmt.Errorf("pubsub.NewClient: %v", err)
 	}
 
-	sub := client.Subscription(subName)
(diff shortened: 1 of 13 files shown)


pubsub/subscriptions/add_users.go
```diff
@@ -26,7 +26,7 @@ import (
 // addUsers adds all IAM users to a subscription.
 func addUsers(projectID, subName string) error {
 	// projectID := "my-project-id"
-	// subName := projectID + "-example-sub"
+	// subName := "my-sub" // subscription id
```
tbpg: Instead of the comment, would you mind renaming the variable `subID`? That way, we don't need the comment at all. Same with `topicName`->`topicID`.

Elgarni: That's actually better. Done.

PR #1021, opened 2019-10-15
README.md: add note to not depend directly on samples

tbpg:


tritone:
I noticed we already have this in the contributing guide:  https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#dont-export-sample-functions  Is this not always being followed?

Regardless, this warning seems good as well.

tbpg:
Some samples need to be exported (I'm writing one right now). Also, I've been meaning to put something from the user's perspective (rather than just sample authors).

PR #1038, opened 2019-10-17
kms: don't say to base64 encode when decrypting

### LINKED ISSUE #1037: kms: no need to encode with base64
https://github.com/GoogleCloudPlatform/golang-samples/blob/ab8c5d453091dbbba3f3b8ddf910a4e7969f013f/kms/decrypt_rsa.go#L34

In the comment, the code encodes the cipherBytes using base64 but it is not needed when decrypting.

tbpg:
Also updates `TestMain` to check if the test context was successfully retrieved.

Fixes #1037.


tbpg:
Please take another look. I renamed `cipherbytes` to `ciphertext`.

PR #1047, opened 2019-10-23
translate/v3: add samples for translate v3

davidcavazos:
Adding tests for [translate v3](https://cloud.google.com/translate/docs/intro-to-v3).

[Python canonicals](https://github.com/munkhuushmgl/python-docs-samples/tree/345425637eceb3b46795e39babe09133cc1cf4ae/translate/cloud-client)

> **Note:** Tests involving custom models will be failing until we create the AutoML models on all the testing projects in GCP.
> In the meantime, the code *should* be working (tested locally successfully) so it's ready for review.
> We can merge after creating the AutoML models and confirm all tests pass.

### DIFF
--- a/translate/v3/batch_translate_text.go
+++ b/translate/v3/batch_translate_text.go
@@ -35,7 +35,7 @@ func batchTranslateText(
 	sourceLang string,
 	targetLang string,
 ) error {
-	// projectID := "your-google-cloud-project-id"
+	// projectID := "my-project-id"
 	// location := "us-central1"
 	// inputURI := "gs://cloud-samples-data/text.txt"
 	// outputURI := "gs://YOUR_BUCKET_ID/path_to_store_results/"
--- a/translate/v3/batch_translate_text.go
+++ b/translate/v3/batch_translate_text.go
@@ -26,15 +26,7 @@ import (
 )
 
 // batchTranslateText translates a large volume of text in asynchronous batch mode.
-func batchTranslateText(
-	w io.Writer,
-	projectID string,
-	location string,
-	inputURI string,
-	outputURI string,
-	sourceLang string,
-	targetLang string,
-) error {
+func batchTranslateText(w io.Writer, projectID string, location string, inputURI string, outputURI string, sourceLang string, targetLang string) error {
 	// projectID := "my-project-id"
 	// location := "us-central1"
 	// inputURI := "gs://cloud-samples-data/text.txt"
@@ -47,7 +39,7 @@ func batchTranslateText(
 	if err != nil {
 		return fmt.Errorf("NewTranslationClient: %v", err)
 	}
-	defer c.Close()
+	defer client.Close()
 
 	req := &translatepb.BatchTranslateTextRequest{
 		Parent:              fmt.Sprintf("projects/%s/locations/%s", projectID, location),
@@ -72,7 +64,7 @@ func batchTranslateText(
 	}
 
 	// The BatchTranslateText operation is async.
-	op, err := c.BatchTranslateText(ctx, req)
+	op, err := client.BatchTranslateText(ctx, req)
 	if err != nil {
 		return fmt.Errorf("BatchTranslateText: %v", err)
 	}
--- a/translate/v3/batch_translate_text.go
+++ b/translate/v3/batch_translate_text.go
@@ -43,7 +43,7 @@ func batchTranslateText(
 	// targetLang := "ja"
 
 	ctx := context.Background()
-	c, err := translate.NewTranslationClient(ctx)
+	client, err := translate.NewTranslationClient(ctx)
 	if err != nil {
 		return fmt.Errorf("NewTranslationClient: %v", err)
 	}
--- a/translate/v3/batch_translate_text.go
+++ b/translate/v3/batch_translate_text.go
@@ -1,0 +1,92 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package v3 contains samples for Google Cloud Translation API v3.
+package v3
+
+// [START translate_v3_batch_translate_text]
+import (
+	"context"
+	"fmt"
+	"io"
+
+	translate "cloud.google.com/go/translate/apiv3"
+	translatepb "google.golang.org/genproto/googleapis/cloud/translate/v3"
+)
+
+// batchTranslateText translates a large volume of text in asynchronous batch mode.
+func batchTranslateText(
+	w io.Writer,
+	projectID string,
+	location string,
+	inputURI string,
+	outputURI string,
+	sourceLang string,
+	targetLang string,
+) error {
+	// projectID := "your-google-cloud-project-id"
+	// location := "us-central1"
+	// inputURI := "gs://cloud-samples-data/text.txt"
+	// outputURI := "gs://YOUR_BUCKET_ID/path_to_store_results/"
+	// sourceLang := "en"
+	// targetLang := "ja"
+
+	ctx := context.Background()
+	c, err := translate.NewTranslationClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewTranslationClient: %v", err)
+	}
+	defer c.Close()
+
+	req := &translatepb.BatchTranslateTextRequest{
+		Parent:              fmt.Sprintf("projects/%s/locations/%s", projectID, location),
+		SourceLanguageCode:  sourceLang,
+		TargetLanguageCodes: []string{targetLang},
+		InputConfigs: []*translatepb.InputConfig{
+			{
+				Source: &translatepb.InputConfig_GcsSource{
+					GcsSource: &translatepb.GcsSource{InputUri: inputURI},
+				},
+				// Optional. Can be "text/plain" or "text/html".
+				MimeType: "text/plain",
+			},
+		},
+		OutputConfig: &translatepb.OutputConfig{
+			Destination: &translatepb.OutputConfig_GcsDestination{
+				GcsDestination: &translatepb.GcsDestination{
+					OutputUriPrefix: outputURI,
+				},
+			},
+		},
+	}
+
+	// The BatchTranslateText operation is async.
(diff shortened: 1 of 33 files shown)


davidcavazos:
These are **a lot** of files created, please let me know if you prefer them to be divided into 1 PR per sample to make it more manageable for review.

translate/v3/batch_translate_text.go
```diff
@@ -0,0 +1,92 @@
...
+// batchTranslateText translates a large volume of text in asynchronous batch mode.
+func batchTranslateText(
+	w io.Writer,
+	projectID string,
+	location string,
+	inputURI string,
+	outputURI string,
+	sourceLang string,
+	targetLang string,
+) error {
```
tbpg: Please do this on a single line.

translate/v3/batch_translate_text.go
```diff
@@ -0,0 +1,92 @@
...
+	// projectID := "your-google-cloud-project-id"
```
tbpg:
```suggestion
	// projectID := "my-project-id"
```

(Matches the pattern of other samples.)

translate/v3/batch_translate_text.go
```diff
@@ -0,0 +1,92 @@
...
+	// location := "us-central1"
+	// inputURI := "gs://cloud-samples-data/text.txt"
+	// outputURI := "gs://YOUR_BUCKET_ID/path_to_store_results/"
+	// sourceLang := "en"
+	// targetLang := "ja"
+
+	ctx := context.Background()
+	c, err := translate.NewTranslationClient(ctx)
```
tbpg:
```suggestion
	client, err := translate.NewTranslationClient(ctx)
```

translate/v3/batch_translate_text_test.go
```diff
@@ -0,0 +1,70 @@
...
+	// Translate a sample text and check the number of translated characters.
+	if err := batchTranslateText(
+		&buf,
+		tc.ProjectID,
+		location,
+		inputURI,
+		outputURI,
+		sourceLang,
+		targetLang,
+	); err != nil {
```
tbpg: Single line.

translate/v3/batch_translate_text_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Fatalf("batchTranslateText: %v", err)
+	}
+	if got := buf.String(); !strings.Contains(got, "Total characters: 13") {
```
tbpg: Translations change relatively frequently. Are we sure this will _always_ be 13 characters? For en->fr, we've used "Hello" in the past and it still broke ("Bonjour" vs "Salut"). "Thanks" -> "Merci" seems stable.

translate/v3/batch_translate_text_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Fatalf("Got '%s', expected to contain 'Total characters: 13'", got)
```
tbpg:
```suggestion
		t.Fatalf("batchTranslateText got\n----\n%s\n----\nWant to contain:\n----\n%s\n----, got, want)
```

translate/v3/batch_translate_text_test.go
tbpg:
```suggestion
	if got, want := buf.String(), "Total characters: 13"; !strings.Contains(got, want) {
```

translate/v3/batch_translate_text_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"cloud.google.com/go/storage"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+	"github.com/google/uuid"
+)
+
+func TestBatchTranslateText(t *testing.T) {
+	tc := testutil.SystemTest(t)
+
+	var buf bytes.Buffer
```
tbpg: Minor: would prefer if this was right before the function call.

translate/v3/batch_translate_text_test.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		t.Fatalf("storage.NewClient: %v", err)
+	}
+	defer c.Close()
+
+	bucket := c.Bucket(bucketName)
+	if err := bucket.Create(ctx, tc.ProjectID, nil); err != nil {
+		t.Fatalf("bucket.Create: %v", err)
+	}
+	defer bucket.Delete(ctx)
```
tbpg: Check this error? I suspect it never gets deleted because the bucket will have objects in it.

See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/storage/objects/main_test.go#L363. Separate PR: we might want to move that function into a shared package so we can reuse it rather than copy it.

translate/v3/batch_translate_text_with_glossary.go
```diff
@@ -0,0 +1,99 @@
...
+	// projectID := "your-google-cloud-project-id"
+	// location := "us-central1"
+	// inputURI := "gs://cloud-samples-data/text.txt"
+	// outputURI := "gs://YOUR_BUCKET_ID/path_to_store_results/"
+	// sourceLang := "en"
+	// targetLang := "ja"
+	// glossaryID := "your-glossary-id"
+
+	ctx := context.Background()
+	c, err := translate.NewTranslationClient(ctx)
```
hongalex: Same as above, replace `c` with `client` in samples.

translate/v3/translate_text_with_model.go
```diff
@@ -0,0 +1,74 @@
...
+) error {
+	// projectID := "your-google-cloud-project-id"
+	// location := "us-central1"
+	// sourceLang := "en"
+	// targetLang := "fr"
+	// text := "Hello, world!"
+	// modelID := "your-model-id"
+
+	ctx := context.Background()
+	c, err := translate.NewTranslationClient(ctx)
```
hongalex: Replace `c` with `client`

translate/v3/create_and_delete_glossary_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+	"github.com/google/uuid"
+)
+
+func TestCreateAndDeleteGlossary(t *testing.T) {
+	tc := testutil.SystemTest(t)
+
+	var buf bytes.Buffer
+	var got string
```
hongalex: I think this doesn't have to be predeclared.

`got := buf.String()` on L42 should work. Needs to be fixed in other locations as well.

translate/v3/batch_translate_text.go
davidcavazos: Got it, changed in all files

translate/v3/batch_translate_text_test.go
davidcavazos: Changed

translate/v3/batch_translate_text_test.go
davidcavazos: I was following what the canonical was testing for, but I guess it would make sense to not test the actual number.

translate/v3/batch_translate_text_test.go
davidcavazos: Good idea, changed

translate/v3/batch_translate_text_test.go
davidcavazos: Good suggestion, I didn't knew you could assign to variables at the same time

translate/v3/batch_translate_text_test.go
davidcavazos: Changing that, thanks!

davidcavazos:
I commented out the 4 tests that used AutoML models since we want to merge this as soon as possible. The samples themselves should work and the docs shouldn't need updating afterwards, but I'll uncomment those 4 tests in another PR once the setup is ready in the testing projects.

davidcavazos:
@tbpg @hongalex This is ready for review

translate/v3/batch_translate_text_with_glossary_and_model_test.go
```diff
@@ -0,0 +1,73 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v3
+
+// TODO: uncomment this test once the AutoML model is set in the testing projects.
```
tbpg: What's the `gcloud` command to do this? I'll run it.

translate/v3/batch_translate_text_with_model_test.go
```diff
@@ -0,0 +1,64 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v3
+
+// TODO: uncomment this test once the AutoML model is set in the testing projects.
```
tbpg: Same for this.

translate/v3/create_and_delete_glossary_test.go
```diff
@@ -0,0 +1,63 @@
...
+	glossaryID := fmt.Sprintf("create_and_delete_glossary-%v", uuid.New().ID())
+	glossaryInputURI := "gs://cloud-samples-data/translation/glossary_ja.csv"
+
+	// Create a glossary.
+	var buf bytes.Buffer
+	if err := createGlossary(&buf, tc.ProjectID, location, glossaryID, glossaryInputURI); err != nil {
+		t.Fatalf("createGlossary: %v", err)
+	}
+	if got, want := buf.String(), "Created"; !strings.Contains(got, want) {
+		t.Fatalf("createGlossary got:\n----\n%s----\nWant to contain:\n----\n%s\n----", got, want)
```
tbpg: Is `t.Fatalf` correct here? Usually, the guidance is to "keep going," so you would use `t.Errorf`. But, it depends on the error whether or not you want to continue or not.

translate/v3/batch_translate_text_with_glossary_and_model_test.go
davidcavazos: Unfortunately, it's not a single command. It's following [these instructions](https://cloud.google.com/translate/automl/docs/before-you-begin). It's a lot of manual labor and training the model takes several hours that will have to be done on all testing projects.

I was writing a script to do it automatically, but it I also got into [this bug](https://buganizer.corp.google.com/issues/143495873) when importing the data using the REST API. I'll give it another try with the Python client library.

In any case, we wanted to publish the samples for the GA of translation v3, so I was thinking on merging this PR and having a follow up PR once the AutoML models are set up since that will take more time.

translate/v3/create_and_delete_glossary_test.go
davidcavazos: Mmm I didn't know, but it makes sense. I'll make them `t.Errorf` instead of `t.Fatalf`.

PR #1051, opened 2019-10-28
testing: don't test .md or deleted dirs

tbpg:
- Don't run Go tests when we only change .md or .github files.
- Don't run tests on deleted files.
- Clean the cache in the trap. Before, a test timeout would cause cache
issues.
- Properly handle changes only in the root dir.


tbpg:
Thanks!

PR #1052, opened 2019-10-28
.github: renovate all dependencies at the same time

tbpg:
See https://docs.renovatebot.com/faq/#use-a-single-branchpr-for-all-dependency-upgrades.


tbpg:
I'm updating `system_tests.sh` at the same time to give better error handling. The original test run for this failed with a cryptic error and `system_tests.sh` didn't give enough detail.

But, I want to make sure the `git diff` command compares only the `.github` changes. So, I modified `system_tests.sh` to only `git diff` the last commit, marked this as do not merge, and will rebase+force push to remove that change once it's working.

PR #1054, opened 2019-10-29
testing: more logging, check changes before trampoline

tbpg:
This reduces the test time of markdown-only changes by ~60 seconds (~62 seconds -> ~2 seconds). There is some extra time for Kokoro to start up and everything, though, so we won't see a 2 second status update on the PR.

### DIFF
--- a/testing/kokoro/system_tests.sh
+++ b/testing/kokoro/system_tests.sh
@@ -16,6 +16,8 @@
 
 set -e
 
+set -x
+
 go version
 date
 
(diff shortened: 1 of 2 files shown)


testing/kokoro/trampoline.sh
```diff
@@ -14,4 +14,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -x
+
+date
+
+cd github/golang-samples
+
+# SIGNIFICANT_CHANGES="$(git --no-pager diff --name-only HEAD..master | egrep -v '(\.md$|^\.github)' || true)"
```
tritone: Remove commented-out line?

testing/kokoro/system_tests.sh
```diff
@@ -67,6 +64,8 @@ echo "Running tests in project $GOLANG_SAMPLES_PROJECT_ID";
 # it when exiting.
 trap "go clean -modcache; gimmeproj -project golang-samples-tests done $GOLANG_SAMPLES_PROJECT_ID" EXIT
 
+set +x
```
tritone: What is the reason for switching this on and off?

testing/kokoro/trampoline.sh
tbpg: Whoops. Fixed.

testing/kokoro/system_tests.sh
tbpg: Don't want to accidentally print a secret environment variable. We don't have any right now, but we've had issues in the past.

tritone: Ahh makes sense-- might want to note that in a comment so someone doesn't remove it in the future.

PR #1060, opened 2019-10-31
all: check PORT env var, explicit ListenAndServe err check, fix testing change check

### LINKED ISSUE #1042: appengine: please make your examples dyslexic friendly
I've recently switched from a previous version of GAE, back when the applications relied on `app` function and there was no `main` package. I glanced over the example and thought that `main()` just replaced `init()`, without noticing that in this line

https://github.com/GoogleCloudPlatform/golang-samples/blob/57656a1b0868ff961125829b725fb5802b434f30/appengine/go11x/helloworld/helloworld.go#L42

there is a `http.ListenAndServe` nested within a `log`, which meant I spent a few hours debugging my code that just returned `exit(1)` with no clear error log or anything.

It would be really good if the examples provided were more dyslexic friendly - avoiding nesting things, avoid being clever in code writing, etc. I know golang has a lot of boring error handling and what have you, but it's better that the example file has a few extra newlines than someone missing something important...

tbpg:
Fixes #1042.

Extra fix: update the testing script to properly check changed directories.


testing/kokoro/system_tests.sh
```diff
@@ -95,15 +95,15 @@ fi
 # CHANGED_DIRS will be empty when run on master.
 # Also see trampoline.sh - system_tests.sh is only run when there are
 # significant changes.
-CHANGED_DIRS=$(git --no-pager diff --name-only HEAD..master | egrep -v '(\.md$|^\.github)' | grep "/" | cut -d/ -f1 | sort -u)
+CHANGED_DIRS=$(git --no-pager diff --name-only HEAD..master | egrep -v '(\.md$|^\.github)' | grep "/" | cut -d/ -f1 | sort -u | tr '\n' ' ')
```
grayside: Is this change intended as part of the PR?

tbpg: Yes. Before this change the for loop below didn't work because the directories were separated by newlines instead of spaces. I updated the PR title.

endpoints/getting-started/app.go
```diff
@@ -47,7 +47,9 @@ func main() {
 	if portStr := os.Getenv("PORT"); portStr != "" {
 		port, _ = strconv.Atoi(portStr)
```
grayside: The strconv seems unnecessary here.

tbpg: Done.

tbpg:
Please take another look.

PR #1063, opened 2019-11-01
getting-started/gce: initial commit

tbpg:
You can run `deploy.sh`, wait for the instance to start, check it works, then run `teardown.sh` to clean everything up.

This will be used for the Go version of https://cloud.google.com/java/docs/tutorials/getting-started-on-compute-engine.

~I left a `DO NOT SUBMIT` in there for testing purposes -- to `git checkout gce` (the branch this PR is from) when getting the code. Once this PR is merged (or about to be), we can delete that line.~


getting-started/gce/startup-script.sh
```diff
@@ -0,0 +1,66 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START getting_started_gce_startup_script]
+set -ex
+
+# Talk to the metadata server to get the project id
+export GOOGLE_CLOUD_PROJECT=$(curl -s "http://metadata.google.internal/computeMetadata/v1/project/project-id" -H "Metadata-Flavor: Google")
```
rakyll: Is this required for fluentd? Can you comment if so.

getting-started/gce/startup-script.sh
```diff
@@ -0,0 +1,66 @@
...
+echo "Project ID: ${GOOGLE_CLOUD_PROJECT}"
+
+# Install logging monitor. The monitor will automatically pickup logs sent to syslog.
+curl -s "https://storage.googleapis.com/signals-agents/logging/google-fluentd-install.sh" | bash
+service google-fluentd restart &
+
+# Install Go and git with apt-get.
+apt-get install -yq ca-certificates git software-properties-common
+add-apt-repository -y ppa:longsleep/golang-backports
+apt-get install -yq golang-go
```
rakyll: I'd not expect users to install go toolchain on production machines. They build the binary on an environment and only will deploy the final binary.

getting-started/gce/startup-script.sh
```diff
@@ -0,0 +1,66 @@
...
+
+go version
+
+# Get the code.
+mkdir /code
+cd /code
+git clone https://github.com/GoogleCloudPlatform/golang-samples.git
```
grayside: I'd add `--depth=1` here since it's part of a script. Assuming this part is being kept.

getting-started/gce/startup-script.sh
broady: +1, I remember writing or reviewing some sample that had a separate script to build and upload to GCS, then the GCE startup script pulled the binary from GCS.

If that's too much indirection for a sample like this, then maybe separating into two scripts could work out better.

Or, if the actual code doesn't matter much, then we can upload a pre-built binary for them.

getting-started/gce/startup-script.sh
```diff
@@ -0,0 +1,66 @@
+#! /bin/sh
```
broady: It looks like this script would fail if it's run twice. Can you confirm?

If startup script is touching the filesystem, e.g., checking out code or downloading something, they should first check existence of those files/resources and decide whether to overwrite or leave untouched.

tbpg: Updated. No issues no (the `gsutil cp` will overwrite the file).

getting-started/gce/startup-script.sh
tbpg: Removed.

getting-started/gce/startup-script.sh
tbpg: Switched to use Cloud Build to build, tar, and upload to GCS. Seems better than a one-off script.

I also moved the systemd service file to a standalone file that gets included in the tar file.

That has the added benefit of making it clear where to add static assets later if needed.

getting-started/gce/startup-script.sh
tbpg: Removed.

badfiles_test.go
```diff
@@ -94,6 +94,9 @@ var allowList = []string{
 
 	// Renovate configuration.
 	".github/renovate.json",
+
+	// Getting Started on GCE systemd service file.
+	"getting-started/gce/my-app.service",
```
broady: `*.service` maybe?

or `**/gce/*.service`

tbpg: Done.

getting-started/gce/startup-script.sh
broady: Nice, that sounds great.

getting-started/gce/deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+
+if [ -z "$DEPLOY_DIR" ]; then
+  echo "Must set \$DEPLOY_DIR. For example: DEPLOY_DIR=gs://my-bucket"
+  exit 1
+fi
+
+if [ -z "$DEPLOY_FILENAME" ]; then
+  echo "Must set \$DEPLOY_FILENAME. For example: DEPLOY_FILENAME=app.tar.gz"
+  exit 1
+fi
```
broady: Maybe these should be positional arguments? People seem to have problems with environment variables.

But maybe that's more about Windows and we're already really excluding them if they're not using WSL.

tbpg: I don't think we'll be telling users to run this script in the tutorial, so I'm going to leave as-is for now. But, happy to change if it ends up being more clear.

Agreed about Windows. Maybe we'll add Powershell scripts later.

PR #1064, opened 2019-11-02
opencensus: add sleep to make sure all latency updates are exported

rghetia:
The example records latency every second and expects to report all the updates to stackdriver backend by calling Flush at the end. The Flush is only meant to flush the exporter buffer. The actual metrics are read periodically and exported.
Hence update the example to sleep for reporting interval to make sure at least one additional export is performed.
So update includes.
1. Reduce the ReportingInterval to 10 seconds (default is 60 sec).
2. Remove Flush (as it is not working as expected).
3. Add sleep equal to ReportingInterval.

### DIFF
--- a/opencensus/metrics_quickstart/main.go
+++ b/opencensus/metrics_quickstart/main.go
@@ -59,38 +59,28 @@ func main() {
 	// Exporters use Application Default Credentials to authenticate.
 	// See https://developers.google.com/identity/protocols/application-default-credentials
 	// for more details.
-	exporter, err := stackdriver.NewExporter(stackdriver.Options{
-		// For demonstration the ReportingInterval is set to 10 seconds.
-		// The default value is 60 seconds. Choose the interval appropriately as it
-		// can affect the cost.
-		ReportingInterval: 10 * time.Second,
-	})
+	exporter, err := stackdriver.NewExporter(stackdriver.Options{})
 	if err != nil {
 		log.Fatal(err)
 	}
-	// Flush must be called before main() exits to ensure metrics are recorded.
-	defer exporter.Flush()
 
 	if err := exporter.StartMetricsExporter(); err != nil {
 		log.Fatalf("Error starting metric exporter: %v", err)
 	}
-
-	defer func() {
-		// Sleep 10 seconds (equal to ReportingInterval) to ensure updated metrics
-		// are exported at least once.
-		fmt.Println("Waiting for one more export interval to finish")
-		time.Sleep(10 * time.Second)
-		exporter.StopMetricsExporter()
-	}()
 	// [END monitoring_opencensus_setup_exporter]
 
-	// Record 100 fake latency values between 0 and 5 seconds.
-	for i := 0; i < 100; i++ {
+	// Record 50 fake latency values between 0 and 5 seconds.
+	for i := 0; i < 50; i++ {
 		ms := float64(5*time.Second/time.Millisecond) * rand.Float64()
 		fmt.Printf("Latency %d: %f\n", i, ms)
 		stats.Record(ctx, latencyMs.M(ms))
 		time.Sleep(1 * time.Second)
 	}
+
+	// Metrics are reported every 60 seconds to the backend. If number of
+	// iteration is changed above or if the interval between iteration is changed
+	// then adjust the sleep below to allow for all iterations to be exported.
+	time.Sleep(15 * time.Second)
 
 	fmt.Println("Done recording metrics")
 }
--- a/opencensus/metrics_quickstart/main.go
+++ b/opencensus/metrics_quickstart/main.go
@@ -59,7 +59,12 @@ func main() {
 	// Exporters use Application Default Credentials to authenticate.
 	// See https://developers.google.com/identity/protocols/application-default-credentials
 	// for more details.
-	exporter, err := stackdriver.NewExporter(stackdriver.Options{})
+	exporter, err := stackdriver.NewExporter(stackdriver.Options{
+		// For demonstration the ReportingInterval is set to 10 seconds.
+		// The default value is 60 seconds. Choose the interval appropriately as it
+		// can affect the cost.
+		ReportingInterval: 10 * time.Second,
+	})
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -69,7 +74,14 @@ func main() {
 	if err := exporter.StartMetricsExporter(); err != nil {
 		log.Fatalf("Error starting metric exporter: %v", err)
 	}
-	defer exporter.StopMetricsExporter()
+
+	defer func() {
+		// Sleep 10 seconds (equal to ReportingInterval) to ensure updated metrics
+		// are exported at least once.
+		fmt.Println("Waiting for one more export interval to finish")
+		time.Sleep(10 * time.Second)
+		exporter.StopMetricsExporter()
+	}()
 	// [END monitoring_opencensus_setup_exporter]
 
 	// Record 100 fake latency values between 0 and 5 seconds.


rghetia:
@tbpg can you please review this?

opencensus/metrics_quickstart/main.go
```diff
@@ -59,7 +59,12 @@ func main() {
 	// Exporters use Application Default Credentials to authenticate.
 	// See https://developers.google.com/identity/protocols/application-default-credentials
 	// for more details.
-	exporter, err := stackdriver.NewExporter(stackdriver.Options{})
+	exporter, err := stackdriver.NewExporter(stackdriver.Options{
+		// For demonstration the ReportingInterval is set to 10 seconds.
+		// The default value is 60 seconds. Choose the interval appropriately as it
+		// can affect the cost.
+		ReportingInterval: 10 * time.Second,
```
tbpg: I was under the impression Stackdriver's minimum reporting interval was 1 minute. Are we sure it's OK to do it even more?

opencensus/metrics_quickstart/main.go
```diff
@@ -59,7 +59,12 @@ func main() {
 	// Exporters use Application Default Credentials to authenticate.
 	// See https://developers.google.com/identity/protocols/application-default-credentials
 	// for more details.
-	exporter, err := stackdriver.NewExporter(stackdriver.Options{})
+	exporter, err := stackdriver.NewExporter(stackdriver.Options{
+		// For demonstration the ReportingInterval is set to 10 seconds.
+		// The default value is 60 seconds. Choose the interval appropriately as it
+		// can affect the cost.
```
tbpg: Delete the last sentence? We aren't giving any guidance about how to choose appropriately.

opencensus/metrics_quickstart/main.go
```diff
@@ -69,7 +74,14 @@ func main() {
 	if err := exporter.StartMetricsExporter(); err != nil {
 		log.Fatalf("Error starting metric exporter: %v", err)
 	}
-	defer exporter.StopMetricsExporter()
+
+	defer func() {
+		// Sleep 10 seconds (equal to ReportingInterval) to ensure updated metrics
+		// are exported at least once.
+		fmt.Println("Waiting for one more export interval to finish")
+		time.Sleep(10 * time.Second)
```
tbpg: We sleep for ~100 seconds below. Why do we need this sleep in addition? Should we all `StopMetricsExporter` after the 100 fake values?

opencensus/metrics_quickstart/main.go
rghetia: I added the last sentence to make sure people are not surprised by the cost as it may be higher if they choose 10 seconds vs 60 seconds.
The only reason I used 10 second is that user doesn't have to wait too long when the are trying it out. If we decided to revert back to 60 seconds then I can remove the comment entirely.

opencensus/metrics_quickstart/main.go
rghetia: I was under the same impression until someone reported the issue with the example and they tried it with 10 seconds.  Let me check with stackdriver team.

opencensus/metrics_quickstart/main.go
rghetia: The problem is as follows. 
The for loop updates the metric every 1 second with different value. While exporter reads every 60 second to report back to SD backend. So after 60 seconds first report will be sent containing updates from 60 iterations.
After 100 iterations it stops metric exporter and exits. The remaining 40 updates are not reported.

Now, the problem could be fixed by reading all metrics and reporting them to the backend when the exporter is stopped. However, it needs to make sure that it waits for enough time since last report so that the backend doesn't reject the report (due to it being too soon). Typically, Stop* is not called. Exporter is started and keeps running until the life of the application. 

Hence, I added the additional sleep equal to ReportingInterval to guarantee reporting the remaining updates.

opencensus/metrics_quickstart/main.go
tbpg: I'm OK with a sentence about cost, but saying to choose appropriately without saying what is appropriate is confusing. Is 10 seconds appropriate, given it's in this sample? (I don't think so.)

opencensus/metrics_quickstart/main.go
tbpg: https://cloud.google.com/monitoring/custom-metrics/creating-metrics#writing-ts says not to call more than once per minute. Linked to from https://opencensus.io/exporters/supported-exporters/go/stackdriver/.

opencensus/metrics_quickstart/main.go
tbpg: Can we pick one, rather than doing both? Make the for loop only 60 iterations?

Or is the `defer func()` best practice that you should do every time you set up an exporter? For a production app, you probably don't want to / can't pause the application for 60 seconds before shutting it down, but you also don't want to be missing data.

rghetia: With 60 iterations it would work without having to have defer func. However, if a user changes either iteration or sleep between iteration then they will not see the correct data.
Because of the restriction on SD you cannot avoid a pause if you don't want to loose the data. Two equally bad choice, loose data or wait (rather a long one) on termination.

defer func() is not required if missing data is tolerable when the application exits.

opencensus/metrics_quickstart/main.go
rghetia: given that official document refers to 60 second, I will revert the 10 second change.

opencensus/metrics_quickstart/main.go
rghetia: I will remove the 10 second change as per the next comment. So the comment will be removed as well.

opencensus/metrics_quickstart/main.go
tbpg: Maybe add a comment describing the situation, switch to 60, and don't do the defer? That way, users can add the defer if they need it, but it won't be needed for this quickstart.

opencensus/metrics_quickstart/main.go
```diff
@@ -63,23 +63,25 @@ func main() {
...
+	for i := 0; i < 50; i++ {
 		ms := float64(5*time.Second/time.Millisecond) * rand.Float64()
 		fmt.Printf("Latency %d: %f\n", i, ms)
 		stats.Record(ctx, latencyMs.M(ms))
 		time.Sleep(1 * time.Second)
 	}
 
+	// Metrics are reported every 60 seconds to the backend. If number of
+	// iteration is changed above or if the interval between iteration is changed
+	// then adjust the sleep below to allow for all iterations to be exported.
```
tbpg:
```
	// By default, the Stackdriver exporter exports metrics to Stackdriver every 60 seconds.
	// If the application exits before metrics are exported, they will be lost.
	// In this sample, we recorded 50 metrics in ~50 seconds, so we sleep for a bit to make
	// sure all of the metrics are exported.
```

opencensus/metrics_quickstart/main.go
```diff
@@ -63,23 +63,25 @@ func main() {
...
+	time.Sleep(15 * time.Second)
```
tbpg: Call `exporter.Flush()` to force a flush right now (and edit the comment to say why we're calling `Flush`)?

Before, we did `defer exporter.Flush()` -- did that not work? If not, it sounds like a bug.

https://godoc.org/contrib.go.opencensus.io/exporter/stackdriver#Exporter.Flush

opencensus/metrics_quickstart/main.go
```diff
@@ -63,23 +63,25 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	// Flush must be called before main() exits to ensure metrics are recorded.
-	defer exporter.Flush()
 
 	if err := exporter.StartMetricsExporter(); err != nil {
 		log.Fatalf("Error starting metric exporter: %v", err)
 	}
-	defer exporter.StopMetricsExporter()
```
tbpg: This call is in the GoDoc for StartMetricsExporter: https://godoc.org/contrib.go.opencensus.io/exporter/stackdriver#Exporter.StartMetricsExporter

Why are we deleting it?

rghetia: I meant to remove only Flush(). I'll add StopMetricsExporter() back

opencensus/metrics_quickstart/main.go
rghetia: SG

opencensus/metrics_quickstart/main.go
rghetia: Flush() was creating a confusion. Flush() only takes care of flushing the internal buffer of the exporter. It doesn't read all the metrics again and export them. For example,
At time T0, Exporter reads all (let say 10) the metrics and queues them in a buffer to export at a later time.
At time T1, application update few (let say 5) metrics.
At time T2, application terminates, calling Flush.
Flush will only make sure that internal buffer is flushed. It will not read the metrics again.
Some may expect that Flush read all metrics again and then export them, which is going to be tricky as discussed in other comments in this PR.
Alternatively, I can restore the Flush() in the example, and clearly put a comment as follows
```
// The exporter reads metrics periodically and queues them to an internal buffer.
// Flush() will ensure that all metric queued in the buffer are exported to the backend. 
```

tbpg: That sounds like a bug in `Flush`? The docs say to call it to avoid losing data, but you can still lose data. Can we update `Flush` to read all of the data?

Earlier, you said
> However, it needs to make sure that it waits for enough time since last report so that the backend doesn't reject the report (due to it being too soon).

Sorry I'm so confused here. If the backend can reject it if its sent data too soon, it seems like you should never call `Flush`?

tbpg:
Marking do not merge per our conversation -- some updates needed for the exporter.

opencensus/metrics_quickstart/main.go
grayside: Note that platforms such as Cloud Run create a need for a solid flush() method and if samples don't include it there will be:

* A need for platform-specific metrics samples
* An increased number of developers losing metrics because they never look at the platform-specific sample

This comment almost exceeds scope of the PR, and I appreciate the goal of keeping this simple.

tbpg:
Closing this for now, but we can reopen when ready.

PR #1069, opened 2019-11-05
storage: update BPO samples to use UBLA

JesseLovelace:
Updates bucket policy only samples to use uniform bucket level access


storage/buckets/main.go
```diff
@@ -397,59 +397,59 @@ func setDefaultKMSkey(c *storage.Client, bucketName string, keyName string) erro
 	return nil
 }
 
-func enableBucketPolicyOnly(c *storage.Client, bucketName string) error {
+func enableUniformBucketLevelAccess(c *storage.Client, bucketName string) error {
```
tbpg: Would you mind making these samples idiomatic at the same time? Would mean moving each sample to its own file, moving the region tag start above the imports, moving the region tag end below the end of the function, initializing a client inside each sample function, and documenting sample values for each argument.

See the full list here: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#style

The tests don't need to change.

JesseLovelace: I see why this is important, but we're targeting today to get these samples out and I don't think I'd be able to get it done by EOD. Would you be okay with merging as-is and making a follow-up issue to update the format?

PR #1070, opened 2019-11-06
run: add facility for e2e testing

### LINKED ISSUE #921: run: System Testing
Now that we have Cloud Run samples merged in this repository, we should define a plan for adding system test coverage. Here are some "use cases" to consider:

1. In order to confirm that containerized samples will be deployable, we must confirm they can be used to build a container image.

2. In order to confirm that a containerized sample can be run locally, we must confirm that a `docker run` command will start up an HTTP service that will response to requests.

3. In order to confirm that a containerized sample remains compatible with Cloud Run hosting, we must deploy to Cloud Run successfully and confirm it will start up an HTTP service that will respond to requests.

4. Point 3, but for Cloud Run on GKE as applicable

grayside:
Fixes #921 

This is a first take on adding support for end-to-end testing of Cloud Run samples, using the System Package sample app as a trial case.

internal/cloudrunci was originally planned as a standalone package and may become one in the future.


internal/testutil/testutil.go
```diff
@@ -88,11 +86,11 @@ func testContext() (Context, error) {
...
-	pkg, err := build.Import("github.com/GoogleCloudPlatform/golang-samples", "", build.FindOnly)
-	if err != nil {
-		return tc, fmt.Errorf("Could not find golang-samples on GOPATH: %v", err)
-	}
-	tc.Dir = pkg.Dir
+	// pkg, err := build.Import("github.com/GoogleCloudPlatform/golang-samples", "", build.FindOnly)
+	// if err != nil {
+	// 	return tc, fmt.Errorf("Could not find golang-samples on GOPATH: %v", err)
+	// }
+	tc.Dir = "."
```
grayside: This code blocks a Cloud Run sample using go.mod from working. Not sure what should be done as part of this PR, if anything.

run/system_package/Dockerfile
```diff
@@ -21,7 +21,8 @@ WORKDIR /app
 COPY . .
 
 # Build the release artifact.
-RUN CGO_ENABLED=0 GOOS=linux go build -v -o graphviz-web
+# -mod=vendor is necessary until cloudrunci is merged.
+RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -v -o graphviz-web
```
grayside: Vendoring is necessary until cloudrunci package is available in master branch.

internal/testutil/testutil.go
tbpg: `Dir` is used by the GAE E2E tests.

https://github.com/GoogleCloudPlatform/golang-samples/blob/master/internal/e2e/standard_test.go#L35 calls `tc.Path`, which uses `tc.Dir` (https://github.com/GoogleCloudPlatform/golang-samples/blob/master/internal/testutil/testutil.go#L36).

If we can get rid of that `GOPATH` logic, lets go for it. But, unfortunately we can't just delete it/comment it out.

run/system_package/e2e_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	defer service.Clean()
+
+	requestPath := "/diagram.png?" + url.QueryEscape("digraph G { A -> {B, C, D} -> {F} }")
+	req, err := service.NewRequest("GET", requestPath)
+	if err != nil {
+		t.Errorf("service.NewRequest: %q", err)
+		return
+	}
+	resp, err := http.DefaultClient.Do(req)
```
tbpg: Might want to create a client and set a timeout. If everything is working, it should be fast. So, a short timeout seems reasonable (e.g. 5-10 seconds).

run/system_package/e2e_test.go
```diff
@@ -0,0 +1,60 @@
...
+		t.Errorf("Get: %v", err)
+		return
+	}
+	defer resp.Body.Close()
+
+	if got := resp.StatusCode; got != 200 {
+		t.Errorf("response status: got %d, want %d", got, 200)
+	}
+
+	want := "image/png"
```
tbpg: Consider inlining `want` declarations. I haven't totally decided which I prefer, so it's up to you.

```
if got, want := resp.Header.Get("Content-Type"), "image/png"; got != want { ... }
```

run/system_package/go.mod
```diff
@@ -1,3 +1,7 @@
-module graphviz-web
+module github.com/GoogleCloudPlatform/golang-samples/graphviz-web
 
 go 1.12
+
+require github.com/GoogleCloudPlatform/golang-samples v0.0.0
+
+replace github.com/GoogleCloudPlatform/golang-samples => ../../
```
tbpg: Add a TODO to remove this? Does this work in deploys?

run/system_package/e2e_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestDiagramService(t *testing.T) {
+	tc := testutil.EndToEndTest(t)
+	service := cloudrunci.NewService("diagram", tc.ProjectID)
+	if err := service.Deploy(); err != nil {
+		t.Fatalf("could not deploy %s: %v", service.Name, err)
+	}
+	defer service.Clean()
```
tbpg: This `NewService` -> `Deploy` -> `Clean` API LGTM.

run/system_package/e2e_test.go
```diff
@@ -0,0 +1,60 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/cloudrunci"
```
tbpg: Might need to `git add` `cloudrunci`?

internal/testutil/testutil.go
grayside: Alright, I've expanded scope on this PR to try fixing it, as without a fix this PR is blocked. Pushed an update just now with a proposed use of golang.org/x/tools/go/packages as suggested in https://github.com/golang/go/issues/26433, let's see if it works.

run/system_package/go.mod
grayside: I'm currently caught in a place where I can run tests or build the container, not both. It might be more effective to complete the PR review in the context of the sample changes, then before final merge remove the sample and do a fast-follow with clean ups.

run/system_package/e2e_test.go
grayside: Oops. Fixed

run/system_package/e2e_test.go
grayside: Done

internal/cloudrunci/example_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/cloudrunci"
+)
+
+func init() {
+	os.Setenv("BUILD_ID", "12345")
+}
+
+// ExampleService_Build shows how to build once, deploy many.
+func ExampleService_Build() {
```
tbpg: Probably OK if we don't have these, given there will be other usages. But, I'm not opposed.

internal/cloudrunci/gcloud.go
```diff
@@ -0,0 +1,67 @@
...
+	"bytes"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// GcloudBin is the path to the gcloud executable.
+var GcloudBin string
```
tbpg: Need to be exported?

tbpg:
Need the license and E2E checks in a few places, too (see Kokoro failure).

internal/cloudrunci/example_test.go
grayside: This dates to plans for standalone, where I wanted to illustrate some more how-to guidance. Since no objections, would like to leave it in for now but if maintenance is an issue we can remove.

internal/cloudrunci/cloudrunci.go
```diff
@@ -0,0 +1,325 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cloudrunci facilitates end-to-end testing against the production Cloud Run.
+//
+// This is a specialized tool that could be used in addition to unit tests. It
+// calls the `gcloud beta run` command directly.
+//
+// gcloud(https://cloud.google.com/sdk) must be installed. You must be authorized via
```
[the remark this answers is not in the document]
tbpg: Looks like the space is still missing?

grayside:
Looks like a lack of `gcloud` is causing the cloudrunci tests to fail.

1. It seems like we need a facility such as testutil.SkipWithoutGcloud(t) which can cache an `os.Stat()` check. This will speed up and make test reports more concise around this problem.

2. Not sure why gcloud is not available. I know we discussed this but it seems like the `bootstrap.sh` still installs it.

tbpg:
`gcloud` is only added to the `PATH` for E2E tests: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/testing/kokoro/system_tests.sh#L87

tbpg:
Will wait for you to manage merging, etc. since this will need another PR to get testing working.

tbpg:
Could also `t.Skip` then remove it in the followup PR? Up to you.

grayside:
@tbpg could you clarify your expectation around this merging and the system packages follow-up as a first sample using this facility?

Also, if you want me to manage merging we should talk about giving me merge powers :)

tbpg:
Added you as an admin!

I'd obviously prefer to not intentionally break the build. So, if there's a way to avoid it and make this work, let's go with that. But, if we need to merge and quickly follow up, that's OK, too.

grayside:
Adding more skips in the code, and identifying more gcloud dependencies. Filed #1093 as an unrelated problem.

PR #1071, opened 2019-11-06
run/grpc-ping: add sample

grayside:
This PR is ready for review of idiomatic go & gRPC practices.
By it's nature, automated tests are best done primarily via #1070.

Architecture

This sample consists of 2 services: a gRPC ping service, and a gRPC or HTTP "relay" service. The intention is the relay might be a public service and passes the ping request onto a private gRPC-only service.

This arrangement is slightly convoluted, but provides an end-to-end testable combination of the features worth exploring in a how-to page on Cloud Run gRPC support.

Pending decisions include:

* Should the sample use github.com/soheilhy/cmux?
* Testing this in Cloud Run for Anthos / replacing Data Plane IAM with connectivity setting.


ahmetb:
This code is longer than most OSS tools I release so I won't be able to find time to review.

My recommendation is to keep the sample bare minimum to something like grpc-ping-go, which is a really small 2-tier app. I'll unsubscribe but please @ me if need input specifically from me.

grayside:
Per Ahmet's suggestion to make this smaller and the lack of a clear need, I have removed support for streaming. Happy for any other suggestions to simplify, but goals of the sample include gRPC server, multiplexing gRPC server, and authenticated requests.

grayside:
Looks like multiplexing has to go, https://github.com/soheilhy/cmux looks fairly unmaintained and https://godoc.org/google.golang.org/grpc#Server.ServeHTTP is marked experimental.

PR #1072, opened 2019-11-06
run/pubsub: update to go 1.13 and new Dockerfile standard

grayside:
Following #981, this updates the Pub/Sub sample to go 1.13 and the new Dockerfile template.


run/pubsub/go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/GoogleCloudPlatform/golang-samples/run/pubsub
 
-go 1.12
+go 1.13
```
tbpg: This signifies the Go version this module is compatible with. Are we sure we want to bump it?

grayside: Undone per suggestion that this not be bumped.

PR #1077, opened 2019-11-09
tasks: Update Cloud Tasks samples to new library version

averikitsch:


tasks/create_http_task.go
```diff
@@ -40,11 +40,11 @@ func createHTTPTask(projectID, locationID, queueID, url, message string) (*tasks
...
 
 	// Build the Task payload.
-	// https://godoc.org/google.golang.org/genproto/googleapis/cloud/tasks/v2beta3#CreateTaskRequest
+	// https://godoc.org/google.golang.org/genproto/googleapis/cloud/tasks/v2#CreateTaskRequest
 	req := &taskspb.CreateTaskRequest{
 		Parent: queuePath,
 		Task: &taskspb.Task{
-			// https://godoc.org/google.golang.org/genproto/googleapis/cloud/tasks/v2beta3#HttpRequest
+			// https://godoc.org/google.golang.org/genproto/googleapis/cloud/tasks/v2#HttpRequest
 			PayloadType: &taskspb.Task_HttpRequest{
```
tbpg: Looks like `PayloadType` was removed? Failing to compile right now.

averikitsch: It should have been merged in this PR https://github.com/googleapis/go-genproto/pull/249

tbpg: Looks like it was renamed to `MessageType`:
* https://godoc.org/google.golang.org/genproto/googleapis/cloud/tasks/v2?tab=doc#Task
* https://github.com/googleapis/go-genproto/pull/249/files#diff-1eca8348633ddb48f4ee4605e29fae3aL104

You can confirm the version of a package used by looking at the `go.mod` for the file you're looking at. In this case, the root `go.mod` of golang-samples: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/go.mod#L51 (the last part of the "pseudo-version" is the short commit hash, which matches the latest commit of go-genproto).

averikitsch: Thank you for your help

PR #1079, opened 2019-11-11
storage: add list gcs objects with s3 library

frankyn:


storage/s3_sdk/list_gcs_objects.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"io"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+func listGCSObjects(w io.Writer, bucketName string, googleAccessKeyID string, googleAccessKeySecret string) ([]*s3.Object, error) {
```
tbpg: Mind adding sample values for the arguments? Sorry, might also need this for the other sample.

tbpg: Consider dropping the `[]*s3.Object` return value, since it's not used or tested.

frankyn: done

frankyn: done

storage/s3_sdk/list_gcs_buckets.go
```diff
@@ -27,7 +27,10 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 )
 
-func listGCSBuckets(w io.Writer, googleAccessKeyID string, googleAccessKeySecret string) ([]*s3.Bucket, error) {
+func listGCSBuckets(w io.Writer, googleAccessKeyID string, googleAccessKeySecret string) error {
+	// googleAccessKeyID := fmt.Sprintf("Your Google Access Key ID")
```
tbpg: I don't think we need to use `fmt.Sprintf` for a single string?

storage/s3_sdk/list_gcs_buckets.go
```diff
@@ -42,15 +45,15 @@ func listGCSBuckets(w io.Writer, googleAccessKeyID string, googleAccessKeySecret
 
 	result, err := client.ListBucketsWithContext(ctx, &s3.ListBucketsInput{})
 	if err != nil {
-		return nil, err
+		return err
```
tbpg: Mind adding context to the error?

```
return fmt.Errorf("ListBucketsWithContext: %v", err)
```

That way, you know where the errors came from.

storage/s3_sdk/s3_gcs_test.go
```diff
@@ -42,7 +42,7 @@ func TestBucketList(t *testing.T) {
 	buf := new(bytes.Buffer)
 	err := listGCSBuckets(buf, googleAccessKeyID, googleAccessKeySecret)
 	if err != nil {
-		t.Errorf("listGCSBuckets: %v", err)
+		t.Errorf("%v", err)
```
tbpg: Please keep these the same.

tbpg:
I updated the objects list test to follow the pattern of the buckets list test and create the test key on the fly. This also has the benefit of not needing the Keystore secrets. Will wait for Kokoro then merge. :)

PR #1081, opened 2019-11-12
firestore: add samples for in and array-contains-any

tritone:
Add samples for these two query operators.

I compiled and ran locally to test.

### DIFF
--- a/firestore/firestore_snippets/query.go
+++ b/firestore/firestore_snippets/query.go
@@ -249,11 +249,31 @@ func createMultipleStartAtQuery(client *firestore.Client) {
 	// [END fs_start_at_multiple]
 }
 
+func createInQuery(ctx context.Context, client *firestore.Client) error {
+	cities := client.Collection("cities")
+	// [START fs_in_query]
+	query := cities.Where("regions", "in", [][]string{{"west_coast"}, {"east_coast"}}).Documents(ctx)
+	// [END fs_in_query]
+
+	_ = query
+	return nil
+}
+
 func createArrayContainsQuery(ctx context.Context, client *firestore.Client) error {
 	cities := client.Collection("cities")
 	// [START fs_array_contains_query]
 	query := cities.Where("regions", "array-contains", "west_coast").Documents(ctx)
 	// [END fs_array_contains_query]
+
+	_ = query
+	return nil
+}
+
+func createArrayContainsAnyQuery(ctx context.Context, client *firestore.Client) error {
+	cities := client.Collection("cities")
+	// [START fs_array_contains_any_query]
+	query := cities.Where("regions", "array-contains-any", []string{"west_coast", "east_coast"}).Documents(ctx)
+	// [END fs_array_contains_any_query]
 
 	_ = query
 	return nil
--- a/firestore/firestore_snippets/query.go
+++ b/firestore/firestore_snippets/query.go
@@ -250,10 +250,20 @@ func createMultipleStartAtQuery(client *firestore.Client) {
 }
 
 func createInQuery(ctx context.Context, client *firestore.Client) error {
+	// [START fs_query_filter_in]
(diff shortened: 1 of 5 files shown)


firestore/firestore_snippets/query.go
```diff
@@ -259,6 +269,16 @@ func createArrayContainsQuery(ctx context.Context, client *firestore.Client) err
 	return nil
 }
 
+func createArrayContainsAnyQuery(ctx context.Context, client *firestore.Client) error {
+	cities := client.Collection("cities")
+	// [START fs_array_contains_any_query]
```
BenWhitehead:
```suggestion
	// [START fs_query_filter_array_contains_any]
```

firestore/firestore_snippets/query.go
```diff
@@ -259,6 +269,16 @@ func createArrayContainsQuery(ctx context.Context, client *firestore.Client) err
...
+	query := cities.Where("regions", "array-contains-any", []string{"west_coast", "east_coast"}).Documents(ctx)
+	// [END fs_array_contains_any_query]
```
BenWhitehead:
```suggestion
	// [END fs_query_filter_array_contains_any]
```

firestore/firestore_snippets/query.go
```diff
@@ -249,6 +249,16 @@ func createMultipleStartAtQuery(client *firestore.Client) {
 	// [END fs_start_at_multiple]
 }
 
+func createInQuery(ctx context.Context, client *firestore.Client) error {
+	cities := client.Collection("cities")
+	// [START fs_in_query]
```
BenWhitehead:
```suggestion
	// [START fs_query_filter_in]
```

firestore/firestore_snippets/query.go
```diff
@@ -249,6 +249,16 @@ func createMultipleStartAtQuery(client *firestore.Client) {
...
+	query := cities.Where("regions", "in", [][]string{{"west_coast"}, {"east_coast"}}).Documents(ctx)
+	// [END fs_in_query]
```
BenWhitehead:
```suggestion
	// [END fs_query_filter_in]
```

firestore/firestore_snippets/query.go
```diff
@@ -249,6 +249,16 @@ func createMultipleStartAtQuery(client *firestore.Client) {
...
+
+	_ = query
+	return nil
+}
+
```
BenWhitehead: Can you also add a function for In query with an array of arrays? For example in java: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/firestore/src/main/java/com/example/firestore/snippets/QueryDataSnippets.java#L464-L475 This scenario is specifically called out in the docs.

tritone: This is an array of arrays (I directly imitated the example in the docs here: https://firebase.google.com/docs/firestore/query-data/queries#in_and_array-contains-any )

Can also add one mirroring https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/firestore/src/main/java/com/example/firestore/snippets/QueryDataSnippets.java#L455 as well if you like.

BenWhitehead: Yeah, we want both scenarios. Sorry for the confusion.

firestore/firestore_snippets/query.go
BenWhitehead: Can you also move this tag start above `cities := ...` We want that ref in the snippet.

firestore/firestore_snippets/query.go
BenWhitehead: Can you move the start tag above the `cities := ...` line, we want that in the snippet.

tritone: The array-contains snippet in this file does not include that, but happy to include it in these.

PR #1088, opened 2019-11-19
functions: add cloudbuild.yaml example

tbpg:
Will be used on https://cloud.google.com/functions/docs/testing/test-cicd#functions-testing-frameworks-go.

### DIFF
--- a/functions/ci_cd/cloudbuild.yaml
+++ b/functions/ci_cd/cloudbuild.yaml
@@ -1,0 +1,25 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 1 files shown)


functions/ci_cd/cloudbuild.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+# [START functions_ci_cd_cloud_build]
+steps:
+  - name: 'mirror.gcr.io/library/golang'
+    args: ['go', 'version']
+  - name: 'mirror.gcr.io/library/golang'
+    args: ['go', 'test']
+    dir: 'functions/autodeploy'
+  - name: 'gcr.io/cloud-builders/gcloud'
+    args: ['functions', 'deploy', '[YOUR_DEPLOYED_FUNCTION_NAME]', '[YOUR_FUNCTION_TRIGGER]', '--runtime', '[YOUR_RUNTIME]', '--entry-point', '[YOUR_FUNCTION_NAME_IN_CODE]']
```
grant: Can't we add  the runtime and other args? I assume we are using `go111`?

grant:
Ah, this PR looks related to the friction log you wrote.
https://cloud.google.com/functions/docs/testing/test-cicd#functions-testing-frameworks-go

grant:
I suggest we add a region tag for this yaml: `functions_ci_cd_cloud_build_go`

functions/ci_cd/cloudbuild.yaml
tbpg: Following the existing example. Would be nice not to have to update it for each runtime version as we update. Can update if you feel strongly.

We could also rework this sample at a later date, doing Node.js and Go (and others) at the same time.

PR #1104, opened 2019-11-26
internal/testutil: get absolute filepath

### LINKED ISSUE #1103: internal/e2e: GAE E2E tests failing
```
--- FAIL: TestGo11xStatic (0.08s)
    standard_test.go:62: could not deploy static: could not read service: open github.com/GoogleCloudPlatform/golang-samples/appengine/go11x/static/app.yaml: no such file or directory
--- FAIL: TestHelloWorld (0.08s)
    standard_test.go:62: could not deploy hw: could not read service: open github.com/GoogleCloudPlatform/golang-samples/appengine/go11x/helloworld/app.yaml: no such file or directory
FAIL
FAIL	github.com/GoogleCloudPlatform/golang-samples/internal/e2e	0.168s
```

tbpg:
Fixes #1103.

### DIFF
--- a/internal/testutil/testutil.go
+++ b/internal/testutil/testutil.go
@@ -21,6 +21,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"golang.org/x/tools/go/packages"
@@ -89,11 +90,22 @@ func testContext() (Context, error) {
 		return tc, noProjectID
 	}
 
-	pkgs, err := packages.Load(&packages.Config{Mode: packages.NeedName}, "github.com/GoogleCloudPlatform/golang-samples")
-	if err != nil {
-		return tc, fmt.Errorf("Could not find golang-samples: %v", err)
+	cfg := &packages.Config{
+		Mode:  packages.NeedName | packages.NeedFiles,
+		Tests: true,
 	}
-	tc.Dir = pkgs[0].PkgPath
+	pkgs, err := packages.Load(cfg, "github.com/GoogleCloudPlatform/golang-samples")
+	if err != nil {
+		return tc, fmt.Errorf("could not find golang-samples: %v", err)
+	}
+	for _, pkg := range pkgs {
+		if len(pkg.GoFiles) > 0 && strings.HasSuffix(pkg.GoFiles[0], ".go") {
+			tc.Dir = filepath.Dir(pkg.GoFiles[0])
+		}
+	}
+	if tc.Dir == "" {
+		return tc, fmt.Errorf("could not find golang-samples directory")
+	}
 
 	return tc, nil
 }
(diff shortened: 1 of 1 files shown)


internal/testutil/testutil.go
```diff
@@ -89,11 +90,22 @@ func testContext() (Context, error) {
...
+	cfg := &packages.Config{
+		Mode:  packages.NeedName | packages.NeedFiles,
+		Tests: true,
+	}
+	pkgs, err := packages.Load(cfg, "github.com/GoogleCloudPlatform/golang-samples")
 	if err != nil {
-		return tc, fmt.Errorf("Could not find golang-samples: %v", err)
+		return tc, fmt.Errorf("could not find golang-samples: %v", err)
+	}
+	for _, pkg := range pkgs {
```
grayside: Given we expect `pkgs` to always be a slice of one package, is the loop best practice vs. just checking the first element? Or perhaps I've misunderstood what `packages.Load` might do?

internal/testutil/testutil.go
```diff
@@ -89,11 +90,22 @@ func testContext() (Context, error) {
...
+		if len(pkg.GoFiles) > 0 && strings.HasSuffix(pkg.GoFiles[0], ".go") {
```
grayside: Which other kinds of files does this categorize as GoFiles?

internal/testutil/testutil.go
```diff
@@ -89,11 +90,22 @@ func testContext() (Context, error) {
...
+			tc.Dir = filepath.Dir(pkg.GoFiles[0])
```
grayside: So the problem with the current code is that instead of getting the path of the golang-samples directory we need the path to the first file? I'm not clear why this is needed, a code comment would be useful for future reference.

internal/testutil/testutil.go
tbpg: Added a comment. It returns multiple packages (`foo`, `foo.test`, and another `foo`?)

internal/testutil/testutil.go
tbpg: There is one package with a file without the `.go` extension.

internal/testutil/testutil.go
tbpg: The file is in the package, which is in the root of the repo. So, if we get the `Dir` of that file, it will be the directory of the package. I hope that makes sense.

grayside: But why is that different from the PkgPath?

tbpg: `PkgPath` is the Go package path (`github.com/GoogleCloudPlatform/golang-samples`). This is getting the file path (`/home/.../golang-samples`).

tbpg:
Going to merge TBR to hopefully get the build green before the holiday.

PR #1110, opened 2019-12-02
automl: add generic batch predict sample

nnegrey:
Once merged, I'll update docs and remove language_batch_predict and vision_batch_predict


automl/batch_predict.go
```diff
@@ -0,0 +1,81 @@
...
+	automl "cloud.google.com/go/automl/apiv1beta1"
+	automlpb "google.golang.org/genproto/googleapis/cloud/automl/v1beta1"
+)
+
+// batchPredict does a batch prediction.
+func batchPredict(w io.Writer, projectID string, location string, modelID string, inputURI string, outputURI string) error {
+	// projectID := "my-project-id"
+	// location := "us-central1"
+	// modelID := "ICN123456789..."
+	// inputURI := "gs://BUCKET_ID/path_to_your_input_csv_or_jsonl"
```
tbpg: Delete `1` at the end?

automl/batch_predict.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	automl "cloud.google.com/go/automl/apiv1beta1"
+	automlpb "google.golang.org/genproto/googleapis/cloud/automl/v1beta1"
+)
+
+// batchPredict does a batch prediction.
+func batchPredict(w io.Writer, projectID string, location string, modelID string, inputURI string, outputURI string) error {
```
tbpg: Can we add a test for this sample?

automl/batch_predict.go
nnegrey: `l` is intentional it's the type of file AutoML expects .jsonl from the docs.

automl/batch_predict.go
nnegrey: @davidcavazos , are you working on the tests for the rest of these samples?

davidcavazos: I was working on them but paused them due to our discussion that we shouldn't be creating datasets and/or models on all of the testing projects.

PR #1113, opened 2019-12-03
vision: add set_endpoint sample

davidcavazos:
Adding code sample for vision to set a custom endpoint.

Python canonical: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2569

### DIFF
--- a/vision/detect/set_endpoint.go
+++ b/vision/detect/set_endpoint.go
@@ -18,46 +18,23 @@ package main
 import (
 	"context"
 	"fmt"
-	"io"
 
 	vision "cloud.google.com/go/vision/apiv1"
 	"google.golang.org/api/option"
-	visionpb "google.golang.org/genproto/googleapis/cloud/vision/v1"
 )
 
 // setEndpoint changes your endpoint.
-func setEndpoint(w io.Writer, endpoint string) error {
+func setEndpoint(ctx context.Context, endpoint string) (*vision.ImageAnnotatorClient, error) {
 	// endpoint := "eu-vision.googleapis.com:443"
 
-	// Create a client with a custom endpoint.
-	ctx := context.Background()
+	// ctx := context.Background()
 	client, err := vision.NewImageAnnotatorClient(ctx, option.WithEndpoint(endpoint))
 	if err != nil {
-		return fmt.Errorf("NewImageAnnotatorClient: %v", err)
+		return nil, fmt.Errorf("NewImageAnnotatorClient: %v", err)
 	}
-	defer client.Close()
+	// defer client.Close()
 
-	// Use the client with custom endpoint to detect texts on an image.
-	image := &visionpb.Image{
-		Source: &visionpb.ImageSource{
-			GcsImageUri: "gs://cloud-samples-data/vision/text/screen.jpg",
-		},
-	}
-	maxResults := 3
-	texts, err := client.DetectTexts(ctx, image, nil, maxResults)
-	if err != nil {
-		return fmt.Errorf("DetectTexts: %v", err)
-	}
-
-	fmt.Fprintf(w, "Texts:\n")
-	for _, text := range texts {
-		fmt.Fprintf(w, "%v\n", text.GetDescription())
-		for _, vertex := range text.GetBoundingPoly().GetVertices() {
-			fmt.Fprintf(w, "  bounding vertex: %v, %v\n", vertex.GetX(), vertex.GetY())
-		}
-	}
-
-	return nil
+	return client, nil
 }
 
 // [END vision_set_endpoint]
--- a/vision/detect/set_endpoint.go
+++ b/vision/detect/set_endpoint.go
@@ -1,0 +1,63 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


vision/detect/set_endpoint.go
```diff
@@ -0,0 +1,40 @@
...
+
+	vision "cloud.google.com/go/vision/apiv1"
+	"google.golang.org/api/option"
+)
+
+// setEndpoint changes your endpoint.
+func setEndpoint(ctx context.Context, endpoint string) (*vision.ImageAnnotatorClient, error) {
+	// endpoint := "eu-vision.googleapis.com:443"
+
+	// ctx := context.Background()
```
tbpg: Please create the context inside the function, rather than passing it in.

https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#declare-a-contextcontext-as-needed

vision/detect/set_endpoint.go
```diff
@@ -0,0 +1,40 @@
...
+	client, err := vision.NewImageAnnotatorClient(ctx, option.WithEndpoint(endpoint))
+	if err != nil {
+		return nil, fmt.Errorf("NewImageAnnotatorClient: %v", err)
+	}
+	// defer client.Close()
+
+	return client, nil
```
tbpg: Don't return the client and do `defer client.Close()`.

vision/detect/set_endpoint_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	visionpb "google.golang.org/genproto/googleapis/cloud/vision/v1"
+)
+
+func TestSetEndpoint(t *testing.T) {
+	ctx := context.Background()
+
+	const endpoint = "eu-vision.googleapis.com:443"
+
+	client, err := setEndpoint(ctx, endpoint)
```
tbpg: Create another client for the tests. It's less efficient, but more clear.

vision/detect/set_endpoint_test.go
```diff
@@ -0,0 +1,63 @@
...
+		t.Fatalf("setEndpoint: %v", err)
+	}
+	defer client.Close()
+
+	image := &visionpb.Image{
+		Source: &visionpb.ImageSource{
+			GcsImageUri: "gs://cloud-samples-data/vision/text/screen.jpg",
+		},
+	}
+	texts, err := client.DetectTexts(ctx, image, nil, 1)
```
tbpg: Does the output of this change based on the endpoint? Is there a more direct way we can test the endpoint was actually set?

vision/detect/set_endpoint_test.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		t.Fatalf("DetectTexts: %v", err)
+	}
+
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, "Texts:\n")
+	for _, text := range texts {
+		fmt.Fprintf(&buf, "%v\n", text.GetDescription())
```
tbpg: This is the test already. So, there is no need to write to a buffer. The result values can be checked directly.

vision/detect/set_endpoint.go
davidcavazos: Alright

vision/detect/set_endpoint.go
davidcavazos: Changing

vision/detect/set_endpoint_test.go
davidcavazos: Alright

vision/detect/set_endpoint_test.go
davidcavazos: I don't think the output changes, this is what they're doing in the canonical

vision/detect/set_endpoint_test.go
davidcavazos: Got it, changing

vision/detect/set_endpoint_test.go
davidcavazos: @nnegrey do you know if there is a more direct way to test if the endpoint was actually used? Or is it completely transparent for users?

nnegrey: Not that I know of for this API, you can pass an invalid endpoint and get an error. But no resource is created. 

Natural Language for example has a different set of endpoints they support as documented here: https://cloud.google.com/natural-language/docs/locations

davidcavazos:
@tbpg do you think this is ready?

vision/detect/set_endpoint_test.go
tbpg: Could we remove this portion of the test if it's not testing the sample? Or is it needed to make sure it still works after changing the endpoint?

nnegrey: Mostly to just verify the endpoint works.

davidcavazos: @tbpg how do you feel about this? Would you still want to change anything or should we go ahead and merge?

shuesc1:
Ping me when this is merged so I can add it to the docs. Thanks.

vision/detect/set_endpoint_test.go
tbpg: Sounds good to me.

broady: In the test, you could set the endpoint to a dummy gRPC server.

davidcavazos:
@tbpg Can we merge this if you think it's ready? (I'm not authorized to merge)

tbpg:
Will wait for Kokoro then merge. Thanks!

tbpg:
Test failure is unrelated. Merging.

PR #1114, opened 2019-12-03
testing: check go.mod and go.sum are tidy, stop using GOPATH

### LINKED ISSUE #1111: testing: add go.mod and go.sum tidy check
Right now, tests are run as root in the container. If the `go.mod` or `go.sum` files are modified, root takes over ownership and the files can't be cached, so you get an obscure error, like https://github.com/GoogleCloudPlatform/golang-samples/pull/1105#issuecomment-560939706.

We should check the `go.mod` and `go.sum` are tidy before running anything else.

tbpg:
Fixes #1111.


testing/kokoro/system_tests.sh
```diff
@@ -21,16 +21,22 @@ set -x
...
-cd $target/golang-samples
+cd github/golang-samples
 
 export GO111MODULE=on # Always use modules.
 export GOPROXY=https://proxy.golang.org
 
+# Fail if a dependency was added without the necessary go.mod/go.sum change
+# being part of the commit.
+# Do this before reserving a project since this doens't need a project.
+for i in `find . -name go.mod`; do
```
jadekler: Are there multiple go.mod? If not you can just use the inner block

tbpg: Yes. Some samples need to be standalone modules.

PR #1120, opened 2019-12-10
Update create_glossary.go

nnegrey:


translate/v3/create_glossary.go
```diff
@@ -29,7 +29,7 @@ import (
 func createGlossary(w io.Writer, projectID string, location string, glossaryID string, glossaryInputURI string) error {
 	// projectID := "my-project-id"
 	// location := "us-central1"
-	// glossaryID := "glossary-id"
+	// glossaryID := "your-glossary-display-name"
```
davidcavazos: As per the style in other Go files, `"my-glossary-display-name"`, see `projectID`.

translate/v3/create_glossary.go
```diff
@@ -29,7 +29,7 @@ import (
 func createGlossary(w io.Writer, projectID string, location string, glossaryID string, glossaryInputURI string) error {
 	// projectID := "my-project-id"
 	// location := "us-central1"
-	// glossaryID := "glossary-id"
+	// glossaryID := "my-glossary-display-name"
```
tbpg: Should we change the variable name? Or is the ID the display name?

nnegrey:
(Can't reply directly on GitHub, will try here)
The ID is the display name, I wish it was different in the proto, but
trying to clarify it is actually a display name for users cus there has
been confusion.

PR #1121, opened 2019-12-10
speech: infinite streaming sample code

wangbill-google:
I have re-uploaded the infinite streaming sample code and run go mod tidy already.


wangbill-google:
kokoro:force-run

grayside:
Please do not create new PR's unless completely changing the approach. This interrupts reviews and discussions in progress.

1. To facilitate the review process, could you clarify how this sample will be used?
2. Thank you for running `go mod tidy`, it appears there should be changes in the existing `/go.mod` and `/go.sum` files, please add those files to this branch.

tbpg:
Bill, can you clarify how this sample will be used? It will help make sure we do the right review and style.

wangbill-google:
This is part of our speech to text public client library, we don't have corresponding infinite streaming code samples for Golang before.

wangbill-google:
That sample we already have is for streaming recognition and could work only for 5 minutes due to API limitation.
So I write the new infinite streaming sample to conduct recognition continuously. 
That would be pretty practical and helpful to our customers and clients.

tbpg:
Do we need both samples? Can we edit the existing one? If not, would you mind doing a new region tag? If they are distinct samples on different pages, they should have different region tags tracked in the sample tracker.

speech/infinite_streaming_from_file/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Run the sample
+
+Before running any example you must first install the Speech API client and the Color Pen:
+
+```bash
+go get -u cloud.google.com/go/speech/apiv1
+go get -u google.golang.org/genproto/googleapis/cloud/speech/v1
+go get -u github.com/fatih/color
+```
```
tbpg: No need to do this. Modules will be downloaded by `go build` automatically`.

speech/infinite_streaming_from_file/README.md
```diff
@@ -0,0 +1,57 @@
+# Google Cloud Speech API Go example
+
+## Authentication
```
tbpg: Do we need a full README for this? If this is going to be a tutorial on cloud.google.com, that should explain all of this. In general, we should avoid duplicating information about how to run the sample. READMEs, unfortunately, fall out of date very easily.

speech/infinite_streaming_from_file/go.mod
```diff
@@ -0,0 +1,11 @@
+module infinite/audio
```
tbpg:
```suggestion
module github.com/GoogleCloudPlatform/golang-samples/speech/infinite_streaming_from_file
```

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,236 @@
...
+// Command infinite_audio pipes the stdin audio data to
+// Google Speech API and outputs the transcript.
+
+// Before executing this program, the sampleRate parameter needs to be adjusted
+// to match your audio file, 16000 is used in this program as an example
+
+// To run this program, you could use the following command:
+// cat [your_audio_filename].wav | go run infinite_audio.go
+
+package main
```
tbpg: Please update the comment to be a "package comment": https://github.com/golang/go/wiki/CodeReviewComments#package-comments

```
// The infinite_streaming_from_file command pipes ...
// ...
package main
```

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,236 @@
...
+	speechpb "google.golang.org/genproto/googleapis/cloud/speech/v1"
+)
+
+const sampleRate int = 16000                                      //Audio sample rate hertz
+const inputChannel int = 1                                        //Number of audio input channel
+const outputChannel int = 0                                       //Number of audio output channel
+const bytesPerSample int = 2                                      //Number of bytes each sample consists of
+const bytesPerSecond = sampleRate * inputChannel * bytesPerSample //Number of bytes each second audio consists of
+const streamTimeLimit = 290 * time.Second                         //Streaming API Limit( ≈ 5 mins)
+const sampleDuration int64 = 1000                                 //The duration of each sample is 1000 ms
```
tbpg:
```
const (
	sampleRate = 16000
	// ...
)
```

I don't think you need the types.

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,236 @@
...
+	"os"
+	"sync"
+	"time"
+
+	speech "cloud.google.com/go/speech/apiv1"
+	color "github.com/fatih/color"
+	speechpb "google.golang.org/genproto/googleapis/cloud/speech/v1"
+)
+
+const sampleRate int = 16000                                      //Audio sample rate hertz
```
tbpg: Please add a space between the `//` and the text and add a period at the end.

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,236 @@
...
+const inputChannel int = 1                                        //Number of audio input channel
+const outputChannel int = 0                                       //Number of audio output channel
+const bytesPerSample int = 2                                      //Number of bytes each sample consists of
+const bytesPerSecond = sampleRate * inputChannel * bytesPerSample //Number of bytes each second audio consists of
+const streamTimeLimit = 290 * time.Second                         //Streaming API Limit( ≈ 5 mins)
+const sampleDuration int64 = 1000                                 //The duration of each sample is 1000 ms
+
+func main() {
```
tbpg: Can we split this into more than one function? It's a bit difficult for me to follow all of the indentation and everything.

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,236 @@
...
+							//Remove the first sample(audio chunk) from the head of processingBuffer like Queue(FIFO)
+							processingBuffer = processingBuffer[1:]
+						}
+
+					}
+				}
+
+			}
+			//If there is no more data to read from input, then the program will terminate
+			if inputIsNull {
```
tbpg: Can `return inputIsNull` directly. No need for the `if`. But, I'm not sure I understand how `inputIsNull` works here. Might be better after it's split into more than one function.

speech/infinite_streaming_from_mic/README.md
```diff
@@ -0,0 +1,57 @@
+# Google Cloud Speech API Go example
```
tbpg: Same comments for this README.

speech/infinite_streaming_from_mic/go.mod
```diff
@@ -0,0 +1,11 @@
+module infinite/mic
```
tbpg:
```suggestion
module github.com/GoogleCloudPlatform/golang-samples/speech/infinite_streaming_from_mic
```

speech/infinite_streaming_from_mic/infinite_mic.go
```diff
@@ -0,0 +1,246 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Command infinite_mic pipes the audio data from microphone
+// input to Google Speech API and outputs the transcript.
+
+// To run this program, SoX could be used to capture mic input:
+// sox -d -e signed -b 16 -c 1 -q mic.wav | go run infinite_mic.go
+
+package main
```
tbpg: Same comments for this file.

speech/infinite_streaming_from_file/README.md
wangbill-google: This is only the instruction in the readme about how to configure pre-requisite packages before running the program.  I refer to all the other sample files under the same directory and they all have this part in the detailed readme. So I keep this part to make the same format.

speech/infinite_streaming_from_file/README.md
wangbill-google: This is only a concise readme which includes the key points to run program instead of a full tutorial on website. All the other samples have the same readme so I just keep the same format to make it easier for our customers who is the first time to try to run this program.

speech/infinite_streaming_from_file/infinite_audio.go
wangbill-google: The reason why I clarify the type for specific variable is that different variables have different functions in the program.  I don't want to use the default int32 type for all the variables because some variable needs to participate in high-precision calculation and some variable has special data type like time.second which must be declared explicitly.

speech/infinite_streaming_from_file/infinite_audio.go
wangbill-google: inputIsNull is a flag variable which will be changed by a concurrent function.  It is used to determine if the program needs to terminate.  We need "if" clause to return right decision whether to exit the for loop or not.

speech/infinite_streaming_from_file/infinite_audio.go
wangbill-google: I try to split into different functions but it is rather difficult to implement the correct logic because some functions need to modify the parameter inside other functions mutually to determine skipping especially when they are running at the same time. So to keep it simple I just combine them into one function and write comment for each sentence of code. Actually there is only one function to implement the main logic and one concurrent function to implement the bi-directional communication. All the other parts of code are auxiliary.

wangbill-google:
I have already tested both programs on my laptop and they could work very well.  Our customers will refer to these samples and change corresponding parameters to write their customized program.

speech/infinite_streaming_from_file/README.md
tbpg: With Go modules, there is no need to configure dependencies before running the program. They're downloaded as needed by `go build`. So, this can be deleted.

speech/infinite_streaming_from_file/infinite_audio.go
tbpg: You should be able to leave off `int`, as that is the default type for untyped integer constants. https://golang.org/ref/spec#Constants

speech/infinite_streaming_from_file/infinite_audio.go
tbpg: Will you consolidate the comment with the comments above? No need to have both.

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,240 @@
...
+	speechpb "google.golang.org/genproto/googleapis/cloud/speech/v1"
+)
+
+const sampleRate int = 16000                                      // Audio sample rate hertz.
+const inputChannel int = 1                                        // Number of audio input channel.
+const outputChannel int = 0                                       // Number of audio output channel.
+const bytesPerSample int = 2                                      // Number of bytes each sample consists of.
+const bytesPerSecond = sampleRate * inputChannel * bytesPerSample // Number of bytes each second audio consists of.
+const streamTimeLimit = 290 * time.Second                         // Streaming API Limit( ≈ 5 mins).
+const sampleDuration int64 = 1000                                 // The duration of each sample is 1000 ms.
```
tbpg: Const group, please (also, the `int` type comment applies here).

speech/infinite_streaming_from_file/infinite_audio.go
tbpg:
```
return inputIsNull
```

is equivalent to:

```
if inputIsNull {
	return true
} else {
	return false
}
```

tbpg:
Also, looks like you need to run `gofmt -w -s .` in the `golang-samples` repo.

tbpg:
Sharing variables between goroutines is generally bad practice. Normally, Go uses channels to communicate between goroutines.

speech/infinite_streaming_from_file/README.md
wangbill-google: Pre-configuration is part of our customer documentation, we have those detailed configuration instructions in all the other language versions for convenience to our customers. This part only supplements documentation and not every customer has good familiarity with Golang.

speech/infinite_streaming_from_file/infinite_audio.go
wangbill-google: I have already added the package comment just above package main between /* and */ 
Did you see that? Long sentence package comment needs to be split into several lines.

wangbill-google:
All done for review changes.

speech/infinite_streaming_from_file/README.md
tbpg: This sample uses Go modules. When using Go modules, you do not run `go get -u ...` to download dependencies. The `go` command looks at the `go.mod` and downloads whatever code is needed when you run `go build` or `go test`.

Running this command could result in the sample not working for the user.

See https://blog.golang.org/using-go-modules for a tutorial for Go modules.

speech/infinite_streaming_from_file/infinite_audio.go
tbpg: Yes, I saw that. Please make it all one comment. If you need multiple lines in the package comment, you can add a newline in the block comment (`/* ... */`).

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,235 @@
...
+// To run this program, you could use the following command:
+// cat [your_audio_filename].wav | go run infinite_audio.go
+
+/*
+The infinite_streaming_from_file program reads audio data from file
+and pipes it to Google Speech API to output the transcript infinitely.
+*/
+package main
+
+// [START speech_transcribe_streaming]
```
tbpg: Please move this above the package declaration (the region should include the package comment and package declaration). There should be an empty line between this comment and the start of the package comment.

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,235 @@
...
+	// There are three functions in the loop:  (Initialize Stream) ->   [ Read Audio | Receive Responses ].
+	for !exitFlag {
+		// Exit from loop if exitFlag condition is met.
+
+		// Wait for Goroutines to finish before going to next loop.
+		var wg sync.WaitGroup
+		wg.Add(1)
+
+		// Start stream.
+		if streamIsNull {
```
tbpg: `streamIsNull` is always true here. So, I don't think we need the `if`.

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,235 @@
...
+			// Initialize speech stream.
+			ctx = context.Background()
+
+			client, err = speech.NewClient(ctx)
+			if err != nil {
+				log.Fatal(err)
```
tbpg: Please include context for what the error is from.

```go
log.Fatalf("speech.NewClient: %v", err)
```

Otherwise, it's sometimes hard to know where the error came from.

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,235 @@
...
+			restartCounter++
+			color.Red("Start New Stream %v!\n", restartCounter)
+
+			// Send the initial configuration message.
+			if err := stream.Send(&speechpb.StreamingRecognizeRequest{
+				StreamingRequest: &speechpb.StreamingRecognizeRequest_StreamingConfig{
+					StreamingConfig: &speechpb.StreamingRecognitionConfig{
+						Config: &speechpb.RecognitionConfig{
+							Encoding:        speechpb.RecognitionConfig_LINEAR16,
+							SampleRateHertz: 16000,
```
tbpg: Use `sampleRate` variable?

speech/infinite_streaming_from_file/infinite_audio.go
```diff
@@ -0,0 +1,235 @@
...
+			// Start timing.
+			streamIsNull = false
+			startTime := time.Now()
+			streamDeadline = startTime.Add(streamTimeLimit)
+
+			restartCounter++
+			color.Red("Start New Stream %v!\n", restartCounter)
+
+			// Send the initial configuration message.
+			if err := stream.Send(&speechpb.StreamingRecognizeRequest{
```
tbpg: Normally, we declare the request as a separate variable, `req`.

https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#prefer-inline-proto-declarations

speech/infinite_streaming_from_file/infinite_audio.go
wangbill-google: Do you really read the code?   At the beginning it will be true to initialize the stream and when the program proceeds into the subsequent for loop it will be false to skip initialization of the stream.

speech/infinite_streaming_from_file/infinite_audio.go
wangbill-google: The content of error itself will explain where the error come from under what kind of context.
All the initializations of speech client in official Golang documentation write like this.

speech/infinite_streaming_from_file/infinite_audio.go
wangbill-google: All the declarations of request in other official  Speech Golang documentations write like this.
Also other code samples under the same directory write in this way.
If you have dual criterion or different opinions, please talk with them.

wangbill-google:
(1)The logic for how to implement infinite streaming correctly is quite complicated indeed, otherwise other people would finish it earlier without a lot of efforts.
I have spent a lot of time trying to use channels to implement the correct logic but I gave it up finally because channels could communicate between goroutines but they will block the goroutine until the sender and receiver has finished their work. This logic is not what we want here in infinite streaming because we have bidirectional streams and they need to work independently without synchronization. If you think you have a better idea to implement the correct logic and I could tell customers you would write a new version for them.

(2)There is no need to add test because it couldn't be tested in github repo using CI/CD tools.
One of the programs will read the specific audio file from your local environment and you need to change some parameters according to the corresponding audio features such as Sample Rate to run the program successfully. Also you need to pipe the data reading from audio file into the program because the program will only receive the binary audio data from stdin in memory. It could only be tested in your local environment with corresponding parameters to fit your specific audio files.
Another program will read live data from your microphone in your local environment, can you show me how to add the test?
This is not the production code and it is only an example documentation to show how to implement infinite streaming. Customers would refer to part of the example to write their customized version and modify parameters to fit their specific needs. Some of our customers have tested and utilized the program in their local environment already. The only reason to upload it here is to make it easier for our customers to download and refer to.

speech/infinite_streaming_from_file/infinite_audio.go
tbpg: Yes. Sorry, it's very difficult for me to follow.

wangbill-google:
xxxx xxx

PR #1129, opened 2019-12-16
videointelligence: make analyze test E2E and add sleep

tbpg:
Getting quota errors: https://source.cloud.google.com/results/invocations/1a2aabd5-6fdd-4224-bec6-82e71c1c0f4d/targets/cloud-devrel%2Fgo%2Fgolang-samples%2Fsystem-tests%2Fgo112/log

If this doesn't work reliably enough, we should get the quota increased.


videointelligence/video_analyze/video_analyze_test.go
```diff
@@ -47,9 +48,9 @@ func TestAnalyze(t *testing.T) {
 				t.Fatal("gcs not set")
 			}
 
+			time.Sleep(5 * time.Second) // Give buffer between tests for quota.
```
tritone: Just out of curiosity-- what is the quota? Why does it seem better to add sleep after every test rather than retry in the case of error?

tbpg: Retry definitely seems better! Switched.

tbpg: I also updated all of the videointelligence tests, not just this one, since they all seem to run into the out of quota error.

videointelligence/annotate/object_tracking_gcs_test.go
```diff
@@ -18,21 +18,25 @@ import (
...
-	testutil.SystemTest(t)
+	testutil.EndToEndTest(t)
 
 	gcsURI := "gs://cloud-samples-data/video/cat.mp4"
 
-	var buf bytes.Buffer
-	if err := objectTrackingGCS(&buf, gcsURI); err != nil {
-		t.Fatal(err)
-	}
+	testutil.Retry(t, 10, 20*time.Second, func(r *testutil.R) {
```
tritone: 20s sleep between retries seems long, but fine if that's what you think makes sense to avoid the quota issue.

PR #1131, opened 2019-12-17
secretmanager: add initial samples

sethvargo:
@tbpg 🎁


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/GoogleCloudPlatform/golang-samples
 go 1.11
 
 require (
-	cloud.google.com/go v0.49.0
+	cloud.google.com/go v0.49.1-0.20191216225856-71c1cb005d22
```
sethvargo: Not sure the best way to handle this. Secret Manager is in the main repo, but not in a tagged version yet.

secretmanager/access_secret_version.go
```diff
@@ -0,0 +1,58 @@
...
+	secretmanager "cloud.google.com/go/secretmanager/apiv1beta1"
+	secretspb "google.golang.org/genproto/googleapis/cloud/secrets/v1beta1"
+)
+
+// accessSecretVersion accesses the payload for the given secret version if one
+// exists. The version can be a version number as a string (e.g. "5") or an
+// alias (e.g. "latest").
+func accessSecretVersion(w io.Writer, name string) error {
+	// name := "projects/my-project/secrets/my-secret/versions/5"
+	// name := "projects/my-project/secrets/my-secret/versions/latest"
```
sethvargo: Departure: I'm trying to show that this is polymorphic by providing two sample inputs.

secretmanager/access_secret_version.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// Create the client.
+	ctx := context.Background()
+	client, err := secretmanager.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create secretmanager client: %w", err)
```
sethvargo: Should we be using `%w`? Happy to update everywhere if we just want `%v` instead.

secretmanager/create_secret.go
```diff
@@ -0,0 +1,63 @@
...
+	"io"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1beta1"
+	secretspb "google.golang.org/genproto/googleapis/cloud/secrets/v1beta1"
+)
+
+// createSecret creates a new secret with the given name. A secret is a logical
+// wrapper around a collection of secret versions. Secret versions hold the
+// actual secret material.
+func createSecret(w io.Writer, parent, name string) error {
```
ptone: IMO - since we deal with both names and ids for these resources, it is clearer to use id in the sample signature, for a value that gets passed to an ID struct member - this is just the final part of the resource name.

sethvargo: I don't disagree, but the contributing guide clearly says we should use "name" here: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#common-identifiers

sethvargo:
This is missing IAM samples, but that's blocked on https://code-review.googlesource.com/c/gocloud/+/49650. I'll add those in a followup PR once that's merged.

secretmanager/create_secret.go
ptone: That says use `name` for fully qualified names, but this param maps to just the id 'my-secret' - not the name of 'project/my-project/secrets/my-secret', the full resource `name`

secretmanager/access_secret_version.go
grayside: The repo-at-large is currently maintaining compatibility with go 1.11, should use %v.

secretmanager/secretmanager_test.go
```diff
@@ -0,0 +1,344 @@
...
+	defer cleanup()
+	version := testSecretVersion(t, secret.Name, payload)
+
+	var b bytes.Buffer
+	if err := accessSecretVersion(&b, version.Name); err != nil {
+		t.Fatal(err)
+	}
+
+	if act, exp := b.String(), string(payload); !strings.Contains(act, exp) {
+		t.Errorf("expected %q to contain %q", act, exp)
```
tbpg: Please include the function that failed. https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

secretmanager/secretmanager_test.go
```diff
@@ -0,0 +1,344 @@
...
+	if err != nil {
+		tb.Fatalf("failed to create client: %v", err)
+	}
+	return client, ctx
+}
+
+func testName(tb testing.TB) string {
+	tb.Helper()
+
+	u, err := uuid.NewV4()
```
tbpg: @broady just filed #1133 to avoid using UUIDs. Can we use a consistent name and avoid any leaking of resources? Maybe only create a single secret and use it for all of the tests, rather than a secret for every test?

sethvargo: If we do that, tests can't run in parallel (one could be destroying while another is enabling), which is less than ideal. I can update the test helper to use a common prefix to make it easy to track down issues, but I think UUIDs are unavoidable if we want fully isolated tests and predictable tests for each sample.

secretmanager/create_secret.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	// Build the request.
+	req := &secretspb.CreateSecretRequest{
+		Parent:   parent,
+		SecretId: id,
+		Secret: &secretspb.Secret{
+			Replication: &secretspb.Replication{
+				Replication: &secretspb.Replication_Automatic_{
+					Automatic: &secretspb.Replication_Automatic{},
```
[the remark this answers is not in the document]
sethvargo: No, and it can't be nil. It explicitly has to be the empty automatic replication policy. For regional replication, there's fields inside the Replication_UserManager struct.

secretmanager/secretmanager_test.go
broady: Our CI system provides a level of isolation at the cloud project level. That is, when the tests for golang-samples run, the test run has exclusive access to the project that its running in (tc.ProjectID)

So you only need to use a resource name that includes tc.ProjectID plus a pre/suffix unique to golang-samples.

sethvargo: Yes, but the tests themselves (the tests in `secretmanager_test.go`) run in parallel. I can add the project ID prefix, but the secrets are also scoped to the project ID, so that seems redundant. 

For example, `TestListSecrets` runs in parallel to `TestCreateSecret`, so the results are unpredictable if we use hard-coded values. Sometimes `TestCreateSecret` would run before `TestListSecrets`, sometimes after. Similarly, some tests actually destroy secrets, so they wouldn't exist in the next run. The tests then become dependent on the project being in a certain state instead of running in isolation.

sethvargo: As far as leaking resources, each function `defer` cleanup. It's theoretically possible that we'd leak if a quota issue or outage arose between the `testSecret` action and `testCleanupSecret`, but that should be incredibly rare.

Each test as written here is fully isolated and independent from all other tests and prior state of the system, by design.

sethvargo: Talked with @broady about this in chat and worked through it. Worked through quota issues by switching to `SystemTest`.

sethvargo:
Hey @tbpg - I think @jadekler was working on a proto update? Not sure if that landed yet.

tbpg:
Thanks. Marking `do not merge` until then.

sethvargo:
@tbpg updated

tbpg:
Hmm...

```
    secretmanager_test.go:369: testCleanupSecret: failed to delete secret: rpc error: code = ResourceExhausted desc = Quota exceeded for quota metric 'Write requests' and limit 'Write requests per minute' of service 'secretmanager.googleapis.com' for consumer 'project_number:263638463498'.
```

Quota errors are tricky since they are hard to reproduce. Any ideas for reducing the number of calls, like combining some tests so they share a secret? Alternatively, we can request an increased quota for the 10 test projects.

PR #1134, opened 2019-12-18
run/logging-manual: strip timestamp prefix from log instead of test

grayside:
Instead of stripping the timestamp prefix from the logs, I had them stripped from the tests. That means the tests are false positives, as the logs from the snippet itself are not valid JSON.

This corrects that by adding `log.SetFlags(0)` to the HTTP handler to remove the timestamp. While doing so there is a bit unusual, I did not want to add another sample, this errs towards keeping the code copyable and working. Happy to adjust for recommendations.


run/logging-manual/main.go
```diff
@@ -79,6 +79,9 @@ func (e Entry) String() string {
 // [START run_manual_logging]
 
 func indexHandler(w http.ResponseWriter, r *http.Request) {
+        // Disable log timestamps, these are added automatically.
+        log.SetFlags(0)
```
tbpg: Would expect this to be in `main`. But, do we need it here to make it work as a Cloud Function? Perhaps an `init`?

grayside: I agree. The reason I did it this way was to avoid adding the overhead of another snippet. I could wrap this in an `init` just above the handler to keep it grouped together.

tbpg: I think `init` makes more sense than setting it inside the handler? Neither is perfect.

run/logging-manual/main.go
```diff
@@ -78,6 +78,12 @@ func (e Entry) String() string {
 // [END run_manual_logging_object]
 // [START run_manual_logging]
 
+func init() {
+	// Disable log timestamps, these are added automatically.
+	// Prefix text leads to malformed JSON.
```
tbpg: Sorry I missed this earlier -- the comment isn't clear to me. What adds timestamps automatically? What JSON?

grayside: The structured logging the sample produces is a JSON object. If the timestamp prefix is added, we end up with something like this:

```
2020/01/05 15:41:22 {"message":"Storing link in /dev/null"}
```

This is read as plain text, not JSON, and as a result the structure is not processed for Stackdriver.

tbpg: Will you update the comment? Not sure the best way to phrase it. "These" is ambiguous in the first sentence. And maybe expand the second sentence?

grayside: What about this?

```
// Disable log prefixes such as the default timestamp.
// Prefix text prevents the message from being parsed as JSON.
// A timestamp is added by the process of shipping logs to Stackdriver.
```

tbpg: "... in the process ..."?

SGTM.

grayside: Going to change to this: 

// Disable log prefixes such as the default timestamp.
// Prefix text prevents the message from being parsed as JSON.
// A timestamp is added when shipping logs to Stackdriver.

PR #1136, opened 2019-12-18
healthcare: add FHIR resource search sample and test

noerog:


noerog:
> Looks like you need to run `gofmt -w -s` with Go 1.13. Otherwise, looks good!

Done.

PR #1144, opened 2019-12-26
bigquery: run CMEK tests conditionally.

shollyman:
CMEK snippet testing necessitates that principals have access to the
underlying keychain we use in the samples, which is not allowed to
have a public ACL.  Therefore, to simplify testing for external users
we only run CMEK tests conditionally if we detect that the environment
appears to be our CI environment or if the RUN_CMEK_TESTS env 
variable is set in the environment.


bigquery/snippets/bqtestutil/bqtestutil.go
```diff
@@ -52,3 +53,18 @@ func sanitize(s string, allowedSeparator string) string {
 	}
 	return reg.ReplaceAllString(s, "")
 }
+
+// RunCMEKTests probes whether CMEK-based tests should run.
+func RunCMEKTests() bool {
+	// TRAMPOLINE_BUILD_FILE is set by the CI testing we use, and is a quick
+	// heuristic for testing whether this is a CI-based build.
+	_, ok := os.LookupEnv("TRAMPOLINE_BUILD_FILE")
```
tbpg: Would you mind using another variable? We might disable Trampoline at some point.

bigquery/snippets/job/integration_test.go
```diff
@@ -93,8 +93,12 @@ func TestCopiesAndExtracts(t *testing.T) {
 		t.Errorf("copyTable(%s): %v", testDatasetID, err)
 	}
 
-	if err := copyTableWithCMEK(tc.ProjectID, testDatasetID, "copycmek"); err != nil {
-		t.Errorf("copyTableWithCMEK(%s): %v", testDatasetID, err)
+	if bqtestutil.RunCMEKTests() {
+		if err := copyTableWithCMEK(tc.ProjectID, testDatasetID, "copycmek"); err != nil {
```
tbpg: Consider a subtest with a `t.Skip` call, rather than a `t.Log` message. Same with the other tests (especially the subtest ones). Could also pass `t` to `RunCMEKTests` and have it call `Skip`.

bigquery/snippets/bqtestutil/bqtestutil.go
shollyman: Switched this to one of the default variables KOKORO_BUILD_ID

bigquery/snippets/job/integration_test.go
shollyman: Switched this around.  I also grouped the copy and tests here into groups that can run in parallel.

PR #1145, opened 2019-12-27
testing: publish test log to Build Cop Bot

tbpg:
This will automatically open/close issues for failing/passing tests ([example](https://github.com/GoogleCloudPlatform/golang-samples/issues/1146)). See https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop.

### DIFF
--- a/testing/kokoro/system_tests.sh
+++ b/testing/kokoro/system_tests.sh
@@ -141,4 +141,36 @@ date
 OUTFILE=gotest.out
 2>&1 go test -timeout $TIMEOUT -v . $TARGET | tee $OUTFILE
 
+set +x
+
 cat $OUTFILE | /go/bin/go-junit-report -set-exit-code > sponge_log.xml
+EXIT_CODE=$?
+
+# If we're running system tests, send the test log to the Build Cop Bot.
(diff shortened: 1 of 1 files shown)


tbpg:
cc @bcoe @sofisl @JustinBeckwith

testing/kokoro/system_tests.sh
```diff
@@ -141,4 +141,36 @@ date
 OUTFILE=gotest.out
 2>&1 go test -timeout $TIMEOUT -v . $TARGET | tee $OUTFILE
 
+set +x
```
bcoe: I would be careful about setting this, because if any secrets are passed as arguments they will show up in public logs.

bcoe:
@tbpg awesome work, I'm excited to see this work end to end.

testing/kokoro/system_tests.sh
tbpg: `set -x` turns logging on, `set +x` turns it off. But, whoops, I meant to turn off `set -e` (set at the top of the file) so a test failure wouldn't stop execution. Fixed.

PR #1151, opened 2020-01-07
cloudsql/mysql: add support for Cloud Run

grayside:
Updates the Cloud SQL sample for simultaneous compatibility with App Engine and Cloud Run.


grayside:
This PR shifts the app to be a go module.

I think the failure is:

```
+ go vet ./cloudsql/...
go: warning: "./cloudsql/..." matched no packages
no packages to vet
```

Could there be a problem with go vet discovering a nested module?

grayside:
@tbpg that doesn't change that tests are failing, do you have any insights on the go vet hypothesis?

tbpg:
Yes, it seems `go vet` doesn't check modules outside the main module...

```
➜  go vet ./cloudsql/...          
go: warning: "./cloudsql/..." matched no packages
no packages to vet
➜  go vet ./cloudsql/mysql/database-sql/...
go: directory cloudsql/mysql/database-sql is outside main module
➜  cd ./cloudsql/mysql/database-sql 
➜  go vet ./...
➜
```

I'd vote merge this and file an issue to make sure it gets fixed.

grayside:
Looks like the revised go vet behavior supports all directories, but the `/...` suffix removal is failing for product subdirectories. I will fix that is part of this PR so we have a practical test case.

I thought I had verified this but it's not working today:

```
target_dir="${TARGET%/...}"
```

Looks like some of the light regex magic in BASH with this statement is choking on the lack of alphanumeric characters.

On the other hand, this seems to work fine and not clobber `.`:

```
echo "$TARGET" | sed -r 's/\/\.\.\.$//'
```

Will add this and see how testing goes.

testing/kokoro/system_tests.sh
```diff
@@ -138,7 +138,7 @@ if [ $GOLANG_SAMPLES_GO_VET ]; then
   diff -u <(echo -n) <(gofmt -d -s .)
 
   # Remove the golang-only triple-dot suffix.
-  target_dir="${TARGET%/...}"
+  target_dir=$(echo "$TARGET" | sed -r 's/\/\.\.\.$//')
```
tbpg: Could we use `TARGET_DIRS`, which doesn't have the `./...` at all?

grayside: Somehow I missed that TARGET_DIRS could be a space-delimited list of directories. I can use it, but will need to make other modifications.

grayside: TIL find can take a space-delimited list of directories

PR #1163, opened 2020-01-16
dataproc: clean clusters before running tests

tbpg:
I also updated some of the error messages and pass project IDs directly, rather than a test context.

### DIFF
--- a/dataproc/create_cluster_test.go
+++ b/dataproc/create_cluster_test.go
@@ -27,23 +27,28 @@ import (
 	dataprocpb "google.golang.org/genproto/googleapis/cloud/dataproc/v1"
 )
 
-func teardown(t *testing.T, tc testutil.Context, clusterName, region string) {
-	t.Helper()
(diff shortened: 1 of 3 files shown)


dataproc/create_cluster_test.go
```diff
@@ -52,7 +57,8 @@ func TestCreateCluster(t *testing.T) {
 	clusterName := fmt.Sprintf("go-cc-test-%s", tc.ProjectID)
 	region := "us-central1"
 
-	defer teardown(t, tc, clusterName, region)
+	deleteCluster(tc.ProjectID, clusterName, region) // Delete the cluster if it already exists.
```
bradmiro: This might throw an error if it doesn't exist, but it seems like that's fine since we aren't capturing the error, if I understand correctly.

tbpg: Yes. Updated the comment.

PR #1165, opened 2020-01-17
iot: update inconsistent comment on JWT algorithm

edorivai:


edorivai:
@googlebot I signed it!

PR #1177, opened 2020-01-28
secretmanager: add IAM samples

sethvargo:


secretmanager/secretmanager_test.go
```diff
@@ -106,6 +106,10 @@ func testCleanupSecret(tb testing.TB, name string) {
 	}
 }
 
+func testIamUser(tb testing.TB) string {
+	return "user:sethvargo@google.com"
```
tbpg: Users may clone this repo and run the tests. They probably shouldn't grant you access to their project at all. :) Also, this email may one day stop working. We usually add an environment variable to configure what the value should be and skip the test if the variable is not set. https://github.com/GoogleCloudPlatform/golang-samples/blob/master/testing/kokoro/system_tests.sh#L44 (can use one of the service accounts referenced in that file).

sethvargo: Good point. I updated it to use the service account already defined in system_test.sh: `GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL`

sethvargo:
@tbpg updated

secretmanager/secretmanager_test.go
```diff
@@ -107,7 +108,14 @@ func testCleanupSecret(tb testing.TB, name string) {
 }
 
 func testIamUser(tb testing.TB) string {
-	return "user:sethvargo@google.com"
+	tb.Helper()
+
+	v := os.Getenv("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL")
+	if v == "" {
+		tb.Fatalf("testIamUser: missing GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL")
```
tbpg: Please skip instead. Tests should ideally pass if someone clones the repo and runs the tests.

sethvargo: Done

sethvargo:
@tbpg updated to skip instead.

PR #1183, opened 2020-01-30
appengine: added closure of client to avoid memory spike

deepnirmal:
While using golang create task example, I experienced there was a spike in memory since the client was not closed.

Added closure of client to allow go clean the memory and to avoid production memory spike.

Thanks.

### DIFF
--- a/appengine/go11x/tasks/create_task/create_task.go
+++ b/appengine/go11x/tasks/create_task/create_task.go
@@ -61,6 +61,9 @@ func createTask(projectID, locationID, queueID, message string) (*taskspb.Task, 
 		return nil, fmt.Errorf("cloudtasks.CreateTask: %v", err)
 	}
 
+	// Closing the client is important, if not done then it will result in Memory spike
+	defer client.Close()
+
 	return createdTask, nil
 }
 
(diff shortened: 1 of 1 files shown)


appengine/go11x/tasks/create_task/create_task.go
```diff
@@ -61,6 +61,9 @@ func createTask(projectID, locationID, queueID, message string) (*taskspb.Task,
 		return nil, fmt.Errorf("cloudtasks.CreateTask: %v", err)
 	}
 
+	// Closing the client is important, if not done then it will result in Memory spike
+	defer client.Close()
```
tbpg: Will you move this line to right below the error check for `NewClient` above? It's possible there will be some other issue between creating the client and here, so we want to make sure the `defer` is actually set.

appengine/go11x/tasks/create_task/create_task.go
```diff
@@ -61,6 +61,9 @@ func createTask(projectID, locationID, queueID, message string) (*taskspb.Task,
 		return nil, fmt.Errorf("cloudtasks.CreateTask: %v", err)
 	}
 
+	// Closing the client is important, if not done then it will result in Memory spike
```
tbpg: I don't think we need the comment? I could be convinced, though.

appengine/go11x/tasks/create_task/create_task.go
deepnirmal: Yes, sure

appengine/go11x/tasks/create_task/create_task.go
deepnirmal: Sure yeah

PR #1212, opened 2020-02-04
profiler: add shakesapp sample

nolanmar511:
This is a more complicated application that can be used to demonstrate how Stackdriver Profiler can be used to optimize an application.

After this change is sumbitted, we will add progressively optimized versions in of the code in subdirectories within the shakesapp directory called "rev_1" and "rev_2"


nolanmar511:
@kalyanac and @aalexand  -- PTAL

profiler/shakesapp/shakesapp_main.go
```diff
@@ -0,0 +1,73 @@
...
+func main() {
+	flag.Parse()
+
+	if err := profiler.Start(profiler.Config{
+		Service:        "shakesapp",
+		ServiceVersion: *version,
+		ProjectID:      *projectID,
+		MutexProfiling: true,
+	}); err != nil {
+		log.Fatalf("failed to start profiler: %v", err)
```
aalexand: Use log.Exit/Exitf in main().

profiler/shakesapp/shakesapp/shakesapp.proto
```diff
@@ -0,0 +1,36 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// To update shakesapp.pb.go, run
+// protoc --go_out=plugins=grpc:. shakesapp.proto
+// --plugin=/path/to/protoc-gen-go
```
aalexand: Please make it so the command can be copy-pasted. You can either escape the newline, but the easiest is make it one line.

profiler/shakesapp/README.md
```diff
@@ -0,0 +1,32 @@
+# Profiler Shakespeare Application in Go
```
aalexand: Is it intentional that the casing of this heading is not consistent with the next one below?

profiler/shakesapp/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+This application starts a server which will return the number of times a word or
```
aalexand: Do we want to mention that this app is intentionally non-optimal and the goal of it is demonstrate how to make a set of progressive steps towards optimizing it? I would also include the exact progressive versions in subdirectories, like we do in the internal version. (just in case - let's not use Git branches for that).

profiler/shakesapp/README.md
```diff
@@ -0,0 +1,32 @@
...
+phrase appears in the works of Shakespeare, and sends queries to that server to
+create load.
+
+Profiler is enabled for this application, and can be used to identify
+opportunities to optimize the server code.
+
+## Running the application
+
+This application should be run with Go 1.14. Once that version of Go is
```
aalexand: This will obsolete when Go 1.15 is out. Do we care to be that prescriptive?

profiler/shakesapp/README.md
```diff
@@ -0,0 +1,32 @@
...
+installed:
+
+*   Get the application code:
+
+    ```sh
+    git clone https://github.com/GoogleCloudPlatform/golang-samples.git
+    cd golang-samples/getting-started/profiler/shakesapp
```
aalexand: I am not sure why this is under getting-started. This is a how-to app, not a getting started I think. There is already https://github.com/GoogleCloudPlatform/golang-samples/tree/master/profiler/profiler_quickstart, that one is getting started. I would put it just under profiler/. I think it is also preferred that all profiler apps are in a single place, in a single nice flat list of directories to avoid future diversion of apps.

profiler/shakesapp/README.md
```diff
@@ -0,0 +1,32 @@
...
+*   Install dependencies:
+
+    ```sh
+    go get .
+    ```
+
+*   Run the application:
+
+    ```sh
+    go run . -project_id="your-project-id" -version="application-version"
```
aalexand: Specifying project ID should never be required when running on GCP and these instructions should favor that as the default case. So we should never ask user to specify the project ID in any public profiler doc, except the "running outside GCP" doc section.

profiler/shakesapp/shakesapp/client.go
```diff
@@ -0,0 +1,68 @@
...
+
+	respErrs := make(chan error)
+	inFlightCh := make(chan bool, reqsInFlight)
+	for i := 0; i < numReqs; i++ {
+		go func() {
+			inFlightCh <- true
+			q := queries[rand.Intn(len(queries))]
+			resp, err := client.GetMatchCount(ctx, &ShakespeareRequest{Query: q.query})
+			if resp.MatchCount != q.wantMatchCount {
+				panic(fmt.Sprintf("GetMatchCount(%q): got %d matches, want %d", q.query, resp.MatchCount, q.wantMatchCount))
```
aalexand: Is it intentionally a panic instead of returning an error? It would be good to add a "why" comment if so.

profiler/shakesapp/shakesapp_main.go
nolanmar511: Done.

profiler/shakesapp/shakesapp_main.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"cloud.google.com/go/profiler"
+	"google.golang.org/grpc"
+
+	"shakesapp/shakesapp"
+)
+
+var (
+	version      = flag.String("version", "1.0.0", "application version")
+	projectID    = flag.String("project_id", "", "project ID to run profiler with; required when running outside of GCP.")
```
aalexand: "required" -> "only required"

aalexand: or "required when" -> "required when and only when"

profiler/shakesapp/shakesapp_main.go
```diff
@@ -0,0 +1,73 @@
...
+	reqsInFlight = flag.Int("requests_in_flight", 1, "number of requests to run in parallel")
+	numRounds    = flag.Int("num_rounds", 0, "number of simulation rounds (0 for infinite)")
+)
+
+func main() {
+	flag.Parse()
+
+	if err := profiler.Start(profiler.Config{
+		Service:        "shakesapp",
+		ServiceVersion: *version,
```
aalexand: It would be good to pre-code the version names for the versions of the app corresponding to each optimization step (which I think we should include) so that we can guide people directly as "compare version X with version Y using the diff feature".

profiler/shakesapp/shakesapp_main.go
nolanmar511: Done.

profiler/shakesapp/README.md
nolanmar511: Done.

profiler/shakesapp/shakesapp/client.go
nolanmar511: Done.

profiler/shakesapp/README.md
nolanmar511: Done.

profiler/shakesapp/README.md
nolanmar511: Removed.

profiler/shakesapp/README.md
nolanmar511: Updated so cases match.

profiler/shakesapp/shakesapp/shakesapp.proto
nolanmar511: Done.

profiler/shakesapp/shakesapp_main.go
nolanmar511: Done.

profiler/shakesapp/README.md
nolanmar511: Stated that this application is non-optimal.

profiler/shakesapp/shakesapp/client.go
```diff
@@ -52,8 +52,8 @@ func SimulateClient(ctx context.Context, addr string, numReqs, reqsInFlight int)
 			inFlightCh <- true
 			q := queries[rand.Intn(len(queries))]
 			resp, err := client.GetMatchCount(ctx, &ShakespeareRequest{Query: q.query})
-			if resp.MatchCount != q.wantMatchCount {
-				panic(fmt.Sprintf("GetMatchCount(%q): got %d matches, want %d", q.query, resp.MatchCount, q.wantMatchCount))
+			if err == nil && (resp.MatchCount != q.wantMatchCount) {
```
aalexand: I think

```
inFlightCh <- true
respErrs <- func() error {
  q := queries[rand.Intn(len(queries))]
  resp, err := client.GetMatchCount(ctx, &ShakespeareRequest{Query: q.query})
  if err != nil {
    return err
  }
  if resp.MatchCount != q.wantMatchCount {
    return fmt.Errorf("GetMatchCount(%q): got %d matches, want %d", q.query, resp.MatchCount, q.wantMatchCount)
  }
}()
<-inFlightCh
```

might be a bit more convenient to read in terms of how control flow is usually structured in Go programs ("if err != nil { return err }").

profiler/shakesapp/shakesapp_main.go
```diff
@@ -56,18 +57,18 @@ func main() {
 	shakesapp.RegisterShakespeareServiceServer(server, shakesapp.NewServer())
 	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", *port))
 	if err != nil {
-		log.Fatalf("failed to listen: %v", err)
+		glog.Exitf("failed to listen: %v", err)
```
aalexand: nit: failed -> Failed

Please make a pass to check that log messages start with capital letter, error strings - with lower.

aalexand:
> though, maintaining this code will require a bit more care

Yes, but I think it's required to make sure the user experience is reproducible and smooth. Trying to express optimization steps with code snippets or explanations elsewhere (like documentation) is even more error-prone. A more maintainable alternative I can think of is a set of diffs that are automatically applied before each commit to generate the optimized versions, but that has its disadvantages too (potentially more fragile).

By the way (since it's related), I think we should limit the total number of apps / samples that we have. I think with this one we've pretty much reached the limit of the number of sample apps we can maintain in a decent shape.

profiler/shakesapp/shakesapp/client.go
nolanmar511: Done.

profiler/shakesapp/shakesapp_main.go
nolanmar511: Done.

nolanmar511:
PTAL

profiler/shakesapp/README.md
```diff
@@ -0,0 +1,29 @@
+# Profiler Shakespeare application
+
+This application starts a server which will return the number of times a word
+or phrase appears in the works of Shakespeare, and sends queries to that server
```
aalexand: sends -> send (because "will")

profiler/shakesapp/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+    ```sh
+    git clone https://github.com/GoogleCloudPlatform/golang-samples.git
+    cd golang-samples/profiler/shakesapp
+    ```
+
+*   Install dependencies:
+
+    ```sh
+    go get .
```
aalexand: I think this is not needed with modules.

profiler/shakesapp/README.md
nolanmar511: Done.

profiler/shakesapp/README.md
nolanmar511: Done.

nolanmar511:
I have verified that follow instructions here produce interesting profiles.

nolanmar511:
PTAL

After this change is merged, I will have a follow-up change in which I make a copy of the code for each step of the fixes. Each copy will also set a unique version to allow for easy comparison.

profiler/shakesapp/shakesapp_main.go
aalexand: Not addressed fwiw.

profiler/shakesapp/shakesapp/client.go
```diff
@@ -0,0 +1,73 @@
...
+	conn, err := grpc.Dial(addr, grpc.WithInsecure())
+	if err != nil {
+		return err
+	}
+	client := NewShakespeareServiceClient(conn)
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	respErrs := make(chan error)
```
tbpg: Consider using `errgroup`. The semantics would be slightly different because `Wait` waits for all functions to return. It looks like this will return (and cancel via the context) as soon as one of the calls returns an error.

profiler/shakesapp/shakesapp/server.go
```diff
@@ -0,0 +1,121 @@
...
+	"google.golang.org/api/option"
+)
+
+// NewServer returns an implementation of the server for ShakespeareService
+// (defined in shakesapp.proto).
+func NewServer() ShakespeareServiceServer {
+	return &server{}
+}
+
+type server struct{}
```
tbpg: Might move this above `NewServer` and give it a comment. Can look a little odd to someone who isn't used to gRPC services.

profiler/shakesapp/shakesapp/server.go
```diff
@@ -0,0 +1,121 @@
...
+// readFiles reads the content of files within the specified bucket with the
+// specified prefix path in parallel and returns their content. It fails if
+// operations to find or read any of the files fails.
+func readFiles(ctx context.Context, bucketName, prefix string) ([]string, error) {
+	type resp struct {
+		s   string
+		err error
+	}
+
+	client, err := storage.NewClient(ctx, option.WithoutAuthentication())
```
tbpg: Why without auth?

profiler/shakesapp/shakesapp/server.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to iterate over files in %s starting with %s: %v", bucketName, prefix, err)
+		}
+		if attrs.Name != "" {
+			paths = append(paths, attrs.Name)
+		}
+	}
+
+	resps := make([]chan resp, len(paths))
```
tbpg: Consider using `errgroup` again: https://pkg.go.dev/golang.org/x/sync/errgroup?tab=doc#example-Group-Parallel

profiler/shakesapp/shakesapp/shakesapp.pb.go
```diff
@@ -0,0 +1,210 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
tbpg: Probably need to add a license exception to https://github.com/GoogleCloudPlatform/golang-samples/blob/master/license_test.go#L42.

profiler/shakesapp/shakesapp_main.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020 Google LLC
```
tbpg: Filename can probably just be main.go, given this is in the shakesapp directory?

profiler/shakesapp/shakesapp_main.go
```diff
@@ -0,0 +1,73 @@
...
+	"flag"
+	"fmt"
+	"net"
+	"time"
+
+	"cloud.google.com/go/profiler"
+	"github.com/golang/glog"
+	"google.golang.org/grpc"
+
+	"shakesapp/shakesapp"
```
tbpg: Should use the full import path. Otherwise, this won't work for others.

profiler/shakesapp/shakesapp_main.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+var (
+	version      = flag.String("version", "1.0.0", "application version")
+	projectID    = flag.String("project_id", "", "project ID to run profiler with; only required when running outside of GCP.")
+	port         = flag.Int("port", 7788, "service port")
+	numReqs      = flag.Int("num_requests", 20, "number of requests to simulate")
+	reqsInFlight = flag.Int("requests_in_flight", 1, "number of requests to run in parallel")
```
tbpg: Consider naming this something like `concurrency`. `reqsInFlight` sounds like it would be a list of the requests that are currently in flight.

profiler/shakesapp/shakesapp_main.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"net"
+	"time"
+
+	"cloud.google.com/go/profiler"
+	"github.com/golang/glog"
```
tbpg: Please don't use `glog`. It registers global flags that can interfere with other libraries. Works for internal stuff, but open source or external stuff is not as nice. Maybe for a personal project, but we shouldn't use it in a sample.

tbpg:
For the optimized copies, what directory/naming scheme are you planning on? Might change where we put the first one.

profiler/shakesapp/shakesapp/client.go
aalexand: I think using errgroup here is a distraction. This sample is to demonstrate some performance aspects that profiler can be helpful with, so I think using language primitives makes more sense than using libraries that would be more standard in the actual production code.

profiler/shakesapp/shakesapp/server.go
aalexand: Perhaps because it's a loopback (the server is spawned by the same process)?

profiler/shakesapp/shakesapp/client.go
nolanmar511: I'd also like to avoid the complexity of using an errgoup here. 
I'm marking this as resolved.

profiler/shakesapp/shakesapp/server.go
nolanmar511: Marking this as resolved. Given that this is a loopback, I don't think we need to introduce the complexity of authentication.

profiler/shakesapp/shakesapp/server.go
nolanmar511: Done.

profiler/shakesapp/shakesapp_main.go
nolanmar511: Done.

profiler/shakesapp/shakesapp_main.go
nolanmar511: I'm using go modules, so I believe this is standard notation; I've verified that the instructions provided in the README work in cloud shell (in addition to working locally for me), and other members of our team have successfully started this application.

profiler/shakesapp/shakesapp_main.go
nolanmar511: Done.

nolanmar511:
PTAL

profiler/shakesapp/go.mod
```diff
@@ -0,0 +1,12 @@
+module shakesapp
```
tbpg: Would you mind making this the full path?

`github.com/GoogleCloudPlatform/golang-samples/profiler/shakesapp`

Otherwise, it can run into trouble using other packages from this repo, being imported, etc. Neither of those are important for this sample, but it's good to do in general (this has bitten us in the past).

profiler/shakesapp/go.mod
```diff
@@ -0,0 +1,12 @@
...
+
+go 1.14
+
+require (
+	cloud.google.com/go v0.52.0
+	cloud.google.com/go/storage v1.5.0
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
```
tbpg: Looks like you need to run `go mod tidy`.

profiler/shakesapp/shakesapp/client.go
```diff
@@ -0,0 +1,73 @@
...
+				resp, err := client.GetMatchCount(ctx, &ShakespeareRequest{Query: q.query})
+				if err != nil {
+					return err
+				}
+				if resp.MatchCount != q.wantMatchCount {
+					return fmt.Errorf("GetMatchCount(%q): got %d matches, want %d", q.query, resp.MatchCount, q.wantMatchCount)
+				}
+				return nil
+			}()
+			<-inFlightCh
```
tbpg: If an error is returned in one goroutine above, this line won't run, and other goroutines cloud block on the `<- true` line, when they try to reserve a spot in `inFlightCh`. Move this to a `defer` right after the `<- true` line?

profiler/shakesapp/shakesapp/server.go
```diff
@@ -0,0 +1,123 @@
...
+// operations to find or read any of the files fails.
+func readFiles(ctx context.Context, bucketName, prefix string) ([]string, error) {
+	type resp struct {
+		s   string
+		err error
+	}
+
+	client, err := storage.NewClient(ctx, option.WithoutAuthentication())
+	if err != nil {
+		return []string{}, fmt.Errorf("failed to create storage client: %s", err)
```
tbpg:
```suggestion
		return nil, fmt.Errorf("failed to create storage client: %s", err)
```

profiler/shakesapp/shakesapp/server.go
tbpg: Resolving. Sounds good to not use.

profiler/shakesapp/shakesapp/server.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to iterate over files in %s starting with %s: %v", bucketName, prefix, err)
+		}
+		if attrs.Name != "" {
+			paths = append(paths, attrs.Name)
+		}
+	}
+
+	resps := make([]chan resp, len(paths))
```
tbpg: Why a slice of channels instead of a single channel?

profiler/shakesapp/go.mod
nolanmar511: Thanks, done.

profiler/shakesapp/go.mod
nolanmar511: Thanks! done.

profiler/shakesapp/shakesapp/client.go
nolanmar511: Thanks! Done.

profiler/shakesapp/shakesapp/server.go
nolanmar511: Thanks! Done.

profiler/shakesapp/shakesapp/server.go
nolanmar511: Thanks -- using a single channel now.

nolanmar511:
PTAL

profiler/shakesapp/shakesapp/client.go
```diff
@@ -0,0 +1,73 @@
...
+				if resp.MatchCount != q.wantMatchCount {
+					return fmt.Errorf("GetMatchCount(%q): got %d matches, want %d", q.query, resp.MatchCount, q.wantMatchCount)
+				}
+				return nil
+			}()
+		}()
+	}
+	for i := 0; i < numReqs; i++ {
+		if err := <-respErrs; err != nil {
+			return err
```
tbpg: If this returns an error, will the remaining goroutines block when trying to send to `respErrors`?

nolanmar511: Updated so an error is returned only after waiting for all responses.

nolanmar511:
PTAL

nolanmar511:
Friendly ping?

nolanmar511:
@tbpg -- If you could take a look, that would be awesome!

profiler/shakesapp/shakesapp/client.go
tbpg: Looks like this still returns as soon as there is any error? I might be misunderstanding.

nolanmar511:
Sorry -- I'd originally only fixed how errors were returned in server.go, not client,go.

Now should be fixed in both locations. And thanks for reviewing! (and PTAL)

PR #1216, opened 2020-02-05
all, run: go mod tidy, don't use log.Writer (added in Go 1.13)

tbpg:
Example `log.Writer()` failure: https://source.cloud.google.com/results/invocations/eaa8e2f4-93f2-45d0-87ba-70d8bbabe8a6/targets/cloud-devrel%2Fgo%2Fgolang-samples%2Fpresubmit%2Fgo112/log

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,6 @@ require (
 	cloud.google.com/go/spanner v1.1.0
 	cloud.google.com/go/storage v1.3.0
 	contrib.go.opencensus.io/exporter/stackdriver v0.12.9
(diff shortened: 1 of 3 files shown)


grayside:
Seems a bit risky to train folks to assume Stderr. Is there a more ideal practice we should be using? If it's been removed, why is it visible without a note in https://pkg.go.dev/log?

tbpg:
Whoops! It was _added_ in Go 1.13. It's not part of the sample, so I think it's OK. I looked a bit and stderr seems to be the default, but I'm open to other ways of getting it.

PR #1217, opened 2020-02-06
storage: fix broken s3_sdk test

### LINKED ISSUE #1213: storage: s3_sdk tests failing
```
list_gcs_buckets_test.go:45: listGCSBuckets: MalformedSecurityHeader: Your request has a malformed header.
        	status code: 400, request id: , host id: 
    list_gcs_buckets_test.go:50: listGCSBuckets got
        ----
        
        ----
        Want to contain
        ----
        Buckets:
        ----
```

Going to send a PR to skip.

tritone:
Changed this so it creates and cleans up its own HMAC key for
testing purposes instead of relying on secrets in Keystore. Also
changed the test to fail rather than skip if the key is missing.

Fixes #1213


storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,29 +17,78 @@ package s3sdk
...
 }
 
+// Create a key for testing purposes and set environment variables
+func createTestKey(projectID string) (*storage.HMACKey, error) {
+	ctx := context.Background()
+	key, err := storageClient.CreateHMACKey(ctx, projectID, serviceAccountEmail)
+	if err != nil {
+		return nil, fmt.Errorf("storage.NewClient: %v", err)
+	}
+	if err = os.Setenv("STORAGE_HMAC_ACCESS_KEY_ID", key.AccessID); err != nil {
```
frankyn: Instead of passing the values through environment variables, can you pass them directly to the test?

tritone: Yeah will do-- I thought the sample itself used the env variables for some reason but that is not the case.

tritone:
This didn't work in the CI, so I need to figure out why that is!

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,29 +17,78 @@ package s3sdk
...
 	"os"
 	"strings"
 	"testing"
+
+	"cloud.google.com/go/storage"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
 )
 
+var serviceAccountEmail = os.Getenv("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL")
+var storageClient *storage.Client
```
tbpg: In general, try to avoid global variables in Go. This isn't a bad case since it's a test. But, it would be better to get the email in where it's needed and pass the client as an argument where it's needed. It makes the dependencies and control flow very clear.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,29 +17,78 @@ package s3sdk
...
+
 func TestMain(m *testing.M) {
 	log.SetOutput(ioutil.Discard)
 	s := m.Run()
 	log.SetOutput(os.Stderr)
+
+	ctx := context.Background()
+	storageClient, _ = storage.NewClient(ctx)
```
tbpg: Check the error. Also, s/storageClient/client/ -- don't need to disambiguate clients here since we are only using storage.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,29 +17,78 @@ package s3sdk
...
+	defer storageClient.Close()
+
+	tc, _ := testutil.ContextMain(m)
```
tbpg: Check `ok`. `GOLANG_SAMPLES_PROJECT_ID` might not be set.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,29 +17,78 @@ package s3sdk
...
+
 	os.Exit(s)
 }
 
+// Create a key for testing purposes and set environment variables
+func createTestKey(projectID string) (*storage.HMACKey, error) {
+	ctx := context.Background()
+	key, err := storageClient.CreateHMACKey(ctx, projectID, serviceAccountEmail)
+	if err != nil {
+		return nil, fmt.Errorf("storage.NewClient: %v", err)
```
tbpg:
```suggestion
		return nil, fmt.Errorf("CreateHMACKey: %v", err)
```

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,29 +17,78 @@ package s3sdk
...
+	"cloud.google.com/go/storage"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
 )
 
+var serviceAccountEmail = os.Getenv("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL")
+var storageClient *storage.Client
+
 func TestMain(m *testing.M) {
 	log.SetOutput(ioutil.Discard)
 	s := m.Run()
```
tbpg: This seems to be running the tests here, before the key is created.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,29 +17,78 @@ package s3sdk
...
+	if err != nil {
+		log.Fatalf("error setting up HMAC key: %v", err)
+	}
+	defer deleteTestKey(key)
+
 	os.Exit(s)
 }
 
+// Create a key for testing purposes and set environment variables
+func createTestKey(projectID string) (*storage.HMACKey, error) {
```
tbpg: Is there any sort of limit on the number of keys? If we delete, does it delete forever or go into a "disabled" state that still counts against a limit?

We've had issues with other key type resources where they don't/can't get completely cleaned up and we run into a limit.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,29 +17,78 @@ package s3sdk
...
 	log.SetOutput(ioutil.Discard)
 	s := m.Run()
 	log.SetOutput(os.Stderr)
+
+	ctx := context.Background()
+	storageClient, _ = storage.NewClient(ctx)
+	defer storageClient.Close()
+
+	tc, _ := testutil.ContextMain(m)
+	key, err := createTestKey(tc.ProjectID)
```
tbpg: Does this need to be in `TestMain`? If there is only one test, we can create/delete the key in there.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,29 +17,78 @@ package s3sdk
...
+
+	"cloud.google.com/go/storage"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
 )
 
+var serviceAccountEmail = os.Getenv("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL")
+var storageClient *storage.Client
+
 func TestMain(m *testing.M) {
 	log.SetOutput(ioutil.Discard)
```
tbpg: The sample shouldn't be logging anything, so I don't think this is necessary. Can probably delete `TestMain` (see other comment).

storage/s3_sdk/list_gcs_buckets_test.go
tritone: Done

storage/s3_sdk/list_gcs_buckets_test.go
tritone: Not necessary to check if we use SystemTest instead.

tritone:
Don't fully understand the Kokoro failure here, in particular why it's passing or failing depending on the version of Go. @tbpg any ideas?

tbpg:
Seems like something is flaky about the test. I don't think it's dependent on the Go version -- probably just happened to be those two. The error is about the signature not matching?

tritone:
Followed up with the GCS team and identified the source of the test flakiness-- HMAC keys may take up to 15s to propagate before they can be used. I added a retry/timeout to the test and it looks to be passing now. Will remove logging statements and then hopefully we can get this merged.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,37 +17,83 @@ package s3sdk
...
+	}
+	defer deleteTestKey(ctx, client, key)
+
+	log.Printf("ProjectID: %v", tc.ProjectID)
+	log.Printf("key: %v", key.AccessID)
+	log.Printf("service account: %v", key.ServiceAccountEmail)
+
+	var buf *bytes.Buffer
+	// New HMAC key may take up to 15s to propagate.
+	timeout := time.After(15 * time.Second)
```
tbpg: Please use `testutil.Retry`. Be sure to update the failure to `r.Errorf` instead of calling `t.Fatalf` directly.

tritone: Done, thanks for noting this.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,37 +17,73 @@ package s3sdk
...
+// Deactivate and delete the given key. Should operate as a teardown method.
+func deleteTestKey(ctx context.Context, client *storage.Client, key *storage.HMACKey) {
+	handle := client.HMACKeyHandle(key.ProjectID, key.AccessID)
+	if key.State == "ACTIVE" {
+		handle.Update(ctx, storage.HMACKeyAttrsToUpdate{State: "INACTIVE"})
+	}
+	if key.State != "DELETED" {
+		handle.Delete(ctx)
+	}
+}
```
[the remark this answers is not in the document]
tritone: Done

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,37 +17,73 @@ package s3sdk
...
-		t.Errorf("listGCSBuckets: %v", err)
+		log.Fatalf("error setting up HMAC key: %v", err)
 	}
+	defer deleteTestKey(ctx, client, key)
+
+	var buf *bytes.Buffer
+	// New HMAC key may take up to 15s to propagate, so we need to retry for up
+	// to that amount of time.
+	testutil.Retry(t, 75, time.Millisecond*200, func (r *testutil.R) {
+		buf = new(bytes.Buffer)
```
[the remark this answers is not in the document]
tritone: Didn't know about this, thanks! Done.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,37 +17,73 @@ package s3sdk
...
 	}
+	defer client.Close()
 
-	buf := new(bytes.Buffer)
-	_, err := listGCSBuckets(buf, googleAccessKeyID, googleAccessKeySecret)
+	// Set up service account HMAC key to use for this test.
+	key, err := createTestKey(ctx, client, tc.ProjectID)
 	if err != nil {
-		t.Errorf("listGCSBuckets: %v", err)
+		log.Fatalf("error setting up HMAC key: %v", err)
```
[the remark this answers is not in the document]
tritone: Whoops, my mistake. Will fix.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -17,37 +17,73 @@ package s3sdk
...
-		t.Errorf("listGCSBuckets got\n----\n%s\n----\nWant to contain\n----\n%s\n----", got, want)
+		t.Fatalf("listGCSBuckets got\n----\n%s\n----\nWant to contain\n----\n%s\n----", got, want)
 	}
 }
+
+// Create a key for testing purposes and set environment variables
+func createTestKey(ctx context.Context, client *storage.Client, projectID string) (*storage.HMACKey, error) {
+	email := os.Getenv("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL")
+	if email == "" {
+		return nil, fmt.Errorf("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL must be defined in the environment")
```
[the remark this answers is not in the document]
tritone: I can change this but I'm really not a fan of this pattern. It means that you could inadvertently remove an env variable from kokoro and never realize that you're accidentally skipping tests. I know the python samples repo for example will fail if you're missing env variables.

storage/s3_sdk/list_gcs_buckets_test.go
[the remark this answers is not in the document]
tritone: Actually this causes a seg fault as written, but works if I use new above when defining buf.

storage/s3_sdk/list_gcs_buckets_test.go
```diff
@@ -64,10 +63,10 @@ func TestList(t *testing.T) {
 }
 
 // Create a key for testing purposes and set environment variables
-func createTestKey(ctx context.Context, client *storage.Client, projectID string) (*storage.HMACKey, error) {
+func createTestKey(ctx context.Context, t *testing.T, client *storage.Client, projectID string) (*storage.HMACKey, error) {
 	email := os.Getenv("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL")
 	if email == "" {
-		return nil, fmt.Errorf("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL must be defined in the environment")
+		t.Skip("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL must be defined in the environment")
```
tbpg:
```suggestion
		t.Skip("GOLANG_SAMPLES_SERVICE_ACCOUNT_EMAIL must be defined in the environment")
		return
```

PR #1219, opened 2020-02-07
bigquery: add defers to explicitly close all clients

### LINKED ISSUE #1218: bigquery: confirm all clients are closed
P1 bug since it leads to a memory leak and has affected a customer.

Spot check shows we're leaking at least one client: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/bigquery/simpleapp/simpleapp.go#L57

shollyman:
Our samples implicitly leveraged short lifetimes, and thus didn't
 include explicit closures.  We've seen examples in the wild of
 people missing this for longer-lived clients, so this makes all samples
 more explicit about client lifecycle.

Fixes #1218


bigquery/simpleapp/simpleapp.go
```diff
@@ -55,6 +55,7 @@ func query(proj string) (*bigquery.RowIterator, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer client.Close()
```
tbpg: This returns a `*bigquery.RowIterator`. Are we sure we can close the client when this returns? Also, the comment says it returns a slice, but it doesn't.

bigquery/simpleapp/simpleapp.go
```diff
@@ -55,6 +55,7 @@ func query(proj string) (*bigquery.RowIterator, error) {
 	if err != nil {
 		return nil, err
```
tbpg: Update to `fmt.Errorf`?

bigquery/bigquery_quickstart/main.go
```diff
@@ -36,6 +36,7 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to create client: %v", err)
```
tbpg: Update to `bigquery.NewClient`?

bigquery/simpleapp/simpleapp.go
shollyman: Refactored simpleapp so we construct client in main and it persists through the lifecycle.

shollyman: And addressed comment.

bigquery/bigquery_quickstart/main.go
shollyman: done as part of refactor.

bigquery/simpleapp/simpleapp.go
shollyman: done.

bigquery/simpleapp/simpleapp.go
```diff
@@ -46,17 +56,8 @@ func main() {
...
-	client, err := bigquery.NewClient(ctx, proj)
-	if err != nil {
-		return nil, err
-	}
-	defer client.Close()
-	// [END bigquery_simple_app_client]
+// query returns a row iterator suitable for reading query results.
+func query(ctx context.Context, client *bigquery.Client) (*bigquery.RowIterator, error) {
 
 	// [START bigquery_simple_app_query]
```
tbpg: Might want to include the function declaration in the sample? Up to you.

shollyman: I'd defer that until we have a chance to address all languages, since they all use this sparse style: https://cloud.google.com/bigquery/docs/quickstarts/quickstart-client-libraries#bigquery_simple_app_client-go

PR #1227, opened 2020-02-11
testing: go version compatibility for module constraints

### LINKED ISSUE #1224: testing: support new go version language features per go submodule
## Current Behavior

Currently all samples are tested against all supported go language versions.
If an individual sample wants to use newer language features (e.g., errors.Is from go1.13), it will fail testing in go1.11 and go1.12.

## Preferred Behavior

When the [test-runner](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/testing/kokoro/system_tests.sh) descends into a go module to run tests, it checks to see if the go.mod language version constraint is higher than the current runtime version. If so, the module is skipped.

## Implementation Consideration

Filter out invalid modules from `$GO_CHANGED_MODULES`, `$GO_CHANGED_SUBMODULES`, and `$GO_TEST_MODULES` before use.

grayside:
Fixes #1224 

This is a first experiment at preventing tests from running for a given go version with incompatible sub-modules. It uses the go version in the go.mod as a minimum compatibility, if the runtime environment of the test is less than that version the module is skipped.

If anything goes wrong in the version analysis, the test is run.
The most recent version of go is always run to reduce the risk of bugs creating false positives.

The initial implementation is using `go run` per submodule twice in each system_tests.sh execution.


grayside:
It looks like the test failure is go 1.12 having a filesystem issue.
This PR should now be close enough for at least concept review.

testing/kokoro/version.go
```diff
@@ -0,0 +1,126 @@
...
+	if v.GreaterThan(max) {
+		log.Printf("always run tests for most advanced go version: go%s", maxVersionStr)
+		return true, nil
+	}
+
+	return compare(v, module)
+}
+
+// compare determines if the module has a newer version than the runtime.
+// This command should fail "successful".
```
tbpg: Fail successful?

testing/kokoro/version.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	var module struct {
+		Version string `json:"Go"`
+	}
+	if err := json.Unmarshal(out, &module); err != nil {
+		return "", err
+	}
+
+	return module.Version, nil
+}
```
tbpg: Newline.

testing/kokoro/version.go
```diff
@@ -0,0 +1,126 @@
...
+
+// valid determines if the module should be tested.
+func valid(max *version.Version, v *version.Version, module string) (bool, error) {
+	if v.GreaterThan(max) {
+		log.Printf("always run tests for most advanced go version: go%s", maxVersionStr)
+		return true, nil
+	}
+
+	return compare(v, module)
```
tbpg: Maybe inline?

testing/kokoro/system_tests.sh
```diff
@@ -194,8 +204,16 @@ date
...
+  set +x
+  echo "Running all tests compatible with current go version"
+  # Test with all go modules compatible with go version under test.
+  for i in $(find . -name go.mod); do
+    mod="$(dirname $i)"
+    echo "Running 'version compatibility check' for '$mod'"
+    go run $ROOT_DIR/testing/kokoro/version.go -module=$mod && GO_TEST_MODULES="$GO_TEST_MODULES $mod"
+  done
+  set -x
 elif [[ -z "${TARGET_DIRS// }" ]]; then
```
tbpg: I guess the root is always compatible with all versions?

testing/kokoro/version.go
grayside: I broke it out because the whole PR is intended as art for design discussion. Do you like the current approach or do you think we need to arrange this so there's a single `go run` to assemble the list of modules?

testing/kokoro/version.go
grayside: In this design, any error is the program should lead to the module being tested. This promotes false negatives instead of false positives.

testing/kokoro/system_tests.sh
grayside: My assumption is that the root module will always be the longest tail. It is checked where the root module is part of a list of all/changed modules.

FWIW, I don't understand the purpose of the root tests.

tbpg: The root tests do stuff like region tag checks and license header checks. So, we should always run them, regardless of the PR. I think it's safe to assume the oldest Go version will work with them.

testing/kokoro/version.go
```diff
@@ -0,0 +1,126 @@
...
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/hashicorp/go-version"
+)
+
+// maxVersion is the most advanced go version in testing.
+const maxVersionStr="1.13"
```
tbpg: Should probably document somewhere this needs to be updated. Maybe we can auto-detect from listing the `*.cfg` files in this directory?

testing/kokoro/version.go
tbpg: I think this makes sense to me. "Run this binary to get the modules compatible with this version of Go." Right?

testing/kokoro/version.go
tbpg: Maybe drop the error return value? Can then update this comment to say, "If there is an error comparing the versions, compare returns true, meaning the tests should be run."

testing/kokoro/version.go
```diff
@@ -0,0 +1,126 @@
...
+	return true, nil
+}
+
+// moduleVersion extracts the minimum go version from the go.mod
+func moduleVersion(m string) (string, error) {
+	p, err := filepath.Abs(m)
+	if err != nil {
+		return "", fmt.Errorf("filepath.Abs: (%s) %v", m, err)
+	}
+	log.Printf("parsing module %q", p)
```
tbpg: Not sure how much logging we want in the test logs.

testing/kokoro/version.go
grayside: We definitely want to log the error. I'll try to avoid the double return value once the final design is clarified.

testing/kokoro/version.go
grayside: Not exactly, right now it's "run this binary against every module you want to test to see if it should be tested". This means one run per module.

As an alternative, we could say "here's a list of modules, give me the list we should test". This means one two runs per test pipeline and moving more code into go.

testing/kokoro/system_tests.sh
grayside: Ah, and those tests traverse the entire repo so only need to be run from root.

testing/kokoro/version.go
grayside: Decision: Polish up the current approach in this PR. Follow-up to consolidate on more go code with fewer go invocations.

grayside: Ack

testing/kokoro/version.go
grayside: Ack

testing/kokoro/version.go
grayside: Ack

testing/kokoro/version.go
grayside: Maybe we could have a metadata.go file at the repo root which defines some constants like this? I'd rather not have testing code dependent on the CI system config.

testing/kokoro/version.go
grayside: Done.

grayside:
Currently reporting "version.NewVersion: Malformed version" errors. I think this needs a test case.

tbpg:
What's the next step for this? Hoping to clean up the open PRs in the repo.

grayside:
Starting a design/approach discussion outside this PR. May resume or drop this specific branch depending on outcome.

tbpg:
Superseded by #1552.

PR #1229, opened 2020-02-13
speech: add speech adaptation sample

sirtorry:


sirtorry:
i have no idea what i'm doing and am rather skeptical tests are passing.

speech/snippets/README.md
```diff
@@ -10,6 +10,12 @@ For more details, see the [Getting Punctuation](https://cloud.google.com/speech-
 
 [Go Code](auto_punctuation.go)
 
+### Speech Adaptation
+
+For more details see the [Using Speech Adaptation](https://cloud.google.com/speech-to-text/docs/context-strength#speech-context-strength-python) tutorial in the docs.
```
shollyman: drop the python anchor from the link?

speech/snippets/context_classes.go
```diff
@@ -0,0 +1,72 @@
...
+// contextClasses provides "hints" to the speech recognizer
+// to favour specific classes of words in the results
+func contextClasses(w io.Writer, gcsURI string) error {
+	ctx := context.Background()
+
+	client, err := speech.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewClient: %v", err)
+	}
+
```
shollyman: Add a `defer client.Close()` after the error check block to release client resources after you're done with them.

speech/snippets/context_classes.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+)
+
+// [START speech_context_classes]
+
+// contextClasses provides "hints" to the speech recognizer
+// to favour specific classes of words in the results
+func contextClasses(w io.Writer, gcsURI string) error {
```
shollyman: context is a bit overloaded in go land.  Maybe rename this (and update references) to be clearer that you're talking about speech adaptation or a speech context?

speech/snippets/context_classes.go
```diff
@@ -0,0 +1,72 @@
...
+	client, err := speech.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewClient: %v", err)
+	}
+
+	// SpeechContext: to configure your speech_context see:
+	// https://cloud.google.com/speech-to-text/docs/reference/rpc/google.cloud.speech.v1#speechcontext
+	// Full list of supported phrases (class tokens) here:
+	// https://cloud.google.com/speech-to-text/docs/class-tokens
+	SpeechContext := &speechpb.SpeechContext{Phrases: []string{"$TIME"}}
```
shollyman: Perhaps use this comment block to explain you're demonstrating use of the "$TIME" class to favor time of day detections, as it wasn't clear on first readthru that there wasn't some kind of indirect variable substitution.

speech/snippets/context_classes_test.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2019 Google LLC
```
shollyman: 2020

speech/snippets/context_classes.go
```diff
@@ -0,0 +1,72 @@
+// Copyright 2019 Google LLC
```
shollyman: 2020

speech/snippets/context_classes_test.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func TestContextClasses(t *testing.T) {
+	testutil.SystemTest(t)
+
+	var buf bytes.Buffer
+	err := contextClasses(&buf, "gs://cloud-samples-data/speech/commercial_mono.wav")
+
+	if got := buf.String(); !strings.Contains(got, "Alternative") {
+		t.Fatalf(`contextClassesgs://cloud-samples-data/speech/commercial_mono.wav) = %q; want "Alternative"`, got)
```
shollyman: Consider putting your desired string in a "want" variable so you can use that instead of the string literal multiple times.  Same with the gcs URI, so you can just do things like:

`t.Fatalf("contextClasses(%q): got %q, want %q", gcsURI, got, want)`

speech/snippets/context_classes.go
czahedi: Speech unfortunately using both words to describe the same thing: https://cloud.google.com/speech-to-text/docs/speech-adaptation and https://cloud.google.com/speech-to-text/docs/reference/rpc/google.cloud.speech.v1#speechcontext

And I believe they're repurposing the docs to use "Context" throughout, so IMO we should call it contexts throughout. 

I don't know anything about Go though so maybe we need to address this

speech/snippets/README.md
sirtorry: bwahaha oops

speech/snippets/context_classes.go
sirtorry: oof

speech/snippets/context_classes.go
sirtorry: happy to do anything. deferring decision to @czahedi

czahedi: We should keep this region tag as-is because all the docs will be changing to "context".

speech/snippets/context_classes.go
sirtorry: hmm should this be after the 2nd err != nil?

speech/snippets/context_classes.go
```diff
@@ -0,0 +1,74 @@
...
+// limitations under the License.
+
+// Package snippets contains speech examples.
+package snippets
+
+// [START speech_context_classes]
+
+import (
+	speech "cloud.google.com/go/speech/apiv1"
+	speechpb "google.golang.org/genproto/googleapis/cloud/speech/v1"
```
tbpg: Totally minor, but `goimports` usually puts the standard library imports in a group above the other imports.

speech/snippets/context_classes.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+)
+
+// contextClasses provides "hints" to the speech recognizer
+// to favour specific classes of words in the results
```
tbpg: Nit: period at the end.

speech/snippets/context_classes.go
tbpg: Create client -> check err -> defer close. So, right where this comment is.

speech/snippets/context_classes_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestContextClasses(t *testing.T) {
+	testutil.SystemTest(t)
+
+	var buf bytes.Buffer
+	gcsURI := "gs://cloud-samples-data/speech/commercial_mono.wav"
+	want := "Alternative"
+	err := contextClasses(&buf, gcsURI)
```
tbpg: Check error and fail the test if there is one.

speech/snippets/context_classes_test.go
```diff
@@ -27,9 +27,12 @@ func TestContextClasses(t *testing.T) {
 
 	var buf bytes.Buffer
 	gcsURI := "gs://cloud-samples-data/speech/commercial_mono.wav"
-	want := "Alternative"
 	err := contextClasses(&buf, gcsURI)
+	if err != nil {
+		t.Fatal(err)
```
tbpg:
```suggestion
		t.Fatalf("contextClasses got err: %v", err)
```

Issue #1157, opened 2020-01-13
storage: separate each sample into a separate file


crwilcox:
Each sample should be in a standalone file. This workitem is to track that work.

IlyaFaer:
Me and @AlisskaPie will do this

IlyaFaer:
We separated the samples (with our own separator-script, just some manual work left)

I suppose style and formatting rules in [CONTRIBUTING.md](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md) are actual, and we should follow them!? Including `Don't initialize one client for the entire set of samples and pass it as an argument. Each sample should initialize its own client/service`?

cc/ @tritone, @frankyn

tbpg:
Yes, please!

IlyaFaer:
@tbpg, I suppose for now all the samples are separated. Do we have anything more to do for this issue, or it can be closed?

P.S. we can probably take part in writing missing samples (which we have on other languages), but these are gonna be separate issues I assume.

tbpg:
Thank you! I'll close this. Please work with @tritone, @frankyn, and @JesseLovelace (sorry if I'm missing anyone) on any additional work.

PR #1249, opened 2020-02-26
refactor(storage): separate samples into standalone files (buckets part)

AlisskaPie:
Separate samples from buckets directory into standalone files
Towards #1157


IlyaFaer:
@googlebot I consent

AlisskaPie:
@tbpg, please review
There is no error check for `deleteBucket()` function in some places. It seems like it's used like a  
cleanup, but maybe it’s worth adding a check anyway?

storage/buckets/addBucketIamMember.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2020 Google LLC
```
tbpg: Please name the files with underscores instead of camelCase. Sorry, missed this in the other review.

storage/buckets/buckets_test.go
```diff
@@ -0,0 +1,251 @@
...
+	bucketName := tc.ProjectID + "-storage-buckets-tests"
+
+	buckets, err := listBuckets(ioutil.Discard, tc.ProjectID)
+	if err != nil {
+		t.Fatalf("listBuckets: %v", err)
+	}
+
+	var ok bool
+outer:
+	for attempt := 0; attempt < 5; attempt++ { // for eventual consistency
```
tbpg: Please use `testutil.Retry`. Be sure to fail with `r.Errorf`, rather than `t` directly.

storage/buckets/buckets_test.go
```diff
@@ -0,0 +1,251 @@
...
+		t.Errorf("got bucket list: %v; want %q in the list", buckets, bucketName)
+	}
+}
+
+func TestGetBucketMetadata(t *testing.T) {
+	tc := testutil.SystemTest(t)
+	bucketName := tc.ProjectID + "-storage-buckets-tests"
+
+	buf := new(bytes.Buffer)
+	if _, err := getBucketMetadata(buf, bucketName); err != nil {
```
tbpg: Does this bucket always exist? Same with the other tests.

storage/buckets/getBucketMetadata.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2020 Google LLC
```
tbpg: If we delete the old files, the existing docs that embed the old samples will not work anymore. So, we either need to freeze the versions in those docs or keep the old files around (being sure to avoid duplicated names or issues) until the docs are updated.

storage/buckets/buckets_test.go
AlisskaPie: Oh, right!
WDYT if I'll add TestMain with creating bucket before all tests run and delete it after tests passed? And, I suppose, for tests TestCreate/TestDelete we can create and delete different buckets

storage/buckets/get_bucket_metadata.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
```
tbpg: Don't update copyright years for existing files. New files should be 2020, though.

storage/buckets/main.go
```diff
@@ -1,500 +0,0 @@
-// Copyright 2019 Google LLC
```
tbpg: Looks like this file is still being deleted. Are we sure that won't break any docs?

storage/buckets/buckets_test.go
tbpg: I'm inclined to say create and delete a bucket for each test. We might want separate names for each test's bucket to reduce the likelihood of collision issues. @tritone and @frankyn, will you work with @AlisskaPie to make sure we do this in a way that (1) is not flaky and (2) does not use up all of our API quota?

tbpg:
I renamed the existing storage sample files in #1270 -- be sure to merge the latest changes if there is any overlap.

storage/buckets/get_bucket_metadata.go
AlisskaPie: Done

storage/buckets/main.go
AlisskaPie: Oops, deleted it by mistake

frankyn:
@JesseLovelace is on point from GCS DPE for sample gap related work (including splitting samples into separate files).

AlisskaPie:
Friendly ping =)

tbpg:
Looks good assuming tests are consistent (@JesseLovelace to make sure we're set up well). But, build failure:

```
can't load package: package github.com/GoogleCloudPlatform/golang-samples/storage/buckets: found packages buckets (add_bucket_iam_member.go) and main (main.go) in /home/runner/work/golang-samples/golang-samples/storage/buckets
```

(From the un-deleted file.)

storage/buckets/main.go
```diff
@@ -15,7 +15,8 @@
 // Sample buckets creates a bucket, lists buckets and deletes a bucket
 // using the Google Storage API. More documentation is available at
 // https://cloud.google.com/storage/docs/json_api/v1/.
-package main
+
+package buckets
```
tbpg:
```suggestion
// +build ignore

package buckets
```

This will make it so this file doesn't build and hopefully fix the compilation issue.

AlisskaPie: Thanks, it helps

frankyn:
Short heads up, @AlisskaPie, in PR #1251 @jkwlui updated and added 2 new samples that should be reflected in the split work. Want to make sure that's captured in this larger PR.

AlisskaPie:
@frankyn thank you for mentioning!
I'll add them

storage/buckets/buckets_test.go
```diff
@@ -102,6 +102,34 @@ func TestIAM(t *testing.T) {
...
+	if err := enableUniformBucketLevelAccess(ioutil.Discard, bucketName); err != nil {
+		t.Fatalf("enableUniformBucketLevelAccess:  %v", err)
+	}
+
+	role := "roles/storage.objectViewer"
+	member := "group:cloud-logs@google.com"
+	title := "title"
+	description := "description"
+	expression := "resource.name.startsWith(\"projects/_/buckets/bucket-name/objects/prefix-a-\")"
+	err := addBucketConditionalIamBinding(
```
tbpg: Please do this on a single line -- Go doesn't have a line length limit. You can also declare `err` inline in the if statement.

This may be an indicator there are just too many args for the sample.

storage/buckets/add_bucket_conditional_iam_binding.go
```diff
@@ -0,0 +1,73 @@
...
+	"io"
+	"time"
+
+	"cloud.google.com/go/storage"
+	iampb "google.golang.org/genproto/googleapis/iam/v1"
+	"google.golang.org/genproto/googleapis/type/expr"
+)
+
+// addBucketConditionalIamBinding adds bucket conditional IAM binding.
+func addBucketConditionalIamBinding(w io.Writer, bucketName, role, member, title, description, expression string) error {
```
tbpg: s/Iam/IAM/

AlisskaPie: Done

storage/buckets/buckets_test.go
AlisskaPie: Done

tbpg:
All tests pass. :tada: Merging! Thanks!

PR #1251, opened 2020-02-26
storage: use V3 IAM API

broady:


storage/buckets/main.go
```diff
@@ -208,15 +212,26 @@ func removeUser(c *storage.Client, bucketName string) error {
...
-	if err := bucket.IAM().SetPolicy(ctx, policy); err != nil {
+	for _, binding := range policy.Bindings {
+		if binding.Role == "roles/storage.objectViewer" {
+			// Filter out member.
+			i := 0
+			for _, member := range binding.Members {
+				if member != "group:cloud-logs@google.com" {
+					binding.Members[i] = member
+				}
+				binding.Members = binding.Members[:i]
```
tbpg: Thought maybe we could find the index of the member then slice it out (https://play.golang.org/p/grn8ypJz0cx), but I think this is better because it would handle multiple instances of the member in the slice? A little hard to wrap my mind around assigning to `binding.Members[:i]` on every loop iteration.

storage/buckets/main.go
```diff
@@ -226,6 +243,80 @@ func removeUser(c *storage.Client, bucketName string) error {
...
+		},
+	})
+
+	if err := bucket.IAM().V3().SetPolicy(ctx, policy); err != nil {
+		return err
+	}
+	// NOTE: It may be necessary to retry this operation if IAM policies are
+	// being modified concurrently. SetPolicy will return an error if the policy
+	// was modified since it was retrieved.
+	// [END add_bucket_conditional_iam_binding]
```
frankyn: Region tag: storage_add_bucket_conditional_iam_binding

storage/buckets/main.go
```diff
@@ -226,6 +243,80 @@ func removeUser(c *storage.Client, bucketName string) error {
...
+	return nil
+}
+
+func removeBucketConditionalIamBinding(c *storage.Client, bucketName string, role string, title string, description string, expression string) error {
+	// [START remove_bucket_conditional_iam_binding]
```
frankyn: Region tag: storage_remove_bucket_conditional_iam_binding

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/GoogleCloudPlatform/golang-samples
 go 1.11
 
 require (
-	cloud.google.com/go v0.55.0
+	cloud.google.com/go v0.53.1-0.20200226181226-323381ea063a
```
tbpg: Please keep this at v0.55.0.

storage/buckets/main.go
tbpg: Maybe lets just create a new filtered slice rather than editing this one in place? This logic is difficult to follow, especially for a sample. Copying might not be as efficient, but it's easier to read.

storage/buckets/main.go
```diff
@@ -208,15 +213,27 @@ func removeUser(c *storage.Client, bucketName string) error {
...
 	bucket := c.Bucket(bucketName)
-	policy, err := bucket.IAM().Policy(ctx)
+	policy, err := bucket.IAM().V3().Policy(ctx)
 	if err != nil {
 		return err
 	}
 	// Other valid prefixes are "serviceAccount:", "user:"
 	// See the documentation for more values.
 	// https://cloud.google.com/storage/docs/access-control/iam
-	policy.Remove("group:cloud-logs@google.com", "roles/storage.objectViewer")
```
jkwlui: We should probably make these arguments to the sample, WDYT? @frankyn

tbpg: We generally try to have as few arguments for sample functions. So, if it's not needed for testing, generally prefer to not make it an argument.

https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#function-arguments-for-snippets

frankyn: It's not necessary, but existing samples introduce role, member as parameters.

https://cloud.google.com/storage/docs/access-control/using-iam-permissions

If it's a hard requirement not to then that's okay, but would be more consistent for GCS samples.

jkwlui: 👍

tbpg: Consistent within the product sounds good to me. Nice when the samples on a single page are consistent. Whatever y'all think. :)

storage/buckets/main.go
```diff
@@ -208,15 +213,32 @@ func removeUser(c *storage.Client, bucketName string) error {
...
-	if err := bucket.IAM().SetPolicy(ctx, policy); err != nil {
+	for _, binding := range policy.Bindings {
+		// Only remove unconditional bindings matching role
+		if binding.Role == "roles/storage.objectViewer" && binding.Condition == nil {
+			// Filter out member.
+			i := -1
+			for j, member := range binding.Members {
+				if member == "group:cloud-logs@google.com" {
+					i = j
```
tbpg: Can there be more than one instance with this as the member? I think that was part of the more complex logic before.

jkwlui: There can only be one binding matching role with no conditions. We use the same logic in other samples as well: https://github.com/googleapis/nodejs-storage/blob/master/samples/removeBucketIamMember.js#L46

PR #1254, opened 2020-02-27
testing: output separate sponge_log.xml files

tbpg:


codyoss:
Looks like the build failed. But appears stuck here on Github.... probably all of the issues they were having this morning.

tbpg:
Started a new build and it passed (and GH is showing the result).

Issue #1247, opened 2020-02-25
testing: add GitHub Actions


tbpg:
Let's add GitHub Actions for a few tests:

- [ ] Code is formatted with ~~`gofmt`~~ [edit:] `goimports` (does formatting + imports).
- [ ] Code compiles.
- [ ] Proper license headers.

These all come up relatively frequently, especially with new contributors. An Action is very clear (and fast!) about what the issue was and we should be able to include the command to fix it. On the other hand, test logs are not clear for new devs when one of these issues pop up.

The license check can probably be general and apply to any Google repo. @JustinBeckwith has brought this up a few times.

The others would be helpful for our Go repos.

@bcoe, is this something you could help out with?

cc @codyoss, who may also be interested.

codyoss:
I have been meaning to take a look into this space more, would love to give it a shot.

tbpg:
Thank you!

PR #1267, opened 2020-03-02
all: support Github Actions for linting and tests

codyoss:
Fixes: #1247


tbpg:
Not sure tests will work -- need secrets. So, we should just do the "other stuff" with Actions.

.github/workflows/go.yaml
```diff
@@ -0,0 +1,45 @@
...
+      run: test -z "$($(go env GOPATH)/bin/goimports -l .)"
+        
+    - name: Fix goimports on fail
+      run: echo Run goimports -w .
+      if: failure()
+
+    - name: License check
+      run: |
+        test "$( \
+          grep -riL "Copyright .* Google LLC" --include=\*.{go,sh} . | \
```
grayside: This regex seems loose on the copyright year and overly strict on the file extensions.

.github/workflows/go.yaml
```diff
@@ -0,0 +1,45 @@
...
+        go-version: '^1.14'
+        
+    - name: Install goimports
+      run: go get golang.org/x/tools/cmd/goimports
+
+    - name: Check code
+      uses: actions/checkout@v2
+    
+    - name: goimports
+      run: test -z "$($(go env GOPATH)/bin/goimports -l .)"
```
grayside: Does this recurse into submodules? The existing CI implementation recently added lots of extra bash logic for this.

codyoss: Yes, I don't think goimports respects modules.

.github/workflows/go.yaml
codyoss: I can make the year part better. What would you recommend to include to files?

grayside: Looking at https://github.com/GoogleCloudPlatform/golang-samples/blob/master/license_test.go and contrasting to feedback I've received from @tbpg in the past, we should check:

* Dockerfile
* .html
* .go
* .sh

We also need exemptions for several files as noted in the linked test file for auto-generated code.

Instead of `.*` for the regex, can we do `\d\d\d\d` or equivalent?

.github/workflows/go.yaml
grayside: Confirmed. It also finds fault with generated `*pb.go` files, so we may need exemption logic.

tbpg: Perhaps:

```
find . -name go.mod -execdir $(go env GOPATH)/bin/goimports -l . \;
```

I confirmed this finds a formatting issue in a submodule. It also doesn't say anything about `*pb.go` files.

.github/workflows/go.yaml
tbpg: We should include all filetypes then add exceptions for ones we don't need (`go.sum`, `pb.go` (I think)).

`\d\d\d\d` SGTM.

.github/workflows/go.yaml
```diff
@@ -0,0 +1,45 @@
...
+        
+    - name: Fix goimports on fail
+      run: echo Run goimports -w .
+      if: failure()
+
+    - name: License check
+      run: |
+        test "$( \
```
tbpg: Use `test -z` instead `wc -l = 0`?

.idea/workspace.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tbpg: Delete.

.github/workflows/go.yaml
codyoss: I will propose a blacklist, but it is much longer than whitelist which is why I went that route

.idea/workspace.xml
codyoss: ok

.github/workflows/go.yaml
codyoss: I don't think we have to do this. goimports does not seem to respect modules from the test I ran.

.github/workflows/go.yaml
codyoss: With a blacklist the command would need to be something more like this `grep -riLI "Copyright \d\d\d\d Google LLC" --exclude=\*.{mod,sum,pb.go,md,dot,json,txt,ssml,cfg,csv,svg,LICENSE} .`

.github/workflows/go.yaml
tbpg: Ah. Even better!

.github/workflows/go.yaml
```diff
@@ -0,0 +1,42 @@
...
+      uses: actions/checkout@v2
+    
+    - name: goimports
+      run: test -z "$($(go env GOPATH)/bin/goimports -l .)"
+        
+    - name: Fix goimports on fail
+      run: echo Run goimports -w .
+      if: failure()
+
+    - name: License check
```
tbpg: Looks like this is failing for a bunch of files that I think are OK?

tbpg: This seems tricky to get right. We should be able to reuse an existing check. So, let's just remove the license check for now.

PR #1268, opened 2020-03-03
bigquery/bigquery_storage_quickstart: update to show v1 endpoint

shollyman:


bigquery/bigquery_storage_quickstart/main.go
```diff
@@ -72,31 +72,30 @@ func main() {
 	}
 
 	// This example uses baby name data from the public datasets.
-	readTable := &bqStoragepb.TableReference{
-		ProjectId: "bigquery-public-data",
-		DatasetId: "usa_names",
-		TableId:   "usa_1910_current",
-	}
+	readTable := fmt.Sprintf("projects/%s/datasets/%s/tables/%s",
```
tbpg: Inline values and don't use `fmt.Sprintf`?

shollyman: I didn't inline here as it's more common to have project/dataset/table in variables rather than pure literals as seen here.  Can drop it, but wanted a sample that's more quickly adapted for a new user.

tbpg: In that case, we should probably stick them in variables right above this. As-is, it looks a little odd. :)

shollyman: done

PR #1276, opened 2020-03-11
feat: adding dataproc instantiate inline workflows sample

bradmiro:
Adding a sample showing how to instantiate an inline workflow template with the client library.

### DIFF
--- a/dataproc/instantiate_inline_workflow_template.go
+++ b/dataproc/instantiate_inline_workflow_template.go
@@ -49,8 +49,8 @@ func instantiateInlineWorkflowTemplate(w io.Writer, projectID, region string) er
 				{
 					JobType: &dataprocpb.OrderedJob_HadoopJob{
 						HadoopJob: &dataprocpb.HadoopJob{
-							JarFileUris: []string{
-								"file:///usr/lib/hadoop-mapreduce/hadoop-mapreduce-examples.jar",
+							Driver: &dataprocpb.HadoopJob_MainJarFileUri{
+								MainJarFileUri: "file:///usr/lib/hadoop-mapreduce/hadoop-mapreduce-examples.jar",
 							},
 							Args: []string{
 								"teragen",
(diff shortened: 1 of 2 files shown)


dataproc/instantiate_inline_workflow_template.go
```diff
@@ -0,0 +1,115 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package dataproc shows how you can use the Cloud Dataproc Client library to manage
+// Cloud Dataproc clusters. In this example, we'll show how to instantiate an inline
+// workflow template.
```
tbpg: Only need the package comment in one file. So, you should be able to delete this.

dataproc/instantiate_inline_workflow_template_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	buf := new(bytes.Buffer)
+
+	if err := instantiateInlineWorkflowTemplate(buf, tc.ProjectID, region); err != nil {
+		t.Fatalf("instantiateInlineWorkflowTemplate got err: %v", err)
+	}
+
+	got := buf.String()
+	if want := fmt.Sprintf("successfully"); !strings.Contains(got, want) {
+		t.Fatalf("InstantiateInlineWorkflowTemplate: got %s, want %s", got, want)
```
tbpg: Super minor:

```suggestion
		t.Fatalf("instantiateInlineWorkflowTemplate got %q, want %q", got, want)
```

dataproc/instantiate_inline_workflow_template.go
bradmiro: If one is looking to learn more about the package, how does one differentiate if they should check this sample or the other one? Should these all be separate packages?

tbpg: https://pkg.go.dev/github.com/GoogleCloudPlatform/golang-samples/dataproc?tab=doc

If there are many files (like for samples), it's common to have a `doc.go` file that has this comment, the `package` line, and nothing else.

PR #1278, opened 2020-03-13
spanner: create admin client for emulator

hengfengli:


skuruppu:
@tbpg we would like your help in merging in this PR since we don't have the permissions to do so.

spanner/spanner_snippets/snippet.go
```diff
@@ -1820,11 +1822,27 @@ func queryWithHistory(ctx context.Context, w io.Writer, client *spanner.Client)
 }
 
 func createClients(ctx context.Context, db string) (*database.DatabaseAdminClient, *spanner.Client) {
-	adminClient, err := database.NewDatabaseAdminClient(ctx)
+	// [START spanner_create_admin_client_for_emulator]
```
tbpg: Would prefer if we migrating the Spanner samples to the new Go sample style (each sample in a different file and the region tags including the import block), but all good to do this for now.

PR #1281, opened 2020-03-16
testing: run goimports from root, run tests not in a submodule

tbpg:
Changes in a directory that wasn't a submodule led to not running any
tests at all, or running goimports on them.

### DIFF
--- a/testing/kokoro/system_tests.sh
+++ b/testing/kokoro/system_tests.sh
@@ -57,6 +57,7 @@ TARGET_DIRS=$(echo "$TARGET_DIRS" | xargs)
 # List all modules in changed directories.
 # If running on master will collect all modules in the repo, including the root module.
 GO_CHANGED_MODULES=$(find ${TARGET_DIRS:-.} -name go.mod)
+GO_CHANGED_MODULES=${$GO_CHANGED_MODULES:-./go.mod}
 # Exclude the root module if present.
 GO_CHANGED_SUBMODULES=${GO_CHANGED_MODULES#./go.mod}
 
@@ -76,6 +77,10 @@ fi
 set +x
 
 if [ $GOLANG_SAMPLES_GO_VET ]; then
+  echo "Running 'goimports compliance check'"
+  set -x
+  diff -u <(echo -n) <(goimports -d .)
+  set +x
   for i in $GO_CHANGED_MODULES; do
     mod="$(dirname $i)"
     pushd $mod > /dev/null;
@@ -86,11 +91,6 @@ if [ $GOLANG_SAMPLES_GO_VET ]; then
       go mod tidy;
       git diff go.mod | tee /dev/stderr | (! read)
       [ -f go.sum ] && git diff go.sum | tee /dev/stderr | (! read)
-      set +x
-
-      echo "Running 'goimports compliance check' in '$mod'..."
-      set -x
-      diff -u <(echo -n) <(goimports -d .)
       set +x
     popd > /dev/null;
   done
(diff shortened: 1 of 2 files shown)


testing/kokoro/system_tests.sh
```diff
@@ -196,8 +196,8 @@ elif [[ -z "${TARGET_DIRS// }" ]]; then
   GO_TEST_MODULES="./go.mod"
   echo "Only running root tests"
 else
-  GO_TEST_TARGET="./..."
-  GO_TEST_MODULES="$GO_CHANGED_SUBMODULES"
+  GO_TEST_TARGET=$TARGET_DIRS
+  GO_TEST_MODULES="./go.mod"
```
grayside: This seems like it would prevent tests running in modules.

tbpg: I _think_ (from running `go test ./storage/...`), this will run tests in different modules. I could be wrong? Or maybe it uses the wrong `go.mod`?

tbpg: Oh wait. I'm wrong. `storage` isn't a different module (it is for cloud.google.com/go).

```
go test ./run/image-processing/...
go: directory run/image-processing is outside main module
go: directory run/image-processing/imagemagick is outside main module
```

So, this needs to change.

tbpg: Updated this. I removed the indirection between defining the target/modules and running the test. That makes it a bit easier to run the necessary tests directly. To make sure we use consistent `go test` invocations (including XML translation), I put the `go test` invocation into a shared function.

tbpg:
All tests are passing, so it's ready for another look.

I'll keep an eye on it to make sure it works for normal PRs, too.

testing/kokoro/system_tests.sh
```diff
@@ -39,25 +39,18 @@ TIMEOUT=45m
 
 # Also see trampoline.sh - system_tests.sh is only run for PRs when there are
 # significant changes.
-SIGNIFICANT_CHANGES=$(git --no-pager diff --name-only HEAD..master | egrep -v '(\.md$|^\.github)' || true)
-# CHANGED_DIRS is the list of significant top-level directories that changed.
+SIGNIFICANT_CHANGES=$(git --no-pager diff --name-only master..HEAD | egrep -v '(\.md$|^\.github)' || true)
+# CHANGED_DIRS is the list of significant top-level directories that changed,
+# but weren't deleted by the current PR.
```
grayside: Nice refinement!

testing/kokoro/system_tests.sh
```diff
@@ -87,31 +84,15 @@ if [ $GOLANG_SAMPLES_GO_VET ]; then
...
-      diff -u <(echo -n) <(goimports -d .)
-      set +x
     popd > /dev/null;
   done
 
-  # Generate a list of all go files not inside a go submodule.
-  # go vet throws an error if you run it and no files are found to process.
-  # If find has anything go vet will be run.
-  # Risk: Any unexpected find output could falsely trigger go vet.
+  # Always run 'go vet' from the root, which does not look at sub-modules.
```
grayside: I don't recall "always running from the repo root" as an optional previously. Is this a testing process change to increase consistency? No objections.

testing/kokoro/system_tests.sh
tbpg: Thanks. :)

testing/kokoro/system_tests.sh
tbpg: Increasing consistency. If `go vet` is broken anywhere, I'm OK blocking all new PRs.

PR #1284, opened 2020-03-17
SecretManager bump version v1beta1->v1

gwhitehawk:
Client lib has been updated: https://pkg.go.dev/cloud.google.com/go@v0.55.0/secretmanager/apiv1


gwhitehawk:
Should I request a review from someone explicitly? Who would be a good person?

Issue #1275, opened 2020-03-11
testing: testutils does not accurately print out error 


bradmiro:
When running a test that lives in a product subdirectory, `testutils.go` will throw the following error if an invalid `.go` file exists at the top level of `$GOPATH/src/github.com/GoogleCloudPlatform/golang-samples/`:
 `testutil.go:62: could not find golang-samples directory`

codyoss:
Thanks for the report!

PR #1291, opened 2020-03-19
jobs: remove v2 samples

tbpg:


tbpg:
Test failure is only in Go 1.11 and caused by #1275.

PR #1293, opened 2020-03-19
internal/testutil: use os.Getwd to calculate Dir

### LINKED ISSUE #1275: testing: testutils does not accurately print out error

tbpg:
I also made a couple small style fixes.

I ran `go build ./...` and it worked. So, I don't think we're using this anywhere (was necessary for `GOPATH` before modules).

It's possible we'll want something like this for E2E AppEngine testing in the future, but it's causing issues right now, so let's just remove it.

Fixes #1275.

### DIFF
--- a/internal/testutil/testutil.go
+++ b/internal/testutil/testutil.go
@@ -17,26 +17,16 @@ package testutil
 
 import (
 	"errors"
-	"fmt"
 	"log"
 	"os"
-	"path/filepath"
-	"strings"
 	"testing"
-
-	"golang.org/x/tools/go/packages"
 )
 
-var noProjectID = errors.New("GOLANG_SAMPLES_PROJECT_ID not set")
+var errNoProjectID = errors.New("GOLANG_SAMPLES_PROJECT_ID not set")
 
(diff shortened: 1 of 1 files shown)


broady:
LGTM

tbpg:
If we need it, we can look back in history. Dead code and all. :)

tbpg:
Turns out, we did need `Dir` for AppEngine E2E tests. Not sure why `go build ./...` from the root passed... Maybe using a previous `testutil` module version?

I added it back using `os.Getwd`. Possibly not as correct as modules give no guarantee the directory path contains "golang-samples" but it works for now. :)

broady:
go build ./... doesn't capture carved out modules

tbpg:
`internal/testutil`, `internal/aeintegrate`, and `internal/e2e` are all in the same module (the root one).

PR #1295, opened 2020-03-19
dialogflow: remove unused code samples

Strykrol:


Strykrol:
Test files unaltered on purpose! 

Considering packages are grouped by Create/Delete/List functions, it looks like this might be easier to execute by fully deleting the packages (code & tests) versus altering individual functions and tests. I say this because 'List'-related functions are not part of this clean-up because they aren't wrapped in [START/END] blocks.

Happy to adjust by fully deleting directories/packages. Also prepared to alter '_test.go' files one-by-one!

PR #1305, opened 2020-03-23
testing: move set +e before running tests

tbpg:
This way, we run all tests and actually publish them to the Build Cop Bot.


broady:
Add a comment on the line that exit_code is used for still exiting when tests fail.

Or use `trap` to send to build cop bot?

LGTM otherwise

tbpg:
Thanks. Added more comments.

PR #1308, opened 2020-03-24
pubsub: update publish with batch settings sample to queue 10 messages 

### LINKED ISSUE #1294: Pub/Sub: update publish with batch settings sample to queue 10 messages
Please update [this Go sample](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/pubsub/topics/publish_settings.go) to match other publish with batch settings samples [here](https://cloud.google.com/pubsub/docs/publisher#batching), by publishing x number of messages instead of 1.

AlisskaPie:
Closes #1294

Also I can't find tests for functions `listSubscriptions` and `publishSingleGoroutine`


AlisskaPie:
@tbpg @anguillanneuf please review

pubsub/topics/publish_settings.go
```diff
@@ -39,14 +39,19 @@ func publishWithSettings(w io.Writer, projectID, topicID, msg string) error {
...
-	id, err := result.Get(ctx)
-	if err != nil {
-		return fmt.Errorf("Get: %v", err)
+	for i := 0; i < 10; i++ {
+		result := t.Publish(ctx, &pubsub.Message{
+			Data: []byte("Message " + strconv.Itoa(i)),
+		})
+		// Block until the result is returned and a server-generated
+		// ID is returned for the published message.
+		id, err := result.Get(ctx)
```
hongalex: This currently blocks after every single Publish. Instead, we should probably use the same method as https://github.com/GoogleCloudPlatform/golang-samples/blob/master/pubsub/topics/publish_scale.go

anguillanneuf: Does resolving the publish futures this way prevent batching? 
If so, the Python sample needs a rewrite as well.

hongalex: I believe so. Since we're guaranteed to get `id` populated after `Get`, the message has to have been published already.
I'm not sure about the Python library.

AlisskaPie: Done

anguillanneuf: Thank you! I will update the Python sample to use `.add_done_callback` today.

pubsub/topics/publish_settings.go
```diff
@@ -19,34 +19,56 @@ import (
...
-	// ID is returned for the published message.
-	id, err := result.Get(ctx)
-	if err != nil {
-		return fmt.Errorf("Get: %v", err)
+	for i := 0; i < 10; i++ {
+		result := t.Publish(ctx, &pubsub.Message{
+			Data: []byte("Message " + strconv.Itoa(i)),
+		})
+
+		wg.Add(1)
```
tbpg: What if we appended the results to a slice, then iterate through that slice calling `res.Get`? If we go that route, I don't think we need any goroutines, waitgroup, or atomic.

We can also append errors to a slice or only keep the last non-nil error.

pubsub/topics/publish_settings.go
```diff
@@ -35,9 +33,8 @@ func publishWithSettings(w io.Writer, projectID, topicID string) error {
 	if err != nil {
 		return fmt.Errorf("pubsub.NewClient: %v", err)
 	}
-
-	var wg sync.WaitGroup
-	var totalErrors uint64
+	var resultSlice []*pubsub.PublishResult
```
tbpg, line 36:
```suggestion
	var results []*pubsub.PublishResult
```

Generally, you don't need to include "Slice" in the name for slices.

pubsub/topics/publish_settings.go
```diff
@@ -35,9 +33,8 @@ func publishWithSettings(w io.Writer, projectID, topicID string) error {
...
+	var errors []error
```
tbpg, line 37: `errors` is the name of a standard library package (http://pkg.go.dev/errors). So, maybe `resultErrors`?

pubsub/topics/publish_settings.go
```diff
@@ -47,26 +44,20 @@ func publishWithSettings(w io.Writer, projectID, topicID string) error {
...
+	for i, res := range resultSlice {
+		id, err := res.Get(ctx)
+		if err != nil {
+			errors = append(errors, err)
+			fmt.Fprintf(w, "Failed to publish: %v", err)
+		}
+		fmt.Fprintf(w, "Published message %d; msg ID: %v\n", i, id)
+	}
+	if errors != nil {
+		return fmt.Errorf("%d of 10 messages did not publish successfully", len(errors))
```
tbpg, line 60: We should include an actual error in this result. Probably just the last one.

pubsub/topics/publish_settings.go
```diff
@@ -47,26 +44,20 @@ func publishWithSettings(w io.Writer, projectID, topicID string) error {
...
+		resultSlice = append(resultSlice, result)
 	}
-
-	wg.Wait()
-
-	if totalErrors > 0 {
-		return fmt.Errorf("%d of 10 messages did not publish successfully", totalErrors)
+	// The Get method blocks until a server-generated ID or
+	// an error is returned for the published message.
+	for i, res := range resultSlice {
```
tbpg, line 51: No more goroutines?

pubsub/topics/publish_settings.go
AlisskaPie: Ok, done

pubsub/topics/publish_settings.go
AlisskaPie: Right!
Done

pubsub/topics/publish_settings.go
AlisskaPie: Done

pubsub/topics/publish_settings.go
```diff
@@ -19,34 +19,47 @@ import (
...
+		})
+		results = append(results, result)
+	}
+	// The Get method blocks until a server-generated ID or
+	// an error is returned for the published message.
+	for i, res := range results {
+		id, err := res.Get(ctx)
+		if err != nil {
+			resultErrors = append(resultErrors, err)
+			fmt.Fprintf(w, "Failed to publish: %v", err)
```
tbpg, line 55: `continue` to avoid printing "Published message" when there was an error.

pubsub/topics/publish_settings.go
```diff
@@ -19,34 +19,47 @@ import (
...
+		}
+		fmt.Fprintf(w, "Published message %d; msg ID: %v\n", i, id)
+	}
+	if resultErrors != nil {
```
tbpg, line 59:
```suggestion
	if len(resultErrors) != 0 {
```

In Go, `len` and `range` treat `nil` slices and empty slices the same way -- length 0. https://play.golang.org/p/rMkImSrkmp9

So, it's generally good to check `len` instead of `nil` because it's less likely to lead to bugs later.

pubsub/topics/topics_test.go
```diff
@@ -126,6 +127,17 @@ func TestPublishThatScales(t *testing.T) {
 	}
 }
 
+func TestPublishWithSettings(t *testing.T) {
+	ctx := context.Background()
+	tc := testutil.SystemTest(t)
+	setup(t)
+	client := setup(t)
```
tbpg, line 134: Calling `setup` twice? I see the test below does this, too. I'm not sure why.

pubsub/topics/publish_settings.go
AlisskaPie: Done

pubsub/topics/publish_settings.go
AlisskaPie: Thank you for the explanation!
Done

PR #1319, opened 2020-03-27
Add OpenTelemetry code snippet

willgoldby:
Create OpenTelemetry directory and add OpenTelemetry code sample.

### DIFF
--- a/trace/trace_opentelemetry/main.go
+++ b/trace/trace_opentelemetry/main.go
@@ -1,0 +1,39 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START trace_opentelemetry]
+
+func main() {
+
+    // Create exporter
+    exporter, err := stackdriver.NewExporter(
+        stackdriver.WithProjectID("PROJECT-ID"),
+        )
+
+    // Create trace provider with the exporter
+    tp, err :=
+    sdktrace.NewProvider(
+    sdktrace.WithConfig(
+    // AlwaysSample() is used here to make sure traces are available for
+    // observation and analysis. In a production environment or high QPS
+    // setup please use ProbabilitySampler set at the desired probability.
+    // Example: DefaultSampler:sdktrace.ProbabilitySampler(0.0001)
+    sdktrace.Config{DefaultSampler:sdktrace.AlwaysSample()}),
+    sdktrace.WithSyncer(exporter))
+    if err != nil {
+            log.Fatal(err)
+    }
+	global.SetTraceProvider(tp)
+	
+// [END trace_opentelemetry]


trace/trace_opentelemetry/main.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2020 Google LLC
```
tbpg: File name comment: we can drop `trace_` in `trace_opentelemetry` because we are in the `trace` directory.

If OpenTelemetry samples will have more than just tracing, it may make more sense to create a top-level `opentelemetry` directory for those samples. We usually have one top-level directory per product.

tbpg: How is this sample intended to be used? That informs the type of sample we should write. Quickstarts are usually `package main` and have a `func main()`. That is the entrypoint to the program and what is run when you build+run the executable.

trace/trace_opentelemetry/main.go
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START trace_opentelemetry]
+
+func main() {
```
tbpg: Assuming this is a quickstart, you'll want something like the following to fix the compilation issue.

`trace_quickstart` is the name of the binary, which defaults to the name of the directory this is in. So, that will need to be updated based on the directory naming. See https://github.com/golang/go/wiki/CodeReviewComments#package-comments.

```suggestion
// The trace_opentelemetry command is an example of setting up OpenTelemetry to export traces to Stackdriver.
package main

func main() {
```

trace/trace_opentelemetry/main.go
```diff
@@ -0,0 +1,39 @@
...
+
+    // Create exporter
+    exporter, err := stackdriver.NewExporter(
+        stackdriver.WithProjectID("PROJECT-ID"),
+        )
```
tbpg: Please read the project ID on a separate line and assign it to a `projectID` variable.

```
```

Alternatively, you could make a `flag.String` so it can be passed as an argument. But, the env var is more common.

tbpg: Please do this on a single line. Go doesn't have a line length limit and recommends against breaking in the middle of function calls.

trace/trace_opentelemetry/main.go
```diff
@@ -0,0 +1,39 @@
...
+
+    // Create trace provider with the exporter
+    tp, err :=
```
tbpg: Do this on a single line. Can declare the `sdktrace.Config` separately so you can include the comment.

trace/trace_opentelemetry/main.go
tbpg: Please run `goimports -w .`. Your editor should be able to be configured to run this automatically. That will automatically format your code and add imports.

Note that if there are new OpenTelemetry packages that you're using, you'll want to run this from inside this directory:

```
go get go.opentelemetry.io/otel
go mod tidy
```

Here are instructions for installing `goimports` manually if your editor doesn't want to work: https://pkg.go.dev/golang.org/x/tools/cmd/goimports?tab=doc

trace/trace_opentelemetry/main.go
```diff
@@ -0,0 +1,39 @@
...
+    // observation and analysis. In a production environment or high QPS
+    // setup please use ProbabilitySampler set at the desired probability.
+    // Example: DefaultSampler:sdktrace.ProbabilitySampler(0.0001)
+    sdktrace.Config{DefaultSampler:sdktrace.AlwaysSample()}),
+    sdktrace.WithSyncer(exporter))
+    if err != nil {
+            log.Fatal(err)
+    }
+	global.SetTraceProvider(tp)
+
```
tbpg: Looks like you need to close `func main`.
```suggestion
}
```

trace/trace_opentelemetry/main.go
willgoldby: The sample shows how to install and import OpenTelemetry.

grayside: Wouldn't that be `monitoring/opentelemetry` and so on to keep the existing directory taxonomy?

tbpg:
Closing this as I think it was subsumed by the other PR.

PR #1324, opened 2020-04-01
profiler: add patch with optimization for shakesapp application

nolanmar511:


nolanmar511:
@aalexand  and @kalyanac -- PTAL

profiler/shakesapp/v1/shakesapp/shakesapp.proto
```diff
@@ -0,0 +1,35 @@
+// Copyright 2020 Google LLC
```
aalexand, line 1: Does this file need to be duplicated? I'd maybe only include changed files in the versioned subdirectories and ask users to copy it over the base version.

tbpg: Could there be a single module at `profiler/shakesapp` that holds this proto? Then I don't think it would need to be duplicated or copied, since it can be imported directly?

Also, I'm not sure about naming the directory `v1`. `vN` directories are used by modules to look for major version implementations, so that could get confusing for modules or humans if it's not actually intended for modules. I'm not sure.

nolanmar511: I've switched to using a patch file within a directory "optimizations".

Using a patch file seems simpler for users and as maintainable for us as copying only changed versions.

nolanmar511:
PTAL

nolanmar511:
Closing this PR -- our current HOW-TO just explains what modifications need to be changed.

PR #1326, opened 2020-04-04
Update README.md

LukaGiorgadze:


tbpg:
Thank you for writing the example and sending this PR. However, we can't accept this change because we try not to link to third parties unless it's very necessary. This may be a good topic for a http://cloud.google.com/community tutorial, if you're interested in writing one.

PR #1331, opened 2020-04-07
spanner: backupfunction sample

ashvds:
A sample function that can be run as a CLI with cron to create scheduled backups
Structured so that it could potentially be deployed to a cloud function.


hengfengli:
> Thank you for sending this! Are we planning a page on cloud.google.com that uses this sample? Also, we generally don't include CLIs that aren't part of a sample (like a quickstart) because we've had issues keeping them up to date and working.

Yes, we are planning to have a page where we can point customers to. 

We might need to think a better way how to include this as a sample. I found that we actually have cloud functions for spanner in: https://github.com/GoogleCloudPlatform/golang-samples/tree/master/functions/spanner. 

Should we put this over there? 

CC: @skuruppu

ashvds:
At this stage it is more of a CLI with the ability to make it a cloud
function.

tbpg:
I'm not too concerned with where in the repo we put it -- `/spanner` or `/functions/spanner`. If these are going in the Spanner docs, I'd prefer `/spanner`.

However, we should limit the sample to what is shown in the documentation. Users shouldn't be depending directly on this repo. When we've included more than necessary in the past (e.g. CLIs for samples), it has caused maintenance issues (READMEs need to be kept up to date, CLIs need to be tested, and so on).

hengfengli:
This PR is closed because we are moving this script to github.com/cloudspannerecosystem as an individual repo. It needs to go through a different review process. 

@ashvds Just let you know that the commit history can't be kept due to the movement. We will amend the first commit in the new repo to add you as an author.

ashvds:
Thanks Hengfeng, that is very much appreciated, though it looks like you
are doing most of the work!
I do not have the vanity to insist on it, so if it goes under your name
that is fine, otherwise I will thank you for the gift.

hengfengli:
@ashvds Just let you know that https://github.com/cloudspannerecosystem/scheduled-backups is live now. Thank you for starting the initial work.

skuruppu:
Woohoo, thanks @hengfengli and @ashvds for completing this work :)

ashvds:
Thanks @hengfengli , you certainly carried it across the line!

PR #1344, opened 2020-04-14
healthcare API: update all samples from v1beta1 to v1

noerog:
Update dates to 2020.

FHIR conditional methods have been deleted for v1 because they are only available in v1beta1.


noerog:
Looks like tests are failing for the reason you mentioned (v1 isn't included in the latest release).

noerog:
Looks like failures are in CI build and unrelated?

tbpg:
Looks like you need to run `go mod tidy`. Also, `TestFHIRStore` is failing?

noerog:
> Looks like you need to run `go mod tidy`. Also, `TestFHIRStore` is failing?

Ran go mod tidy. Yeah it looks like there was a regression in the FHIR store create operation. I fixed it.

noerog:
I'm not seeing any healthcare-related failures in the CI build. Any thoughts?

tbpg:
https://source.cloud.google.com/results/invocations/af75b8c4-1f0f-412c-9549-8ad333388605/targets/github%2Fgolang-samples/tests shows the files don't have the right license header.

The other runs couldn't find a free test project. So, the next run shouldn't be a problem.

noerog:
Maybe the regex doesn't like the comma in the first line? Trying again.

noerog:
> The extra year in the copyright is what is causing the issue.
> 
> You can run the license test locally by changing to the project root directory and running `go test`.
I just pushed a commit with the correct header. Can we at some point change the regex check to allow a range of years? That's what we're supposed to be doing.

healthcare/internal/fhir-resource-conditional-delete/fhir_resource_conditional_delete.go
```diff
@@ -1,72 +0,0 @@
-// Copyright 2019 Google LLC
```
tbpg, line 1: Looks like this file is still being referenced in docs. Are we sure we want to delete it? The docs won't be publishable as-is if we merge this PR right now.

noerog: Undeleted but left as v1beta1.

noerog:
Is this blocked on anything? Or can it be merged?

PR #1358, opened 2020-04-18
bigquery: address streaming error logic in storage sample

shollyman:
Noticed by an astute reader that we don't correctly break out
of the inner loop on error to restart a new streaming RPC call.


tbpg:
Looks like there was a flaky test failure:

```
integration_test.go:49: failed to create test dataset: googleapi: Error 400: IAM setPolicy failed for Dataset golang-samples-tests-8:snippet_table_tests_d99e3c846e0847b093f9c893e43d1cc3: There were concurrent policy changes. Please retry the whole read-modify-write with exponential backoff., invalid
```

Haven't seen this before.

shollyman:
Looking at the failure, seems like a flake from IAM propagating through BQ.  We don't retry 400 responses as they indicate we've done something wrong, and the sample under test doesn't involve policy mutations.

PR #1378, opened 2020-04-27
iap: add samples using new idtoken package

codyoss:


codyoss:
This PR will need to be updated next week when `google.golang.org/api v0.23.0` is released.

iap/request.go
```diff
@@ -0,0 +1,43 @@
...
+	"google.golang.org/api/idtoken"
+)
+
+// makeIAPRequest makes a request to an application protected by Identity-Aware
+// Proxy with the given iapClientID.
+func makeIAPRequest(request *http.Request, iapClientID string) error {
+	ctx := context.Background()
+	client, err := idtoken.NewClient(ctx, iapClientID)
+	if err != nil {
+		return err
```
tbpg, line 31: Add context to the error.

iap/request.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	response, err := client.Do(request)
+	if err != nil {
+		return err
+	}
+
+	// Use the response
+	_ = response
```
tbpg, line 39: Do something with the response.

iap/request.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+	"net/http"
+
+	"google.golang.org/api/idtoken"
+)
+
+// makeIAPRequest makes a request to an application protected by Identity-Aware
+// Proxy with the given iapClientID.
+func makeIAPRequest(request *http.Request, iapClientID string) error {
```
tbpg, line 27: I like this, since it could be copied and used as a helper. But, I wonder if we should show making a request to example.com or google.com, or something. Maybe the sample value is enough?

tbpg, line 27: Give sample values.

https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#function-arguments-for-snippets

codyoss: Ack.

iap/request.go
codyoss: Ack.

iap/request.go
codyoss: Ack.

iap/request.go
codyoss: I am leaning toward just providing sample values.

.gitignore
```diff
@@ -0,0 +1,18 @@
+# Created by .ignore support plugin (hsz.mobi)
```
tbpg, line 1: I think this needs a license header.

iap/request.go
```diff
@@ -0,0 +1,54 @@
...
+	"net/http"
+
+	"google.golang.org/api/idtoken"
+)
+
+// makeIAPRequest makes a request to an application protected by Identity-Aware
+// Proxy with the given iapClientID.
+func makeIAPRequest(w io.Writer, request *http.Request, iapClientID string) error {
+	// request, err := http.NewRequest("GET", "http://example.com", nil)
+	// iapClientID := "IAP_CLIENT_ID.apps.googleusercontent.com"
```
tbpg, line 33: Should we name this variable something else? Seems like this is a URL that uses the client ID.

iap/request.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return fmt.Errorf("idtoken.NewClient: %v", err)
+	}
+
+	response, err := client.Do(request)
+	if err != nil {
+		return fmt.Errorf("client.Do: %v", err)
+	}
+	defer response.Body.Close()
+	b, err := ioutil.ReadAll(response.Body)
```
tbpg, line 45: `io.Copy`?

iap/validate.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+	"io"
+
+	"google.golang.org/api/idtoken"
+)
+
+// validateJWTFromAppEngine validates a JWT found in the
+// "x-goog-iap-jwt-assertion" header.
+func validateJWTFromAppEngine(w io.Writer, iapJWT, projectNumber, projectID string) error {
+	// iapJWT := "base64.encoded.jwt"
```
tbpg, line 29: Make this a fake value? Could add a comment after the value saying where to find that value. Maybe we should add an HTTP response as an argument to show pulling it from the header?

.gitignore
codyoss: Ack.

iap/request.go
codyoss: I was going for consistency with other samples, but I think you are right. Will switch to audience as that is what it actually is.

iap/request.go
codyoss: Good idea, ack.

iap/validate.go
codyoss: Added a comment to the comment for now. If you don't think this is clear we can add a parameter though. Between this example and the Go doc above the function, I think the point is made.

iap/request.go
tbpg: Might be missing a push?

tbpg:
Looks like there is a merge conflict.

codyoss:
Yeah, I will rebase and fix next week once we have a really release of the apiaries.

codyoss:
@tbpg Is it okay if I allow .gitignore as a "good" file:
```
    badfiles_test.go:139: Likely bad file checked in: .gitignore. MIME type: {  }
```

Or would you prefer it not be checked in?

codyoss:
@tbpg If this still looks good to you, can you merge. I don't have "writes" on this repo 😉

PR #1380, opened 2020-04-27
functions: add sample calling a function with an ID token

codyoss:


functions/security/idtoken.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"google.golang.org/api/idtoken"
+)
+
+// callFunction makes a request to the provided functionURL with an
+// authenticated client.
+func callFunction(w io.Writer, functionURL string) error {
+	// functionURL := "https://REGION-PROJECT.cloudfunctions.net/RECEIVING_FUNCTION"
+	ctx := context.Background()
+	client, err := idtoken.NewClient(ctx, functionURL)
```
tbpg, line 33: I wonder if we should include a comment (also in the other PR) explaining that this returns an authenticated http.Client? If you aren't familiar, it's not clear what the type of `idtoken.NewClient` is.

functions/security/idtoken.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	"google.golang.org/api/idtoken"
+)
+
+// callFunction makes a request to the provided functionURL with an
+// authenticated client.
+func callFunction(w io.Writer, functionURL string) error {
```
tbpg, line 30: Please add a test for this.

https://github.com/GoogleCloudPlatform/golang-samples/blob/98312c1ac4900a276d3ff2af286f6f9e0fe5af7b/functions/http/cors_auth_test.go#L72-L80

Filed #1381 to track enabling those tests in Kokoro.

functions/security/idtoken.go
codyoss: Great idea.

functions/security/idtoken.go
codyoss: Followed a similar pattern as laid out in the example above.

functions/security/hello.go
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+package security
+
+import (
+	"fmt"
+	"net/http"
+)
+
+// HelloHTTP is an HTTP Cloud Function with a request parameter.
```
tbpg, line 22: Update comment.

codyoss: Ack.

codyoss:
cc @grayside

functions/security/hello.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+	"net/http"
+)
+
+// HelloHTTP is an HTTP Cloud Function.
+func HelloHTTP(w http.ResponseWriter, r *http.Request) {
+	fmt.Fprint(w, "Hello, World!")
+}
```
grayside, line 25: Seems unfortunate to add yet another hello world function to the repo. Is there a reasonable way to reuse one of the existing samples?

functions/security/idtoken.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	"google.golang.org/api/idtoken"
+)
+
+// callFunction makes a request to the provided functionURL with an
+// authenticated client.
+func callFunction(w io.Writer, functionURL string) error {
```
grayside, line 29: If this does not need to conform to the GCF function signature, would be great to update and use github.com/GoogleCloudPlatform/golang-samples/tree/master/run/authenticating for both Cloud Run and Cloud Functions.

functions/security/hello.go
codyoss: I will give the other functions a look over quick to see.

functions/security/idtoken.go
codyoss: I would be fine updating that sample and using it for both.

codyoss: To stay inline with other languages, I think I would like to keep this a separate sample if that is okay. I think it will provide a clearer picture to consumers of the docs.

grayside: We have other cases where we have the same sample used in multiple places in the docs, would like to understand the differences other than "typical practice" that degrade the clarity of information.

functions/security/README.md
```diff
@@ -0,0 +1,11 @@
+# Function Security Samples
+
+## System Tests
+
+1. `gcloud alpha functions deploy HelloHTTP --region=us-central1 --runtime=go111 --trigger-http --source=../helloworld`
```
tbpg, line 5:
```suggestion
1. `gcloud functions deploy HelloHTTP --region=us-central1 --runtime=go111 --trigger-http --source=../helloworld`
```

tbpg:
Does the block this PR? Looks like we already have Java, Python, and Node.js versions of this sample. We can merge and publish this one then come back and improve after.

grayside:
I was not aware those had gone through. Yes we can unblock this PR.

PR #1384, opened 2020-04-28
datastore: make dst slice for GetMulti

### LINKED ISSUE #1362: datastore: GetMulti example gives an error
When using this line to batch get entities from datastore it generates this error:

datastore: keys and dst slices have different length

This should be changed to
var tasks []*Task
tasks := make([]*Task, len(taskKeys))

https://github.com/GoogleCloudPlatform/golang-samples/blob/fe6277fb1aa032d9fd61206138fd0fc9ce8fb4fc/datastore/snippets/snippet_test.go#L265

tbpg:
I moved `taskKeys` into the snippet since it's referenced by the new `tasks` declaration.

Thanks to @rumsrami for finding & reporting this!

Fixes #1362.

### DIFF
--- a/datastore/snippets/snippet_test.go
+++ b/datastore/snippets/snippet_test.go
@@ -260,9 +260,9 @@ func SnippetClient_PutMulti() {
 func SnippetClient_GetMulti() {
 	ctx := context.Background()
 	client, _ := datastore.NewClient(ctx, "my-proj")
-	var taskKeys []*datastore.Key // Populated with incomplete keys.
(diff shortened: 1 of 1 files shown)


rumsrami:
Thanks for the prompt fix. Kindly please update the documentation as it still shows the old code.

https://cloud.google.com/datastore/docs/concepts/entities

tbpg:
Thanks for letting us know. It should be updated today or tomorrow.

PR #1385, opened 2020-04-28
functions: update slack sample to use environment variables

codyoss:
This is to bring sample in line with other languages.

Related: b/145664680


codyoss:
@labtopia will give the final all clear once necessary docs changes have taken place.

ace-n:
@labtopia LGTMed internally, merging. ✅

PR #1390, opened 2020-05-01
cloudsql: SQL Server sample

dmahugh:
This sample demonstrates connecting to a Cloud SQL for SQL Server instance from Golang. It's a port of the MySQL version previously published at https://github.com/GoogleCloudPlatform/golang-samples/tree/master/cloudsql/mysql/database-sql, with changes required by SQL Server syntax and data types, as well as changes for default port number and driver used.


cloudsql/sqlserver/database-sql/README.md
```diff
@@ -0,0 +1,120 @@
+# Connecting to Cloud SQL (SQL Server) from a Go web app
```
tbpg, line 1: We generally try to avoid having in depth READMEs in this repository. Instead, we prefer having the canonical documentation on cloud.google.com. We've had issues in the past with READMEs falling out of date and not working. So, we've been switching to the README only having a link to the corresponding cloud.google.com tutorial.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,302 @@
...
+	"math"
+	"net/http"
+	"os"
+	"strconv"
+
+	_ "github.com/denisenkom/go-mssqldb"
+)
+
+// db is the global database connection pool.
+var db *sql.DB
```
tbpg, line 35: Optional: other apps have followed a pattern of having a `type app struct { ... }` that contains all of the state that would otherwise be global. Then, you create the `app` in `func main` and the handlers can be defined as methods of that type to get access to the fields. That way, you avoid having any global state.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,302 @@
...
+			return votes, fmt.Errorf("Rows.Scan: %v", err)
+		}
+		votes = append(votes, nextVote)
+	}
+	return votes, nil
+}
+
+// currentTotals returns a templateData structure for populating the web page.
+func currentTotals() (templateData, error) {
+
```
tbpg, line 153: Nit: Extra newline.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return templateData{}, fmt.Errorf("recentVotes: %v", err)
+	}
+	return templateData{tabVotes, spaceVotes, voteDiffStr, latestVotesCast}, nil
+
+}
+
+// showTotals renders an HTML template showing the current vote totals.
+func showTotals(w http.ResponseWriter, r *http.Request) error {
+
```
tbpg, line 177: Newline?

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	var team string
+	if teamprop, ok := r.Form["team"]; ok {
+		team = teamprop[0]
+	} else {
+		return fmt.Errorf("team property missing from form submission")
+	}
+
+	// [START cloud_sql_sqlserver_databasesql_connection]
```
tbpg, line 202: Normally, samples include the required imports, as it can cause confusion for users when they aren't sure what to import.

https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#include-imports-and-flags-in-region-tags

For some tutorials it makes sense to have a separate import block and include it at the top of the page, not in every snippet.

We also try to include the function declaration. Otherwise, if users copy the code into a file (or look at it embedded on cloud.google.com), it doesn't compile.

Is there any way we can follow those patterns here?

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,302 @@
...
+	return dbPool, nil
+	// [END cloud_sql_sqlserver_databasesql_create_socket]
+}
+
+// initTcpConnectionPool initializes a TCP connection pool for a Cloud SQL
+// instance of SQL Server.
+func initTcpConnectionPool() (*sql.DB, error) {
+	// [START cloud_sql_sqlserver_databasesql_create_tcp]
+	var (
+		dbUser    = mustGetenv("DB_USER")
```
tbpg, line 259: Seems like it should be an error if one of these isn't set? Or is logging enough?

cloudsql/sqlserver/database-sql/Dockerfile
```diff
@@ -0,0 +1,49 @@
+# Copyright 2019 Google LLC
```
kurtisvg, line 1:
```suggestion
# Copyright 2020 Google LLC
```

Please update elsewhere as well.

cloudsql/sqlserver/database-sql/README.md
kurtisvg: This REAME is the canonical instructions. These samples deviate a bit from the traditional approach - the snippets being shown on c.g.c (in multiple places) are a small part of the overall app, but provide a link to the README for their context in a complete web application. 

We previously [discussed it for the MySQL sample as well](https://github.com/GoogleCloudPlatform/golang-samples/pull/1085#discussion_r347005538).

cloudsql/sqlserver/database-sql/cloudsql.go
kurtisvg: If you decide to do this, please update the MySQL so they don't drift too much.

cloudsql/sqlserver/database-sql/cloudsql.go
kurtisvg: Normally I would agree with the preference, but in the original design for these samples we opted not to include imports since they were part of larger app (which was important because we wanted to emphasize connection pooling in a "common use case"). Specifically, the reasoning was:
1. We might either include unnecessary or unused dependencies, which wouldn't be caught be testing
2. User's copy/pasting was less likely for these samples, since most connections to a database are done in the context of larger application

This is consistent with the [MySQL Go sample, as well as the samples for the other languages](https://cloud.google.com/sql/docs/mysql/manage-connections#go).

cloudsql/sqlserver/database-sql/cloudsql.go
kurtisvg: +1 - `log.Fatalf` is probably preferred.

cloudsql/sqlserver/database-sql/templates/index.html
```diff
@@ -0,0 +1,96 @@
+<!--
+Copyright 2018 Google LLC
```
kurtisvg, line 2:
```suggestion
Copyright 2020 Google LLC
```

cloudsql/sqlserver/database-sql/cloudsql.go
dmahugh: I've added the App struct and used its db property for global database connection, with the index handler as a method of the struct. @tbpg let me know if this looks OK for the general approach, and if so I'll update the MySQL sample to match.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,306 @@
...
+	TabsCount   uint
+	SpacesCount uint
+	VoteMargin  string
+	RecentVotes []vote
+}
+
+// App struct contains global state.
+type App struct {
+	// db is the global database connection pool.
+	db *sql.DB
```
tbpg, line 66: Can also store the parsed template here.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,306 @@
...
+func main() {
+
+	var err error
+
+	parsedTemplate, err = template.ParseFiles("templates/index.html")
+	if err != nil {
+		log.Fatalf("unable to parse template file: %s", err)
+	}
+
+	app := &App{}
```
tbpg, line 96: This will change to something like:
```suggestion
	app := &App{
		tmpl: tmpl,	
	}
```

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	// Drop the votes table if it already exists.
+	if _, err = app.db.Exec(`DROP TABLE IF EXISTS votes;`); err != nil {
+		log.Fatalf("DB.Exec: unable to drop votes table: %s", err)
+	}
+	// Create the votes table.
+	if _, err = app.db.Exec(`CREATE TABLE votes
+	( vote_id int IDENTITY(1,1) PRIMARY KEY, time_cast DATETIME NOT NULL,
+	candidate CHAR(6) NOT NULL );`); err != nil {
```
tbpg, line 121: Might split this up since it's a bit hard to follow the multi-line inline declaration.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,306 @@
...
+	return votes, nil
+}
+
+// currentTotals returns a templateData structure for populating the web page.
+func currentTotals(app *App) (templateData, error) {
+	// get total votes for each candidate
+	var tabVotes, spaceVotes uint
+	err := app.db.QueryRow(`SELECT count(vote_id) FROM votes WHERE candidate='TABS'`).Scan(&tabVotes)
+	if err != nil {
+		return templateData{}, fmt.Errorf("DB.QueryRow: %v", err)
```
tbpg, line 163: Can return a pointer to avoid having to be able to switch this to `nil, fmt.Errorf(...)`. The value shouldn't be used when the error is not nil.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	var voteDiffStr string = voteDiff(int(math.Abs(float64(tabVotes) - float64(spaceVotes)))).String()
+
+	latestVotesCast, err := recentVotes(app)
+	if err != nil {
+		return templateData{}, fmt.Errorf("recentVotes: %v", err)
+	}
+	return templateData{tabVotes, spaceVotes, voteDiffStr, latestVotesCast}, nil
+
```
tbpg, line 177: Delete extra newline.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,306 @@
...
+		return templateData{}, fmt.Errorf("DB.QueryRow: %v", err)
+	}
+
+	var voteDiffStr string = voteDiff(int(math.Abs(float64(tabVotes) - float64(spaceVotes)))).String()
+
+	latestVotesCast, err := recentVotes(app)
+	if err != nil {
+		return templateData{}, fmt.Errorf("recentVotes: %v", err)
+	}
+	return templateData{tabVotes, spaceVotes, voteDiffStr, latestVotesCast}, nil
```
tbpg, line 176: Generally a good practice to use named initialization for structs. That way, if the fields change at all, this declaration won't break. Would probably make sense to declare a separate variable at that point since it will be a bit longer.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+	}
+	return nil
+	// [END cloud_sql_sqlserver_databasesql_connection]
+}
+
+// mustGetEnv is a helper function for getting environment variables.
+// Displays a warning if the environment variable is not set.
+func mustGetenv(k string) string {
+	v := os.Getenv(k)
```
tbpg, line 223: Consider `os.LookupEnv` which returns a bool indicating if the variable is set, in case it's set to the empty string. If that doesn't matter here, feel free to ignore.

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,306 @@
...
+	if v == "" {
+		log.Fatalf("Warning: %s environment variable not set.\n", k)
+	}
+	return v
+}
+
+// initSocketConnectionPool initializes a Unix socket connection pool for
+// a Cloud SQL instance of SQL Server.
+func initSocketConnectionPool() (*sql.DB, error) {
+	// [START cloud_sql_sqlserver_databasesql_create_socket]
```
tbpg, line 233: Any particular reason to not include the function declaration?

cloudsql/sqlserver/database-sql/go.mod
```diff
@@ -0,0 +1,5 @@
+module sample-sqlserver
```
tbpg, line 1: Please use the full module path. Will be something like:

```suggestion
module github.com/GoogleCloudPlatform/golang-samples/cloudsql/sqlserver/database-sql
```

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,306 @@
...
+// templateData struct is used to pass data to the HTML template.
+type templateData struct {
+	TabsCount   uint
+	SpacesCount uint
+	VoteMargin  string
+	RecentVotes []vote
+}
+
+// App struct contains global state.
+type App struct {
```
tbpg, line 64:
```suggestion
// app struct contains application state.
type app struct {
```

No need to export. Also, this doesn't end up as a global -- you could theoretically have more than one app.

cloudsql/sqlserver/database-sql/cloudsql.go
dmahugh: For the env vars used in this sample, setting to an empty string would never be correct, so I'll leave this with os.Getenv(). But good to know, I hadn't seen lookupEnv.

cloudsql/sqlserver/database-sql/cloudsql.go
dmahugh: The function declaration is omitted for consistency with the snippets from other languages. For example, see this Python snippet: https://cloud.google.com/sql/docs/mysql/manage-connections#pools-python

cloudsql/sqlserver/database-sql/cloudsql.go
dmahugh: Thanks, addressed in this commit: https://github.com/GoogleCloudPlatform/golang-samples/pull/1390/commits/849f3f7f099bc0f6e7d8e0d0e25f74beeb0bcb33

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,311 @@
...
+	latestVotesCast, err := recentVotes(app)
+	if err != nil {
+		return nil, fmt.Errorf("recentVotes: %v", err)
+	}
+
+	var pageData templateData
+	pageData.TabsCount = tabVotes
+	pageData.SpacesCount = spaceVotes
+	pageData.VoteMargin = voteDiffStr
+	pageData.RecentVotes = latestVotesCast
```
tbpg, line 180:
```suggestion
	pageData := templateData{
		TabsCount: tabVotes,
		SpacesCount: spaceVotes,
		VoteMargin: voteDiffStr,
		RecentVotes: latestVotesCast,
	}
```

cloudsql/sqlserver/database-sql/cloudsql.go
```diff
@@ -0,0 +1,311 @@
...
+	if err := r.ParseForm(); err != nil {
+		return fmt.Errorf("Request.ParseForm: %v", err)
+	}
+
+	var team string
+	if teamprop, ok := r.Form["team"]; ok {
+		team = teamprop[0]
+	} else {
+		return fmt.Errorf("team property missing from form submission")
+	}
```
tbpg, line 209:
```suggestion
	team := r.FormValue("team")
	if team == "" {
		return fmt.Errorf("team property missing from form submission")
	}
```

Uses the `FormValue` helper and follows https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow.

dmahugh: Very cool, I didn't know about the FormValue helper!

PR #1392, opened 2020-05-04
storage: add PostPolicyV4 sample

tritone:
Add sample for GenerateSignedPostPolicyV4 method, which generates
a signed post policy using V4 signing and creates an upload form
based on the policy.

Depends on https://code-review.googlesource.com/c/gocloud/+/55330
which still needs to be merged and released.


storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,71 @@
...
+	"io"
+	"io/ioutil"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"golang.org/x/oauth2/google"
+)
+
+// generateSignedPostPolicyV4 generates a signed post policy.
+func generateSignedPostPolicyV4(w io.Writer, bucket, object, serviceAccountJSON string) (*storage.PostPolicyV4, error) {
```
odeke-em, line 29: This signature is a little confusing given that the body of this form manually creates a form then also returns the storage.PostPolicyV4. I'd think that perhaps we'd make it return:
a) Generated form string
b) URL string
c) error

or perhaps show the complete upload as we have in example_test.go

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,71 @@
...
+	fmt.Fprintf(w, "<form action='%v' method='POST' enctype='multipart/form-data'>\n", policy.URL)
+
+	// Include all fields returned in the HTML form as they're required.
+	for k, v := range policy.Fields {
+		fmt.Fprintf(w, "  <input name='%v' value='%v' type='hidden'/>\n", k, v)
+	}
+
+	fmt.Fprint(w, "  <input type='file' name='file'/><br />\n")
+	fmt.Fprint(w, "  <input type='submit' value='Upload File' name='submit'/><br />\n")
+	fmt.Fprint(w, "</form>")
```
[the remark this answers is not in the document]
frankyn: It's an example @odeke-em. It's intended only as an example of how to use the returned policy object.

storage/objects/generate_signed_post_policy_v4.go
frankyn: The return in this case is added for integration tests. @tritone and I discussed this further and it was a tradeoff to make it easier to verify in tests that the form values are valid.

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil, fmt.Errorf("google.JWTConfigFromJSON: %v", err)
+	}
+	metadata := map[string]string{
+		"x-goog-meta-test": "data",
+	}
+	opts := &storage.PostPolicyV4Options{
+		GoogleAccessID: conf.Email,
+		PrivateKey:     conf.PrivateKey,
+		Expires:        time.Now().Add(10 * time.Minute),
+		Fields:         &storage.PolicyV4Fields{Metadata: metadata},
```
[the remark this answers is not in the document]
tritone: Done

storage/objects/generate_signed_post_policy_v4.go
tritone: Just to add additional context, samples in other languages return the form itself but lack integration tests to actually verify that the post will work as expected against GCS. I discussed this with Frank and we think it's best to add that integration test in other languages as well. But that work can come later.

storage/objects/generate_signed_post_policy_v4.go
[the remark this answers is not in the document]
tritone: This matches what we have in other languages, but we should certainly make future changes across languages if this is a bad pattern to even give as an example.

storage/objects/objects_test.go
```diff
@@ -256,6 +259,81 @@ func TestV4SignedURL(t *testing.T) {
...
+	policy, err := generateSignedPostPolicyV4(putBuf, bucketName, objectName, serviceAccount)
+	if err != nil {
+		t.Fatalf("generateSignedPostPolicyV4: %v", err)
+	}
+	got := putBuf.String()
+	if want := "<form action="; !strings.HasPrefix(got, want) {
+		t.Errorf("got output %q, should start with %q", got, want)
+	}
+
+	// Now create a form using the returned post policy.
```
tbpg, line 286: I'm not sure I understand the rest of the test. Can you clarify or maybe add additional comments?

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"cloud.google.com/go/storage"
+	"golang.org/x/oauth2/google"
+)
+
+// generateSignedPostPolicyV4 generates a signed post policy.
+func generateSignedPostPolicyV4(w io.Writer, bucket, object, serviceAccountJSON string) (*storage.PostPolicyV4, error) {
+	// bucket := "bucket-name"
+	// object := "object-name"
+	// serviceAccountJSON := "service_account.json"
```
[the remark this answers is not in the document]
tritone: Done

storage/objects/objects_test.go
```diff
@@ -256,6 +259,81 @@ func TestV4SignedURL(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("storage.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	bucketName := tc.ProjectID + "-post-policy-bucket-name"
+	objectName := "foo.txt"
+	serviceAccount := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
+
+	cleanBucket(t, ctx, client, tc.ProjectID, bucketName)
```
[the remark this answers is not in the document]
tritone: Done. I can make a separate change to clean this up in the rest of this test.

storage/objects/objects_test.go
```diff
@@ -256,6 +259,81 @@ func TestV4SignedURL(t *testing.T) {
...
+	req.Header.Set("Content-Type", mw.FormDataContentType())
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		t.Fatalf("client.Do: %v", err)
+	}
+	if g, w := res.StatusCode, 204; g != w {
+		blob, _ := httputil.DumpResponse(res, true)
+		t.Errorf("status code in response mismatch: got %d want %d\nBody: %s", g, w, blob)
+	}
+	io.Copy(ioutil.Discard, res.Body)
```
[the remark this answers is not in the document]
tritone: Done.

storage/objects/objects_test.go
tritone: The signed post policy allows an unauthenticated client to make a POST using the form values from policy.Fields. The remainder of the test is verifying that this actually works as expected against the live service, using the code from the sample. I can add a bit more of that explanation as a comment.

storage/objects/generate_signed_post_policy_v4.go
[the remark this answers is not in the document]
tritone: I figured out how to use a template but the way I did it seems a bit janky and I'm not 100% sure if this was what you had in mind. Let me know what you think @odeke-em @tbpg

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,87 @@
...
+	// serviceAccountJSONPath := "service_account.json"
+	jsonKey, err := ioutil.ReadFile(serviceAccountJSONPath)
+	if err != nil {
+		return nil, fmt.Errorf("ioutil.ReadFile: %v", err)
+	}
+	conf, err := google.JWTConfigFromJSON(jsonKey)
+	if err != nil {
+		return nil, fmt.Errorf("google.JWTConfigFromJSON: %v", err)
+	}
+	metadata := map[string]string{
```
tbpg, line 42: Inline?

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+	}
+
+	policy, err := storage.GenerateSignedPostPolicyV4(bucket, object, opts)
+	if err != nil {
+		return nil, fmt.Errorf("storage.GenerateSignedPostPolicyV4: %v", err)
+	}
+
+	// Create an HTML form with the provided policy.
+	raw := fmt.Sprint("<form action='{{ index .url 0 }}' method='POST' enctype='multipart/form-data'>\n")
```
tbpg, line 60: Please make the template based on a static string that doesn't change. Something like this (completely untested, the variable substitutions won't work).

```go
var form = `<form action='{{ .URL }}' method='POST' enctype='multipart/form-data'>
	{{- range .Fields }}
	<input name='{{ .Name %v }}' value='{{ .Value %v }}' type='hidden'/>
	{{- end }}
	<input type='file' name='file'/><br />
	<input type='submit' value='Upload File' name='submit'/><br />
</form>`

var tmpl *html.Template

func init() {
	tmpl = template.Must(template.New("policyV4").Parse(form))
}
```

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Include all fields returned in the HTML form as they're required.
+	var data []string
```
tbpg, line 63: Based on the suggestion above, this type would be:

```
var data struct{
	URL string
	Fields []struct{
		Name string
		Value string
	}
}

File: storage/objects/generate_signed_post_policy_v4.go (hunk 8 above)
Comment (tritone) [reply storage/objects/generate_signed_post_policy_v4.go:42]: Done

File: storage/objects/generate_signed_post_policy_v4.go (hunk 9 above)
Comment (tritone) [reply storage/objects/generate_signed_post_policy_v4.go:60]: Ahh, much better. Will test this to figure out a working implementation.

File: storage/objects/generate_signed_post_policy_v4.go (hunk 10 above)
Comment (tritone) [reply storage/objects/generate_signed_post_policy_v4.go]: Done.

File: storage/objects/generate_signed_post_policy_v4.go [hunk 11]
```diff
@@ -0,0 +1,94 @@
...
+	var form = `<form action='{{ .URL }}' method='POST' enctype='multipart/form-data'>
+	{{- range .Fields }}
+  <input name='{{ .Name }}' value='{{ .Value }}' type='hidden'/>
+	{{- end }}
+	<input type='file' name='file'/><br />
+	<input type='submit' value='Upload File' name='submit'/><br />
+</form>`
+
+	var tmpl *template.Template
+	tmpl = template.Must(template.New("policyV4").Parse(form))
```
tbpg, line 69: Please do this in the global scope. Parsing a template is slow, so you don't want to do it on every request. While this isn't actually a request handler, we're more likely to nudge people to a better design if we don't do this inline.

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,94 @@
...
+	var data struct {
+		URL    string
+		Fields []struct {
+			Name  string
+			Value string
+		}
+	}
+	data.URL = policy.URL
+	for k, v := range policy.Fields {
+		data.Fields = append(data.Fields, struct {
```
tbpg, line 81: Maybe define a `field` type to avoid duplicating this?

tritone: Good idea, done.

storage/objects/generate_signed_post_policy_v4.go
odeke-em: Indeed as Tyler mentioned.
Perhaps also inclined in the global value.
var tmpl = template.Must(....Parse(`<form....
`)

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,94 @@
...
+	"io"
+	"io/ioutil"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"golang.org/x/oauth2/google"
+)
+
+// generateSignedPostPolicyV4 generates a signed post policy.
+func generateSignedPostPolicyV4(w io.Writer, bucket, object, serviceAccountJSONPath string) (*storage.PostPolicyV4, error) {
```
odeke-em, line 30: Minor nit.
Perhaps rename “w” to “wForm”

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,94 @@
...
+  <input name='{{ .Name }}' value='{{ .Value }}' type='hidden'/>
+	{{- end }}
+	<input type='file' name='file'/><br />
+	<input type='submit' value='Upload File' name='submit'/><br />
+</form>`
+
+	var tmpl *template.Template
+	tmpl = template.Must(template.New("policyV4").Parse(form))
+
+	// Include all fields returned in the HTML form as they're required.
```
broady, line 71: What does this comment mean?

Why not pass the policy in directly?

```go
tmplHTML = `
{{.URL}}
{{range .Fields}}
...
`

tmpl := template.Must(...)

tmpl.Execute(w, policy)
```

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,94 @@
...
+		},
+	}
+
+	policy, err := storage.GenerateSignedPostPolicyV4(bucket, object, opts)
+	if err != nil {
+		return nil, fmt.Errorf("storage.GenerateSignedPostPolicyV4: %v", err)
+	}
+
+	// Create an HTML form with the provided policy.
+	var form = `<form action='{{ .URL }}' method='POST' enctype='multipart/form-data'>
```
broady, line 60: prefer double quotes for attributes

https://google.github.io/styleguide/htmlcssguide.html#HTML_Quotation_Marks

storage/objects/generate_signed_post_policy_v4.go
tritone: I added this using an init function as @tbpg suggested previously. Let me know if this looks right to you.

storage/objects/generate_signed_post_policy_v4.go
tritone: All the other samples in this directory seem to use w for the writer that's passed in, so I think I'll leave this.

storage/objects/generate_signed_post_policy_v4.go
tritone: I just copied the canonical sample here, but I can fix this.

broady: Yes please, can you fix that across all languages' samples?

storage/objects/generate_signed_post_policy_v4.go
tritone: Was just my lack of experience with this templating language-- this is way nicer!! Done.

storage/objects/generate_signed_post_policy_v4.go
tritone: Will do!

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,83 @@
...
+// post policy.
+var form = `<form action="{{ .URL }}" method="POST" enctype="multipart/form-data">
+	{{- range $name, $value := .Fields }}
+  <input name="{{ $name }}" value="{{ $value }}" type="hidden"/>
+	{{- end }}
+	<input type="file" name="file"/><br />
+	<input type="submit" value="Upload File" name="submit"/><br />
+</form>`
+
+var tmpl *template.Template
```
broady, line 39:
```
var tmpl = template.Must(...)
```

no need for init

tritone: Done, thanks!

storage/objects/generate_signed_post_policy_v4.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"cloud.google.com/go/storage"
+	"golang.org/x/oauth2/google"
+)
+
+// form is a template for an HTML form that will use the data from the signed
+// post policy.
+var form = `<form action="{{ .URL }}" method="POST" enctype="multipart/form-data">
+	{{- range $name, $value := .Fields }}
+  <input name="{{ $name }}" value="{{ $value }}" type="hidden"/>
```
tbpg, line 33: Minor: spacing looks off here?

tritone: oof, tabs vs. spaces issue here. will fix.

Issue #1402, opened 2020-05-08
testing: audit sample test coverage


tbpg:
Indirect function calls aren't covered by testing/sampletests. How often does that happen? For example, some tests use string-> function indirection.

broady:
It's definitely used a bunch in vision tests I wrote.

```
$ git grep 'func(' | grep _test.go | grep -v testing.T | grep -v testutil.R| grep -v filepath.Walk | grep -v RunInTrans | grep -v '\:func'
bigtable/filters/filters_test.go:		filter func(io.Writer, string, string, string) error
bigtable/reads/reads_test.go:		filter func(io.Writer, string, string, string) error
container_registry/container_analysis/samples_test.go:		go func() {
dlp/snippets/jobs/jobs_test.go:	err = s.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
firestore/firestore_snippets/save_test.go:	must := func(f func(context.Context, *firestore.Client) error) {
functions/http/request_test.go:	s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
functions/spanner/spanner_test.go:	defer func() {
internal/testutil/retry_test.go:	Retry(t, 5, time.Millisecond, func(r *R) {
internal/testutil/retry_test.go:	Retry(t, 10, time.Millisecond, func(r *R) {
iot/mqttsnippets/subscribe_gateway_device_topic_test.go:		go func() {
logging/simplelog/simplelog_test.go:	defer func() {
logging/simplelog/simplelog_test.go:	defer func() {
logging/simplelog/simplelog_test.go:	client.OnError = func(err error) {
pubsub/subscriptions/subscription_test.go:	once.Do(func() {
pubsub/topics/topics_test.go:	once.Do(func() {
regiontag_test.go:	go func() {
regiontag_test.go:	go func() {
regiontag_test.go:	output := func(c <-chan RegionTag) {
regiontag_test.go:	go func() {
run/authentication/auth_test.go:	metadata := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
run/authentication/auth_test.go:	target := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
[...]

PR #1406, opened 2020-05-12
testing/sampletests: work with function references

tbpg:
Results:
Directory | Before | After | Diff
------------ | --------- | ------ | -----
root | `547/1161` | `613/1162` | `+66ish`
`vision` | `26/53` | `49/53` | `+23`
`bigtable` | `11/44` | `34/44` | `+23`

The root denominator went up by 1 because I added a new (tested) region tag to test against.

Spanner stays the same.

Updates #1402.

### DIFF
--- a/testing/sampletests/fakesamples/hello.go
+++ b/testing/sampletests/fakesamples/hello.go
@@ -44,3 +44,12 @@ func notTested() string {
 }
 
 // [END fakesamples_not_tested]
+
+// [START fakesamples_indirect_test]
+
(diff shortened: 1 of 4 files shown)


tbpg:
Here is what this looks like for Bigtable from the tests for this PR: https://source.cloud.google.com/results/invocations/2eca7f0c-9085-4037-8bb2-520e2e914ea3/targets/github%2Fgolang-samples/tests;group=garbagecollection;test=TestGarbageCollection;row=101

PR #1409, opened 2020-05-14
datastore: samples for Datastore Admin API

AlisskaPie:
The samples related to the generated [Datastore Admin API](https://github.com/q-logic/google-cloud-go/pull/1). 

This code doesn't have tests for import and export indexes. Is there a way to create/delete indexes with client methods? 

Also, for the same reason it is not clear how to test `indexGet()`. [The documentation](https://cloud.google.com/datastore/docs/reference/admin/rpc/google.datastore.admin.v1#createindexrequest) has `CreateIndex`, and it looks like what we want, but it seems like it's not implemented yet.
So I tested it with my credentials on existing index. Got the first one from the list of indexes. Just to make sure that everything works fine.


AlisskaPie:
Here's the building error: `module cloud.google.com/go/datastore@latest found (v1.1.0), but does not contain package cloud.google.com/go/datastore/admin/apiv1`
I had the same problem when writing samples. And the problem was locally resolved by adding `admin` folder with generated files to `pkg` folder with modules (to `datastore@v1.1.0` in that case).

AlisskaPie:
@BenWhitehead, @tbpg review it please

datastore/admin/datastore_admin_client_create.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// clientCreate creates a new Datastore admin client.
+func clientCreate(w io.Writer) (*admin.DatastoreAdminClient, error) {
+	ctx := context.Background()
+	client, err := admin.NewDatastoreAdminClient(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("admin.NewDatastoreAdminClient: %v", err)
+	}
+	// Close client when done with
```
tbpg, line 33:
```suggestion
	// Close client when done using it.
```

datastore/admin/datastore_admin_entities_export.go
```diff
@@ -0,0 +1,55 @@
...
+	req := &adminpb.ExportEntitiesRequest{
+		ProjectId:       projectID,
+		OutputUrlPrefix: outputURLPrefix,
+	}
+	op, err := client.ExportEntities(ctx, req)
+	if err != nil {
+		return fmt.Errorf("ExportEntities: %v", err)
+	}
+	_, err = op.Wait(ctx)
+	if err != nil {
```
tbpg, line 48:
```suggestion
	if _, err := op.Wait(ctx); err != nil {
```

datastore/admin/datastore_admin_entities_export.go
```diff
@@ -0,0 +1,55 @@
...
+		return fmt.Errorf("Wait: %v", err)
+	}
+	fmt.Fprintf(w, "Entities were exported\n")
```
tbpg, line 51:
```suggestion
	fmt.Fprintln(w, "Entities were exported")
```

datastore/admin/datastore_admin_entities_export.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	admin "cloud.google.com/go/datastore/admin/apiv1"
+	adminpb "google.golang.org/genproto/googleapis/datastore/admin/v1"
+)
+
+// entitiesExport exports a copy of all or a subset of entities from Google
+// Cloud Datastore to another storage system, such as Google Cloud Storage.
```
tbpg, line 28:
```suggestion
// entitiesExport exports a copy of all or a subset of entities from Datastore
// to another storage system, such as Cloud Storage.
```

(Follows brand guidelines.)

datastore/admin/datastore_admin_entities_import.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+
+	admin "cloud.google.com/go/datastore/admin/apiv1"
+	adminpb "google.golang.org/genproto/googleapis/datastore/admin/v1"
+)
+
+// entitiesImport imports entities into Google Cloud Datastore.
```
tbpg, line 27:
```suggestion
// entitiesImport imports entities into Datastore.
```

datastore/admin/datastore_admin_entities_import.go
```diff
@@ -0,0 +1,54 @@
...
+	req := &adminpb.ImportEntitiesRequest{
+		ProjectId: projectID,
+		InputUrl:  inputURL,
+	}
+	op, err := client.ImportEntities(ctx, req)
+	if err != nil {
+		return fmt.Errorf("ImportEntities: %v", err)
+	}
+	err = op.Wait(ctx)
+	if err != nil {
```
tbpg, line 47:
```suggestion
	if err = op.Wait(ctx); err != nil {
```

datastore/admin/datastore_admin_index_get.go
```diff
@@ -0,0 +1,50 @@
...
+	ctx := context.Background()
+	client, err := admin.NewDatastoreAdminClient(ctx)
+	if err != nil {
+		return fmt.Errorf("admin.NewDatastoreAdminClient: %v", err)
+	}
+	defer client.Close()
+
+	req := &adminpb.GetIndexRequest{
+		ProjectId: projectID,
+		IndexId:   indexID}
```
tbpg, line 40:
```suggestion
		IndexId:   indexID,
	}
```

This way, if you add another value at the end, you only modify that single line, rather than both. Also makes it clear when scanning the code where the end of the declaration is.

datastore/admin/datastore_admin_index_list.go
```diff
@@ -0,0 +1,58 @@
...
+	client, err := admin.NewDatastoreAdminClient(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("admin.NewDatastoreAdminClient: %v", err)
+	}
+	defer client.Close()
+
+	req := &adminpb.ListIndexesRequest{
+		ProjectId: projectID}
+	it := client.ListIndexes(ctx, req)
+	var indexArr []*adminpb.Index
```
tbpg, line 41:
```suggestion
	var indices []*adminpb.Index
```

Technically, this is a slice, not an array. A slice is dynamically sized and stores data in arrays under the hood. https://blog.golang.org/slices-intro

datastore/admin/datastore_admin_test.go
```diff
@@ -0,0 +1,48 @@
...
+	tc := testutil.SystemTest(t)
+	if _, err := indexList(ioutil.Discard, tc.ProjectID); err != nil {
+		t.Fatalf("indexList: %v", err)
+	}
+}
+
+func TestIndexGet(t *testing.T) {
+	tc := testutil.SystemTest(t)
+	// Get first index from the list of indexes.
+	indexes, _ := indexList(ioutil.Discard, tc.ProjectID)
```
tbpg, line 42:
```suggestion
	indices, _ := indexList(ioutil.Discard, tc.ProjectID)
```

datastore/admin/datastore_admin_test.go
```diff
@@ -0,0 +1,48 @@
...
+	indexID := indexes[0].IndexId
```
tbpg, line 43: Should check `len(indices) == 0` and `t.Skip` the test if so.

BenWhitehead:
The calls for create & delete aren't yet available in the googleapis protos so don't worry about them for the time being. All of these operations are long running, I'm not sure if we want them running every CI. @tbpg how have other APIs handled lots of LRO in their samples?

tbpg:
Did we get this client configured to auto-generate? This PR is blocked until then.

codyoss:
I will look into this today.

AlisskaPie:
The configuration was merged

AlisskaPie:
I've seen that new Datastore release was made recently, so I rerun checks again and they've passed :fireworks:

datastore/admin/datastore_admin_test.go
```diff
@@ -0,0 +1,50 @@
...
+	tc := testutil.SystemTest(t)
+	if _, err := indexList(ioutil.Discard, tc.ProjectID); err != nil {
+		t.Fatalf("indexList: %v", err)
+	}
+}
+
+func TestIndexGet(t *testing.T) {
+	tc := testutil.SystemTest(t)
+	// Get first index from the list of indexes.
+	indices, _ := indexList(ioutil.Discard, tc.ProjectID)
```
tbpg, line 42: Please check the error and fail the test if it fails.

datastore/admin/datastore_admin_test.go
```diff
@@ -0,0 +1,50 @@
...
+	indexID := indices[0].IndexId
+	if len(indices) == 0 {
```
tbpg, line 44: Check `len(indices)` before doing `indices[0].IndexId`. Otherwise, this will panic if indices is empty.

datastore/admin/datastore_admin_test.go
```diff
@@ -0,0 +1,50 @@
...
+		t.Skip("Skipping datastore test. At least one index should present in database.")
```
tbpg, line 45: Should we combine combine these tests into a single test? That way, we could create the index then get it in a single test.

datastore/admin/datastore_admin_test.go
AlisskaPie: Done

datastore/admin/datastore_admin_test.go
AlisskaPie: Done

datastore/admin/datastore_admin_test.go
AlisskaPie: Thanks, good idea

tbpg: Could we also merge `TestCreate` in here? Is the create operation guaranteed to be faster?

AlisskaPie: Practically, It takes the same time. Ok, I'll merge it too

tbpg: If we create an index above, then `indexList` returns an empty slice, the test should fail, right? Sorry I was unclear -- I meant is list guaranteed to contain the index that was just created? We might need a retry if not.

AlisskaPie: Sorry, I'm not sure, I'm understand you correctly.  We aren't creating index here, 'cause `Create` method for index not implemented yet. I could now only take an index from the project. In this case I need a retry too?

tbpg: Oh, sorry! Yeah, that makes sense. No, I don't think we need to add a retry. This looks good.

tbpg:
Thank you!

PR #1417, opened 2020-05-20
trace/trace_quickstart: use Google Cloud propagation in server

stevencl1013:
Enables tracing of incoming requests

### DIFF
--- a/trace/trace_quickstart/main.go
+++ b/trace/trace_quickstart/main.go
@@ -78,10 +78,10 @@ func main() {
 	}
 	log.Printf("Listening on port %s", port)
 
-    // Use this handler in server in order to instrument OpenCensus for 
-	// incoming requests
-	httpHandler := &ochttp.Handler {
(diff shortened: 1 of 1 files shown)


trace/trace_quickstart/main.go
```diff
@@ -77,7 +77,10 @@ func main() {
 		port = "8080"
 	}
 	log.Printf("Listening on port %s", port)
-	if err := http.ListenAndServe(":"+port, nil); err != nil {
+	if err := http.ListenAndServe(":"+port, &ochttp.Handler {
+        // Use Google Cloud propagation format.
+        Propagation: &propagation.HTTPFormat{},
+        }); err != nil {
```
tbpg, line 83: Would you mind declaring a separate variable for the `ochttp.Handler`? Some developers have had issues with the inline calling of `ListenAndServe`, so probably good to not make the line more complex.

Might also be good to add a comment saying what this results in.

PR #1424, opened 2020-05-28
storage: enable bucket object lifecycle management

jkwlui:


storage/buckets/buckets_test.go
```diff
@@ -272,6 +274,30 @@ func TestUniformBucketLevelAccess(t *testing.T) {
 	}
 }
 
+func TestLifecycleManagement(t *testing.T) {
+	tc := testutil.SystemTest(t)
+	bucketName := tc.ProjectID + "-storage-buckets-tests"
```
tbpg, line 279: Do we need to make sure this bucket exists? Can use `testutil.CleanBucket`.

storage/buckets/enable_bucket_lifecycle_management.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"cloud.google.com/go/storage"
+)
+
+// enableBucketLifecycleManagement adds a lifecycle delete rule with the
+// condition that the bucket is 100 days old.
```
tritone, line 28: I think it would be when the object is 100 days old, not the bucket, correct?

storage/buckets/enable_bucket_lifecycle_management.go
```diff
@@ -0,0 +1,64 @@
...
+				},
+			},
+		},
+	}
+
+	attrs, err := bucket.Update(ctx, bucketAttrsToUpdate)
+	if err != nil {
+		return fmt.Errorf("Bucket(%q).Update: %v", bucketName, err)
+	}
+	fmt.Fprintf(w, "Lifecycle management is enabled for bucket %v\n and the rules are:\n%+v", bucketName, attrs.Lifecycle.Rules)
```
tritone, line 59: Would it be clearer to iterate through rules here? I see that the C++ sample does this.

storage/buckets/buckets_test.go
```diff
@@ -272,6 +274,30 @@ func TestUniformBucketLevelAccess(t *testing.T) {
...
+
+	if err := enableBucketLifecycleManagement(ioutil.Discard, bucketName); err != nil {
+		t.Errorf("enableBucketLifecycleManagement: %v", err)
+	}
+
+	// verify lifecycle is set
+	m, err := getBucketMetadata(ioutil.Discard, bucketName)
```
tritone, line 286: Is there a potential delay before the rule propagates such that it will be visible? Just wondering if this could flake.

Also, seems preferable to call the library directly here rather than using another sample.

storage/buckets/enable_bucket_lifecycle_management.go
jkwlui: I think metadata updates are strongly consistent: https://cloud.google.com/storage/docs/consistency

storage/buckets/buckets_test.go
jkwlui: I think metadata updates are strongly consistent: https://cloud.google.com/storage/docs/consistency

tritone: I see the following:

```
For buckets, while metadata updates are strongly consistent for read-after-metadata-update operations, the resulting configuration changes may take time to propagate. For example, if you enable object versioning on a bucket, you should wait at least 30 seconds before deleting or overwriting objects.
```

...not sure which category this falls into? I guess the propagation from bucket configuration to individual objects is what takes a while.

jkwlui: I think you're right, the metadata update itself is consistent, but the effect on objects might be delayed.

storage/buckets/buckets_test.go
```diff
@@ -278,22 +279,29 @@ func TestLifecycleManagement(t *testing.T) {
...
 		t.Errorf("enableBucketLifecycleManagement: %v", err)
 	}
 
 	// verify lifecycle is set
-	m, err := getBucketMetadata(ioutil.Discard, bucketName)
+	client, err := storage.NewClient(ctx)
 	if err != nil {
-		t.Errorf("getBucketMetadata: %#v", err)
+		t.Errorf("storage.NewClient: %v", err)
+	}
```
tbpg, line 293:
```suggestion
	}
	defer client.Close()
```

storage/buckets/buckets_test.go
tritone: Okay, sounds good! Will assign to you if we see a flake anytime soon though. 😸

tbpg: Can wrap with `testutil.Retry` if we're concerned.

storage/buckets/buckets_test.go
```diff
@@ -278,22 +279,29 @@ func TestLifecycleManagement(t *testing.T) {
...
+		t.Errorf("Bucket(%q).Attrs: %v", bucketName, err)
 	}
 
 	want := storage.LifecycleRule{
 		Action:    storage.LifecycleAction{Type: "Delete"},
 		Condition: storage.LifecycleCondition{AgeInDays: 100},
 	}
 
-	if r := m.Lifecycle.Rules[0]; !reflect.DeepEqual(r, want) {
+	if r := attrs.Lifecycle.Rules[0]; !reflect.DeepEqual(r, want) {
```
tritone, line 304: If attr.Lifecycle.Rules is empty, this will panic. I would check the length first (we would want to know that there is exactly 1 rule) so we can cause a test failure instead.

storage/buckets/buckets_test.go
```diff
@@ -278,22 +279,29 @@ func TestLifecycleManagement(t *testing.T) {
...
 	// verify lifecycle is set
-	m, err := getBucketMetadata(ioutil.Discard, bucketName)
+	client, err := storage.NewClient(ctx)
 	if err != nil {
-		t.Errorf("getBucketMetadata: %#v", err)
+		t.Errorf("storage.NewClient: %v", err)
+	}
+	attrs, err := client.Bucket(bucketName).Attrs(ctx)
+	if err != nil {
+		t.Errorf("Bucket(%q).Attrs: %v", bucketName, err)
```
tritone, line 296: Probably want `t.Fatalf` here and above. No need to continue if there was an error retrieving the attrs.

storage/buckets/buckets_test.go
tritone: I think we can leave as is for now, but yes if we see a flake that's what we should do.

jkwlui:
Thank you @tritone! Bother you for another look?

storage/buckets/buckets_test.go
tritone: +1 to this suggestion

storage/buckets/buckets_test.go
```diff
@@ -272,6 +275,42 @@ func TestUniformBucketLevelAccess(t *testing.T) {
...
+		t.Fatalf("Bucket(%q).Attrs: %v", bucketName, err)
+	}
+
+	want := storage.LifecycleRule{
+		Action:    storage.LifecycleAction{Type: "Delete"},
+		Condition: storage.LifecycleCondition{AgeInDays: 100},
+	}
+
+	r := attrs.Lifecycle.Rules
+	if len(r) == 0 {
```
tritone, line 305: Would len > 1 also indicate unexpected behavior? In this case I would check that as well.

PR #1430, opened 2020-06-01
spanner: split code samples into separate files

hengfengli:


hengfengli:
@tbpg Thanks for your comments. I will fix the return errors and create non-main package. I am planning to do it all together. However, the doc comment for each function will be done in a separate PR.

hengfengli:
@tbpg I just realized that there might be some issue in cloud spanner with the new style guide: 

1) https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#initialize-clients-and-services-in-every-sample

This will be very expensive in Cloud Spanner client library. By default, client will create 100 sessions. This means when we close the client, we also need to close 100 sessions. This will slow down the entire test run. Is it possible to still just pass a pre-created client into each code sample function? So all code samples can share a single client. 

2) Currently, I divide it into different packages: backup, batch, client, dbadmin, dml, mutation, query, read. However, for each package, we need to create a db before the tests and drop it after the tests. We can solve this by putting everything into a single non-main package, but it is not neat. Also, with multiple packages, we have to use UUID for database names. Otherwise, tests in different packages will run in parallel and conflict with each other.

tbpg:
1. Sounds reasonable to me. We can add a comment showing how to initialize it inside the function (like the other sample values).
1. A single package sounds good. For some other APIs, we've prefixed filenames with what would have been the package (e.g. obj_create.go & obj_delete.go rather than create_obj.go and delete_obj.go) to make it a little easier to organize.

hengfengli:
@tbpg I have fixed most of the issues. For annotating errors, I would like to leave it as a separate PR because this PR is already very big. In terms of debugging, we have added the failure messages on the caller level (integration_test.go): https://github.com/GoogleCloudPlatform/golang-samples/pull/1430/files#diff-919a6f9b5fc4d1f2fed9b0853f6a1c0aR183. 

For [initialize-clients-and-services-in-every-sample](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#initialize-clients-and-services-in-every-sample), I follow the style guide because when I run the integration test, the runtime is okay and it doesn't add too much overhead.

skuruppu:
@tbpg we were talking to the docs team and we realized that if we submit this as is, we would break the samples in the Cloud docs since we need to modify the file path of each sample. Docs get auto-published so users will notice. So would you mind if we don't delete `snippets.go` in this PR and do is as a separate step?

That would allow us to have an intermediate time period where we can make the change on the docs side to point to the newly written samples and see it in staging to make sure that nothing is broken before updating the docs.

tbpg:
Keeping the existing samples is totally good with me. You can add `// +build ignore` near the top to avoid it being being built/tested.

The other option is freezing the version embedded in the docs. I'm good either way.

hengfengli:
@tbpg I have added `// +build ignore`. Are we able to merge this in?

PR #1439, opened 2020-06-02
bigquery: add reservation quickstart

shollyman:
The BigQuery reservation API has the possibility to generate nontrivial
costs when creating resources, so this initial quickstart demonstrates
simple enumerations and does no resource creations.

### DIFF
--- a/bigquery/bigquery_reservation_quickstart/main.go
+++ b/bigquery/bigquery_reservation_quickstart/main.go
@@ -1,0 +1,114 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START bigqueryreservation_quickstart]
+
+// The bigquery_reservation_quickstart application demonstrates usage of the
+// BigQuery reservation API by enumerating some of the resources that can be
+// associated with a cloud project.
+package main
+
+import (
+	"bytes"
+	"context"
+	"flag"
+	"fmt"
+	"log"
+
+	bqReservation "cloud.google.com/go/bigquery/reservation/apiv1"
+	"google.golang.org/api/iterator"
+	reservationpb "google.golang.org/genproto/googleapis/cloud/bigquery/reservation/v1"
+)
+
+// Define two command line flags for controlling the behavior of this quickstart.
+var (
+	projectID = flag.String("project_id", "",
+		"Cloud Project ID, used for session creation.")
+	location = flag.String("location", "US",
+		"BigQuery location used for interactions")
+)
+
+func main() {
+	flag.Parse()
+	ctx := context.Background()
+	bqResClient, err := bqReservation.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("NewClient: %v", err)
+	}
+	defer bqResClient.Close()
+
+	b, err := printCapacityCommitments(ctx, bqResClient)
+	if err != nil {
+		log.Fatalf("printCapacityCommitments: %v", err)
+	}
+	fmt.Println(string(b))
+
+	b, err = printReservations(ctx, bqResClient)
+	if err != nil {
+		log.Fatalf("printReservations: %v", err)
+	}
+	fmt.Println(string(b))
+}
+
+// printCapacityCommitments iterates through the capacity commitments and returns a byte buffer with details.
+func printCapacityCommitments(ctx context.Context, client *bqReservation.Client) ([]byte, error) {
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, "Capacity commitments in project %s in location %s:\n", *projectID, *location)
+
+	req := &reservationpb.ListCapacityCommitmentsRequest{
+		Parent: fmt.Sprintf("projects/%s/locations/%s", *projectID, *location),
+	}
+	totalCommitments := 0
+	it := client.ListCapacityCommitments(ctx, req)
+	for {
+		commitment, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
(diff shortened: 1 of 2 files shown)


bigquery/bigquery_reservation_quickstart/main.go
```diff
@@ -0,0 +1,114 @@
...
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		fmt.Fprintf(&buf, "\tCommitment %s in state %s\n", commitment.GetName(), commitment.GetState().String())
+		totalCommitments++
+	}
+	fmt.Fprintf(&buf, "\n%d commitments processed.\n", totalCommitments)
+	return buf.Bytes(), nil
```
tbpg, line 86: Can return a string directly with `buf.String()`. Could also use a `strings.Builder`.

bigquery/bigquery_reservation_quickstart/main.go
```diff
@@ -0,0 +1,114 @@
...
+
+	b, err = printReservations(ctx, bqResClient)
+	if err != nil {
+		log.Fatalf("printReservations: %v", err)
+	}
+	fmt.Println(string(b))
+}
+
+// printCapacityCommitments iterates through the capacity commitments and returns a byte buffer with details.
+func printCapacityCommitments(ctx context.Context, client *bqReservation.Client) ([]byte, error) {
```
tbpg, line 65: These don't print directly -- maybe there is a better name?

bigquery/bigquery_reservation_quickstart/main.go
```diff
@@ -0,0 +1,114 @@
...
+	"google.golang.org/api/iterator"
+	reservationpb "google.golang.org/genproto/googleapis/cloud/bigquery/reservation/v1"
+)
+
+// Define two command line flags for controlling the behavior of this quickstart.
+var (
+	projectID = flag.String("project_id", "",
+		"Cloud Project ID, used for session creation.")
+	location = flag.String("location", "US",
+		"BigQuery location used for interactions")
```
tbpg, line 39: Can we declare these inside `func main` and pass them as arguments to avoid having the global variables?

Also, can we declare on a single line to avoid a break in the middle of a function call? Or would that make the lines too long to embed?

bigquery/bigquery_reservation_quickstart/main.go
```diff
@@ -0,0 +1,114 @@
...
+package main
+
+import (
+	"bytes"
+	"context"
+	"flag"
+	"fmt"
+	"log"
+
+	bqReservation "cloud.google.com/go/bigquery/reservation/apiv1"
```
tbpg, line 29: Package names, even when renamed, should be all lower case. https://golang.org/doc/effective_go.html#package-names

bigquery/bigquery_reservation_quickstart/main.go
```diff
@@ -0,0 +1,114 @@
...
+// Define two command line flags for controlling the behavior of this quickstart.
+var (
+	projectID = flag.String("project_id", "",
+		"Cloud Project ID, used for session creation.")
+	location = flag.String("location", "US",
+		"BigQuery location used for interactions")
+)
+
+func main() {
+	flag.Parse()
```
tbpg, line 43: Check if `projectID` is empty?

bigquery/bigquery_reservation_quickstart/main.go
shollyman: Switched to string-based signature and usage.  Kept the existing bytes buffer rather than a string builder.

bigquery/bigquery_reservation_quickstart/main.go
shollyman: went with report as the verb.

bigquery/bigquery_reservation_quickstart/main.go
shollyman: done and done.

bigquery/bigquery_reservation_quickstart/main.go
shollyman: added a minimal validateFlags() func.

bigquery/bigquery_reservation_quickstart/main.go
shollyman: done.

bigquery/bigquery_reservation_quickstart/main.go
```diff
@@ -26,48 +26,57 @@ import (
...
+	s, err = reportReservations(ctx, bqResClient, *projectID, *location)
 	if err != nil {
 		log.Fatalf("printReservations: %v", err)
 	}
-	fmt.Println(string(b))
+	fmt.Println(s)
+}
+
+// validateFlags does some basic validation of the command line flags.
+func validateFlags() {
```
tbpg, line 64: I think this could be inlined? Up to you. Also, generally not a fan from calling `log.Fatal` in a function, which is another benefit of inlining (can still call `log.Fatal` directly).

bigquery/bigquery_reservation_quickstart/main.go
```diff
@@ -26,48 +26,57 @@ import (
 	"fmt"
 	"log"
 
-	bqReservation "cloud.google.com/go/bigquery/reservation/apiv1"
+	reservationapi "cloud.google.com/go/bigquery/reservation/apiv1"
```
tbpg, line 29: Why not just `reserveration`?

bigquery/bigquery_reservation_quickstart/main.go
```diff
@@ -26,48 +26,57 @@ import (
...
 )
 
 // Define two command line flags for controlling the behavior of this quickstart.
 var (
-	projectID = flag.String("project_id", "",
-		"Cloud Project ID, used for session creation.")
-	location = flag.String("location", "US",
-		"BigQuery location used for interactions")
+	projectID = flag.String("project_id", "", "Cloud Project ID, used for session creation.")
+	location  = flag.String("location", "US", "BigQuery location used for interactions")
```
tbpg, line 37: Would you mind moving these declarations inside `func main`?

bigquery/bigquery_reservation_quickstart/main.go
shollyman: inlined

bigquery/bigquery_reservation_quickstart/main.go
shollyman: done

bigquery/bigquery_reservation_quickstart/main.go
shollyman: done

PR #1442, opened 2020-06-02
all: remove legacy health check handlers

### LINKED ISSUE #977: appengine/flexible: use split health checks
From deploying helloworld:

> GAE Flex legacy health check is deprecated and its support will end soon. Please migrate to updated health checks. See https://cloud.google.com/appengine/docs/flexible/python/reference/app-yaml#updated_health_checks for instructions.

tbpg:
Legacy health checks migrated to split health checks, which are now the
default. Split health checks are not sent to apps by default, so these
handlers are not working as intended.

None of the app.yaml files have health checks configured, so I chose not
to add new split handlers & configuration. I'm open to adding them
if we decide we should.

Fixes #977.

### DIFF
--- a/appengine_flexible/helloworld/helloworld.go
+++ b/appengine_flexible/helloworld/helloworld.go
@@ -24,7 +24,6 @@ import (
 
 func main() {
 	http.HandleFunc("/", handle)
-	http.HandleFunc("/_ah/health", healthCheckHandler)
 	port := os.Getenv("PORT")
 	if port == "" {
 		port = "8080"
@@ -41,8 +40,4 @@ func handle(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	fmt.Fprint(w, "Hello world!")
-}
-
-func healthCheckHandler(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprint(w, "ok")
 }
--- a/appengine_flexible/websockets/main.go
+++ b/appengine_flexible/websockets/main.go
@@ -18,7 +18,6 @@
 package main
 
 import (
-	"fmt"
 	"log"
 	"net/http"
 	"os"
@@ -29,7 +28,6 @@ import (
 func main() {
 	http.Handle("/", http.FileServer(http.Dir("static")))
 	http.HandleFunc("/ws", socketHandler)
-	http.HandleFunc("/_ah/health", healthCheckHandler)
 
 	port := os.Getenv("PORT")
 	if port == "" {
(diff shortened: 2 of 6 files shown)


appengine_flexible/helloworld/helloworld.go
```diff
@@ -42,7 +41,3 @@ func handle(w http.ResponseWriter, r *http.Request) {
 	}
 	fmt.Fprint(w, "Hello world!")
 }
-
-func healthCheckHandler(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprint(w, "ok")
-}
```
grayside, line 48: Used in the quickstart, healthchecks can be skipped.

appengine_flexible/websockets/main.go
```diff
@@ -70,9 +68,4 @@ func socketHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-// healthCheckHandler is used by App Engine Flex to check instance health.
-func healthCheckHandler(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprint(w, "ok")
-}
-
```
grayside, line 77: Healthchecks are noted in the [docs for websockets](https://cloud.google.com/appengine/docs/flexible/go/using-websockets-and-session-affinity?hl=en) as part of how the platform relocates sessions to preserve connections using session affinity. Migrating to split healthchecks seems valuable.

getting-started/bookshelf/main.go
```diff
@@ -100,13 +100,6 @@ func (b *Bookshelf) registerHandlers() {
...
 
-	// Respond to App Engine and Compute Engine health checks.
-	// Indicate the server is healthy.
-	r.Methods("GET").Path("/_ah/health").HandlerFunc(
-		func(w http.ResponseWriter, r *http.Request) {
-			w.Write([]byte("ok"))
-		})
-
 	r.Methods("GET").Path("/logs").Handler(appHandler(b.sendLog))
 	r.Methods("GET").Path("/errors").Handler(appHandler(b.sendError))
```
grayside, line 105: I'm getting a bit turned around between quickstart, getting started, and bookshelf, but none of these entry-level tutorials seem to need healthchecks.

appengine_flexible/websockets/main.go
tbpg: That doc mentions "If the target instance fails health checks, App Engine moves the session to a healthy instance." That seems very reasonable.

https://cloud.google.com/appengine/docs/flexible/custom-runtimes/configuring-your-app-with-app-yaml#updated_health_checks mentions health checks are on by default and only need to be customized if you want to. So, I'm not sure it's valuable to add health checks here. Maybe we should update the websocket doc to link to the health check docs?

grayside: Agreed, let's remove this and cross-link the docs. I'll send a CL to the docs now.

PR #1443, opened 2020-06-03
functions: document source code location

hatstand:
I'm not sure this is the right place to fix this really but the new directory layout does not seem to be documented anywhere else.


grant:
Hi @hatstand, this change does not reflect the intention of the sample. I'm not sure what issue this PR solves.

The sample intends to read from `./`, the **current directory**, not `./serverless_function_source_code`.

```go
// ListFiles lists the files in the current directory.
func ListFiles(w http.ResponseWriter, r *http.Request) {
	files, err := ioutil.ReadDir("./")
```

Can you create a GitHub issue for the issue you're facing?

grant:
OK. Thanks for the context.

This is defined here:
https://github.com/GoogleCloudPlatform/buildpacks/blob/master/cmd/go/functions_framework/main.go

Can we add a comment like this to the PR? Otherwise it seems like a magic string.
```go
// Uses directory "serverless_function_source_code" as defined in the Go Functions Framework Buildpack
// See https://github.com/GoogleCloudPlatform/buildpacks/blob/master/cmd/go/functions_framework/main.go
```

hatstand:
Added the comment (made the link a permalink in case the code changes).

Thanks for finding the source of that magic string! That also looks go-specific so I don't think any fix is required for the other languages?

grant:
> Thanks for finding the source of that magic string! That also looks go-specific so I don't think any fix is required for the other languages?

I think we're good. I'll ask the eng team about it. This sample will be used in the docs too, so we got the docs covered. Thanks!

functions/tips/file_system.go
```diff
@@ -25,8 +25,11 @@ import (
 )
 
 // ListFiles lists the files in the current directory.
+// Uses directory "serverless_function_source_code" as defined in the Go
+// Functions Framework Buildpack.
+// See https://github.com/GoogleCloudPlatform/buildpacks/blob/56eaad4dfe6c7bd0ecc4a175de030d2cfab9ae1c/cmd/go/functions_framework/main.go#L38.
 func ListFiles(w http.ResponseWriter, r *http.Request) {
-	files, err := ioutil.ReadDir("./")
+	files, err := ioutil.ReadDir("./serverless_function_source_code")
```
tbpg, line 32: Would you mind creating `functions/tips/serverless_function_source_code` and putting a `README.md` file inside explaining it's used for this test?

hatstand: Done

tbpg:
Thank you!

PR #1445, opened 2020-06-03
functions/helloworld: expand logging sample, update README

### LINKED ISSUE #802: functions: create a sample of writing a log with a severity
The [docs](https://cloud.google.com/functions/docs/monitoring/logging#functions-log-retrieve-go) say that 
> Logs to stdout or stderr do not have an associated log level.

Then later say 
> Logs can also be written and retrieved through the Logging API

There is a sample of [retrieving](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/functions/log/list.go) logs, but I cannot find a way to write them with a severity level.

In Kubernetes deployed servers I can set the severity level by printing a JSON object like this.
```golang
fmt.Println({ "message" : "msg", "severity": "info" })
```
But in cloud functions even if the payload is JSON, stackdriver logs always consider it a `textPayload`.

It would be nice if those docs could be updated to include a sample of how to include log levels in Cloud Functions.

tbpg:
Fixes #802.

### DIFF
--- a/functions/helloworld/README.md
+++ b/functions/helloworld/README.md
@@ -8,13 +8,13 @@
 
 1. `gsutil mb gs://$BUCKET_NAME`
 
-1. `gcloud alpha functions deploy HelloGCS --runtime=go111 --entry-point=HelloGCS --trigger-resource=$BUCKET_NAME --trigger-event=providers/cloud.storage/eventTypes/object.change`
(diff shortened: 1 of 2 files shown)


grayside:
This looks like a good improvement, but I want to double-check what we're doing across languages, and whether it might make sense to keep this focused on log vs. fmt while structured logging reuses the Cloud Run sample.

tbpg:
Agreed the samples should be consistent. Thank you for pointing that out.

We can update this to be consistent in the future if we decide there is a better style for this sample.

For now, this is a good incremental improvement.

Issue #1455, opened 2020-06-04
Problem in handling messages from PubSub concurrently


mantishK:
Referring to the code below 
https://github.com/GoogleCloudPlatform/golang-samples/blob/a289aabbb0bb9c8f720e22f61b252874970e57f4/pubsub/subscriptions/pull_concurrency.go#L28

I recently came across a case where the receiver blocked indefinitely. I used the same example with a context timeout of 10s. What happened was, the `ctx.Done()` on line 58 was getting executed before the receiver stopped. So, any messages that came in later (and before receiver stopped) were not getting acknowledged and hence the receiver blocked.

Is it really necessary to handle `ctx.Done()` in the function again? Can't we write the anonymous function in the following way -

```
go func() {
    for {
        msg, ok: = < -cm
        if !ok {
            return
        }
        fmt.Fprintf(w, "Got message :%q\n", string(msg.Data))
        msg.Ack()
    }
}()
```

I am new to concurrency in Go, maybe I am missing something.

hongalex:
Yeah, I think your proposal makes sense and is a bit cleaner. If you want to create a PR to fix this, I'd be happy to review it. Otherwise, I'll get to this when I have some free time (since there are a few other samples that could use updating too).

mantishK:
Raised a PR [here](https://github.com/GoogleCloudPlatform/golang-samples/pull/1467)

PR #1467, opened 2020-06-07
Proper handling of concurrent pubsub pull in Go

mantishK:
"Context.Done" need not be handled separately again. Let the Receiver
close the channel to notify the anonymous function about the end
of pull.

Github Issue: [1455](https://github.com/GoogleCloudPlatform/golang-samples/issues/1455)

### DIFF
--- a/pubsub/subscriptions/pull_concurrency.go
+++ b/pubsub/subscriptions/pull_concurrency.go
@@ -51,13 +51,12 @@ func pullMsgsConcurrenyControl(w io.Writer, projectID, subID string) error {
 	// Handle individual messages in a goroutine.
 	go func() {
 		for {
-			select {
-			case msg := <-cm:
-				fmt.Fprintf(w, "Got message :%q\n", string(msg.Data))
-				msg.Ack()
-			case <-ctx.Done():
+			msg, ok := <-cm
+			if !ok {
 				return
 			}
+			fmt.Fprintf(w, "Got message :%q\n", string(msg.Data))
+			msg.Ack()
 		}
 	}()
 


mantishK:
@googlebot I signed it!

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -51,13 +51,12 @@ func pullMsgsConcurrenyControl(w io.Writer, projectID, subID string) error {
...
-				fmt.Fprintf(w, "Got message :%q\n", string(msg.Data))
-				msg.Ack()
-			case <-ctx.Done():
+			msg, ok := <-cm
+			if !ok {
 				return
 			}
+			fmt.Fprintf(w, "Got message :%q\n", string(msg.Data))
+			msg.Ack()
 		}
```
broady, line 60:
```suggestion
		for msg := range cm {
			fmt.Fprintf(w, "Got message :%q\n", string(msg.Data))
			msg.Ack()
		}
```

hongalex:
> As far as I can tell, this goroutine now never exits.

I think channel is closed no matter what (it happens outside the `err != nil` check. With that said, I think your solution is cleaner so let's go with that.

broady:
The goroutine still never exits. Here, maybe this snippet can help demonstrate:
https://play.golang.org/p/mOKbElsKRgV

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -50,14 +50,9 @@ func pullMsgsConcurrenyControl(w io.Writer, projectID, subID string) error {
 	cm := make(chan *pubsub.Message)
```
broady, line 50:
```suggestion
	cm := make(chan *pubsub.Message)
	defer close(cm)
```

mantishK:
@broady is right. If `sub.receive` returns an error, we are [returning](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/pubsub/subscriptions/pull_concurrency.go#L69) without closing the channel. `defer close(ch)` makes sense.

broady:
Could you remove the existing close() down below? Sorry, couldn't edit it with the suggestion tool.

mantishK:
Done

PR #1475, opened 2020-06-09
.github: add autoformat Action

tbpg:
* Update the CI action to also check `go mod tidy`.
* Update the CI action to also be triggered by the `trigger-ci`, `kokoro:run`, `kokoro:force-run` labels (open to something other than `trigger-ci`).
* Add an Action that runs `goimports` and `go mod tidy` when you add the `autoformat` label to a PR.
  * After the update, the build/lint Action doesn't run (by design of using `GITHUB_TOKEN`). So, we need `trigger-ci` or equivalent because there doesn't seem to be a nice way to trigger the Action otherwise.


.github/workflows/autoformat.yml
```diff
@@ -0,0 +1,46 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git config user.name yoshi-automation
+          git config user.email yoshi-automation@google.com
+          git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
+          git add .
+          if [ -n "$(git status --porcelain)" ]; then
+            git commit --author "yoshi-automation <yoshi-automation@google.com>" --message "Run goimports and go mod tidy"
+            git push
```
broady, line 34: What ref does this push to?

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork

If it's in someone's fork and they don't have "Allow edits from maintainers" on then this'll fail. Is that working as intended?

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

tbpg: Yes, that's intentional. I'm not sure what the failure will look like, exactly, though.

tbpg:
Please take a look. You can see a demo of this running in the PR history. Some of it was debugging issues, but I think those are resolved now.

PR #1490, opened 2020-06-11
spanner: add spanner_create_instance sample.

hengfengli:
Add a new code sample `spanner_create_instance` and be consistent with other languages.

### DIFF
--- a/spanner/spanner_snippets/spanner/integration_test.go
+++ b/spanner/spanner_snippets/spanner/integration_test.go
@@ -25,12 +25,16 @@ import (
 	"time"
 
 	database "cloud.google.com/go/spanner/admin/database/apiv1"
+	instance "cloud.google.com/go/spanner/admin/instance/apiv1"
 	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+	"github.com/google/uuid"
 	"google.golang.org/api/iterator"
 	adminpb "google.golang.org/genproto/googleapis/spanner/admin/database/v1"
+	instancepb "google.golang.org/genproto/googleapis/spanner/admin/instance/v1"
 )
 
 type sampleFunc func(w io.Writer, dbName string) error
+type instanceSampleFunc func(w io.Writer, projectID, instanceID string) error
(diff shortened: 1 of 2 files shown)


spanner/spanner_snippets/spanner/spanner_create_instance.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	defer instanceAdmin.Close()
+
+	op, err := instanceAdmin.CreateInstance(ctx, &instancepb.CreateInstanceRequest{
+		Parent:     fmt.Sprintf("projects/%s", projectID),
+		InstanceId: instanceID,
+		Instance: &instancepb.Instance{
+			Config:      "regional-us-central1",
+			DisplayName: instanceID,
+			NodeCount:   2,
```
tbpg, line 42: I wonder if we should do 1 node to keep the count down?

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -114,6 +118,14 @@ func TestSample(t *testing.T) {
 	defer cleanup()
 
 	var out string
+
+	instanceID := fmt.Sprintf("go-sample-test-%s", uuid.New().String()[:8])
```
tbpg, line 122: Could we create another test for this? Or do they all need to run sequentially in the same test? It might be nice to move away from a single giant test.

hengfengli: Thanks for suggestion. I've made it a separate test. I would like to make the task of refactoring the giant test into a separate PR.

tbpg: Sounds good! I don't think it's urgent -- just nice to have.

Looks like the test is failing right now:
```
integration_test.go:372: failed to create an instance: could not create instance projects/golang-samples-tests-7/instances/go-sample-test-afc7957b: rpc error: code = InvalidArgument desc = Invalid CreateInstance request.
```

hengfengli:
@tbpg It should be fine now.

PR #1491, opened 2020-06-11
.github: expand Blunderbuss config

tbpg:
Copied from https://github.com/googleapis/google-cloud-go/blob/master/.github/blunderbuss.yml.


tbpg:
Merging as the same configuration was fully reviewed for google-cloud-go. Happy to revise in future PRs.

PR #1496, opened 2020-06-11
bigtable: update region tags for reads and filters

billyjacobson:


bigtable/filters/filters.go
```diff
@@ -16,24 +16,7 @@
...
-// [START bigtable_filters_limit_value_regex]
-// [START bigtable_filters_limit_timestamp_range]
-// [START bigtable_filters_limit_block_all]
-// [START bigtable_filters_limit_pass_all]
-// [START bigtable_filters_modify_strip_value]
-// [START bigtable_filters_modify_apply_label]
-// [START bigtable_filters_composing_chain]
-// [START bigtable_filters_composing_interleave]
-// [START bigtable_filters_composing_condition]
+// [START bigtable_filters_imports]
```
tbpg, line 19: I'm still a bit concerned by these all being grouped together into a single snippet, rather than each snippet in it's own file. It seems unlikely to me that all of the snippets use the same exact set of imports. If one of the imports is extra or missing, the sample will fail to compile.

The helper functions would probably only be defined once and defined with the new tags you gave. So, I don't know how to guarantee the imports needed for the helpers are present in each file...

It's up to you what you would prefer to do.

billyjacobson: Hey Tyler, I definitely agree on the imports, but this is a quick fix to help get new documentation out. In the original version it would've made more sense, but this will have to do for the moment. I will probably go through sometime or work with a contractor to get all these samples split up into individual files.

tbpg: Sounds good.

PR #1499, opened 2020-06-12
monitoring: demonstrate creation of a POST uptime check to go along with the GET uptime check

syedashrafulla:
These samples are pulled into https://cloud.google.com/monitoring/uptime-checks#create.

### DIFF
--- a/monitoring/uptime/uptime.go
+++ b/monitoring/uptime/uptime.go
@@ -30,8 +30,8 @@ import (
 
 // [START monitoring_uptime_check_create]
 
-// create creates an example uptime check.
-func create(w io.Writer, projectID string) (*monitoringpb.UptimeCheckConfig, error) {
+// createGet creates an example uptime check on a GET request.
+func createGet(w io.Writer, projectID string) (*monitoringpb.UptimeCheckConfig, error) {
 	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


monitoring/uptime/uptime.go
```diff
@@ -62,9 +63,49 @@ func create(w io.Writer, projectID string) (*monitoringpb.UptimeCheckConfig, err
...
+					Type: "uptime_url",
+					Labels: map[string]string{
+						"host": "example.com",
+					},
+				},
+			},
+			CheckRequestType: &monitoringpb.UptimeCheckConfig_HttpCheck_{
+				HttpCheck: &monitoringpb.UptimeCheckConfig_HttpCheck{
+					RequestMethod: monitoringpb.UptimeCheckConfig_HttpCheck_POST,
+					ContentType:   monitoringpb.UptimeCheckConfig_HttpCheck_URL_ENCODED,
```
slz250, line 95: How about adding a body like `foo=bar`? The field is type bytes so it might be worth doing whatever conversion the language requires.

syedashrafulla: Agreed, considering it was non-trivial for me to figure out how to do it, especially for the URL-encoded content type.

PR #1505, opened 2020-06-16
asset: add samples for SearchAllResources and SearchAllIamPolicies

yuyifan-google:


asset/quickstart/search-all-iam-policies/main.go
```diff
@@ -0,0 +1,70 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START asset_quickstart_search_all_iam_policies]
+
+// Sample search-all-iam-policies search all iam policies within the given scope.
```
tbpg, line 17:
```suggestion
// Sample search-all-iam-policies searches all IAM policies within the given scope.
```

asset/quickstart/search-all-iam-policies/main.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// Command-line flags.
+var (
+	scope = flag.String("scope", "", "Scope of the search.")
+	query = flag.String("query", "", "Query statement.")
+)
+
+func main() {
+	flag.Parse()
```
tbpg, line 38:
```suggestion
func main() {
	scope := flag.String("scope", "", "Scope of the search.")
	query := flag.String("query", "", "Query statement.")
	flag.Parse()
```

asset/quickstart/search-all-iam-policies/main.go
```diff
@@ -0,0 +1,70 @@
...
+	ctx := context.Background()
+	client, err := asset.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("asset.NewClient: %v", err)
+	}
```
tbpg, line 43:
```suggestion
	}
	defer client.Close()
```

asset/quickstart/search-all-iam-policies/main.go
```diff
@@ -0,0 +1,70 @@
...
+
+	pageSize := 0
+	pageToken := ""
+
+	req := &assetpb.SearchAllIamPoliciesRequest{
+		Scope:     *scope,
+		Query:     *query,
+		PageSize:  int32(pageSize),
+		PageToken: pageToken,
```
tbpg, line 52: Do we need to include the page size and token if they are the default values (`0` and `""`)?

asset/quickstart/search-all-iam-policies/main.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	response := client.SearchAllIamPolicies(ctx, req)
```
tbpg, line 54: This is normally named `it`, and `Remaining` isn't checked. Might be good for consistency.

https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines

asset/quickstart/search-all-iam-policies/main.go
yuyifan-google: I was thinking listing all the fields in the API request might be easier for users to copy the sample code and make changes on their own.

tbpg: Maybe link to the iterator guidelines in a comment, if you want? We haven't done this for other iterators. So, I was confused when I first saw this -- I thought this was a different type of iterator or something. I think it's reasonable to not show all request options. If people want more, they can look at the definition of the request.

As is, I'm concerned it looks like these are mandatory or somehow used as part of the iteration process.

asset/quickstart/search-all-iam-policies/main.go
```diff
@@ -51,20 +48,13 @@ func main() {
...
-		if err != nil {
-			log.Fatal(err)
-		}
-		fmt.Println(policy)
-		if response.PageInfo().Remaining() == 0 {
-			break
-		}
+	it := client.SearchAllIamPolicies(ctx, req)
+	var policies []*assetpb.IamPolicySearchResult
+	_, err = iterator.NewPager(it, pageSize, "").NextPage(&policies)
```
tbpg, line 53: Any particular reason not to do the for loop? That's what we do for other iterator samples, so it would be nice to be consistent. Is it all guaranteed to fit in a single page?

yuyifan-google: Updated. And removed to unused default fields in the previous comment.
Sorry I've been confusing to you. I wanted to only print the first page instead of all results, but yeah I should be consistent with other samples;)

PR #1506, opened 2020-06-16
Clarify infinite retries example

ace-n:
Addresses b/158467672


functions/tips/infinite_retries/infinite_retries.go
```diff
@@ -54,5 +54,6 @@ func FiniteRetryPubSub(ctx context.Context, m PubSubMessage) error {
 
 // processTheMessage is a stand-in for the domain logic of your function.
 func processTheMessage(m PubSubMessage) error {
+	// Throw an error to retry this function execution
```
codyoss, line 57: Let's use a return instead of throw.

codyoss: nit: end with a period.

PR #1510, opened 2020-06-16
mod: update google.golang.org/api to v0.27.0

broady:


broady:
abandoning for now, will wait for next rev of gocloud first.

PR #1538, opened 2020-06-26
memorystore: add Dockerfile for Cloud Run deployment

ericschmidtatwork:
I'm working on a tutorial for connecting to Memorystore from Cloud Run via Serverless VPC Access. The existing sample app still works, we just need a Dockerfile to deploy it. I'm also adding back `go.mod` and `go.sum` in the proper location, and including a `.gcloudignore`.


memorystore/redis/cloud_run_deployment/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# Use the official Golang image to create a build artifact.
+# This is based on Debian and sets the GOPATH to /go.
+# https://hub.docker.com/_/golang
+FROM golang:1.13 as builder
```
tbpg, line 4: Should we use 1.13 or 1.14? If this is only for Cloud Run, it seems like we should use 1.14?

grayside: The current "canonical" Dockerfile base image is `golang:1.13-buster`, though #1541 proposed to update that to 1.14. I'm fine rolling that forward. More significantly, this Dockerfile is a copy of an earlier version of the template used in Cloud Run, it would be best to update this from [the Cloud Run Pub/Sub sample Dockerfile](https://github.com/renovate-bot/golang-samples/blob/renovate/docker-all/run/pubsub/Dockerfile)

memorystore/redis/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/GoogleCloudPlatform/golang-samples/memorystore/redis
+
+go 1.13
+
+require github.com/gomodule/redigo v2.0.0+incompatible
```
grayside, line 5: Looks like the redigo community has been spinning on what to do about this incompatible flag, possibly 2.0 is meant to be ignored.

Resolution in https://github.com/gomodule/redigo/issues/366 or https://github.com/gomodule/redigo/pull/440 are not clearly documented.

I wonder if it's since been fixed, and re-generating the go.mod and go.sum would help?

memorystore/redis/cloud_run_deployment/Dockerfile
ericschmidtatwork: I've updated the Dockerfile based on the Cloud Run Pub/Sub sample, thanks

memorystore/redis/go.mod
ericschmidtatwork: Looks like it's fixed - I regenerated `go.mod` and `go.sum`

PR #1539, opened 2020-06-28
Update and rename auth/overview/service_account.go to 认证/概观/service_a…

teejonei:
…ccount.go


tbpg:
Thank you for the PR. But, we keep filenames in English.

PR #1543, opened 2020-06-29
profiler: add WithTelemetryDisabled for OpenCensus users

ymotongpoo:
In Cloud Trace and Cloud Monitoring, we officially recommend to use OpenCensus as the instrumentation framework.

* https://cloud.google.com/trace/docs/setup/go
* https://cloud.google.com/monitoring/custom-metrics/open-census

However, when it comes to use them with Cloud Profiler agent, Trace UI is polluted with a lot of Spans generated from the Profiler agent, because it sends profile data to the backend via gRPC and the gRPC dialer comes with OpenCensus's ClientHandler.

* https://github.com/googleapis/google-api-go-client/blob/00ea0dae70a28d5e659ad288fdee5f7f5ab56a96/transport/grpc/dial.go#L173

In order to avoid the pollution on Cloud Trace UI and to reduce the billing cost to Cloud Trace, we should put some notes in the example for OpenCensus users to disable unnecessary telemetries from the Profiler agent.


profiler/snippets/snippets.go
```diff
@@ -28,7 +29,12 @@ func main() {
 		ServiceVersion: "1.0.0",
 		// ProjectID must be set if not running on GCP.
 		// ProjectID: "my-project",
-	}); err != nil {
+	},
+		// For OpenCensus user for Cloud Trace and Cloud Monitoring:
+		// This option prevents the Profiler agent to send Span data
+		// when the agent send profile data to the backend.
+		option.WithTelemetryDisabled(),
```
tbpg, line 36: It's a little hard for me to follow this `if` declaration with the `Start` call with the inline config and this option. Perhaps we should declare the config and opts separately?

```
cfg := profiler.Config{
	// ...
}
// For OpenCensus...
opts := []option.ClientOption{option.WithTelemetryDisabled()}
if err := profiler.Start(cfg, opts...); err != nil {
	// ...
}
```

profiler/snippets/snippets.go
```diff
@@ -28,7 +29,12 @@ func main() {
 		ServiceVersion: "1.0.0",
 		// ProjectID must be set if not running on GCP.
 		// ProjectID: "my-project",
-	}); err != nil {
+	},
+		// For OpenCensus user for Cloud Trace and Cloud Monitoring:
+		// This option prevents the Profiler agent to send Span data
+		// when the agent send profile data to the backend.
```
tbpg, line 35: A couple minor wording suggestions:
```suggestion
		// For OpenCensus users that use Cloud Trace and Cloud Monitoring:
		// cloud.google.com/go/profiler creates its own spans when making its
		// own requests to send profiling data to the server, which can be
		// noisy. This option prevents the profiler package from adding spans
		// for its own requests.
```

ymotongpoo: Thank you for the correction! Accepted the suggestion.

profiler/snippets/snippets.go
ymotongpoo: I agreed to that. I updated the lines for config and option to have separate declaration.

ymotongpoo:
@tbpg thank you for the review. i updated the sample to follow the suggestions and it should be fine now.

profiler/snippets/snippets.go
```diff
@@ -19,16 +19,26 @@ package main
...
 		// ProjectID: "my-project",
-	}); err != nil {
+	}
+
+	// For OpenCensus users that use Cloud Trace and Cloud Monitoring:
+	// cloud.google.com/go/profiler creates its own spans when making its
+	// own requests to send profiling data to the server, which can be
+	// noisy. This option prevents the profiler package from adding spans
+	// for its own requests.
+	opts := []option.ClientOption{option.WithTelemetryDisabled()}
```
kalyanac, line 38: This needs to be updated to use the flag, and the snippet should show how to turn on the telemetry since default is to turn it off.

profiler/snippets/snippets.go
```diff
@@ -19,16 +19,26 @@ package main
...
-	if err := profiler.Start(profiler.Config{
+	cfg := profiler.Config{
 		Service:        "myservice",
 		ServiceVersion: "1.0.0",
 		// ProjectID must be set if not running on GCP.
 		// ProjectID: "my-project",
-	}); err != nil {
+	}
+
+	// For OpenCensus users that use Cloud Trace and Cloud Monitoring:
```
kalyanac, line 33: We can trim the comment - To see Profiler agent spans in Cloud Trace and Cloud Monitoring, set EnableOCTelemetry to true.

ymotongpoo:
By [the PR to Cloud Profiler agent](https://github.com/googleapis/google-cloud-go/pull/2521), the default behavior will be changed and the agent won't send spans to Trace backend without enabling the flag (`EnableOCTelemetry`).

I'll change the example to follow the changes in Cloud Profiler agent as @kalyanac mentioned.

profiler/snippets/snippets.go
ymotongpoo: Fixed to use EnableOCTelemetry flag.

profiler/snippets/snippets.go
ymotongpoo: Done. Because I noticed that not only Cloud Trace but also other APMs, such as Jaeger and Zipkin will be affected by the change, I changed the wording a bit.

ymotongpoo:
@kalyanac Thank you for the review. I changed the example to use the new flag and removed obsolete options.

@tbpg Thank you for the initial review. If the new change looks fine, could you merge this example change once [the pull request in Profiler agent](https://github.com/googleapis/google-cloud-go/pull/2521) get merged?

tbpg:
We'll have to wait for a new release of the `profiler` package before we can merge this PR (which we'll need to update to use the new version).

ymotongpoo:
Sounds good. I look forward to both of them to be merged accordingly.

ymotongpoo:
@tbpg @broady Because the changes to the profiler agent is merged now, this PR is good to merge, I think.
https://github.com/googleapis/google-cloud-go/pull/2521

PR #1544, opened 2020-06-29
monitoring/irm: delete unused samples

tbpg:


codyoss:
I totally forgot that this should be done as well, thank you 😊. Another reason to get these samples integrated into google-cloud-go sooner!

tbpg:
Yes, that would be ideal for catching stuff like this. Luckily, this is not a "bad" breakage.

PR #1545, opened 2020-06-30
spanner: only keep necessary CLIs in snippet.go

hengfengli:
As discussed, we only keep the used commands in getting started guide.


hengfengli:
@tbpg Is it okay to keep the duplication in `snippet.go`. Because I have already modified the docs to point to the new style code samples: https://cloud.google.com/spanner/docs/getting-started/go. 

To remove the duplication, we have to

1) update the docs to point to `snippet.go` (revert the previous changes). 
2) remove the individual code sample files in `spanner` folder. 

But eventually, we have to delete the `snippet.go` and create individual code samples files. With the current duplication, once we don't support CLIs in getting started guide and we can just delete the `snippet.go` (we don't need to update the docs).

tbpg:
Hey @hengfengli, what's the next step for this PR?

hengfengli:
@tbpg Sorry that we kept missing the chat history. 

For this PR, I am still waiting for the completion of translated pages. I contacted with the docs team and they say that the translations to other languages usually take 2-4 weeks. So that's why we still can find references to snippet.go internally. I will make sure no more references to snippet.go internally.

In addition, I think that moving `snippet.go` into the folder will cause many changes in the docs (it is not worth to do so because these changes will be removed). I prefer to keep it where it is. If we move `snippet.go` into the folder, then we need to do: 

1) update all individual files to `main` package
2) update all commands to `go run . createdatabase ...`
3) update the text in the docs, including pointing to the right folder, and etc. This is not easy because all languages share the same markdown file, which needs to write something like `if it is golang, use this text`. 

Because eventually we will remove the CLIs from the getting started. Once the CLIs are removed, we can easily delete the `snippet.go` file.

tbpg:
@hengfengli, is this ready to go?

hengfengli:
I am keeping watching this all the time. However, there are still some translations of getting-started guide that are not re-generated, such Japanese and Germany, linking to this file. I will contact the docs team to see what we can do.

tbpg:
Sounds good. Thanks for the update, keeping an eye on it, and seeing what we can do.

hengfengli:
@tbpg The issue of translations has been resolved. This is ready to go.

tbpg:
Sounds good! Thanks!

PR #1546, opened 2020-06-30
testing, .github: remove autoformat, move go vet and tidy to Actions

tbpg:
Autoformat only works on branches of GoogleCloudPlatform/golang-samples, not forks. But, forks most commonly need this. My hope was for Renovate CLs, but those come from forks.

The interactions between labels, building CI, and autoformat are not correct right now. CI needs to run after autoformat does its thing, but it doesn't happen automatically. So, we need some sort of other trigger. I tried labels, but then it's weird because do you run CI when you add `autoformat`? No, that wouldn't make sense. So, then you need a `trigger-ci` type label or something?

Either way, it seems like labels aren't the answer for this and some sort of comment system would be better. I'm going to delete autoformat for now and we can come back to it later as desired.

I also added `go vet` to the `Build and lint` action and removed linting from system_test.sh.


testing/kokoro/system_tests.sh
```diff
@@ -65,52 +65,6 @@ elif echo "$SIGNIFICANT_CHANGES" | tr ' ' '\n' | grep "^go.mod$" || [[ $CHANGED_
...
-    pushd "$mod" > /dev/null;
-      # Fail if a dependency was added without the necessary go.mod/go.sum change
-      # being part of the commit.
-      echo "Running 'go.mod/go.sum sync check' in '$mod'..."
-      set -x
-      go mod tidy;
-      # shellcheck disable=SC2162
-      git diff go.mod | tee /dev/stderr | (! read)
-      # shellcheck disable=SC2162
-      [ -f go.sum ] && git diff go.sum | tee /dev/stderr | (! read)
```
broady, line 88: Go mod/sum stuff got dropped

testing/kokoro/system_tests.sh
```diff
@@ -65,52 +65,6 @@ elif echo "$SIGNIFICANT_CHANGES" | tr ' ' '\n' | grep "^go.mod$" || [[ $CHANGED_
...
-      set +x
-    popd > /dev/null;
-  done
-
-  # Always run 'go vet' from the root, which does not look at sub-modules.
-  set +x
-  echo "Running 'go vet' in golang-samples root..."
-  set -x
-  go vet ./...
```
broady, line 97: Does this need to be retained? Or does the new one also check root?

.github/workflows/go.yaml
```diff
@@ -51,12 +23,17 @@ jobs:
...
       run: |
         if [ -n "$(git status --porcelain)" ]; then
           git status # Show the files that failed to pass the check.
           exit 1
         fi
+    - name: go vet
+      # Use find to build all modules. '-execdir ... ;' doesn't set an exit code
+      # based on command results. So, create a file if a build fails and check
+      # if the file exists to set the right exit code.
+      run: find . -name go.mod -execdir sh -c 'go vet ./... || touch vet_failed.txt' \; ; test ! -f vet_failed.txt
```
broady, line 36: consider splitting it out to a shell script in the testing directory if it's going to get complicated

tbpg: I think that will be the way to go in the future. I'd like to keep the automatic parallelism, but this is a gnarly line of Bash.

testing/kokoro/system_tests.sh
tbpg: It's already included in the existing lint action. I updated the PR title to make that more clear.

testing/kokoro/system_tests.sh
tbpg: It also checks root. `find . -name go.mod` includes the `go.mod` in the current directory.

PR #1547, opened 2020-06-30
storage: reuse testutil.CleanBucket

tbpg:
I also removed an extraneous call to clean a bucket not used in the test.

Updates #1549.


tbpg:
@tritone, looks like this is still hitting `Error 429: The project exceeded the rate limit for creating and deleting buckets., rateLimitExceeded`. Should we reconfigure something in the test or adjust quota?

tbpg:
Thanks! Tests passed this time around. I edited the comment to not say this fixes the issue. I think we can move forward with this PR while we investigate the quota issue separately.

PR #1552, opened 2020-07-02
testing: skip modules with a newer version than toolchain

broady:
This allows newer language/toolchain features to be used.
Tests in that module are skipped if the installed Go version in Kokoro
is too old.

### DIFF
--- a/testing/kokoro/system_tests.sh
+++ b/testing/kokoro/system_tests.sh
@@ -185,6 +185,13 @@ set +e # Don't exit on errors to make sure we run all tests.
 # runTests runs the tests in the current directory. If an argument is specified,
 # it is used as the argument to `go test`.
 runTests() {
+  if goVersionShouldSkip; then
+    set +x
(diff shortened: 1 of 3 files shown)


testing/kokoro/system_tests.sh
```diff
@@ -197,6 +204,18 @@ runTests() {
...
+# Returns 0 if the test should be skipped because the current Go
+# version is too old for the current module.
+goVersionShouldSkip() {
+  modVersion="$(go list -m -f '{{.GoVersion}}')"
+  if [ -z "$modVersion" ]; then
+    # Not in a module or minimum Go version not specified, don't skip.
+    return 1
+  fi
+
+  ! (go list -f "{{context.ReleaseTags}}" | grep -q "go$modVersion\b")
```
broady, line 216:
```suggestion
  go list -f "{{context.ReleaseTags}}" | grep -q -v "go$modVersion\b"
```

PR #1563, opened 2020-07-08
cloudsql: add Postgres sample

shubha-rajan:
This is mostly copied from the existing SQL Server sample that was recently merged (#1390) but edited to use [pgx](https://github.com/jackc/pgx) as a driver instead. Other changes include:
* Allowing override of the default `/cloudsql` Unix socket path with an environment variable
* Adding instructions to deploy to App Engine Flexible and Cloud Run


shubha-rajan:
> Other Postgres samples have used https://github.com/lib/pq. Is there a reason for this to use a different package? In general, I'd prefer to be consistent.

From the documentation for pq: "This package is effectively in maintenance mode and is not actively developed. Small patches and features are only rarely reviewed and merged. We recommend using pgx which is actively maintained."

tbpg:
Will wait for @dmahugh before merging.

shubha-rajan:
sgtm. just updated the README to mirror some changes I made in a separate PR for the mysql sample.

PR #1567, opened 2020-07-10
gaming/servers: add samples cloud gaming

Laremere:
This is adding all of the samples we need for cloud gaming.  This will be used as the canonical sample for writing the samples in other languages, though it is roughly based on the gcloud examples in the documentation: https://cloud.google.com/game-servers/docs/how-to/creating-realm

Given that there is a hierarchy of resource types, in the test the root resource types tests are also used to create the parents for the child resource types.

Draft: Need to give the test projects access to the sample cluster.  We're using one cluster to avoid toil on updates.


Laremere:
Running goimports and mod tidy (per presubmit failure's instructions) changed a bunch of files.

gaming/servers/create_config.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+
+	gaming "cloud.google.com/go/gaming/apiv1beta"
+	gamingpb "google.golang.org/genproto/googleapis/cloud/gaming/v1beta"
+)
+
+const fleet = `
```
tbpg, line 28: Perhaps include a comment explaining the format of this?

gaming/servers/create_cluster.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2019 Google LLC
```
tbpg, line 1:
```suggestion
// Copyright 2020 Google LLC
```

Applies to all of the new files.

gaming/servers/create_config.go
Laremere: done.

gaming/servers/create_cluster.go
Laremere: done.

PR #1571, opened 2020-07-10
Run goimports and go mod tidy

Laremere:


tbpg:
I ran `go mod tidy` and `goimports -w .` locally and there were no changes. So, I don't think this PR is necessary.

PR #1584, opened 2020-07-17
all: skip vision, bigquerystorage for mTLS (known bad)

broady:
Adds a test that fails when these endpoints are ready to use and the
test skipping can be removed.

### DIFF
--- a/bigquery/bigquery_storage_quickstart/main_test.go
+++ b/bigquery/bigquery_storage_quickstart/main_test.go
@@ -23,6 +23,8 @@ import (
 )
 
 func TestApp(t *testing.T) {
+	testutil.KnownBadMTLS(t)
+
 	tc := testutil.SystemTest(t)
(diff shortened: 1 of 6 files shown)


broady:
CC= @shinfan

shinfan:
Thanks for putting this up quickly!

shollyman:
I lack context on the mtls efforts, but the skip logic based on an env variable seems reasonable.  Is this something special to the specific API(s)? I'm trying to figure out if we need to think about this when augmenting test/sample surface for other secondary APIs.

broady:
> Is this something special to the specific API(s)?

Yes, they are two APIs that need to be reconfigured to work properly with their mtls hostname.

> I'm trying to figure out if we need to think about this when augmenting test/sample surface for other secondary APIs.

Not sure what you mean. All APIs we have samples for get tested nightly for the mtls functionality. Can follow up if you want!

PR #1590, opened 2020-07-17
testing: skip bad mtls by removing go files

broady:
system_tests.sh is a bit over-aggressive (but rightly so) in deciding
which tests to run.

Skip them by deleting them at the beginning of the CI run.

Also adds functions/imagemagick to the skiplist.


broady:
CC= @shinfan

PR #1595, opened 2020-07-21
use atomic

haoxins:


haoxins:
@googlebot I signed it!

haoxins:
> `mu` is used as a lock, so there shouldn't be any race conditions.

that should be, but not working as expect in my code.
something strange.

`GCP functions go1.13`

PR #1596, opened 2020-07-21
testing: skip cloud run e2e tests for mtls_smoketest

broady:
These are failing because of the mod=readonly go commands in the Docker
build, which conflicts with the "go get @master" we use for mtls_smoketest.

These wouldn't be actually helpful for the mTLS tests, anyway, because
the GOOGLE_API_USE_MTLS isn't set to "always" as part of the deployment.


broady:
CC= @shinfan

broady:
Not deleting all of the tests in those directories because some of them have useful tests.

tbpg:
Superseded by #1612.

Issue #1602, opened 2020-07-24
all: code does not build with go1.11


codyoss:
The main module declares a version of 1.11, but does not build with 1.11.

```bash
$ go version
go version go1.11.13 linux/amd64

$ go build ./...
# golang.org/x/exp/rand
/go/pkg/mod/golang.org/x/exp@v0.0.0-20200513190911-00229845015e/rand/rng.go:63:19: undefined: bits.Add64
/go/pkg/mod/golang.org/x/exp@v0.0.0-20200513190911-00229845015e/rand/rng.go:64:16: undefined: bits.Add64
/go/pkg/mod/golang.org/x/exp@v0.0.0-20200513190911-00229845015e/rand/rng.go:68:12: undefined: bits.Mul64
```

PR #1603, opened 2020-07-24
fix: make sure go.mod matches actual language level

codyoss:
Fixes: #1602


codyoss:
Looks like there are some more in the other modules will, be adding more constraints as I find them

PR #1604, opened 2020-07-24
pubsub: add subscription detach samples

hongalex:


pubsub/subscriptions/detach_subscription.go
```diff
@@ -0,0 +1,46 @@
...
+	// subName := "projects/my-project/subscriptions/my-sub"
+	ctx := context.Background()
+	client, err := pubsub.NewClient(ctx, projectID)
+	if err != nil {
+		return fmt.Errorf("pubsub.NewClient: %v", err)
+	}
+
+	// Call DetachSubscription, which detaches a subscription from
+	// a topic you own. This can only be done if you have the
+	// pubsub.topics.detachSubscription role on the topic.
```
anguillanneuf, line 37: 👍

pubsub/subscriptions/detach_subscription.go
```diff
@@ -0,0 +1,46 @@
...
+	_, err = client.DetachSubscription(ctx, subName)
+	if err != nil {
+		return fmt.Errorf("detach subscription failed: %v", err)
+	}
+	fmt.Fprintf(w, "Detached subscription %s", subName)
```
anguillanneuf, line 42: It will be good to show detached=true in the subscription after detachment.

pubsub/subscriptions/detach_subscription.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func detachSubscription(w io.Writer, projectID, subName string) error {
+	// projectID := "my-project-id"
+	// subName := "projects/my-project/subscriptions/my-sub"
+	ctx := context.Background()
+	client, err := pubsub.NewClient(ctx, projectID)
+	if err != nil {
+		return fmt.Errorf("pubsub.NewClient: %v", err)
+	}
```
tbpg, line 33: `defer client.Close()`

pubsub/subscriptions/detach_subscription.go
hongalex: I would like to do this, but it causes the sample to look strange. `subName` is the fully qualified, globally unique name (required to call `DetachSubscription`). In the Go library, we expose two ways of grabbing a subscription, `client.Subscription` which uses the same `projectID` that was passed into the `NewClient` object. This isn't ideal because the subscription exist in another project. The alternative method is `client.SubscriptionInTopic`, which takes the project and subscription name separately (but requires splitting the subName before calling that method).

I check for `detached = true` in the tests, so I think that would suffice.

hongalex:
CI is failing due to the golang sample test projects not being added to allowlist.

PR #1614, opened 2020-07-29
refactor(run/grpc-ping): use client library to mint identity token

grayside:
This change switches from metadata server to use the new google.golang.org/api/idtoken package.

I ran the tests (including the e2e auth test) locally and it passed.


run/grpc-ping/request_auth.go
```diff
@@ -21,30 +21,35 @@ import (
...
 
 	pb "github.com/GoogleCloudPlatform/golang-samples/run/grpc-ping/pkg/api/v1"
 )
 
-// pingRequestWithAuth mints a new ID Token with the compute metadata server for each request.
+// pingRequestWithAuth mints a new Identity Token for each request.
 // This token has a 1 hour expiry and should be reused.
-// This function will only work on Google Cloud with an available compute metadata server and compute identity.
-// audience is the auto-assigned URL of the Cloud Run service (do not include port number)
+// See https://godoc.org/golang.org/x/oauth2#ReuseTokenSource
```
tbpg, line 33:
```suggestion
// See https://godoc.org/golang.org/x/oauth2#ReuseTokenSource.
```

codyoss: This is auto-refreshed for the user, no need to call this out.

grayside: @codyoss are you saying ReuseTokenSource is not needed for auto-refresh or recommending that it's not worth an inline comment?

codyoss: I am saying that wrapping is done under the hood already and token will be refreshed for the user.

codyoss: So I guess I am saying we don't need to comment of ReuseTokenSource. From the [impl](https://github.com/googleapis/google-api-go-client/blob/f10cc94b4f1ec2f17bebd8f39f16c6e8822f6658/idtoken/idtoken.go#L123).

grayside: Nice! I hadn't noticed that. I will clean up both PRs on this.

PR #1617, opened 2020-07-30
feat(run/markdown-preview): use client library to mint identity tokens

grayside:
This change switched the run/markdown-preview sample to use the client library to mint identity tokens for requests to the private renderer service. As part of this change, we now have built-in token reuse via the caching system prescribed by `oauth2.ReuseTokenSource`.


run/markdown-preview/editor/render.go
```diff
@@ -46,14 +50,24 @@ func (s *RenderService) NewRequest(method string) (*http.Request, error) {
...
+	defer cancel()
+
+	// Create a TokenSource if none exists.
+	if s.tokenSource == nil {
+		tokenSource, err := idtoken.NewTokenSource(ctx, s.URL)
+		if err != nil {
+			return nil, fmt.Errorf("idtoken.NewTokenSource: %w", err)
+		}
+		// Create a caching token source to reuse tokens until expiration.
+		s.tokenSource = oauth2.ReuseTokenSource(nil, tokenSource)
```
codyoss, line 63: This is already handled by the library, no need to also do it here. It will just be a no-op: https://github.com/googleapis/google-api-go-client/blob/9fba589c08a5782269327760a7c342fd1159b8e9/idtoken/idtoken.go#L123

run/markdown-preview/editor/render.go
```diff
@@ -46,14 +50,24 @@ func (s *RenderService) NewRequest(method string) (*http.Request, error) {
...
 	}
 
-	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token))
+	token, err := s.tokenSource.Token()
+	if err != nil {
+		return nil, fmt.Errorf("TokenSource.Token: %w", err)
+	}
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token.AccessToken))
```
codyoss, line 70: Maybe we should use [SetAuthHeader](https://pkg.go.dev/golang.org/x/oauth2?tab=doc#Token.SetAuthHeader) instead?

grayside:
```
datastore: Could not find a free project. Try again soon.
```

PR #1618, opened 2020-07-30
feat(firestore): query watch samples

### LINKED ISSUE #1619: firestore: Query Watch Samples
From QLogic Q3 issues

AlisskaPie:
Samples were added for:
- [x] Listen to a document 
- [x] Listen to multiple documents in a collection
- [x] View changes between snapshots
- [x] Detach a listener
- [x] Handle listen errors

Since there is no context that can be passed to the `Next()` function, I decided not to go through all the snapshots, but to take the first one, as in samples in Java, and provide features on it.

Closes #1619


AlisskaPie:
@BenWhitehead @tbpg @tritone review it, please

firestore/firestore_snippets/listen_test.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	return client, projectID
+}
+
+func TestListen(t *testing.T) {
+	ctx := context.Background()
+	client, projectID := setup(ctx)
+	defer client.Close()
+
+	if projectID == "" {
```
tritone, line 42: Could you pass `t` into `setup` and have this check in there rather than duplicating across all tests?

firestore/firestore_snippets/listen_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Skip("Skipping firestore test. Set GOLANG_SAMPLES_FIRESTORE_PROJECT.")
+	}
+	if err := prepareQuery(ctx, client); err != nil {
```
tritone, line 45: Let's create the query in this test rather than relying on another sample.

firestore/firestore_snippets/listen_stop.go
```diff
@@ -0,0 +1,44 @@
...
+	// projectID := "project-id"
+	ctx := context.Background()
+
+	client, err := firestore.NewClient(ctx, projectID)
+	if err != nil {
+		return fmt.Errorf("firestore.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	qsnap := client.Collection("cities").Where("state", "==", "CA").Snapshots(ctx)
```
tritone, line 37: I notice we don't actually make any reference to the concept of a listener in the cloud.google.com/go/firestore docs. Should that change? Is a query snapshot equivalent? I just worry that this is a bit confusing. @BenWhitehead maybe you have thoughts?

tritone:
DIscussed with @BenWhitehead . We think that these samples should instead use a channel or similar to await `QuerySnapshotIterator.Next`, which will block until there are changes to be picked up. Ben will add an example of what he means to the review.

AlisskaPie:
Thanks for review.
I'm out of office for some time, and will add changes ASAP

BenWhitehead:
I spoke with @tbpg and he's going to take over drafting an example of what we'd like from the channel & go-routine approach

firestore/firestore_snippets/listen_test.go
AlisskaPie: Done

firestore/firestore_snippets/listen_test.go
AlisskaPie: Sounds reasonable, done

AlisskaPie:
> I'd prefer to create a `context.Context` with a timeout and loop over the snapshot iterator until hitting the timeout. No need for channels or goroutines in the sample code, but we should loop "forever" (until a `ctx` runs out).
> ...

Thanks for pretty detailed answer

firestore/firestore_snippets/listen_document.go
```diff
@@ -19,33 +19,42 @@ import (
...
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("Snapshots.Next: %v", err)
+		}
+		if !snap.Exists() {
+			fmt.Fprintf(w, "Document no longer exists\n")
+			return nil
+		}
+		fmt.Fprintf(w, "Received document snapshot: %v\n", snap.Data())
```
tritone, line 56: Should there be a sleep between loop iterations? @tbpg thoughts?

firestore/firestore_snippets/listen_stop.go
```diff
@@ -1,44 +0,0 @@
...
-	defer client.Close()
-
-	qsnap := client.Collection("cities").Where("state", "==", "CA").Snapshots(ctx)
-	// Stop listening for changes.
-	qsnap.Stop()
-
-	return nil
-}
-
-// [END fs_detach_listener]
```
tritone: I think it might make sense to keep this just for completeness; it would be odd if there are Go snippets for everything else on the page except for this.

firestore/firestore_snippets/listen_test.go
```diff
@@ -36,66 +39,54 @@ func setup(ctx context.Context) (*firestore.Client, string) {
...
+
+	for _, c := range cityCollection {
+		if _, err := client.Collection("cities").Doc(c.city).Set(ctx, map[string]string{
+			"name":  c.name,
+			"state": c.state,
+		}); err != nil {
+			log.Fatalf("Set: %v", err)
+		}
 	}
 	if err := listenDocument(ioutil.Discard, projectID); err != nil {
```
tritone, line 61: Does this now take 30s per test? If so, could we make this more configurable, for example by allowing a ctx to be passed into the snippet also?

firestore/firestore_snippets/listen_test.go
```diff
@@ -36,66 +39,54 @@ func setup(ctx context.Context) (*firestore.Client, string) {
...
 		t.Errorf("listenDocument: %v", err)
```
tritone, line 62: I feel like a more thorough test would be to actually introduce changes in the docs in parallel with running the snippet and then validate that output occurs. This might not be practical here, though. @BenWhitehead any thoughts?

firestore/firestore_snippets/listen_document.go
tbpg: I don't think so. Will stop when the context is cancelled.

firestore/firestore_snippets/listen_stop.go
tbpg: "Stop" doesn't make sense as a standalone example, though. You can't listen without stopping. So, we can include something. But, it seems like a square peg in a round hole.

firestore/firestore_snippets/listen_changes.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"cloud.google.com/go/firestore"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// listenChanges listens to a query, returning the list of document changes.
+func listenChanges(w io.Writer, projectID string) error {
+	// projectID := "project-id"
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
tbpg, line 32: Let's pass a `context.Context` in so we can do a shorter timeout for tests. Include this 30 second timeout here so users don't end up with an infinite loop.

firestore/firestore_snippets/listen_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+func setup(ctx context.Context, t *testing.T) (*firestore.Client, string) {
+	projectID := os.Getenv("GOLANG_SAMPLES_FIRESTORE_PROJECT")
+	if projectID == "" {
+		t.Skip("Skipping firestore test. Set GOLANG_SAMPLES_FIRESTORE_PROJECT.")
+	}
+
+	client, err := firestore.NewClient(ctx, projectID)
+	if err != nil {
+		log.Fatalf("firestore.NewClient: %v", err)
```
tbpg, line 35:
```suggestion
		t.Fatalf("firestore.NewClient: %v", err)
```

firestore/firestore_snippets/listen_test.go
tbpg: +1. See other comment.

firestore/firestore_snippets/listen_stop.go
AlisskaPie: I added a comment in listenDocument sample for now

BenWhitehead: The intent with this snippet is to have a stand alone sample showing users how they can detach a registered listener. The idea behind it is that sometimes listeners will be put into the 'background' and at some later time will need to be detached. For example, starting a listener as part of a users session and detach when the session closes/times out. We may not need an entire file to demonstrate this action, but we should ideally keep these region tags (possibly as part of another sample) so that they can be displayed in the docs here https://firebase.google.com/docs/firestore/query-data/listen#detach_a_listener .

firestore/firestore_snippets/listen_test.go
BenWhitehead: Starting the listener before setting the data, then checking the expected document was sent to the listener would indeed be a stronger test.

firestore/firestore_snippets/listen_stop.go
tbpg: +1 to add the region tag to the existing file. When we embed the detach snippet, we can highlight the particular "detach" line(s).

AlisskaPie: Done, put tags in the listenDocument

firestore/firestore_snippets/listen_changes.go
```diff
@@ -0,0 +1,72 @@
...
+	"io"
+	"time"
+
+	"cloud.google.com/go/firestore"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// listenChanges listens to a query, returning the list of document changes.
+func listenChanges(ctx context.Context, w io.Writer, projectID string) ([]firestore.DocumentChange, error) {
```
tbpg, line 30: Do we need to return a list of changes?

firestore/firestore_snippets/listen_changes.go
```diff
@@ -0,0 +1,72 @@
...
+	defer cancel()
+
+	client, err := firestore.NewClient(ctx, projectID)
+	if err != nil {
+		return nil, fmt.Errorf("firestore.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	modified := false
+	changes := make([]firestore.DocumentChange, 0)
```
tbpg, line 42: If we need to return the changes (not sure we do):
```suggestion
	changes := []firestore.DocumentChange{}
```

firestore/firestore_snippets/listen_changes.go
```diff
@@ -0,0 +1,72 @@
...
+	it := client.Collection("cities").Where("state", "==", "CA").Snapshots(ctx)
+	for {
+		snap, err := it.Next()
+		if err != nil {
+			// DeadlineExceeded will be returned when ctx is cancelled.
+			if status.Code(err) == codes.DeadlineExceeded && modified {
+				return changes, nil
+			} else if status.Code(err) == codes.DeadlineExceeded && !modified {
```
tbpg, line 50: It's not an error to not have a modification. Should probably get rid of `modified`. Unless this is part of the sample spec?

firestore/firestore_snippets/listen_test.go
```diff
@@ -0,0 +1,136 @@
...
+func TestListen(t *testing.T) {
+	ctx := context.Background()
+	client, projectID := setup(ctx, t)
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, duration)
+	defer cancel()
+
+	// Delete all docs first to make sure setup works.
+	docs, err := client.Collection("cities").Documents(ctx).GetAll()
```
tbpg, line 51: Can we update this so we can run this test so we can run it twice at the same time in different `GOLANG_SAMPLES_PROJECT_ID`s? The usual way we've done that is by including the project ID in the collection name and making the collection name an argument for the sample functions. That way, you can get do `tc := testutil.SystemTest(t)` and use `tc.ProjectID`. That project ID is only used as an identifier, not as the `GOLANG_SAMPLES_FIRESTORE_PROJECT`.

firestore/firestore_snippets/listen_test.go
```diff
@@ -0,0 +1,136 @@
...
+		{city: "LA", name: "Los Angeles", state: "CA"},
+		{city: "DC", name: "Washington D.C."},
+	}
+
+	for _, c := range cityCollection {
+		if _, err := client.Collection("cities").Doc(c.city).Set(ctx, map[string]string{
+			"name":  c.name,
+			"state": c.state,
+		}); err != nil {
+			log.Fatalf("Set: %v", err)
```
tbpg, line 70:
```suggestion
			t.Fatalf("Set: %v", err)
```

Don't `log.Fatal` in tests.

firestore/firestore_snippets/listen_test.go
```diff
@@ -0,0 +1,136 @@
...
+	client, projectID := setup(ctx, t)
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, duration)
+	defer cancel()
+
+	c := make(chan []firestore.DocumentChange)
+	go func() {
+		defer close(c)
+		changes, err := listenChanges(ctx, ioutil.Discard, projectID)
```
tbpg, line 101: If you don't return the changes, you can just pass a `&bytes.Buffer{}` then check its content instead.

AlisskaPie: Oh, yeah, you're right, thanks!

firestore/firestore_snippets/listen_changes.go
AlisskaPie: Done

firestore/firestore_snippets/listen_changes.go
AlisskaPie: it's been removed

firestore/firestore_snippets/listen_changes.go
AlisskaPie: It seems we can do it without returning changes

firestore/firestore_snippets/listen_test.go
AlisskaPie: Thanks for clarifying it, I've made an update

tritone (approved):

tbpg (approved):

Issue #1566, opened 2020-07-09
run/events_pubsub: service does not start


grayside:
In trying out the events_pubsub container, the following error occurs:

```
docker run --rm -p 8080 pubsub-event
2020/07/09 21:43:07 expected a function returning one or mode of (*event.Event, protocol.Result) ordered; cannot convert parameter 1 from string to *event.Event or transport.Result
```

grant:
Fixed with #1621. Tested on Cloud Run:

```
curl -XPOST https://cloudrun-events-pubsub-q7vieseilq-uc.a.run.app --data '{"message":{"data":"YXNkZg=="}}'
```

PR #1620, opened 2020-07-30
fix: run events pubsub docker fix

grant:
Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/1566.

- Tests `HelloPubSub` function handler
- Added test for sending a CE to a localhost using Go SDK
  - Tested locally with valid CE POST request using Go SDK
- Deployed to Run successfully
- Adds outstanding TODO for using `httptest`. I'm not familiar with this and would rather have that test change in a separate PR.

Note, service does not respond to non-CloudEvent as intended by the CE SDK.

### DIFF
--- a/run/events_pubsub/main.go
+++ b/run/events_pubsub/main.go
@@ -21,45 +21,25 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"net/http"
-	"os"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 )
 
-var (
-	handler = http.DefaultServeMux
-)
-
 func main() {
 	ctx := context.Background()
-	// Create a new HTTP client for CloudEvents
 	p, err := cloudevents.NewHTTP()
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalf("failed to create protocol: %s", err.Error())
 	}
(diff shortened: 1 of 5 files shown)


run/events_pubsub/main.go
```diff
@@ -21,46 +21,26 @@ import (
...
-		port = "8080"
-		log.Printf("Defaulting to port %s", port)
-	}
-	// Start HTTP server.
-	log.Printf("Listening on port %s", port)
-	if err := http.ListenAndServe(":"+port, nil); err != nil {
-		log.Fatal(err)
-	}
+	log.Printf("Listening on http://localhost:8080\n")
+	log.Fatalf("Failed to start receiver: %s", c.StartReceiver(ctx, HelloPubSub))
```
grayside, line 41: I'm concern with the approach of using the CloudEvents SDK to manage the HTTP server. While it makes sense as a sort of CloudEvent version of a Functions Framework, in the context of Cloud Run it prevents us from showing developers how to create a fully-functional web server that happens to have some routes that handle CloudEvents.

This makes sense in a quickstart where the goal is more about getting something running, but if this is code we are going to use to show developers how to build CloudEvents support into a Cloud Run service, I think we need to present more flexibility.

PR #1621, opened 2020-07-31
feat(run/events): Run Events Quickstarts

grant:
Adds 2 basic run_events_* samples without using the Go CE SDK.
There are some opinions in the Go CE SDK that makes it easier for both developers and sample maintainers to not use the SDK for this language.

Storage (only use HTTP headers):

header: `ce-subject`

Pub/Sub HTTP payload:

```JSON
{
  "message": {
    "data": "SGVsbG8gdGhlcmU=", 
    "messageId": "1408577928008405", 
    "publishTime": "2020-08-06T22:31:14.536Z"
  },
  "subscription": "projects/MY-PROJECT/subscriptions/MY-SUB"
} 
```

Fixes #1516, https://github.com/GoogleCloudPlatform/golang-samples/issues/1566


run/events_pubsub/main.go
```diff
@@ -12,80 +12,66 @@
...
 
-// [START run_events_pubsub_handler]
-
-// PubSubMessage is the payload of the Pub/Sub message.
-type PubSubMessage struct {
-	Data []byte `json:"data,omitempty"`
-	ID   string `json:"id"`
+	s := fmt.Sprintf("Hello, %s! ID: %s", name, string(r.Header.Get("Ce-Id")))
+	log.Printf(s)
+	fmt.Printf(s)
```
tbpg, line 60: Why log and print?

run/events_pubsub/main.go
```diff
@@ -12,80 +12,66 @@
...
+	if port == "" {
+		port = "8080"
 	}
-	// Print and return the data from the Pub/Sub CloudEvent.
-	s := fmt.Sprintf("Hello, %s! ID: %s", string(m.Message.Data), event.ID())
-	log.Printf(s)
-	return s, nil
+
+	log.Printf("run_events_pubsub: listening on port %s", port)
+	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
```
tbpg, line 74: Please split this line up -- makes it easier for people newer to Go to follow:

https://github.com/GoogleCloudPlatform/golang-samples/blob/8d5a78f809c11604fa1d3edc2f3f1998eb061b7d/getting-started/helloworld/helloworld.go#L33-L35

run/events_pubsub/main_test.go
```diff
@@ -15,43 +15,85 @@
...
+		}
+		req := httptest.NewRequest("POST", "/", payload)
+		req.Header.Set("ce-id", test.ID)
+
+		rr := httptest.NewRecorder()
+		HelloEventsPubSub(rr, req)
+
+		w.Close()
+		log.SetOutput(os.Stderr)
+		log.SetFlags(originalFlags)
```
tbpg, line 85: Can update these outside the loop, then `defer` these calls so they're right next to the original modifications.

run/events_pubsub/e2e_test.go
```diff
@@ -25,10 +25,9 @@ import (
 )
 
 func TestPubSubEventsService(t *testing.T) {
-	t.Skip("Service is broken: https://github.com/GoogleCloudPlatform/golang-samples/issues/1566")
 	tc := testutil.EndToEndTest(t)
 
-	service := cloudrunci.NewService("pubsub-events", tc.ProjectID)
+	service := cloudrunci.NewService("run_events_pubsub", tc.ProjectID)
```
grayside, line 30: I believe this must be hyphens.

run/events_pubsub/main_test.go
```diff
@@ -15,43 +15,85 @@
...
+		{want: "Hello, World! ID: \n"},
+		{data: "Go", want: "Hello, Go! ID: 1234\n", ID: "1234"},
 	}
-	if want := "Hello, foo! ID: 321-CBA"; got != want {
-		t.Errorf("HelloPubSub: got %q, want %q", got, want)
+	for _, test := range tests {
+		r, w, _ := os.Pipe()
+		log.SetOutput(w)
+		originalFlags := log.Flags()
+		log.SetFlags(log.Flags() &^ (log.Ldate | log.Ltime))
```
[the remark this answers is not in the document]
grant: Testing this, I'm not able to see the ability to remove this log configuration without failing the tests due to date/time output.

run/events_pubsub/main_test.go
grant: I was able to set the flags outside of the loop.

I was not able to successfully update the output outside of the loop. I believe some of this logic is common with other files. The argument for setting this inside or outside of the loop could be argued both ways.

run/events_pubsub/e2e_test.go
grant: I couldn't find that requirement looking at the source and didn't see an error here.

run/events_pubsub/main.go
```diff
@@ -12,38 +12,49 @@
...
-// [START run_events_pubsub_server]
+// [START run_events_pubsub_handler]
 
-// Sample run-pubsub-events is a Cloud Run service which handles Pub/Sub messages.
+// Sample run-events-pubsub is a Cloud Run service which handles Pub/Sub messages.
 package main
 
 import (
-	"context"
+	b64 "encoding/base64"
```
tbpg, line 21:
```suggestion
	"encoding/base64"
```

The `base64` package is not usually renamed.

run/events_pubsub/main.go
```diff
@@ -12,38 +12,49 @@
...
-	ctx := context.Background()
-	// Create a new HTTP client for CloudEvents
-	p, err := cloudevents.NewHTTP()
-	if err != nil {
-		log.Fatal(err)
+// HelloEventsPubSub receives and processes a Pub/Sub push message.
+func HelloEventsPubSub(w http.ResponseWriter, r *http.Request) {
+	var e PubSubMessage
+	if err := json.NewDecoder(r.Body).Decode(&e); err != nil {
+		http.Error(w, "Bad Request", http.StatusBadRequest)
```
tbpg, line 42: Might want to log this error so it shows up somewhere.

run/events_pubsub/main.go
```diff
@@ -12,38 +12,49 @@
...
+		return
 	}
-	handleFn, err := cloudevents.NewHTTPReceiveHandler(ctx, p, HelloPubSub)
-	if err != nil {
-		log.Fatal(err)
+	dataBase64 := e.Message.Data
+	nameBytes, _ := b64.StdEncoding.DecodeString(dataBase64)
```
tbpg, line 46:
```suggestion
	nameBytes, err := base64.StdEncoding.DecodeString(e.Message.Data)
	if err != nil { ... }
```

run/events_pubsub/main.go
```diff
@@ -12,38 +12,49 @@
...
-	cloudevents "github.com/cloudevents/sdk-go/v2"
 )
 
-var (
-	handler = http.DefaultServeMux
-)
+// PubSubMessage is the payload of a Pub/Sub event.
+type PubSubMessage struct {
+	Message struct {
+		Data string `json:"data,omitempty"`
```
tbpg, line 32: Should this be `[]byte` instead of `string`? Would be able to forget about `base64` since it should be handled automatically.

run/events_storage/main.go
```diff
@@ -0,0 +1,48 @@
...
+	"log"
+	"net/http"
+	"os"
+)
+
+// HelloEventsStorage receives and processes a Pub/Sub message via a CloudEvent.
+func HelloEventsStorage(w http.ResponseWriter, r *http.Request) {
+	s := fmt.Sprintf("GCS CloudEvent type: %s", string(r.Header.Get("Ce-Subject")))
+	log.Printf(s)
+	fmt.Printf(s)
```
tbpg, line 31: Should this be `Fprintf`?

run/events_storage/main.go
```diff
@@ -0,0 +1,48 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START run_events_gcs_handler]
+
+// Cloud Run service which handles Audit Logs from Cloud Storage.
```
tbpg, line 17: Update to follow https://github.com/golang/go/wiki/CodeReviewComments#package-comments.

run/events_storage/main.go
```diff
@@ -0,0 +1,48 @@
...
+
+// HelloEventsStorage receives and processes a Pub/Sub message via a CloudEvent.
+func HelloEventsStorage(w http.ResponseWriter, r *http.Request) {
+	s := fmt.Sprintf("GCS CloudEvent type: %s", string(r.Header.Get("Ce-Subject")))
+	log.Printf(s)
+	fmt.Printf(s)
+}
+
+func main() {
+	log.Print("run_events_gcs: starting server...")
```
tbpg, line 35: Avoid "GCS" for branding. Should just be Storage, or perhaps `events_storage` if you want it to follow the binary name?

tbpg, line 35: Use `log.SetPrefix` if you'd like a consistent prefix (I see this is manually copied as a prefix a couple times).

run/events_storage/main.go
```diff
@@ -0,0 +1,48 @@
...
+
+	http.HandleFunc("/", HelloEventsStorage)
+
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = "8080"
+	}
+
+	log.Printf("run_events_gcs: listening on port %s", port)
+	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
```
tbpg, line 45: Split up following the pattern of other `ListenAndServe` calls (we went through and made them all consistent a while back).

run/events_pubsub/main.go
grant: In this case, we really want a string since the HTTP request sends a JSON string for the value. Otherwise we'll just need to convert the `[]byte` to a string later right?

LMK if there's a reason to use a `[]byte`.

run/events_pubsub/main.go
grant: Sure, added log.

run/events_pubsub/e2e_test.go
grayside: Cool, if it deploys and produces a working URL then it's good. I am surprised this is the case.

run/events_pubsub/main.go
grant: Applied.

run/events_storage/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+		want    string
+	}{
+		{subject: "1234", want: "GCS CloudEvent type: 1234\n"},
+	}
+	for _, test := range tests {
+		r, w, _ := os.Pipe()
+		log.SetOutput(w)
+		originalFlags := log.Flags()
+		log.SetFlags(log.Flags() &^ (log.Ldate | log.Ltime))
+
```
grayside, line 39: There are no test assertions on log output, so code to redirect logs can be skipped.

run/events_storage/main.go
grant: Done.

run/events_storage/main.go
grant: Good catch. Fixed.

run/events_storage/main.go
grant: Right now we can't change the `gcs` part for the region tag as it's in the other languages.

We can change this across languages uniformly later.

grant: Fixed this.

run/events_storage/main.go
grant: Fixed.

run/events_storage/main_test.go
grant: Without this, we get the date and time in the `string(out)` like `2020/08/10 12:16:43` which causes an error. Is there something we suggest instead?

grayside: No, you're right. I misread the read below as pulling from the HTTP response instead of the logs. Tyler's comment about using defer to restore the log config is applicable though.

run/events_pubsub/main_test.go
grant: Sorry, I'll need more guidance if we want to go this route. I've tried a few things.

This came from another test originally:
https://github.com/GoogleCloudPlatform/golang-samples/blob/a699cc86997175aa096113486151599df8948a10/functions/helloworld/hello_pubsub_test.go

run/events_pubsub/main.go
tbpg: Nope. Below, you do `nameBytes, err := base64.StdEncoding.DecodeString(e.Message.Data)`. You could delete all of that and just do:

```
name := string(e.Message.Data)
```

if `Data` was of type `[]byte`.

run/events_pubsub/main.go
```diff
@@ -12,38 +12,54 @@
...
+		http.Error(w, "Bad HTTP Request", http.StatusBadRequest)
+		log.Printf("Bad HTTP Request: %v", http.StatusBadRequest)
+		return
 	}
-	handleFn, err := cloudevents.NewHTTPReceiveHandler(ctx, p, HelloPubSub)
+	nameBytes, err := base64.StdEncoding.DecodeString(e.Message.Data)
 	if err != nil {
-		log.Fatal(err)
+		http.Error(w, "Bad Base64 Data", http.StatusBadRequest)
+		log.Printf("Bad Base64 Data: %v", http.StatusBadRequest)
```
tbpg, line 49: This doesn't print the actual error showing why it was bad. Do you want to?

run/events_pubsub/main.go
grant: Interesting. Fixed.

I guess a byte array to string will base 64 decode it?

run/events_pubsub/main.go
grant: Removed this part when using `[]byte` for the data.

run/events_pubsub/main_test.go
grant: Figured it out.

run/events_storage/main_test.go
grant: Fixed the defer part. It's an issue in other samples though.

grant:
Should be g2g @grayside @tbpg

PR #1625, opened 2020-08-01
feat(pubsub): add ordering key samples

hongalex:


pubsub/topics/publish_ordering.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				// Error handling code can be added here.
+				fmt.Printf("Failed to publish: %s\n", err)
+				atomic.AddUint64(&totalErrors, 1)
+				return
+			}
+		}(result)
+	}
+
+	wg.Wait()
```
tbpg, line 91: What is the part that needs to be done in the background here? `Publish` or `Get` or both? I'm wondering if we could split this into more than one loop to make it a little easier to follow.

pubsub/topics/publish_resume_ordering.go
```diff
@@ -0,0 +1,105 @@
...
+		{
+			message:     "message3",
+			orderingKey: "key1",
+		},
+		{
+			message:     "message4",
+			orderingKey: "key2",
+		},
+	}
+	for _, m := range messages {
```
tbpg, line 70: Same question here.

pubsub/topics/publish_ordering.go
hongalex: Yeah, all we want to show is how to publish with message ordering (lines 71-74). I'll try splitting this up until 2 loops.

PR #1642, opened 2020-08-11
fix(functions/console_snippets/http): add error handling and test coverage

### LINKED ISSUE #1639: functions: err suppressed
In the function snippets, the `err` value is not surfaced:

https://github.com/GoogleCloudPlatform/golang-samples/blob/a699cc86997175aa096113486151599df8948a10/functions/console_snippets/http/http.go#L31-L34

If the `err` is not `nil`, I would expect to see that and not see a normal hello world.

grayside:
This fixes the HTTP console snippet to log an error if invalid JSON is parsed.
It also adds test coverage to handle the various cases.

Fixes #1639


functions/console_snippets/http/http.go
```diff
@@ -29,6 +30,7 @@ func HelloWorld(w http.ResponseWriter, r *http.Request) {
 		Message string `json:"message"`
 	}
 	if err := json.NewDecoder(r.Body).Decode(&d); err != nil {
+		log.Printf("json.NewDecoder: %v", err)
 		fmt.Fprint(w, "Hello, World!")
```
grant, line 34: We should probably not print `Hello, World!` in this block if there is an error with decoding the message.

grant: I'd imagine something like:

```go
http.Error(w, "Server error", http.StatusUnprocessableEntity)
```

functions/console_snippets/http/http_test.go
```diff
@@ -0,0 +1,72 @@
...
+			data: `{"message": "Greetings, Ocean!"}`,
+		},
+		{
+			name: "empty",
+			want: "Hello, World!",
+			data: "",
+		},
+		{
+			name: "invalid",
+			want: "Hello, World!",
```
grant, line 32: I'm not sure why we would intentionally want data="" and data="not-valid-JSON" to produce the same result.

A "Data not valid." result or something else would seem fine.

functions/console_snippets/http/http.go
grayside: Ah, I see. I don't believe this had any stakeholders outside DPE, so we can change the design (I believe this was intentional, based on reading the comment.)

grayside: Checking this, it will make the the code several lines longer to support defaulting an empty payload to hello world. Is it worth it?

grant: We can create a GitHub issue or have a TODO comment and address this in a different PR. There are many instances of this logic (err != nil, still print) it looks like.

grayside: We can expand scope here as far as I'm concerned. The question to get to is to what extent we're trying to keep the "API Design".

1. HTTP request without a payload => "Hello, World!"
2. HTTP request with invalid payload => Bad Request
3. HTTP request with JSON payload but no/empty message => "Hello, World!"
4. HTTP request with JSON payload and message property => "$message"

I interpret the issue as asking for (2), do we want to respond bad request for (3)? A side effect of failing the response on invalid JSON is that we need to add ~6 lines of code to address "no body":

* We add code to inspect the error type/message
* We read in the body, check if it's empty, then separately unmarshall

The initial design is what you might consider "failing open" instead of "failing closed", so I see this less about correctness than a question of "fewest developers get distracted when reviewing the simple sample".

functions/console_snippets/http/http_test.go
```diff
@@ -0,0 +1,72 @@
...
+		data string
+	}{
+		{
+			name: "valid",
+			want: "Greetings, Ocean!",
+			data: `{"message": "Greetings, Ocean!"}`,
+		},
+		{
+			name: "empty",
+			want: "Hello, World!",
```
grant, line 27: This should probably also raise the error invalid JSON.

The data seen ("") is invalid JSON and should err.

functions/console_snippets/http/http.go
grant: Yes, that API response sounds good:

1. SG
2. Is what I'm asking for (expected: Bad Request, actual: Hello, World! for this PR)
3. SG
4. SG

I assume 3 and 4 are requests with valid JSON. Invalid JSON should respond with Bad Request. No body is invalid JSON.

I don't want to add super robust error handling, but simply put from above, a non nil error should be not produce a normal result.

grayside:
Badly broke this branch, closing will re-open with change.

PR #1649, opened 2020-08-14
fix(profiler): close connections/clients in shakesapp to fix memory leak

nolanmar511:
TESTED= enabled heap profiler and compared profiles w/ and w/o fix. Profiles w/ fix showed less memory being used.

Since this application is run only for short intervals in Profiler's How-To, I don't think we need to retake screenshots there.


nolanmar511:
@jqll -- Could you review this change?

PR #1681, opened 2020-08-28
refactor(run): move e2e tests to separate testing directory

### LINKED ISSUE #1674: run: Isolate Cloud Run e2e tests to avoid package dependency pollution
Cloud Run samples depend on the `internal` testutil and cloudrunci packages.
As a result, the golang-samples base package ends up as a dependency of the sample module.

* Removing this would speed up sample builds for developers, maintainers, and unit tests.
* Removing this would hide our test infrastructure from the sample, which may be interesting but is ultimately a distraction from sample goals.

## Proposed Solution

Create a `run/testing` directory.
Sample tests grouped such as:

* run/testing/pubsub.e2e_test.go
* run/testing/markdown_preview_editor.e2e_test.go

This has the following advantages:
* Cross-reference of tests with samples are relative direct
* Maximum co-location of tests without distracting from the samples
* Any future testing-only samples have a place to live.

@tbpg could you confirm this plan?

grayside:
Fixes #1674 

I had to create a module for the new /run/testing directory to ensure the e2e tests are only run once instead of in every supported Go version.

More discussion on this in the issue.


run/testing/events_pubsub.e2e_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package testing
```
tbpg, line 15: Might want a more specific package name than this -- at least different from the standard library `testing` package.

run/testing/events_storage.e2e_test.go
```diff
@@ -24,10 +24,11 @@ import (
 	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
 )
 
-func TestPubSubEventsService(t *testing.T) {
+func TestPubSubStorageService(t *testing.T) {
 	tc := testutil.EndToEndTest(t)
 
-	service := cloudrunci.NewService("events-storage", tc.ProjectID)
+	service := cloudrunci.NewService("events_storage", tc.ProjectID)
```
tbpg, line 30: I think this needs to be a dash, not an underscore?

run/testing/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/GoogleCloudPlatform/golang-samples/run/testing
+
+go 1.15
+
+require (
+	github.com/GoogleCloudPlatform/golang-samples v0.0.0-20200828000024-04937a2b81cf
+	github.com/GoogleCloudPlatform/golang-samples/run/grpc-ping v0.0.0-20200828000024-04937a2b81cf
+	github.com/GoogleCloudPlatform/golang-samples/run/grpc-server-streaming v0.0.0-20200828000024-04937a2b81cf
```
tbpg, line 8: Please add `replace` directives to make sure these tests always use the local versions rather than this particular pseudo-version.

run/testing/events_pubsub.e2e_test.go
grayside: Switching to `cloudruntests`

run/testing/go.mod
grayside: Good catch, thanks.

grayside:
I ran enough of the tests to prove the approach. Now with feedback will run through all.

run/testing/events_storage.e2e_test.go
grayside: Nice catch!

grayside:
Fighting with git over these various go.mod/go.sum conflicts. Will need to leave this until later/next week.

tbpg:
Suggested sequence:

```
git checkout master
git pull origin master
git checkout run-split-e2e
git merge master # Conflict!
rm run/testing/go.* # Avoid unknown to git error. We'll bring these back.
git co master -- run/**/go.mod run/**/go.sum
git restore run/testing
find run -name go.mod -execdir go mod tidy \;
git commit # Finish the merge
git push origin run-split-e2e
```

grayside:
Github is continuing to fight me, despite reconciling this multiple times along the lines you laid out. I'm going to force it through via the UI then correct any inconsistencies by layering on one more `go mod tidy` correction.

grayside:
Okay, moved the helloworld tests, I see go.mod stuff is back in conflict, will see if running your suggested steps from a clean state does the trick.

PR #1682, opened 2020-08-28
chore(deps): upgrade all

tbpg:
See second commit for `otel` fix.


tbpg:
Re-working the sample in #1685.

PR #1691, opened 2020-09-02
testing(functions): Add an E2E testing facility

### LINKED ISSUE #1690: functions: create facility to deploy Cloud Functions for e2e testing
In order to enable end-to-end/integration tests of Cloud Functions (#1381) we need a facility such as [cloudrunci](https://github.com/GoogleCloudPlatform/golang-samples/tree/master/internal/cloudrunci).

I propose creating `functest` as a similar internal package. However, where cloudrunci tries to enforce more of a workflow on what operations can be done, and provide facilities with various options for deployment, I think we should keep this simpler. In implementation of tests, the same pattern used in Cloud Run should be followed for Cloud Functions (#1674).

In scope here:
* Create a functest package with tests/docs
* Set up functions/testing with a first test implementation to demonstrate usage.

Not in scope:
* Retrofit all Cloud Functions E2E tests to use this new facility.

grayside:
Fixes #1690 

This is an initial cut of a simplified, GCF variant of `cloudrunci`. It includes a service-to-service test implementation and a package doc to illustrate usage.


grayside:
> 1. The tests that use BASE_URL are used as samples. Can we run those? Perhaps use a hidden TestMain that deploys the function(s)? Would end up depending on golang-samples, which isn't ideal. This part is core to me -- don't want to build this now then not be able to use it for the sample tests.

Ah, I hadn't realized that. I suspect some are just copying around a tests as a usage pattern.

A hidden TestMain seems like a good idea to me. Given a TestMain per package and a BASE_URL per function, not sure how we set up the environment as expected.

> 2. Is there any way we can combine this with cloudrunci? Option to deploy Cloud Run vs deploy Cloud Function? Ideally, that would avoid having to duplicate other components.

I looked into this briefly, and while I agree having a single package would be ideal, the amount of flexibility would require refactoring in cloudrunci or creating a bit of a beast. If we're going to build a solution that flexible, I'd prefer to go further and solve this in a cross-language, multi-platform way, which is one of my goals for the Samples Working Group. Maybe building on the ideas in this prototype: https://github.com/GoogleCloudPlatform/serverless-sample-tester.

My approach here is:

1. What can we use that provides a similar experience to cloudrunci
2. How can we keep the overhead of authenticated functions testing low
3. Gets these tests enabled sooner than later

In holding off for a more complete solution, we might expand on the TestMain idea and use os.Exec to run `gcloud functions deploy` for each function in the package.

tbpg:
> In holding off for a more complete solution, we might expand on the TestMain idea and use os.Exec to run gcloud functions deploy for each function in the package.

Sounds like a good solution to me. Can make it more complex if needed. `os.Setenv` after deploying the function seems like it would do the trick.

PR #1711, opened 2020-09-15
chore(testing/sampletests): make its own module

codyoss:
With the most recent commit, x/tools/packages will require 1.12.
Carve out a new module with proper go version.


codyoss:
Related to https://github.com/googleapis/google-cloud-go/pull/2849

testing/sampletests/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/GoogleCloudPlatform/golang-samples/testing/sampletests
+
+go 1.12
+
+require (
+	github.com/google/go-cmp v0.5.2
+	github.com/jstemmer/go-junit-report v0.9.1
```
tbpg, line 7: I think there is a replace in the root go.mod we should copy over here.

tbpg (changes requested):
Please double-check the installation of sampletests just in case we need to `cd` into that directory first (so we are inside the right module).

testing/sampletests/go.mod
codyoss: Whoops, good catch!

codyoss:
> Please double-check the installation of sampletests just in case we need to cd into that directory first (so we are inside the right module).

You're right, fixed.

tbpg (approved):

PR #1714, opened 2020-09-16
fix(storage): add viewing the CORS configuration

AlisskaPie:
[The documentation ](https://cloud.google.com/storage/docs/configuring-cors#viewing-cors-bucket)says that CORS fields could be viewed in bucket's metadata response, but they were missing.


AlisskaPie:
@codyoss @tritone @tbpg PTAL, this is just a small fix :)

tbpg (approved):

Issue #1707, opened 2020-09-14
storage: missing samples


AlisskaPie:
In [Storage Sample Coverage](https://docs.google.com/spreadsheets/d/1EfdvfkTG0hIYAdFI_Xs3ikqfH9oyFsfKy_5gv5baURo/edit?ts=5e1e6870#gid=2137817657) following samples are missing or have wrong region tag:
- [x] storage_bucket_delete_default_kms_key
- [x] storage_generate_encryption_key 
- [x] storage_view_bucket_iam_members
- [x] storage_add_bucket_label
- [x] storage_compose_file
- [x] storage_remove_bucket_label
- [x] storage_change_file_storage_class
- [x] storage_enable_versioning
- [ ] storage_set_metadata
- [x] storage_set_bucket_public_iam
- [x] storage_define_bucket_website_configuration
- [x] storage_download_public_file
- [x] storage_get_service_account
- [x] storage_cors_configuration
- [x] storage_disable_versioning
- [x] storage_copy_file_archived_generation
- [x] storage_object_csek_to_cmek
- [ ] storage_change_default_storage_class
- [x] storage_delete_file_archived_generation
- [x] storage_remove_cors_configuration

tbpg:
@tritone @crwilcox What is the next step for this issue?

tritone:
Reassigning to @JesseLovelace who is working on sample completeness. Jesse, is there any remaining work to be completed here or should this be closed out?

PR #1722, opened 2020-09-21
feat(storage): sample gap (1/2 object part)

AlisskaPie:
Added samples with following region tags:
- [x] storage_generate_encryption_key 
- [x] storage_enable_versioning
- [x] storage_disable_versioning
- [x] storage_copy_file_archived_generation
- [x] storage_delete_file_archived_generation
- [x] storage_get_service_account

First part(objects) from [this issue](https://github.com/GoogleCloudPlatform/golang-samples/issues/1707)


AlisskaPie:
@tritone @tbpg PTAL

storage/objects/generate_encryption_key.go
```diff
@@ -0,0 +1,36 @@
...
+// [START storage_generate_encryption_key]
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"io"
+)
+
+// generateEncryptionKey generates a base64-encoded encryption key.
+func generateEncryptionKey(w io.Writer) error {
```
tritone, line 26: I notice that the C++ sample here has a lot more information about encryption. Do we want to include that in the other samples (or pull up into the docs)? @frankyn  https://cloud.google.com/storage/docs/encryption/using-customer-supplied-keys#generating_your_own_encryption_key

tritone (commented):
Again, let's wait for @JesseLovelace to review before merging.

storage/objects/generate_encryption_key.go
JesseLovelace: The C++ sample seems a little excessive to me, but some other samples have a little piece of clarification text that I think should be included here:
```
    Generates a 256 bit (32 byte) AES encryption key and prints the
    base64 representation.

    This is included for demonstration purposes. You should generate your own
    key. Please remember that encryption keys should be handled with a
    comprehensive security policy.
```

JesseLovelace (changes requested):
Only note is about the comment in key generation sample, other than that looks good

storage/objects/generate_encryption_key.go
AlisskaPie: Done

storage/service_account/hmac/hmac_test.go
```diff
@@ -134,6 +134,13 @@ func TestGetKey(t *testing.T) {
 	}
 }
 
+func TestGetServiceAccount(t *testing.T) {
```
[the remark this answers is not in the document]
AlisskaPie: Done

storage/objects/copy_file_archived_generation.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"cloud.google.com/go/storage"
+)
+
+// copyOldVersionOfObject copies a noncurrent version of an object.
+func copyOldVersionOfObject(w io.Writer, srcBucket, srcObject, dstObject string, gen int64) error {
```
tritone, line 28: Change `srcBucket` -> `bucket` (since there's only one now).

tritone (commented):
Couple nits, otherwise looks good!

storage/service_account/service_account.go
```diff
@@ -0,0 +1,44 @@
...
+
+// getServiceAccount gets the default Cloud Storage service account email address.
+func getServiceAccount(w io.Writer, projectID string) error {
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("storage.NewClient: %v", err)
+	}
+	defer client.Close()
+
```
tritone, line 34: Keep a context timeout, but set to 10s instead of 1 min.

storage/objects/copy_file_archived_generation.go
AlisskaPie: Done

storage/service_account/service_account.go
AlisskaPie: Done

tritone (approved):

tbpg (approved):

PR #1723, opened 2020-09-22
fix(functions/console_snippets/http): error handling and unit tests

### LINKED ISSUE #1639: functions: err suppressed

grayside:
Fixes #1639 

1. Invalid JSON results in a BadRequest error.
2. No body / no message property results in default message.

This is a retry on #1642, in which I broke the branch. This version has further refinements, so should be preferred regardless.

### DIFF
--- a/functions/console_snippets/http/http.go
+++ b/functions/console_snippets/http/http.go
@@ -19,6 +19,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"html"
+	"io"
+	"log"
 	"net/http"
 )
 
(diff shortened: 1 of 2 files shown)


tbpg (changes requested):
Need a license header then LGTM.

grant (approved):

tbpg (approved):

PR #1724, opened 2020-09-22
docs(appengine): remove master_slave region tags

tbpg:
I updated all of the regions in this file to start with `gae_consistency`.


codyoss (approved):
LGTM

PR #1733, opened 2020-09-28
chore: Updated appengine samples to 1.14

engelke:


appengine/go11x/cloudsql/go.mod
```diff
@@ -1,5 +1,5 @@
 module cloudsql
 
-go 1.11
```
tbpg, line 3: The go.mod specifies the minimum version of Go needed to run the code. So, we may want to keep this at 1.11.

tbpg (changes requested):
I don't expect any problems, but did you happen to deploy these to go114 to confirm they work?

engelke:
Yes, every one was deployed and worked.

appengine/go11x/cloudsql/go.mod
grayside: GAE Go 1.11 is a LTS release, at what point should we consider idiomatic language updates such as pervasive `%w` and raising the minimum version?

engelke: I'll make the change soon, but I'm out for a few days.

engelke: Restored 1.11 in go.mod in this PR

appengine/go11x/warmup/app.yaml
```diff
@@ -1,9 +1,8 @@
-runtime: go111
+runtime: go114
 
 inbound_services:
   - warmup
 
 handlers:
   - url: /_ah/warmup
     script: auto
-    login: admin
```
grayside, line 9: Given auto is the default behavior, is the handlers block needed in Gen2?

engelke: It seems to be ignored in Gen2. But some people may still look at this sample for Gen1, so I left it here.

tbpg (approved):
Thank you!

PR #1753, opened 2020-10-07
pubsub: fix error in the concurrency example.

mindprince:
The current example is processing all the messages in a single goroutine and is NOT concurrent.

 To make tests pass, use a counter that can be incremented concurrently.

### DIFF
--- a/pubsub/subscriptions/pull_concurrency.go
+++ b/pubsub/subscriptions/pull_concurrency.go
@@ -18,14 +18,14 @@ package subscriptions
 import (
 	"context"
 	"fmt"
-	"io"
 	"runtime"
+	"sync/atomic"
 	"time"
 
 	"cloud.google.com/go/pubsub"
 )
 
-func pullMsgsConcurrenyControl(w io.Writer, projectID, subID string) error {
+func pullMsgsConcurrenyControl(counter *int32, projectID, subID string) error {
 	// projectID := "my-project-id"
 	// subID := "my-sub"
 	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


mindprince:
cc @hongalex @tbpg

hongalex:
Thanks for bringing this up. So the sample in question isn't about how to process messages concurrently, but [configuring the subscriber to pull messages concurrently](https://cloud.google.com/pubsub/docs/pull#concurrency_control). Specifically, `NumGoroutines` specifies how many streams are used to pull messages in the `Receive` call.

`MaxOutstandingMessages` configures additional callbacks in process at any one time, but does not configure message pulling itself. I think it could be useful configuring that setting and adding a comment about this as well.

The reason why the message processing is not concurrent is because writing to the buffer makes it easier to run the test associated with the sample by avoiding race conditions. Otherwise, we would have needed to create locks around writing to the buffer and it wouldn't be concurrent anyway. Concurrent message processing was meant to be implemented by developers, but I think a case can be made to refactor the tests to make showing concurrent processing more clear.

mindprince:
I see. I still think it's confusing to do explicit serial processing in a heading titled concurrency control (that's what confused me). Are there use cases where people would pull from multiple threads but do serial processing of messages? IMO, we should update the example to show concurrent processing of messages also.

hongalex:
No, generally no one would pull more than they can handle so I agree we should show concurrent processing. I believe the tests will currently fail due to the concurrency issue I mentioned above so if you want, you can either rework the test or close this PR and I'll fix it when I have some spare cycles.

mindprince:
Updated to fix the tests.

I tested this by:

```
$ export GOLANG_SAMPLES_PROJECT_ID=<MY-PROJECT>

$ go test  ./... -run TestPullMsgsConcurrencyControl -count 5
=== RUN   TestPullMsgsConcurrencyControl
--- PASS: TestPullMsgsConcurrencyControl (17.96s)
=== RUN   TestPullMsgsConcurrencyControl
--- PASS: TestPullMsgsConcurrencyControl (50.92s)
=== RUN   TestPullMsgsConcurrencyControl
--- PASS: TestPullMsgsConcurrencyControl (51.21s)
=== RUN   TestPullMsgsConcurrencyControl
--- PASS: TestPullMsgsConcurrencyControl (51.62s)
=== RUN   TestPullMsgsConcurrencyControl
--- PASS: TestPullMsgsConcurrencyControl (50.98s)
PASS
ok  	github.com/GoogleCloudPlatform/golang-samples/pubsub/subscriptions	222.691s
```

ks1707 (approved):

mindprince:
Ping. Can we get this merged.

pubsub/subscriptions/pull_concurrency.go
```diff
@@ -18,14 +18,14 @@ package subscriptions
...
-	"io"
 	"runtime"
+	"sync/atomic"
 	"time"
 
 	"cloud.google.com/go/pubsub"
 )
 
-func pullMsgsConcurrenyControl(w io.Writer, projectID, subID string) error {
+func pullMsgsConcurrenyControl(counter *int32, projectID, subID string) error {
```
hongalex, line 28: We should stick with writer as the first argument, per https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#print-to-an-iowriter-for-snippets

Instead, initialize a counter as a local variable and print after the end of the `Receive` call. Then, we can just test for a string that's something like "Received 5 messages".

tbpg:
> Ping. Can we get this merged.

Looks like another comment was posted right when you posted this.

mindprince:
I don't think I will have cycles to address the comments anytime soon.

I have seen people in my team get confused by the current examples. So, I think they should be fixed.

Feel free to either merge this and send a follow up to address the comments or close this and send a PR with the correct fix directly.

hongalex:
I created #1825 with the changes I had in mind based on your PR, so feel free to comment on that.

Sorry this process has been taking so long, and I appreciate the time you put into helping improve our samples. Closing in favor of #1825.

PR #1764, opened 2020-10-13
fix(bigquery):  minor refactor to avoid pushback failures

shollyman:
Currently, table snippet tests mutate metadata for a table serially.
Recently, changes to how mutation pushback work in the backend are
causing intermittent failures.  The pushback is significant enough
that retries are insufficient to address this with the table snippet
tests.

This change addresses the problem by directing some of the mutation
examples to modify an additional BigQuery table.

Fixes: #1749

### DIFF
--- a/bigquery/snippets/table/integration_test.go
+++ b/bigquery/snippets/table/integration_test.go
@@ -150,6 +150,12 @@ func TestTables(t *testing.T) {
 	if err := updateTableAddColumn(tc.ProjectID, testDatasetID, testTableID); err != nil {
 		t.Fatalf("updateTableAddColumn(%q %q): %v", testDatasetID, testTableID, err)
 	}
+
+	// Change tables to avoid hitting metadata update limits in a short period.
+	testTableID, err = bqtestutil.UniqueBQName("testtable")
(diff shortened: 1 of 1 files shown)


bigquery/snippets/table/integration_test.go
```diff
@@ -150,6 +150,12 @@ func TestTables(t *testing.T) {
 	if err := updateTableAddColumn(tc.ProjectID, testDatasetID, testTableID); err != nil {
 		t.Fatalf("updateTableAddColumn(%q %q): %v", testDatasetID, testTableID, err)
 	}
+
+	// Change tables to avoid hitting metadata update limits in a short period.
+	testTableID, err = bqtestutil.UniqueBQName("testtable")
+	if err := createTableExplicitSchema(tc.ProjectID, testDatasetID, testTableID); err != nil {
```
tbpg, line 156: Do we need to clean up the extra table?

tbpg (approved):
Thank you!

bigquery/snippets/table/integration_test.go
shollyman: No.  There's a recursive delete of the encapsulating dataset as part of a defer, which cleans up all the resources.

PR #1767, opened 2020-10-14
feat: add eventarc generic snippet

grant:
## feat: add eventarc generic snippet

- Simple Cloud Run service that logs and replies with the HTTP request (headers/body) received by the service.
  - Tests event received string, header, body
- Does not log `Authorization` header.
  - Includes test
- Very useful for debugging Eventarc systems in Go
- Adds a simple test to ensure the headers and body are appropriately logged.

### Testing

- Tested locally
- Deployed to Cloud Run
  - Example: https://helloworld123-q7vieseilq-uc.a.run.app/

### DIFF
--- a/eventarc/generic/main.go
+++ b/eventarc/generic/main.go
@@ -1,0 +1,90 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START eventarc_generic_handler]
+
+// Sample run-events-pubsub is a Cloud Run service which handles Pub/Sub messages.
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+)
+
+func logAndRespond(w http.ResponseWriter, msg string) {
+	log.Println(msg)
+	fmt.Fprintln(w, msg)
+}
+
+// GenericHandler receives and echos a HTTP request's headers and body.
+func GenericHandler(w http.ResponseWriter, r *http.Request) {
+	logAndRespond(w, "Event received!")
+
+	// Log all headers besides authorization header
+	logAndRespond(w, "HEADERS:")
+	delete(r.Header, "Authorization")
+	headerMap := make(map[string]string)
+	for k, v := range r.Header {
+		headerMap[k] = string(v[0])
+		logAndRespond(w, fmt.Sprintf("%q: %q\n", k, v[0]))
+	}
+
+	// Log body
+	logAndRespond(w, "BODY:")
+	bodyBytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		log.Printf("error parsing body: %v", err)
+	}
+	body := string(bodyBytes)
+	logAndRespond(w, body)
+
+	// Format and print full output
+	type result struct {
+		Headers map[string]string `json:"headers"`
+		Body    string            `json:"body"`
+	}
+	res := &result{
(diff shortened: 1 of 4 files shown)


michaelawyu (approved):

eventarc/generic/main_test.go
```diff
@@ -0,0 +1,71 @@
...
+		req := httptest.NewRequest("POST", "/", payload)
+		req.Header.Set("Ce-Id", "1234")
+		req.Header.Set("Authorization", "super-secret-value")
+		req.Header.Set("Ce-Source", "//storage.googleapis.com/projects/YOUR-PROJECT")
+		rr := httptest.NewRecorder()
+		GenericHandler(rr, req)
+
+		w.Close()
+
+		if code := rr.Result().StatusCode; code == http.StatusBadRequest {
```
tbpg, line 56: Should this check for anything other than success, rather than for this particular code?

eventarc/generic/main_test.go
```diff
@@ -0,0 +1,71 @@
...
+			t.Errorf("GenericHandler invalid input, status code (%q)", code)
+		}
+
+		out, err := ioutil.ReadAll(r)
+		if err != nil {
+			t.Fatalf("ReadAll: %v", err)
+		}
+		if got := string(out); test.want != "" && strings.Contains(got, test.want) != true {
```
tbpg, line 64:
```suggestion
		got := string(out)
		if !strings.Contains(got, test.want) {
```

Then, `got` can be reused below rather than being re-converted. Also, every string contains the empty string.

eventarc/generic/main_test.go
```diff
@@ -0,0 +1,71 @@
...
+			t.Errorf("\nGenericHandler: \ngot: %q\nwant to contain: %q", got, test.want)
```
tbpg, line 65:
```suggestion
			t.Errorf("GenericHandler: \ngot: %q\nwant to contain: %q", got, test.want)
```

eventarc/generic/main_test.go
```diff
@@ -0,0 +1,71 @@
...
+		log.SetOutput(w)
+		defer log.SetOutput(os.Stderr)
+
+		jsonStr := fmt.Sprintf(`{"message": "some string"}`)
+		payload := strings.NewReader(jsonStr)
+
+		req := httptest.NewRequest("POST", "/", payload)
+		req.Header.Set("Ce-Id", "1234")
+		req.Header.Set("Authorization", "super-secret-value")
+		req.Header.Set("Ce-Source", "//storage.googleapis.com/projects/YOUR-PROJECT")
```
tbpg, line 50: Consider including the header in the test case. That way, it's clear what is being tested against. On first read, it's unclear where `1234`, `super-secret-value`, etc. come from.

tbpg (commented):
Left some Go readability comments.

eventarc/generic/main.go
```diff
@@ -0,0 +1,90 @@
...
+// GenericHandler receives and echos a HTTP request's headers and body.
+func GenericHandler(w http.ResponseWriter, r *http.Request) {
+	logAndRespond(w, "Event received!")
+
+	// Log all headers besides authorization header
+	logAndRespond(w, "HEADERS:")
+	delete(r.Header, "Authorization")
+	headerMap := make(map[string]string)
+	for k, v := range r.Header {
+		headerMap[k] = string(v[0])
```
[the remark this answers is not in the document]
grant: Added comment saying this sample assumes headers don't have duplicate keys, which is true for the product's headers.

eventarc/generic/main.go
```diff
@@ -0,0 +1,90 @@
...
+		Headers map[string]string `json:"headers"`
+		Body    string            `json:"body"`
+	}
+	res := &result{
+		Headers: headerMap,
+		Body:    body,
+	}
+	out, err := json.Marshal(res)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
grant: Changed.

eventarc/generic/main_test.go
grant: This is fine for now.

eventarc/generic/main_test.go
grant: Thanks.

eventarc/generic/main_test.go
grant: I'm not sure if we're keeping this in the first place, so I think it's satisfactory for now.

PR #1770, opened 2020-10-15
fix(dialogflow): remove the cmd options

### LINKED ISSUE #1302: dialogflow: samples have incorrect imports
For example, dialogflow/detect_intent/detect_intent.go has `"flag"` in the import block, but it isn't used in any of the samples.

We should either delete the command line interfaces or move each sample into its own file so the imports will always be correct. I'd prefer separate files. https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#one-file-per-sample

Also, the sample functions should not be exported (https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#dont-export-sample-functions). For example, `DetectIntentStream`.

munkhuushmgl:
Fixes #1302


dialogflow/intent_management/intent_management.go
```diff
@@ -12,114 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package intent_mgmt
```
tbpg, line 15: `intent` or `intentmanagement` or `intentmgmt` (no underscore).

https://golang.org/doc/effective_go.html#package-names

dialogflow/intent_management/intent_management_test.go
```diff
@@ -12,14 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package intent_mgmt
```
tbpg, line 15: Update to match.

tbpg (changes requested):
Thank you! One small package name comment.

tbpg (approved):
Thanks!

PR #1788, opened 2020-10-21
refactor: improve eventarc go samples

grant:
Slightly improves Eventarc samples.

Addresses many comments from: https://github.com/GoogleCloudPlatform/golang-samples/pull/1767 (Thanks for the suggestions)

Feel free to add more suggestions, or 👍  if all good.

### DIFF
--- a/eventarc/generic/main.go
+++ b/eventarc/generic/main.go
@@ -61,6 +61,7 @@ func GenericHandler(w http.ResponseWriter, r *http.Request) {
 		Body:    body,
 	}
 	if err := json.NewEncoder(w).Encode(res); err != nil {
+		log.Printf("error encoding response: %v", err)
 		http.Error(w, "Could not marshal JSON output", 500)
 	}
 	fmt.Fprintln(w)
--- a/eventarc/generic/main.go
+++ b/eventarc/generic/main.go
@@ -35,9 +35,9 @@ func GenericHandler(w http.ResponseWriter, r *http.Request) {
 	log.Println("HEADERS:")
 	headerMap := make(map[string]string)
 	for k, v := range r.Header {
-		val := strings.Join(v, ",")
 		if k != "Authorization" {
-			headerMap[k] = string(val)
+			val := strings.Join(v, ",")
+			headerMap[k] = val
 			log.Println(fmt.Sprintf("%q: %q\n", k, val))
 		}
 	}
@@ -63,6 +63,7 @@ func GenericHandler(w http.ResponseWriter, r *http.Request) {
 	if err := json.NewEncoder(w).Encode(res); err != nil {
 		log.Printf("error encoding response: %v", err)
 		http.Error(w, "Could not marshal JSON output", 500)
+		return
 	}
 	fmt.Fprintln(w)
 }
--- a/eventarc/generic/main.go
+++ b/eventarc/generic/main.go
@@ -24,37 +24,32 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"strings"
 )
-
-func logAndRespond(w http.ResponseWriter, msg string) {
-	log.Println(msg)
-	fmt.Fprintln(w, msg)
-}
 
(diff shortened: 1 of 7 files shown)


grant:
Can I get a review @tbpg @grayside? Thanks!

eventarc/generic/main.go
```diff
@@ -36,12 +36,15 @@ func GenericHandler(w http.ResponseWriter, r *http.Request) {
 	logAndRespond(w, "Event received!")
 
 	// Log all headers besides authorization header
+	// Assumes headers don't have duplicate keys
```
tbpg, line 39: Why? `strings.Join(v, ",")` can join the values into a single string. Alternatively, keep it as a `[]string`.

eventarc/generic/main.go
```diff
@@ -62,11 +65,10 @@ func GenericHandler(w http.ResponseWriter, r *http.Request) {
 		Headers: headerMap,
 		Body:    body,
 	}
-	out, err := json.Marshal(res)
-	if err != nil {
-		panic(err)
+	if err := json.NewEncoder(w).Encode(res); err != nil {
```
tbpg, line 68: I'm a bit confused -- why write the body and heads above _and_ in JSON here? I might be missing something.

tbpg (changes requested):

eventarc/generic/main.go
```diff
@@ -62,11 +65,10 @@ func GenericHandler(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "Could not marshal JSON output", 500)
```
tbpg, line 69: Should probably `log` the error.

tbpg (changes requested):

eventarc/generic/main.go
grant: I suppose I haven't seen multiple non-unique headers in a HTTP request seen here.
Changed to the suggestion, thanks.

eventarc/generic/main.go
grant: Thanks for catching. You're right, we already respond here so no need to respond above.

Simplified the sample a bit.

grant (commented):
Thanks for the review @tbpg, PTAL.

eventarc/generic/main.go
grant: Now also logs the error.

eventarc/generic/main.go
```diff
@@ -24,34 +24,32 @@ import (
...
+	log.Println("HEADERS:")
 	headerMap := make(map[string]string)
 	for k, v := range r.Header {
-		headerMap[k] = string(v[0])
-		logAndRespond(w, fmt.Sprintf("%q: %q\n", k, v[0]))
+		val := strings.Join(v, ",")
+		if k != "Authorization" {
+			headerMap[k] = string(val)
+			log.Println(fmt.Sprintf("%q: %q\n", k, val))
+		}
```
tbpg, line 42:
```suggestion
		if k != "Authorization" {
			val := strings.Join(v, ",")
			headerMap[k] = val
			log.Println(fmt.Sprintf("%q: %q\n", k, val))
		}
```

You could also make `headerMap` a `map[string][]string` and not have to do `strings.Join`.

eventarc/generic/main.go
```diff
@@ -62,11 +60,11 @@ func GenericHandler(w http.ResponseWriter, r *http.Request) {
 		Headers: headerMap,
 		Body:    body,
 	}
-	out, err := json.Marshal(res)
-	if err != nil {
-		panic(err)
+	if err := json.NewEncoder(w).Encode(res); err != nil {
+		log.Printf("error encoding response: %v", err)
+		http.Error(w, "Could not marshal JSON output", 500)
```
tbpg, line 65:
```suggestion
		http.Error(w, "Could not marshal JSON output", 500)
		return
```

tbpg (changes requested):
A couple small comments.

tbpg (approved):

Issue #1795, opened 2020-10-23
endpoints / getting-started-grpc example Docker build seems to be broken


jmartin127:
When running through the gRPC example for Endpoints, I ran into this error that I have not been able to resolve.  Any insights into this would be greatly appreciated.  It reproduces pretty easily:

Steps to reproduce:

1. `$ git clone git@github.com:GoogleCloudPlatform/golang-samples.git`
1. `$ cd golang-samples/endpoints/getting-started-grpc/`
1. `$ docker build .`

Error:
```
$ docker build .
Sending build context to Docker daemon  32.26kB
Step 1/6 : FROM golang:alpine
 ---> 30df784d6206
Step 2/6 : RUN apk update   && apk add git
 ---> Using cache
 ---> c31075041eb7
Step 3/6 : COPY . /go/src/app
 ---> fb078af80334
Step 4/6 : RUN go get -v app/server
 ---> Running in 08aace27bae2
get "google.golang.org/grpc": found meta tag get.metaImport{Prefix:"google.golang.org/grpc", VCS:"git", RepoRoot:"https://github.com/grpc/grpc-go"} at //google.golang.org/grpc?go-get=1
google.golang.org/grpc (download)
get "golang.org/x/net/trace": found meta tag get.metaImport{Prefix:"golang.org/x/net", VCS:"git", RepoRoot:"https://go.googlesource.com/net"} at //golang.org/x/net/trace?go-get=1
get "golang.org/x/net/trace": verifying non-authoritative meta tag
golang.org/x/net (download)
github.com/golang/protobuf (download)
get "google.golang.org/protobuf/encoding/prototext": found meta tag get.metaImport{Prefix:"google.golang.org/protobuf", VCS:"git", RepoRoot:"https://go.googlesource.com/protobuf"} at //google.golang.org/protobuf/encoding/prototext?go-get=1
get "google.golang.org/protobuf/encoding/prototext": verifying non-authoritative meta tag
google.golang.org/protobuf (download)
get "google.golang.org/protobuf/encoding/protowire": found meta tag get.metaImport{Prefix:"google.golang.org/protobuf", VCS:"git", RepoRoot:"https://go.googlesource.com/protobuf"} at //google.golang.org/protobuf/encoding/protowire?go-get=1
get "google.golang.org/protobuf/encoding/protowire": verifying non-authoritative meta tag
[...]

PR #1797, opened 2020-10-23
fix: #1795 endpoints sample mustEmbedUnimplemented

jmartin127:
Testing:
* Verified that the Docker image can now be built.  
* This solution appears to be the preferred solution from what I have read.
* Had another user verify that this fix works for them as well.


jmartin127:
@googlebot I signed it!

jmartin127:
Going to re-open this with my other (personal) GitHub user.

PR #1798, opened 2020-10-23
fix: #1795 endpoints sample mustEmbedUnimplemented

jmartin127:
Testing:
* Verified that the Docker image can now be built.
* This solution appears to be the preferred solution from what I have read.
* Had another user verify that this fix works for them as well.


tbpg (dismissed):
Thank you!

tbpg (approved):

Issue #1800, opened 2020-10-26
logging: speed up TestSimpleLog


nicoleczhu:
Currently TestSimple log takes 15-37 minutes to run

PR #1801, opened 2020-10-26
test(logging): remove TestSimpleLog flakiness and reduce test run time

nicoleczhu:
fixes: #1559, #1800 

1. Flakiness context: DeleteLog itself creates an audit log entry which cannot be deleted by deleteLog(). This caused previous flakiness issues as the test was expecting only 2 logs even if the test was run/terminated many times (and had multiple audit logs). 

2. [Run time context](https://github.com/googleapis/google-cloud-go/issues/3088)


tbpg (approved):
Thank you!

PR #1805, opened 2020-10-27
feat(logging): advanced log write sample

nicoleczhu:
Fixes: https://github.com/googleapis/google-cloud-go/issues/3074

It will show up here:
https://cloud.google.com/logging/docs/samples/logging-write-log-entry-advanced

Rubric: go/cdpe-snippet-rubric

### DIFF
--- a/logging/advancedlog/main.go
+++ b/logging/advancedlog/main.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -38,14 +38,15 @@ func main() {
 	// Sets the name of the log to write to.
 	logger := client.Logger("my-log")
 
-	// Logs a basic entry
+	// Logs a basic entry.
 	logger.Log(logging.Entry{Payload: "hello world"})
 
-	// Logs an HTTPRequest type entry
+	// Logs an HTTPRequest type entry.
 	httpEntry := logging.Entry{
 		Payload: "optional message",
 		HTTPRequest: &logging.HTTPRequest{
-			// TODO: pass in request
+			// TODO(developer): pass in the request.
+			// Some request metadata will be autopopulated in the log entry.
 			Request: &http.Request{},
 		},
 	}
--- a/logging/advancedlog/main.go
+++ b/logging/advancedlog/main.go
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// [START logging_write_log_entry_advanced]
+
+// Writes an advanced log entry to Cloud Logging.
 package main
 
 import (
@@ -23,7 +26,7 @@ import (
 )
 
 func main() {
-	// [START logging_write_log_entry_advanced]
+	
 	ctx := context.Background()
 
 	// Creates a client.
@@ -48,5 +51,6 @@ func main() {
 		},
 	}
 	logger.Log(httpEntry)
-	// [END logging_write_log_entry_advanced]
 }
+
+// [END logging_write_log_entry_advanced]
--- a/logging/advancedlog/main.go
+++ b/logging/advancedlog/main.go
@@ -41,13 +41,15 @@ func main() {
 	// Logs a basic entry.
 	logger.Log(logging.Entry{Payload: "hello world"})
 
+	// TODO(developer): replace with your request value.
+	r, err := http.NewRequest("GET", "http://example.com", nil)
+
 	// Logs an HTTPRequest type entry.
+	// Some request metadata will be autopopulated in the log entry.
 	httpEntry := logging.Entry{
 		Payload: "optional message",
 		HTTPRequest: &logging.HTTPRequest{
-			// TODO(developer): pass in the request.
-			// Some request metadata will be autopopulated in the log entry.
-			Request: &http.Request{},
+			Request: r,
 		},
 	}
 	logger.Log(httpEntry)
--- a/logging/advancedlog/main.go
+++ b/logging/advancedlog/main.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"log"
 	"net/http"
+	"os"
 
 	"cloud.google.com/go/logging"
 )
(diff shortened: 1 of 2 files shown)


simonz130 (commented):
The link in which the sample shows seem to include nodejs sample that is not identical to the go sample (bar language differences). Could you also check which documentation page will include this sample (besides sample browser)?
Finally, we might need to revise that page to see if rest of the languages samples match yours (or the other way around) so the content of docs and the sample are consistent across 4 languages.

logging/advancedlog/main.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+	"log"
+	"net/http"
+
+	"cloud.google.com/go/logging"
+)
+
+func main() {
+	// [START logging_write_log_entry_advanced]
```
Daniel-Sanche, line 26: In some languages, we started trying to include the imports as part of the sample. Is go doing that?

logging/advancedlog/main.go
```diff
@@ -0,0 +1,52 @@
...
+	httpEntry := logging.Entry{
+		Payload: "optional message",
+		HTTPRequest: &logging.HTTPRequest{
+			// TODO: pass in request
+			Request: &http.Request{},
+		},
+	}
+	logger.Log(httpEntry)
+	// [END logging_write_log_entry_advanced]
+}
```
Daniel-Sanche, line 52: Is it possible to add a test for this?

logging/advancedlog/main.go
```diff
@@ -0,0 +1,52 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"log"
+	"net/http"
+
```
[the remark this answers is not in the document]
nicoleczhu: Yep its on purpose. Not enforced by go but is a internal convention for us

logging/advancedlog/main.go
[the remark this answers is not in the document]
nicoleczhu: done!

Daniel-Sanche (approved):

logging/advancedlog/main.go
nicoleczhu: yes , adding

logging/advancedlog/main.go
```diff
@@ -0,0 +1,52 @@
...
+	// Logs an HTTPRequest type entry
+	httpEntry := logging.Entry{
+		Payload: "optional message",
+		HTTPRequest: &logging.HTTPRequest{
+			// TODO: pass in request
+			Request: &http.Request{},
+		},
+	}
+	logger.Log(httpEntry)
+	// [END logging_write_log_entry_advanced]
```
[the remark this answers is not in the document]
nicoleczhu: there's a test in simplelog_test that covers advanced queries so if anything changes, we'll catch it there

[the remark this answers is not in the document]
nicoleczhu: Update: I started to write the test for this use case and realized why the other complete samples (that include imports) don't have tests. 
I lose context of the client used to write logs after calling this function, so I'm unable to list entries in the tests. I will omit the test for this sample.

nicoleczhu:
> The link in which the sample shows seem to include nodejs sample that is not identical to the go sample (bar language differences). Could you also check which documentation page will include this sample (besides sample browser)?

Tracking it here: https://github.com/googleapis/nodejs-logging/issues/922

logging/advancedlog/main.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2019 Google LLC
```
tbpg, line 1:
```suggestion
// Copyright 2020 Google LLC
```

logging/advancedlog/main.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	defer client.Close()
+
+	// Sets the name of the log to write to.
+	logger := client.Logger("my-log")
+
+	// Logs a basic entry
+	logger.Log(logging.Entry{Payload: "hello world"})
+
+	// Logs an HTTPRequest type entry
```
tbpg, line 44:
```suggestion
	// Logs a basic entry.
	logger.Log(logging.Entry{Payload: "hello world"})

	// Logs an HTTPRequest type entry.
```

logging/advancedlog/main.go
```diff
@@ -0,0 +1,55 @@
...
+	httpEntry := logging.Entry{
+		Payload: "optional message",
+		HTTPRequest: &logging.HTTPRequest{
+			// TODO: pass in request
```
tbpg, line 48:
```suggestion
			// TODO: pass in request.
```

tbpg, line 48: Should we expand this comment to say information will be pulled from the request and attached to the log? Not sure if it's needed.

logging/advancedlog/main.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"context"
+	"log"
+	"net/http"
+
+	"cloud.google.com/go/logging"
+)
+
+func main() {
```
tbpg, line 28: Add a test? See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/internal/testutil/runmain.go (not required to use it, but it may be helpful).

tbpg (changes requested):

logging/advancedlog/main.go
nicoleczhu: Updated. Should this be updated every time I edit an old file as well? Or it's timestamped as of the date of file creation?

logging/advancedlog/main.go
nicoleczhu: Yeah that's a good idea. we should advertise the fact that other httprequest metadata gets autopopulated into httprequest type logs!

logging/advancedlog/main.go
nicoleczhu: Oh neat. Added.

I test that the main executed without error or stderr logs. I did not test get-logs again. Figured it's redundant since I updated simplelog to run the same test. Also, it would take 15 seconds run.

nicoleczhu:
@tbpg mind taking another look? Thank you

tbpg (approved):
Thanks!

PR #1816, opened 2020-11-05
docs(firestore): normalize region tags

### LINKED ISSUE #1817: firestore: normalizing sample region tags
Make sample's region tags consistent across all the languages.
[Inventorying and Normalizing Firestore Sample Region Tags](https://docs.google.com/document/d/1l4oECPCalM0-vCBDGxbvW-_UXzEU0efBsJ346CuKCkk/edit#heading=h.be87cnseibxs)
[Tags tracker](https://docs.google.com/spreadsheets/d/1aRCMW0krFoe2Gabg7_R3WSByoNXqSMox1nx5sdclMkQ/edit#gid=0)

AlisskaPie:
Closes #1817


AlisskaPie:
I couldn't find a sample with this tag `fs_update_multiple_fields` on a site and there was no new region tag proposed in the tag tracker.
@tritone @BenWhitehead could you take a look?

firestore/firestore_quickstart/main.go
```diff
@@ -15,7 +15,7 @@
 // Sample firestore_quickstart demonstrates how to connect to Firestore, and add and list documents.
 package main
 
-// [START fs_initialize]
+// [START firestore_setup_client_create]
```
BenWhitehead, line 18: Thanks for opening this PR @AlisskaPie, we actually need to keep the existing tags and add the new tags alongside right now. At a later time we'll be able to come back and cleanup the existing tags.

Can you update the PR to look like the following:

```suggestion
// [START fs_initialize]
// [START firestore_setup_client_create]
```

BenWhitehead (changes requested):
Can you update the PR so that all tags have their existing and new values similar to the suggestion in my other comment?

firestore/firestore_quickstart/main.go
AlisskaPie: Thank you for quick review, done!
Also, there're 5 samples that were not implemented on a site yet. I did them recently, its query watch samples (
fs_detach_listener, 
fs_listen_document,
fs_listen_errors,
fs_listen_changes,
fs_listen_multiple)
I guess we can safely rename these tags.

BenWhitehead: I think this would be okay. @crwilcox Do you agree?

crwilcox: I agree. Are these a match for a rows in the tracking spreadsheet? If not can we add rows? We are going to want to convert all `fs_name` snippets in the end.

AlisskaPie: Yes, they are the same.
The one region tag left not replaced is `fs_update_multiple_fields` tag. There was no new region tag proposed for it in the tracking spreadsheet.

BenWhitehead (approved):

tritone (approved):
@tbpg The CI failures look like a test infra issue, can you investigate?

tbpg:
Sent https://github.com/GoogleCloudPlatform/golang-samples/pull/1834 which will hopefully fix it.

tmatsuo:
@tbpg 

snippet-bot is complaining because the snippet data still thinks `fs` is the right one.

This migration workflow is probably one of the edge cases that the current drift data can not cover.

I proposed that we have additional product prefix (in this case, `firestore` on top of `fs`), but they don't like the idea.

tbpg:
Thanks, @tmatsuo! Makes sense.

tbpg (approved):

PR #1819, opened 2020-11-05
docs: update Cloud Run region tags

averikitsch:


grayside (approved):
Phase 1 of expected region tag prefix migration.

PR #1821, opened 2020-11-09
chore(opentelemetry): update trace example to latest released APIs

briandealwis:
Update the OpenTelemetry example from OT SDK v0.11.0 to v0.13.0.
I needed to explicitly shutdown the Cloud Trace exporter to cause it to flush any pending spans to Cloud Trace.

I encountered this code when trying to follow the example at:

   https://cloud.google.com/trace/docs/setup/go-ot


opentelemetry/trace/main.go
```diff
@@ -63,6 +60,7 @@ func main() {
 
 		return nil
 	}(ctx)
+	exporter.Shutdown(ctx) // flush any pending spans
```
tbpg, line 63: Should this be called with `defer`?

tbpg (commented):
Thank you!

opentelemetry/trace/main.go
briandealwis: I had wondered, but `ctx` was defined further below.  Hmm, I'm not familiar with this code, but I wonder if we're better using the exporter options?

briandealwis: I don't see [any combination of bundler options][opts] that can be set to disable the bundling.

[opts]: github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace/trace.go

tbpg (approved):
Thank you!

opentelemetry/trace/main.go
tbpg: Defer looks good to me. Thank you for checking!

tbpg (approved):

PR #1823, opened 2020-11-10
fix(spanner): clients are not properly closed

hengfengli:
If clients are not closed properly, this can lead to an issue that sessions are not properly deleted.


tbpg (approved):
Thanks!

PR #1825, opened 2020-11-10
fix(pubsub): remove channel usage in concurrency sample

hongalex:
Originally brought up in #1753, the concurrency sample is misleading since a single unbuffered channel was being used to handle messages in a single goroutine. This is confusing and may mislead developers.

The sample now handles everything in the original callback function, with a note to make sure shared memory is synchronized.


mindprince (approved):

pubsub/subscriptions/subscription_test.go
```diff
@@ -299,10 +299,9 @@ func TestPullMsgsConcurrencyControl(t *testing.T) {
 	publishMsgs(ctx, topic, numMsgs)
 
 	buf := new(bytes.Buffer)
-	if err := pullMsgsConcurrenyControl(buf, tc.ProjectID, subIDConc); err != nil {
+	if err := pullMsgsConcurrenyControl(buf, tc.ProjectID, subIDConc); err != nil jj
 		t.Fatalf("failed to pull messages: %v", err)
```
mindprince, line 303: typo

hongalex: Fixed, thanks

tbpg (approved):

PR #1830, opened 2020-11-13
feat(kms): add integrity verification to cryptographic operations

iamtamjam:


kms/encrypt_symmetric.go
```diff
@@ -40,21 +41,41 @@ func encryptSymmetric(w io.Writer, name string, message string) error {
 	// Convert the message into bytes. Cryptographic plaintexts and
 	// ciphertexts are always byte arrays.
 	plaintext := []byte(message)
+	
+	// Optional, but recommended: Compute plaintext's CRC32C.
+	// See crc32c() function below.
+	plaintextCrc32c = crc32c(plaintext)
```
tbpg, line 47: Might be better to inline the `crc32c` function to keep the sample as simple as possible.

kms/encrypt_symmetric.go
```diff
@@ -40,21 +41,41 @@ func encryptSymmetric(w io.Writer, name string, message string) error {
...
 
 	// Call the API.
 	result, err := client.Encrypt(ctx, req)
 	if err != nil {
 		return fmt.Errorf("failed to encrypt: %v", err)
 	}
+	
+	// Optional, but recommended: perform integrity verification on result.
+        // For more details on ensuring E2E in-transit integrity to and from Cloud KMS visit:
+        // https://cloud.google.com/kms/docs/data-integrity-guidelines
```
tbpg, line 64: Looks like you need to run `gofmt -w .`.

kms/encrypt_symmetric.go
tbpg, line 47:
```suggestion
	plaintextCRC32C = crc32c(plaintext)
```

kms/encrypt_symmetric.go
```diff
@@ -40,21 +41,41 @@ func encryptSymmetric(w io.Writer, name string, message string) error {
...
+	if result.VerifiedPlaintextCrc32C == false {
+		return fmt.Errorf("The request sent to the server was corrupted in-transit.")
```
tbpg, line 66: > Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

```suggestion
		return fmt.Errorf("Encrypt: request corrupted in-transit")
```

`Encrypt` is helpful for debugging because that's the function that was called that led to this error. Otherwise, when looking at logs, it can be unclear where the error came from.

tbpg (changes requested):

kms/encrypt_symmetric.go
iamtamjam: Chatted offline - since the function is used twice in the snippet, we'll keep it as a standalone function.

kms/encrypt_symmetric.go
iamtamjam: Done.

kms/encrypt_symmetric.go
```diff
@@ -53,8 +59,24 @@ func encryptSymmetric(w io.Writer, name string, message string) error {
...
+	// End integrity verification
+
 	fmt.Fprintf(w, "Encrypted ciphertext: %s", result.Ciphertext)
 	return nil
 }
 
+func crc32c(data []byte) uint32 {
+	t := crc32.MakeTable(crc32.Castagnoli)
+	return crc32.Checksum(data, t)
+}
```
tbpg, line 80: This is leading to a compiler error because this function is declared here and in decrypt_symmetric.go. So, perhaps we should inline it. I think you can make a single `Table` then reuse it when calling `Checksum`?

tbpg (changes requested):

kms/encrypt_symmetric.go
iamtamjam: np. I made it into a nested function that lives only within the scope of the sample function. 
I prefer to leave the function intact, since I expect the user will have such a function in a shared util library, and so they can just copy past the function as is without needing the think about how to create one themselves.

kms/decrypt_asymmetric.go
```diff
@@ -18,10 +18,12 @@ package kms
 import (
 	"context"
 	"fmt"
+	"hash/crc32"
 	"io"
 
 	kms "cloud.google.com/go/kms/apiv1"
 	kmspb "google.golang.org/genproto/googleapis/cloud/kms/v1"
+	wrapperspb "google.golang.org/protobuf/types/known/wrapperspb"
```
tbpg, line 26:
```suggestion
	"google.golang.org/protobuf/types/known/wrapperspb"
```

No need to rename if the package name matches the final element of the import path.

kms/decrypt_asymmetric.go
```diff
@@ -37,17 +39,39 @@ func decryptAsymmetric(w io.Writer, name string, ciphertext []byte) error {
 		return fmt.Errorf("failed to create kms client: %v", err)
 	}
 
+	// Optional but recommended: Compute ciphertext's CRC32C.
+	crc32c := func(data []byte) int64 {
+		t := crc32.MakeTable(crc32.Castagnoli)
+		// Convert to int64 to match API's use of Int64Value wrapper for CRC32C fields
```
tbpg, line 45: I'm not sure we need this comment? And it might be better to handle the int64 conversion when building the request, rather than in this helper function?

kms/decrypt_asymmetric.go
```diff
@@ -37,17 +39,39 @@ func decryptAsymmetric(w io.Writer, name string, ciphertext []byte) error {
...
+	// Optional, but recommended: perform integrity verification on result.
+	// For more details on ensuring E2E in-transit integrity to and from Cloud KMS visit:
+	// https://cloud.google.com/kms/docs/data-integrity-guidelines
+	if result.VerifiedCiphertextCrc32C == false {
+		return fmt.Errorf("AsymmetricDecrypt: request corrupted in-transit")
+	}
+	if crc32c(result.Plaintext) != result.PlaintextCrc32C.Value {
+		return fmt.Errorf("AsymmetricDecrypt: response corrupted in-transit")
+	}
+	// End integrity verification
```
tbpg, line 73: Delete? Or add a period at the end.

kms/decrypt_asymmetric.go
iamtamjam: sg. will address in a followup commit.

iamtamjam: Then again, it seems that even for response verification the casting is required. To minimize casting-related clutter, I will cast once in the crc32c() function. 

See fails: https://source.cloud.google.com/results/invocations/19166235-a7aa-4377-9cd6-71767514ead5/targets/cloud-devrel%2Fgo%2Fgolang-samples%2Fpresubmit/log

iamtamjam: Never mind. It is clearer when done at the call site of crc32c, I think. :)

kms/decrypt_asymmetric.go
iamtamjam: Removed comment.

kms/decrypt_asymmetric.go
iamtamjam: Done.

tbpg (approved):

PR #1840, opened 2020-11-18
feat(asset): add samples for AnalyzeIamPolicy and AnalyzeIamPolicyLongrunning

donghez-google:


asset/quickstart/analyze-iam-policy-longrunning-bigquery/main.go
```diff
@@ -0,0 +1,72 @@
...
+				},
+			},
+		},
+	}
+
+	op, err := client.AnalyzeIamPolicyLongrunning(ctx, req)
+	if err != nil {
+		log.Fatalf("AnalyzeIamPolicyLongrunning: %v", err)
+	}
+	fmt.Print(op.Metadata())
```
tbpg, line 68: Should this wait for the operation to finish? Same with other samples.

asset/quickstart/analyze-iam-policy-longrunning-bigquery/main_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"cloud.google.com/go/bigquery"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestMain(t *testing.T) {
+	tc := testutil.SystemTest(t)
```
tbpg, line 29: How long is this test expected to take? If it's more than a few seconds, will you use `testutil.EndToEndTest` to make it so it only runs in one Go version and in nightly jobs? Same with other samples.

tbpg (changes requested):
Looks good! Just a couple comments.

asset/quickstart/analyze-iam-policy-longrunning-bigquery/main_test.go
donghez-google: It will take one second or two, so I guess we are ok.

asset/quickstart/analyze-iam-policy-longrunning-bigquery/main.go
donghez-google: I intended to not wait for that. Here we printed the longrunning operation metadata, and in the test we only validate the metadata.

donghez-google:
> Looks good! Just a couple comments.

asset/quickstart/analyze-iam-policy-longrunning-bigquery/main.go
```diff
@@ -66,6 +66,16 @@ func main() {
...
+
+	// Wait for the longrunning operation complete.
+	resp, err := op.Wait(ctx)
+	if err != nil && !op.Done() {
+		fmt.Println("failed to fetch operation status", err)
+	} else if err != nil && op.Done() {
+		fmt.Println("operation completed with error", err)
+	} else {
+		fmt.Println("operation completed successfully", resp)
+	}
```
tbpg, line 78: Might be good to switch these to returning early to keep indentation at a minimum for the happy path:

```suggestion
	if err != nil && !op.Done(){
		fmt.Println("failed to fetch operation status", err)
		return
	}
	if err != nil && op.Done(){
		fmt.Println("operation completed with error", err)
		return
	}
	fmt.Println("operation completed successfully", resp)
```

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

asset/quickstart/analyze-iam-policy-longrunning-bigquery/main.go
```diff
@@ -66,6 +66,16 @@ func main() {
...
 	fmt.Print(op.Metadata())
+
+	// Wait for the longrunning operation complete.
+	resp, err := op.Wait(ctx)
+	if err != nil && !op.Done() {
+		fmt.Println("failed to fetch operation status", err)
+	} else if err != nil && op.Done() {
+		fmt.Println("operation completed with error", err)
+	} else {
+		fmt.Println("operation completed successfully", resp)
```
tbpg, line 77: Is there something in the response we can print? We usually print a field or two, rather than the entire thing. Depends on the API.

tbpg (changes requested):

asset/quickstart/analyze-iam-policy-longrunning-bigquery/main.go
donghez-google: This response is an empty proto, all actual results are written into GCS or BQ, and useful metadata are in Operation.Metadata.

tbpg (approved):

PR #1850, opened 2020-11-30
feat(dataproc): new sample for job_submit + updating quickstart / create_cluster

bradmiro:
- Adding submit-job sample
- Updating quickstart to reflect API changes
- Update create_cluster docstring


dataproc/quickstart/quickstart_test.go
```diff
@@ -131,7 +131,7 @@ func deleteClusters(ctx context.Context, projectID string) error {
 }
 
 func TestQuickstart(t *testing.T) {
-	tc := testutil.EndToEndTest(t)
+	tc := testutil.SystemTest(t)
```
tbpg, line 134: Please keep this as an `EndToEndTest`. Those only run nightly one version of Go and not on pull requests. They are generally for tests that take a long time.

dataproc/submit_job.go
```diff
@@ -0,0 +1,105 @@
...
+						MainClass: "org.apache.spark.examples.SparkPi",
+					},
+					JarFileUris: []string{"file:///usr/lib/spark/examples/jars/spark-examples.jar"},
+					Args:        []string{"1000"},
+				},
+			},
+		},
+	}
+
+	// Create the job client.
```
tbpg, line 67: Wrong comment?

dataproc/submit_job.go
```diff
@@ -0,0 +1,105 @@
...
+
+	submitJobResp, err := submitJobOp.Wait(ctx)
+	if err != nil {
+		return fmt.Errorf("error submitting job: %v", err)
+	}
+
+	re := regexp.MustCompile("gs://(.*?)/(.*)")
+	matches := re.FindStringSubmatch(submitJobResp.DriverOutputResourceUri)
+
+	// Dataproc job outget gets saved to a GCS bucket allocated to it.
```
tbpg, line 81:
```suggestion
	// Dataproc job output gets saved to a GCS bucket allocated to it.
```

dataproc/submit_job_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"testing"
+	"time"
+
+	dataproc "cloud.google.com/go/dataproc/apiv1"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+	"google.golang.org/api/option"
+	dataprocpb "google.golang.org/genproto/googleapis/cloud/dataproc/v1"
+)
+
+func setupForSubmitJobTest(projectID, region, clusterName string) error {
```
tbpg, line 31: Can we set up a single cluster for all tests to avoid having to set up and tear down for each one? One way to do it would be to have a single test (something like `TestDataproc`) for `createCluster` and `submitJob` (which are called in that order).

tbpg (changes requested):

dataproc/submit_job_test.go
bradmiro: Both `createCluster` and `quickstart` create a cluster as part of the sample itself, not the test. I'd like to avoid co-dependent tests where possible.

dataproc/quickstart/quickstart.go
```diff
@@ -124,83 +124,30 @@ func main() {
...
-
-	resp, err := clusterClient.GetCluster(ctx, getCReq)
+	submitJobResp, err := submitJobOp.Wait(ctx)
 	if err != nil {
-		fmt.Printf("error getting cluster %q: %v\n", clusterName, err)
+		fmt.Printf("error submitting job: %v\n", err)
 		return
 	}
 
+	re := regexp.MustCompile("gs://(.*?)/(.*)")
```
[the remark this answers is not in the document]
bradmiro: The `?` allows the regex to capture the bucket and object separately. For `gs://a/b/c/d/e`, it's the difference between `a` + `b/c/d/e` with the `?` and 'a/b/c/d` + `e` without.

Changing to `+` makes sense.

dataproc/quickstart/quickstart_test.go
bradmiro: I thought I ran into an issue testing this with E2E but I can revert and try again.

bradmiro: So I tried to revert this and the test does not do anything locally as E2E. It will run as it's supposed to if it's a SystemTest.

dataproc/submit_job.go
bradmiro: Ack

dataproc/quickstart/quickstart.go
```diff
@@ -124,83 +124,30 @@ func main() {
...
+	// Dataproc job outget gets saved to a GCS bucket allocated to it.
 	storageClient, err := storage.NewClient(ctx)
 	if err != nil {
 		fmt.Printf("error creating storage client: %v\n", err)
 		return
 	}
 
-	obj := fmt.Sprintf("google-cloud-dataproc-metainfo/%s/jobs/%s/driveroutput.000000000", resp.ClusterUuid, id)
-	reader, err := storageClient.Bucket(resp.Config.ConfigBucket).Object(obj).NewReader(ctx)
+	obj := fmt.Sprintf("%s.000000000", matches[2])
```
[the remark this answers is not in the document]
bradmiro: Ack

dataproc/quickstart/quickstart_test.go
bradmiro: Ok, realized it was because `GOLANG_SAMPLES_E2E_TEST` wasn't set.

dataproc/submit_job_test.go
tbpg: We can leave out the quickstart, since it creates its own cluster. But, `createCluster` and `submitJob` can be tested together -- `createCluster` can create the cluster used by `submitJob`.

My motivation is keeping overall test runtime and flakiness as low as possible. Creating a cluster seems slow and subject to more flakiness.

tbpg (changes requested):

tbpg:
Also, looks like you need a `gofmt -w .` -- that's causing the Build and Lint failure.

dataproc/submit_job_test.go
bradmiro: It feels like we shouldn't be introducing ordered tests here, no? Alternatively, we can leave the `createCluster` test as-is, and future tests that aren't order-dependent can use a separate, shared cluster that is created via a setup function.

tbpg: Sorry, I don't mean ordered tests. I mean a single test that tests both samples.

bradmiro: Done

tbpg (approved):
Thanks!

PR #1854, opened 2020-12-04
chore(.github): disable x/tools updates

tbpg:
The latest changes are incompatible with Go 1.11.


codyoss (approved):
This is good to know. Is there a package in particular to watch out for? Will this be a concern on any other repos we own?

tbpg:
I'm not sure. Will hopefully pop up as a compiler issue if it happens in other repos.

PR #1859, opened 2020-12-09
testing(functions): pattern for complete system tests

### LINKED ISSUE #1690: functions: create facility to deploy Cloud Functions for e2e testing

grayside:
Fixes #1690 

This PR introduces function setup and teardown, using two examples: a simple case for HelloHTTP and a more complex two-function deployment for makeGetRequest.


ace-n (approved):

ace-n:
**Aside:** I think we're [missing](https://cloud.google.com/functions/docs/testing/test-http#functions-testing-http-example-go) an HTTP integration test for Go. (See b/154418159)

functions/security/security_system_test.go
```diff
@@ -16,22 +16,121 @@ package security
...
+	}
+	targetURL := strings.TrimSpace(string(out))
+	relayName := entryPoint + "-relay-" + time.Now().Format("20060102-150405")
+
+	// Setup function for tests.
+	cmd = exec.Command("gcloud", "functions", "deploy", relayName,
+		"--entry-point="+entryPoint,
+		"--runtime="+RuntimeVersion,
+		"--no-allow-unauthenticated",
+		"--update-env-vars", "TARGET="+targetURL,
```
tbpg, line 87: Should this be `TARGET_URL`?

functions/helloworld/hello_http_system_test.go
```diff
@@ -20,20 +20,82 @@
...
+	// Run the tests.
+	return m.Run(), nil
+}
+
 func TestHelloHTTPSystem(t *testing.T) {
 	client := http.Client{
 		Timeout: 10 * time.Second,
 	}
-	urlString := os.Getenv("BASE_URL") + "/HelloHTTP"
+	urlString := os.Getenv("BASE_URL")
```
tbpg, line 98: I think `BASE_URL` intentionally didn't include the name of the function. Maybe it shouldn't? But, we'd need to update the docs here: https://cloud.google.com/functions/docs/testing/test-http#functions-testing-http-system-cmd-go

functions/security/idtokenFixture.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"bytes"
+	"log"
+	"net/http"
+	"os"
+)
+
+// MakeGetRequestFixture is an HTTP Cloud Function to facilitate testing security snippets.
+func MakeGetRequestFixture(w http.ResponseWriter, r *http.Request) {
```
tbpg, line 25: What does `Fixture` mean here?

tbpg (changes requested):
Thank you!

functions/security/security_system_test.go
grayside: This test is deploying two Cloud Functions. This one is deploying the Cloud Function which will be tested.

Toward the end of setupAndRun() os.Setenv is called to set `BASE_URL` for the test. Otherwise this function doesn't use env vars for URLs.

functions/helloworld/hello_http_system_test.go
grayside: I'm a bit confused what our goal is here. I'm not putting this together as a sample of a system test, but an internal system test. I'm arranging it so the full URL comes from a gcloud command, rather than based on the user's understanding of the URL pattern.

functions/security/idtokenFixture.go
grayside: This file is only used to make the system test possible, it's a wrapper for the code in idtoken.go.

tbpg: Got it. Unfortunately, I don't know what `Fixture` means in this context.

functions/helloworld/hello_http_system_test.go
ace-n: +1 to @tbpg here. If we plan on updating any _public-facing_ samples in a similar manner, we'll have to update this for the [other languages](https://cloud.google.com/functions/docs/testing/test-http#system_tests) as well.

(@grayside if these samples are _not_ public-facing, perhaps we should use a different env. var name here - to avoid confusion?)

grayside: Resolved in discussion, going to back out this piece in the expectation that URL pattern changes will then break the tests.

grayside:
This is ready for new review.

functions/security/security_system_test.go
```diff
@@ -34,99 +34,83 @@ func TestMain(m *testing.M) {
 		os.Exit(m.Run())
 	}
 
+	if os.Getenv("GOLANG_SAMPLES_PROJECT_ID") == "" {
+		log.Println("Stopping test execution: GOLANG_SAMPLES_PROJECT_ID not set")
+		os.Exit(1)
```
tbpg, line 39: We usually just skip, so this would be exit 0.

tbpg (approved):
One comment about exit code, otherwise LGTM.

functions/security/security_system_test.go
grayside: done.

PR #1862, opened 2020-12-10
feat(bigquery): add hive partitioning table sample

shollyman:


bigquery/snippets/table/bigquery_create_table_external_hivepartitioned.go
```diff
@@ -0,0 +1,71 @@
...
+		ExternalDataConfig: &bigquery.ExternalDataConfig{
+			SourceFormat: bigquery.Parquet,
+			SourceURIs:   []string{"gs://cloud-samples-data/bigquery/hive-partitioning-samples/autolayout/*"},
+			AutoDetect:   true,
+		},
+	}
+
+	// The layout of the files in here is compatible with the layout requirements for hive partitioning,
+	// so we can add an optional Hive partitioning configuration to leverage the object paths for deriving
+	// partitioning column information.  We'll also require queries against this table to specify a partitioning
```
tswast, line 55: Do we have examples of such queries? It's not clear with all the autodetection what the column names are.

shollyman: That's the interesting bit.  Should we actually perform a predicate query in the same sample, or just add a comment of an example?

tswast (approved):
Much clearer, thanks!

PR #1867, opened 2020-12-11
docs(securitycenter): wrap samples with future prefix

Strykrol:
Standardizing Security Command Center samples to use 'securitycenter' prefixing. 

- Wrapped existing samples to keep published doclinks unbroken
- Replace region tags that aren't published. 

Once this PR is through, published sample inclusions will be updated to use the new prefix, then I'll come through again and remove the unused block wraps.

### DIFF
--- a/securitycenter/assets/add_delete_security_marks.go
+++ b/securitycenter/assets/add_delete_security_marks.go
@@ -14,6 +14,7 @@
 
 package assets
 
+// [START securitycenter_add_delete_security_marks]
 // [START add_delete_security_marks]
 import (
 	"context"
@@ -63,3 +64,4 @@ func addDeleteSecurityMarks(w io.Writer, assetName string) error {
 }
 
 // [END add_delete_security_marks]
+// [END securitycenter_add_delete_security_marks]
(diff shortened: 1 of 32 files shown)


securitycenter/assets/add_delete_security_marks.go
```diff
@@ -14,6 +14,7 @@
 
 package assets
 
+// [START securitycenter_add_delete_security_marks]
```
tbpg, line 17: Is this the right prefix? The snippet-bot doesn't seem to like it. Should it be `scc`?

tbpg (changes requested):

securitycenter/assets/add_delete_security_marks.go
Strykrol: Yes it's correct. Once all offending snippets are wrapped, I'll be moving the prefix to `securitycenter` (in app), then fix doc includes to reference the new prefix. After all of that is submitted, the old "inside" tags can be removed. Snippet bot will not like any of these for now.

tbpg (approved):

PR #1872, opened 2020-12-16
testing(cloudsql): added integration tests for all 3 Cloud SQL flavors

craiglabenz:
Need to successfully integrate with Kokoro, but these tests are passing locally in all flavor/connection type permutations


cloudsql/mysql/database-sql/cloudsql.go
```diff
@@ -84,6 +84,21 @@ func (app *app) indexHandler(w http.ResponseWriter, r *http.Request) {
...
+		port = "8080"
+	}
+
+	log.Printf("Listening on port %s", port)
+	if err := http.ListenAndServe(":"+port, nil); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func startApp() app {
```
tbpg, line 101: Maybe `newApp` since `start` sounds like it will start listening?

tbpg, line 101: Return `*app` instead, since that's what the other functions work with?

cloudsql/mysql/database-sql/cloudsql_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"bytes"
+	"io/ioutil"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestIndex(t *testing.T) {
+	app := startApp()
+	writer := httptest.NewRecorder()
```
tbpg, line 27: This is usually called `rr` for response recorder, from what I've seen.

cloudsql/mysql/database-sql/cloudsql_test.go
```diff
@@ -0,0 +1,56 @@
...
+	request := httptest.NewRequest("GET", "/", nil)
+	app.indexHandler(writer, request)
+	resp := writer.Result()
+	body, _ := ioutil.ReadAll(resp.Body)
```
tbpg, line 31:
```suggestion
	body := writer.Body.String()
```

cloudsql/mysql/database-sql/cloudsql_test.go
```diff
@@ -0,0 +1,56 @@
...
+	request.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	app.indexHandler(writer, request)
+	resp := writer.Result()
+	body, _ := ioutil.ReadAll(resp.Body)
+
+	if resp.StatusCode != 200 {
+		t.Errorf("Expected StatusCode of 200. Got %d", resp.StatusCode)
+	}
+	if !strings.Contains(string(body[:]), "Vote successfully cast for SPACES") {
+		t.Errorf("Expected to see 'Vote successfully cast for SPACES' in response body")
```
tbpg, line 54: Consider creating a `want` variable and using `%q` to print it in the error message.

tbpg (changes requested):
Let me know if you need anything for Kokoro setup.

cloudsql/postgres/database-sql/run_tests
```diff
@@ -0,0 +1,9 @@
+#! /bin/sh
+
+# Run the tests with default values for all
+# environment variable
+go test -timeout $TIMEOUT -v "${1:-./...}" | tee sponge_log.log
+
+# Run the tests again with custom environment
+# variables that force a Unix connection
+DB_HOST=; go test -timeout $TIMEOUT -v "${1:-./...}" | tee sponge_log.log
```
tbpg, line 9: Go tests can set environment variables before calling whatever method is under test. In particular, you can unset `DB_HOST` and run the test, directly from the Go code.

In general, I'm a fan of as little Bash as possible. I'd also like to keep all of the `go test` calls in the same `system_test.sh` script so it's easier to keep them all consistent.

cloudsql/mysql/database-sql/cloudsql_test.go
```diff
@@ -0,0 +1,59 @@
...
+func TestIndex(t *testing.T) {
+	app := newApp()
+	rr := httptest.NewRecorder()
+	request := httptest.NewRequest("GET", "/", nil)
+	app.indexHandler(rr, request)
+	resp := rr.Result()
+	body := rr.Body.String()
+
+	if resp.StatusCode != 200 {
+		t.Errorf("Expected StatusCode of 200. Got %d", resp.StatusCode)
```
tbpg, line 33: Please print got and want: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

cloudsql/mysql/database-sql/cloudsql_test.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	want := "Tabs VS Spaces"
+	if !strings.Contains(string(body[:]), want) {
```
tbpg, line 37:
```suggestion
	if !strings.Contains(body, want) {
```

`body` is already a string.

testing/kokoro/system_tests.sh
```diff
@@ -131,6 +131,17 @@ if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"system-tests"* ]]; then
   ./testing/kokoro/configure_gcloud.bash;
 fi
 
+# Download and prepare Cloud SQL Proxy
```
tbpg, line 134: Please call `testutil.EndToEndTest(t)` in the test cases, and only configure the Cloud SQL Proxy when running E2E tests (see a few lines above).

Might make sense to have a `configure_cloudsql.sh` file to match `configure_gcloud.sh`? Or perhaps rename `configure_gcloud.sh` to `configure_e2e.sh`?

I don't think we need `gcloud` for any non-E2E tests.

testing/kokoro/system_tests.sh
```diff
@@ -131,6 +131,17 @@ if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"system-tests"* ]]; then
...
+wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64
+mv cloud_sql_proxy.linux.amd64 /cloud_sql_proxy
+chmod +x /cloud_sql_proxy
+mkdir /cloudsql && chmod 0777 /cloudsql
+
+# TODO: These variable names need to be populated.
+cloud_sql_proxy -instances="${MYSQL_INSTANCE}"=tcp:3306,${MYSQL_INSTANCE} -dir /cloudsql
+cloud_sql_proxy -instances="${POSTGRES_INSTANCE}"=tcp:5432,${POSTGRES_INSTANCE} -dir /cloudsql
+cloud_sql_proxy -instances="${SQLSERVER_INSTANCE}"=tcp:1433
```
tbpg, line 143: Need the full path to `cloud_sql_proxy`.

tbpg (changes requested):

cloudsql/postgres/database-sql/run_tests
craiglabenz: I share your preference for minimizing bash and taking care of everything in Go. My initial thinking here was to solve two problems with the introduction of this script:

1) Unsetting the environment variable, which can move to the testing layer, but also
2) Run these tests twice, once with the variable set and once without.

Do you have any recommendations on how to preserve that second goal?

cloudsql/mysql/database-sql/cloudsql_test.go
craiglabenz: Are you looking to print "200" from a variable, or are you looking for a specific structure of the output, aka, exactly `"StatusCode == 500, want 200"'`?

cloudsql/postgres/database-sql/run_tests
tbpg: That's usually done with [table driven tests](https://github.com/golang/go/wiki/TableDrivenTests) in Go. Something like this:

```
tests := []struct{
  dbHost string
}{
  {dbHost: "8.8.8.8"},
  {},
}

for _, test := range tests {
  oldDBHost := os.Getenv("DB_HOST")
  os.Setenv(test.dbHost)
  // ...
  os.Setenv(oldDBHost)
}
```

cloudsql/mysql/database-sql/cloudsql_test.go
tbpg: Yeah, got before want. Sorry that was unclear. Something like `indexHandler got status code %d, want 200`.

tbpg (commented):
Sorry for the slow response!

cloudsql/mysql/database-sql/cloudsql_test.go
craiglabenz: This has been incorporated.

cloudsql/mysql/database-sql/cloudsql_test.go
```diff
@@ -0,0 +1,63 @@
...
+	resp := rr.Result()
+	body := rr.Body.String()
+
+	if resp.StatusCode != 200 {
+		t.Errorf("indexHandler got status code %d, want 200", resp.StatusCode)
+	}
+
+	want := "Tabs VS Spaces"
+	if !strings.Contains(body, want) {
+		t.Errorf("Expected to see '%s' in indexHandler response body", want)
```
tbpg, line 41:
```suggestion
		t.Errorf("indexHandler got response body %q, want to contain %q", body, want)
```

testing/kokoro/system_tests.sh
```diff
@@ -166,7 +167,11 @@ runTests() {
   set +x
   echo "Running 'go test' in '$(pwd)'..."
   set -x
-  2>&1 go test -timeout $TIMEOUT -v "${1:-./...}" | tee sponge_log.log
+  if [[ -f "run_tests" ]]; then
```
tbpg, line 170: Looks like the other comment got lost on a previous commit, so I can't seem to continue that conversation. But, please update this to be done purely in the Go test. You can configure environment variables with `os.Getenv`, `os.Setenv`, and `os.Unsetenv`. You can do that directly or using a table driven test (https://github.com/golang/go/wiki/TableDrivenTests).

tbpg (changes requested):

testing/kokoro/system_tests.sh
craiglabenz: Thanks for this recommendation. I've just implemented and pushed the change :)

craiglabenz:
@tbpg Am I reading the output of that test failure correctly that unrelated tests are failing?

tbpg:
> @tbpg Am I reading the output of that test failure correctly that unrelated tests are failing?

Yes. Sorry about that!

cloudsql/sqlserver/database-sql/templates/index.html
```diff
@@ -60,7 +60,7 @@ <h4 class="header center">Recent Votes</h4>
     <ul class="container collection center">
         {{ range .RecentVotes }}
             <li class="collection-item avatar">
-                {{ if eq .Candidate "TABS" }}
+                {{ if eq  .Candidate "TABS  " }}
```
tbpg, line 63: Intentional change?

cloudsql/mysql/database-sql/cloudsql_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestIndex(t *testing.T) {
+	testutil.EndToEndTest(t)
+	tcpApp := createTCPApp()
+	loadIndex(tcpApp, t)
+	unixApp := createUnixApp()
+	loadIndex(unixApp, t)
+}
```
tbpg, line 33: The standard way to do this in Go is using a table driven test (https://github.com/golang/go/wiki/TableDrivenTests). See https://github.com/golang/go/wiki/TestComments#table-driven-tests-vs-multiple-test-functions for a longer form comparison to multiple test functions.

The goal is to make it easy to follow exactly what is being tested and how, including making it easy to debug.

In this case, it would look something like this:

```go
func testIndex(t *testing.T) {
	testutil.endToEndTest(t)
	tests := []struct{
		dbHost string
	}{
		{dbHost: ""},
		{dbHost: os.Getenv("DB_HOST")},
	}
	
	for _, test := range tests {
		oldDBHost := os.Getenv("DB_HOST")
		os.Setenv("DB_HOST", test.dbHost)
		
		a := newApp()
		rr := httptest.NewRecorder()
		// ...
		// Make sure to include the value of test.dbHost in any test failures.
		// Also see https://github.com/golang/go/wiki/TestComments#identify-the-function and Identify the Input right after.
		os.Setenv("DB_HOST", oldDBHost)
	}
}
```

cloudsql/mysql/database-sql/cloudsql_test.go
```diff
@@ -0,0 +1,89 @@
...
+package main
+
+import (
+	"bytes"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
```
tbpg, line 24: This causes the golang-samples requirement to be added to the `go.mod` files. We've generally avoided that for samples like this, which get deployed on their own and we don't _really_ need them to depend on `golang-samples`. But, it's up to you.

To avoid the dependency, you would just inline the logic of checking if you should skip the test.

tbpg (changes requested):

cloudsql/sqlserver/database-sql/templates/index.html
craiglabenz: Unfortunately, yes. The SQLServer column used here mandates _exact_ character length (as opposed to a maximum length), so it pads data with white space 😭

cloudsql/mysql/database-sql/cloudsql_test.go
craiglabenz: Done.

testing/kokoro/configure_cloudsql.bash
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
tbpg, line 1: Would you mind adding a license header at the top of this file?

testing/kokoro/configure_cloudsql.bash
```diff
@@ -0,0 +1,14 @@
...
+# Download and prepare Cloud SQL Proxy
+wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64
+mv cloud_sql_proxy.linux.amd64 /cloud_sql_proxy
+chmod +x /cloud_sql_proxy
+mkdir /cloudsql && chmod 0777 /cloudsql
+
+# TODO: These variable names need to be populated.
+/cloud_sql_proxy -instances="${MYSQL_INSTANCE}"=tcp:3306,${MYSQL_INSTANCE} -dir /cloudsql
+/cloud_sql_proxy -instances="${POSTGRES_INSTANCE}"=tcp:5432,${POSTGRES_INSTANCE} -dir /cloudsql
+/cloud_sql_proxy -instances="${SQLSERVER_INSTANCE}"=tcp:1433
```
tbpg, line 14: Nit: and a newline at the bottom? :smile:

testing/kokoro/configure_cloudsql.bash
```diff
@@ -0,0 +1,14 @@
...
+
+set -ex
+
+# Download and prepare Cloud SQL Proxy
+wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64
+mv cloud_sql_proxy.linux.amd64 /cloud_sql_proxy
+chmod +x /cloud_sql_proxy
+mkdir /cloudsql && chmod 0777 /cloudsql
+
+# TODO: These variable names need to be populated.
```
tbpg, line 11: Is this still a TODO?

tbpg (commented):
A few minor comments on the proxy setup. Overall, looks good!

testing/kokoro/configure_cloudsql.bash
craiglabenz: Nope!

tbpg (approved):

PR #1878, opened 2020-12-18
chore(speech): quickstart now uses a gcs file

b-loved-dreamer:
Updated the sample to now use a gcs uri instead of a local resource file


speech/speech_quickstart/main.go
```diff
@@ -38,13 +37,7 @@ func main() {
 	}
 
 	// Sets the name of the audio file to transcribe.
```
tbpg, line 39: Probably need to update this comment as it's not really the name of a file.

speech/speech_quickstart/main.go
```diff
@@ -38,13 +37,7 @@ func main() {
...
-	filename := "/path/to/audio.raw"
-
-	// Reads the audio file into memory.
-	data, err := ioutil.ReadFile(filename)
-	if err != nil {
-		log.Fatalf("Failed to read file: %v", err)
-	}
+	gcsURI := "gs://cloud-samples-data/speech/brooklyn_bridge.raw"
```
tbpg, line 40: I think the GCS acronym isn't used for branding. Maybe `uri` or `fileURI`?

tbpg (changes requested):
Thank you!

tbpg (approved):
Thanks!

PR #1882, opened 2020-12-24
run: helloworld fix

ali2210:
iter := client.Collection("users").Documents(ctx)
for {
        doc, err := iter.Next()
        if err == iterator.Done {
                break
        }
        if err != nil {
                log.Fatalf("Failed to iterate: %v", err)
        }
        fmt.Println(doc.Data())
}
question is why iterator.Done() not iter?


tbpg:
`iterator` is a package. You can read more about iterators in this project at https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines. I hope that helps!

PR #1886, opened 2020-12-31
feat: add cloud workflows quickstart

grant:
Inits Cloud Workflows API quickstart sample.

- Executes the Cloud Workflows quickstart (`myfirstworkflow`) using the Workflows client library
- Awaits the results with exponential backoff
- Includes test

```sh
GOLANG_SAMPLES_PROJECT_ID=my-project go test
```


grant:
Looks like the Workflows API needs to be enabled for the test runner project:

https://console.developers.google.com/apis/api/workflowexecutions.googleapis.com/overview?project=167833236165

workflows/quickstart/execute_workflow.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+	"log"
+	"os"
+	"time"
+
+	executions "cloud.google.com/go/workflows/executions/apiv1beta"
+	executionspb "google.golang.org/genproto/googleapis/cloud/workflows/executions/v1beta"
+)
+
+// [START workflows_api_quickstart]
```
tbpg, line 28: Is this intended to be a quickstart or a snippet? Quickstarts should have a `func main`, meaning they can be run as a standalone binary. See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#only-quickstarts-have-package-main.

I see you have a `func main` in another file -- that is ordinarily part of the sample. Otherwise, we try to avoid having that code at all.

tbpg, line 28: Please move the region tag above the imports, and above the package declaration if this is a quickstart. See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#include-imports-and-flags-in-region-tags.

workflows/quickstart/execute_workflow.go
```diff
@@ -0,0 +1,86 @@
...
+
+// executeWorkflow executes a workflow and returns the results from the workflow.
+func executeWorkflow(projectID, locationID, workflowID string) (string, error) {
+	ctx := context.Background()
+
+	// Creates a client.
+	client, err := executions.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("Failed to create client: %v", err)
+		return "", err
```
tbpg, line 38: `log.Fatal` exists immediately and should only be used in a `func main`. https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#return-errors is relevant.

workflows/quickstart/execute_workflow.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	name := exe.GetName()
+	fmt.Fprintf(os.Stdout, "Created execution: %v\n", name)
+
+	// Wait for execution to finish, then print results.
+	finished := false
+	backoffDelay := 1 * time.Second // Start wait with delay of 1s
+	fmt.Println("Poll every second for result...")
```
tbpg, line 60: Delete or update with actual delay. Perhaps log a message right before calling `time.Sleep`?

workflows/quickstart/execute_workflow.go
```diff
@@ -0,0 +1,86 @@
...
+	for !finished {
```
tbpg, line 61: What is the goal with the `finished` variable? It appears this will loop until the execution is finished, then return. So, I don't think we need `finished`.

Should there be some sort of timeout?

workflows/quickstart/execute_workflow.go
```diff
@@ -0,0 +1,86 @@
...
+
+		// If we haven't seen the result yet, wait a second.
+		if !finished {
+			fmt.Println("- Waiting for results...")
+			time.Sleep(backoffDelay)
+			backoffDelay *= 2 // Double the delay to provide exponential backoff.
+		} else {
+			fmt.Printf("Execution finished with state: %v\n", e.State)
+			return e.Result, nil
+		}
```
tbpg, line 79: I wonder if we should reverse this logic? If the execution is finished, return the result. Otherwise, sleep for `backoffDelay` and go on with the next loop. We wouldn't need the `else` and it would be very clear when the sleep happens (at the end of the loop iteration if we haven't returned).

workflows/quickstart/execute_workflow.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Fatalf("Failed to start workflow execution: %v", err)
+		return "", err
+	}
+
+	name := exe.GetName()
+	fmt.Fprintf(os.Stdout, "Created execution: %v\n", name)
+
+	// Wait for execution to finish, then print results.
+	finished := false
+	backoffDelay := 1 * time.Second // Start wait with delay of 1s
```
tbpg, line 59: Nit: please add a period at the end.

workflows/quickstart/execute_workflow_test.go
```diff
@@ -0,0 +1,31 @@
...
+package main
+
+import (
+	"testing"
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestExecuteWorkflow(t *testing.T) {
+	tc := testutil.SystemTest(t)
```
tbpg, line 24: How long do you expect the test to take? If it's more than a few seconds, please make this `EndToEndTest` so it only runs during the nightly CI runs and not for PRs.

workflows/quickstart/main.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	res, err := executeWorkflow(projectID, locationID, workflowID)
+	if err != nil {
+		log.Fatalf("executeWorkflow: %v", err)
+	}
+	var jsonStringArr []string
+	err = json.Unmarshal([]byte(res), &jsonStringArr)
+
+	fmt.Print(jsonStringArr)
```
tbpg, line 46: Can we print directly? Do we need to parse?

workflows/quickstart/main.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+)
+
+func main() {
+	if len(os.Args) < 3 {
+		fmt.Println("Usage: Must include 2 arguments for projectID, locationID")
```
tbpg, line 29: And an optional third arg for workflow ID?

tbpg (changes requested):

tbpg (changes requested):
I enabled the API for the test projects.

.gitignore
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
grant: Oops. Not sure how that got in.

workflows/quickstart/execute_workflow.go
grant: Quickstart.

Taken from https://github.com/GoogleCloudPlatform/golang-samples/tree/master/tasks

I can move the contents to main.

grant: Done.

workflows/quickstart/execute_workflow.go
grant: Removed/refactored.

workflows/quickstart/execute_workflow.go
grant: Done everywhere.

workflows/quickstart/execute_workflow.go
grant: Updated to simply inform the user that polling is happening.

workflows/quickstart/execute_workflow.go
grant: Thanks. I refactored the block to remove the `finished` variable.

Re timeout: The Workflow product has a timeout, making an additional timeout here unnecessary.

workflows/quickstart/execute_workflow.go
grant: I believe the updated logic with removed `finished` variable reads well and makes logical sense in terms of reading the program linearly.

workflows/quickstart/execute_workflow_test.go
grant: It looks like the workflow takes on average 0.3 seconds, so the program should probably finish in 1 or 3 seconds.

This is a very lightweight test that should be fast, so I think running on PRs is fine.

workflows/quickstart/main.go
grant: Yes, that is already the logic. 2 required fields, 1 optional field.

workflows/quickstart/main.go
grant: We could, but I believe it would be more convenient for the user to see an example of how to unmarshal a Workflow execution result (a string array here). So I'd prefer keeping this sample code in.

grant (commented):
Thanks for the quick review! I've addressed the great comments. PTAL with the latest commit.

workflows/quickstart/main.go
```diff
@@ -0,0 +1,104 @@
...
+	if workflowID == "" {
+		workflowID = "myFirstWorkflow"
+	}
+	workflowPath := fmt.Sprintf("projects/%s/locations/%s/workflows/%s", projectID, locationID, workflowID)
+
+	exe, err := client.CreateExecution(ctx, &executionspb.CreateExecutionRequest{
+		Parent: workflowPath,
+	})
+	if err != nil {
+		return "", err
```
grayside, line 53:
```suggestion
		return "", fmt.Errorf("client.CreateExecution: %w", err)
```

workflows/quickstart/main.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Wait for execution to finish, then print results.
+	backoffDelay := 1 * time.Second // Start wait with delay of 1s.
+	fmt.Println("Poll for result...")
+	for {
+		e, err := client.GetExecution(ctx, &executionspb.GetExecutionRequest{
+			Name: name,
+		})
+		if err != nil {
+			return "", err
```
grayside, line 67:
```suggestion
			return "", fmt.Errorf("client.GetExecution: %w", err)
```

workflows/quickstart/main.go
```diff
@@ -0,0 +1,104 @@
...
+
+	exe, err := client.CreateExecution(ctx, &executionspb.CreateExecutionRequest{
+		Parent: workflowPath,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	name := exe.GetName()
+	fmt.Fprintf(os.Stdout, "Created execution: %v\n", name)
```
grayside, line 57: nit: specifying stdout seems a bit verbose, we usually use `fmt.Printf`.

workflows/quickstart/main.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Wait for execution to finish, then print results.
+	backoffDelay := 1 * time.Second // Start wait with delay of 1s.
+	fmt.Println("Poll for result...")
+	for {
+		e, err := client.GetExecution(ctx, &executionspb.GetExecutionRequest{
```
grayside, line 63: nit: The variables in this sample include exe, e, and err. That's a bit confusing to follow. Maybe this should overwrite the initial return of the execution since we don't need it anymore?

```suggestion
		exe, err = client.GetExecution(ctx, &executionspb.GetExecutionRequest{
```

workflows/quickstart/main.go
grayside: nit: It's a bit confusing to read unmarshalling a byte slice to a string slice. A comment explaining that the expected result of this specific workflow is a list of strings.

grayside (commented):
I have several suggestions and nit feedbacks that I think would tighten this quickstart up a little, but I don't think as a whole they should necessarily block.

grant (commented):
Will work on tests and re-request review. Removed current requests for review.

grant:
I'm trying to debug an error calling an API (one API works. Seeing error:

## Code

```
_, err = client.CreateWorkflow(ctx, &workflowspb.CreateWorkflowRequest{...})
```

## Current Error

```
golang-samples/workflows/quickstart/main_test.go:37: deployWorkflow(deploy): workflows.CreateWorkflow: rpc error: code = InvalidArgument desc = Malformed collection name: 'workflows/workflows'
```

Shortened:

```
"Malformed collection name: 'workflows/workflows'"
```

I'm not sure where this collection name is coming from, as the variables for the API are set properly. The only similar variable is the `workflowPath`, which is longer than this collection name.

## Node equivalent

https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/workflows/quickstart/test/quickstart.test.js#L56

Any tips for debugging?

---

### Internal notes

```
err := deployWorkflow("serverless-com-demo", locationID, workflowName)
```

tbpg:
@grant I looked through and didn't see anything obvious that would be causing that error message... Did you get it figured out?

grant:
> @grant I looked through and didn't see anything obvious that would be causing that error message... Did you get it figured out?

I'm trying to figure it out with Adam. He's getting the same error locally, I'm not. The error message isn't easy to debug.

grant:
Going to close for now. Hopefully someone can pick this up at a later point.

PR #1888, opened 2021-01-08
fix: relax https:// requirement in license check

ahmetb:
The Apache 2.0 license itself (https://www.apache.org/licenses/LICENSE-2.0)
shows http:// to apply the license.

Therefore our own tooling (https://github.com/google/addlicense) also uses
this http:// string in the boilerplate.

I'm sending a patch to relax this requirement


tbpg (changes requested):
I think I'd prefer to stay consistent within this repo. So, either we switch everything to `http` or leave it all as `https`.

PR #1892, opened 2021-01-12
feat: update service directory samples to v1

katiehochberg:
As our GA launch is approaching, we want to change our samples that populate our documentation to use our v1 api.


tbpg (approved):
Thanks!

PR #1896, opened 2021-01-14
fix: Update opentelemetry version to v0.15.0

dashpole:
This should fix documentation: https://cloud.google.com/trace/docs/setup/go-ot, which references this example.  

A user following the documentation would `go get`s opentelemetry dependencies, it fetches v0.15.0+, which don't contain the `go.opentelemetry.io/otel/api/global` module.  It would fail to find the reference, and the user wouldn't know why it didn't work.


tbpg (approved):
Thank you!

PR #1911, opened 2021-01-21
test(run): increase gcloud retry attempts from 3 to 5

grayside:
#1903 and friends had build failures after 3 attempts, though surrounding builds were fine and as far as I can tell the build should have been fine. To reduce flakiness, trying to boost the retries.

Since we won't know if this is a fix without seeing if there are more flakes, I propose making the change and waiting for three nightly testing cycles to see if things are more robust. At that point we can close the issues.


internal/cloudrunci/gcloud.go
```diff
@@ -37,12 +37,13 @@ func init() {
...
 // gcloud provides a common mechanism for executing gcloud commands.
-// It will attempt to retry failed commands 3 times with 2 second wait intervals.
+// It will attempt to retry failed commands. Use gcloudWithoutRetry() for no retry.
 func gcloud(label string, cmd *exec.Cmd) ([]byte, error) {
 	var out []byte
 	var err error
 
-	success := testutil.RetryWithoutTest(3, 2*time.Second, func(r *testutil.R) {
+	maxAttempts := 5
+	success := testutil.RetryWithoutTest(maxAttempts, 2*time.Second, func(r *testutil.R) {
```
averikitsch, line 46: It might be good to also increase the sleep time to better offset attempts.

grayside: Are you aware of any particular time interval used in other repos?

averikitsch: No. This kicks off Cloud Builds right? I just assume Cloud Builds take longer than other retry requests we do for testing, so if you kick off 11 builds and 1 fails because of the quota, it may keep retrying while the other 10 builds are still running.

grayside: Cloud Build is just one of several gcloud operations taken. In this repo, Cloud Build is only used to build the container image.

What size interval do you think is needed? Is going up to 3 sufficient or do you think we need to create much more space for another build to complete?

averikitsch: If it's just building the docker container then maybe this isn't a concern.

tbpg (approved):

tbpg (approved):

grayside:
Unrelated test failure:
```
=== RUN   TestMain
--- FAIL: TestMain (123.83s)
    main_test.go:36: Bucket.Delete("golang-samples-tests-for-assets"): googleapi: Error 409: The bucket you tried to delete was not empty., conflict
    retry.go:44: FAILED after 10 attempts:
        storage_cleaner.go:45: Bucket.Create("golang-samples-tests-for-assets"): googleapi: Error 409: You already own this bucket. Please select another name., conflict
FAIL
FAIL	github.com/GoogleCloudPlatform/golang-samples/asset/quickstart/export-assets	123.844s
```

PR #1916, opened 2021-01-25
fix(storage): clarify makePublic sample

tritone:
This sample should include the correct values for entity and role
rather than omitting them.


tbpg (approved):
Makes sense to me! Thanks!

frankyn (approved):

codyoss (approved):

codyoss (approved):

PR #1921, opened 2021-01-28
feat: add kuberun eventing samples

eclipselu:


kuberun/eventing/generic/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"strings"
+)
+
+// GenericHandler receives and echos a HTTP request's headers and body.
+func GenericHandler(w http.ResponseWriter, r *http.Request) {
```
tbpg, line 30: Perhaps `EchoHandler` or just `Echo`? Generic is going to mean something else in Go relatively soon. :smiley:

kuberun/eventing/generic/main.go
```diff
@@ -0,0 +1,73 @@
...
+	bodyBytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		log.Printf("error parsing body: %v", err)
+	}
+	body := string(bodyBytes)
+	log.Println(body)
+
+	// send empty reply response
+	w.WriteHeader(http.StatusNoContent)
+	w.Write([]byte(""))
```
tbpg, line 55: Echo to the request back?

tbpg (changes requested):
Overall, looks good to me! How does this relate to the existing eventarc samples (https://github.com/GoogleCloudPlatform/golang-samples/tree/master/eventarc)? It seems they are very similar?

kuberun/eventing/generic/main.go
eclipselu: Thanks, I changed to EventHandler, also see my comment below.

kuberun/eventing/generic/main.go
eclipselu: thanks! We choose to return empty response for now due to our current plan, will change these examples to use cloudevent and google-cloudevents-sdk once they are ready to use.

tbpg:
@eclipselu how does this relate to the existing eventarc samples (https://github.com/GoogleCloudPlatform/golang-samples/tree/master/eventarc)?

eclipselu:
> @eclipselu how does this relate to the existing eventarc samples (https://github.com/GoogleCloudPlatform/golang-samples/tree/master/eventarc)?

sorry for replying late. 
currently in [Events for Cloud Run](https://cloud.google.com/kuberun/docs/events/anthos/quickstart?hl=en) it's referring to eventarc examples, we decided to change this because:

1. Events for Cloud Run is a different product, we want to decouple from using the same folder. 
2. Currently eventarc examples do not work in Events for Cloud Run.

eclipselu:
Removed region tags as we don't need them for now.

kuberun/eventing/audit-pubsub/main.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// HelloAuditLogEventHandler receives and processes a Pub/Sub message via a CloudEvent.
+func HelloAuditLogEventHandler(w http.ResponseWriter, r *http.Request) {
+	s := fmt.Sprintf("Created Pub/Sub topic: %s", string(r.Header.Get("Ce-Subject")))
+	log.Printf(s)
+
+	// send empty reply response
+	w.WriteHeader(http.StatusNoContent)
+	w.Write([]byte(""))
```
tbpg, line 32: I don't think the calls to write an empty string are needed? I think that's the same as not having a call at all.

kuberun/eventing/audit-pubsub/main.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/http"
+	"os"
+)
+
+// HelloAuditLogEventHandler receives and processes a Pub/Sub message via a CloudEvent.
+func HelloAuditLogEventHandler(w http.ResponseWriter, r *http.Request) {
+	s := fmt.Sprintf("Created Pub/Sub topic: %s", string(r.Header.Get("Ce-Subject")))
+	log.Printf(s)
+
+	// send empty reply response
```
tbpg, line 30:
```suggestion
	// Send empty reply response.
```

kuberun/eventing/generic/main.go
```diff
@@ -0,0 +1,68 @@
...
+	"os"
+	"strings"
+)
+
+func EventHandler(w http.ResponseWriter, r *http.Request) {
+	log.Println("Event received!")
+
+	// Log all headers besides authorization header
+	log.Println("HEADERS:")
+	headerMap := make(map[string]string)
```
tbpg, line 32: Doesn't look like `headerMap` is ever used. Delete?

kuberun/eventing/generic/main.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	// Log body
+	log.Println("BODY:")
+	bodyBytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		log.Printf("error parsing body: %v", err)
+	}
+	body := string(bodyBytes)
+	log.Println(body)
```
tbpg, line 48: No need for the `body` variable if we aren't sending it in the response. So, rename `bodyBytes` to `body` and do `log.Println(string(body))`?

kuberun/eventing/generic/main.go
```diff
@@ -0,0 +1,68 @@
...
+			headerMap[k] = val
+			log.Println(fmt.Sprintf("%q: %q\n", k, val))
+		}
+	}
+
+	// Log body
+	log.Println("BODY:")
+	bodyBytes, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		log.Printf("error parsing body: %v", err)
```
tbpg, line 45: Should this return or set a response code?

kuberun/eventing/generic/main_test.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil {
+			t.Fatalf("ReadAll: %v", err)
+		}
+		got := string(out)
+		if !strings.Contains(got, test.want) {
+			t.Errorf("EventHandler: \ngot: %q\nwant to contain: %q", got, test.want)
+		}
+		if test.omit != "" && strings.Contains(got, test.omit) {
+			t.Errorf("EventHandler: \ngot: %q\nwant to omit: %q", got, test.omit)
+		}
```
tbpg, line 70: Consider moving the for loop for the test case to surround just these checks. I was confused at first where the values were coming from.

kuberun/eventing/storage/main.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"fmt"
+	"log"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+)
+
+// Receive simply print the Ce-Subject header out
+func Receive(event cloudevents.Event) {
+	// do something with event.
```
tbpg, line 28:
```suggestion
// Receive prints the Ce-Subject header.
func Receive(event cloudevents.Event) {
	// Do something with event.
```

kuberun/eventing/storage/main.go
```diff
@@ -0,0 +1,39 @@
...
+	fmt.Printf("Detected change in GCS bucket: %s\n", event.Subject())
```
tbpg, line 29: Is this always for a bucket? Also, I don't think "GCS" is used for branding purposes.

kuberun/eventing/audit-pubsub/Dockerfile
```diff
@@ -0,0 +1,47 @@
+# Copyright 2021 Google LLC
```
tbpg, line 1: If these samples are related to Cloud Run, would you mind moving them to the `run/` directory in this repo?

tbpg (changes requested):

eclipselu:
closing as this is not needed

PR #1923, opened 2021-01-29
test: Temporarily disable default deadlines while running Spanner sample tests

### LINKED ISSUE #1883: spanner/spanner_snippets/spanner: TestSample failed
This test failed!

To configure my behavior, see [the Build Cop Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/master/packages/buildcop).

If I'm commenting on this issue too often, add the `buildcop: quiet` label and
I will stop commenting.

---

commit: 555da975282aa3ce627e7f003aaa5bf55fde9e25
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/9115ef13-87fb-441b-970e-17073717bbd9), [Sponge](http://sponge2/9115ef13-87fb-441b-970e-17073717bbd9)
status: failed
<details><summary>Test output</summary><br><pre>    integration_test.go:376: failed to add numeric column: rpc error: code = DeadlineExceeded desc = context deadline exceeded
    integration_test.go:376: failed to update data with numeric: spanner: code = "NotFound", desc = "Column not found in table Venues: Revenue"
    integration_test.go:376: failed to query with numeric parameter: spanner: code = "InvalidArgument", desc = "Unrecognized name: Revenue [at 1:43]\\nSELECT VenueId, Revenue FROM Venues WHERE Revenue < @revenue\\n                                          ^"
    integration_test.go:325: got output ""; want it to contain "4 "
    integration_test.go:326: got output ""; want it to contain "35000"</pre></details>

olavloite:
(Experimental, see considerations in #1883)

Fixes #1883


tbpg:
@codyoss @noahdietz

tbpg (commented):
Seems OK to get the test passing. But, my concern is that if this test is failing, it would also fail for users. I would prefer to fix the underlying issue.

codyoss:
I have not heard of other instances of this happening with other apis but I will try to do a little digging. If it is happening more broadly I think we should evaluate raising the default deadline. Or we may want to consider LROs a special case where we don't want a default deadline at all as  they are more of a foundational piece to the infrastructure of other apis.

skuruppu:
I think this may only happen for Spanner since the backends can sometimes be overloaded and can't return the GetOperation result within that 10 sec window. Personally I think it makes sense to disable this timeout for Spanner.

I also like option #4 from https://github.com/GoogleCloudPlatform/golang-samples/issues/1883#issuecomment-769845604.

noahdietz:
+1 to option 4: pass a long deadline on the `Wait` request context in the test. 

Also +1 to @codyoss's point, the LRO client is used in every client library so maybe we shouldn't have a default deadline there. But if we have no reported issues, I'd prefer to not make a change across the client corpus.

olavloite:
> I think this may only happen for Spanner since the backends can sometimes be overloaded and can't return the GetOperation result within that 10 sec window. Personally I think it makes sense to disable this timeout for Spanner.
> 
> I also like option #4 from [#1883 (comment)](https://github.com/GoogleCloudPlatform/golang-samples/issues/1883#issuecomment-769845604).

One thing that I forgot to mention regarding option 4 is that the test is running the sample code. In order to pass in a longer timeout, we need to change the sample code a little bit and add a `ctx context.Context` parameter:
* Each sample function currently has this signature: `type sampleFunc func(w io.Writer, dbName string) error`
* That would change to: `type sampleFunc func(ctx context.Context, w io.Writer, dbName string) error`

tbpg:
I'd qualify this as a reported issue. If it's happening consistently enough for us to want to increase the deadline, users should probably increase it, too.

Having a context as an argument also goes against [part of the samples style guide](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#declare-a-contextcontext-as-needed).

In that vein, would you mind moving the increased timeout on the context within the sample? That's what we're doing with the Storage samples:
https://github.com/GoogleCloudPlatform/golang-samples/blob/7cc02421347a2ff75d542aed7df8101c6aa11fc0/storage/buckets/add_bucket_iam_member.go#L38

olavloite:
> I'd qualify this as a reported issue. If it's happening consistently enough for us to want to increase the deadline, users should probably increase it, too.
> 
> Having a context as an argument also goes against [part of the samples style guide](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#declare-a-contextcontext-as-needed).
> 
> In that vein, would you mind moving the increased timeout on the context within the sample? That's what we're doing with the Storage samples:
> https://github.com/GoogleCloudPlatform/golang-samples/blob/7cc02421347a2ff75d542aed7df8101c6aa11fc0/storage/buckets/add_bucket_iam_member.go#L38

I considered that as well, but decided against it, as I don't think this should be considered a reported issue. The default timeout values should be good enough for 'normal' usage, and that is already the case for the actual timeout values for Cloud Spanner.

Note: The problem in this case is (as far as I can tell) not caused by the timeout values configured in the Cloud Spanner service definitions, but in the 10 second default timeout value for `GetOperation`. That will in most cases be enough, but on tightly scaled test instances that can get a bunch of parallel (test) requests in short period of time, it seems that it can be too little.
An additional problem in this case is the following:
1. If you use the default timeouts, then `op.Wait(..)` will use the default timeout value for `GetOperation` ([10 seconds](https://github.com/googleapis/google-cloud-go/blob/84dd78c4012dba78e64fb1abd6f6f9375e33b1de/longrunning/autogen/operations_client.go#L251)) and the default value for the operation itself (for example [1 hour](https://github.com/googleapis/google-cloud-go/blob/0b4370a0d397913d932dbbdc2046a958dc3b836a/spanner/admin/database/apiv1/database_admin_client.go#L431) for `UpdateDatabaseDdl`)
2. If you however decide to pass in a context with a deadline, the same deadline will be used for both.

So we can't pass in a separate deadline for both, except by using the default deadlines, so we would need to pass in a very long deadline to be sure.

hengfengli (approved):
LGTM. The only issue is that this doesn't match the style guide. But admin operations have long running time and personally, I think they can have a seperate timeout setting.

hengfengli:
@olavloite As @tbpg's suggestion, can we create a context just for `op.Wait(ctx)` call? 

```
ctx, cancel := context.WithTimeout(ctx, time.Hour)       // or smaller values, such as 1 min, which is longer than 10 secs
defer cancel()

op.Wait(ctx)
```

olavloite:
> @olavloite As @tbpg's suggestion, can we create a context just for `op.Wait(ctx)` call?
> 
> ```
> ctx, cancel := context.WithTimeout(ctx, time.Hour)       // or smaller values, such as 1 min, which is longer than 10 secs
> defer cancel()
> 
> op.Wait(ctx)
> ```
> 
> What would be the problem with this?

@hengfengli 
It would become a part of each sample that is listed in the documentation. I don't think it is good to add a specific timeout to each sample, as it would indicate to users that they need to set a timeout themselves, even when the RPCs already have (or should have) reasonable default timeouts. So we would be showing something in all the samples that users shouldn't really need to care about in most cases.

Note also that the context that is passed in to `op.Wait(ctx)` is used for two different things __if it contains a deadline__.
1. `op.Wait(ctx)` polls the `GetOperation` RPC regularly to get the current status of the operation. If `ctx` does not contain a deadline, it will use its own default deadline, which is 10 seconds.
2. `op.Wait(ctx)` returns the result of the long-running operation when it finishes, or the error that is returned from the operation. If `ctx` does not contain a deadline, it will use the default deadline of the long-running operation for the operation itself. If `ctx` does contain a deadline, then that deadline is also applied to the operation.

So the above means that the context that is passed in to `op.Wait` needs to be long enough for the long-running operation. That is the reason that this PR uses 1 hour instead of for example 1 minute.

hengfengli:
Thanks a lot for explanation. 

For 1, I can see it in https://github.com/googleapis/google-cloud-go/blob/8b4adbf9815e1ec229dfbcfb9189d3ea63112e1b/longrunning/longrunning.go#L96. 

However, for 2, if I pass a different `ctx` to `op.Wait(ctx)`, why will `ctx` affect the LRO itself? The deadline of LRO is affected by the context when we call the method, e.g., 

```
op, err := adminClient.CreateDatabase(ctx, &adminpb.CreateDatabaseRequest{
...
```

olavloite:
> Thanks a lot for explanation.
> 
> For 1, I can see it in https://github.com/googleapis/google-cloud-go/blob/8b4adbf9815e1ec229dfbcfb9189d3ea63112e1b/longrunning/longrunning.go#L96.
> 
> However, for 2, if I pass a different `ctx` to `op.Wait(ctx)`, why will `ctx` affect the LRO itself? The deadline of LRO is affected by the context when we call the method, e.g.,
> 
> ```
> op, err := adminClient.CreateDatabase(ctx1, &adminpb.CreateDatabaseRequest{
> ...
> 
> op.Wait(ctx2) // why does ctx2 affect the deadline of LRO itself? 
> ```

Technically, the second context does not affect the __LRO itself__, but it affects the __waiting for the LRO__ to finish:

1. `op, err := adminClient.CreateDatabase(ctx1, &adminpb.CreateDatabaseRequest{`: This will use `ctx1` and then return a long-running operation. `ctx1` can normally be relatively short, because the RPC will return quickly with an LRO. The returned LRO itself will not do anything with the deadline of `ctx1`.
2. `op.Wait(ctx2)`: `ctx2` will not affect the LRO, as an LRO does not have a (public) deadline. It will just run for as long as it needs, or until it is cancelled. (It might be that a backend automatically cancels an LRO after a certain time, but such a deadline is not something that is part of the [public interface of an LRO](https://cloud.google.com/spanner/docs/reference/rpc/google.longrunning#operation).)
3. `op.Wait(ctx2)` will use the deadline in `ctx2` for two purposes:
3.1. As a deadline for the `GetOperation` RPC when it is polling the LRO for the latest status. If `ctx2` does not contain a deadline, it will use the [default deadline of 10 seconds](https://github.com/googleapis/google-cloud-go/blob/84dd78c4012dba78e64fb1abd6f6f9375e33b1de/longrunning/autogen/operations_client.go#L251).
3.2. As a deadline for how long it will wait for the LRO to actually return a result. That happens in [this loop](https://github.com/googleapis/google-cloud-go/blob/8b4adbf9815e1ec229dfbcfb9189d3ea63112e1b/longrunning/longrunning.go#L156). The `sl(ctx, bo.Pause())` function will return an error if the [context deadline is exceeded](https://github.com/googleapis/gax-go/blob/fbaf9882acf3297573f3a7cb832e54c7d8f40635/v2/invoke.go#L56).

tbpg:
I'm OK not following the style guide for this. Let me know if you need me to approve/merge. But, it looks like you have a merge conflict at the moment.

olavloite:
@tbpg I've fixed the merge conflicts. If you're still OK with these samples not following the style guide regarding passing in the context, then yes please go ahead and merge (I'm not authorized to merge).

PR #1939, opened 2021-02-05
Revert "all: fix region tags"

umairaziz112221:
Reverts GoogleCloudPlatform/golang-samples#1666


tbpg:
Thanks for the PR, but we'd like to keep that commit.

PR #1941, opened 2021-02-07
fix: add missing return statements

tibbes:
This adds two missing 'return' statements after calls to http.Error to avoid execution falling through into the 'success' code path.

In both cases, the calls to http.Error are reporting failed security checks, so if this example code was copied into a production system it would probably cause a security vulnerability.


tbpg (approved):
Thank you!

PR #1959, opened 2021-02-18
fix(bigquery): correct error checking logic in dry run example

shollyman:


bigquery/snippets/querying/bigquery_query_dry_run.go
```diff
@@ -51,7 +51,7 @@ func queryDryRun(w io.Writer, projectID string) error {
 	}
 	// Dry run is not asynchronous, so get the latest status and statistics.
 	status := job.LastStatus()
-	if err != nil {
+	if status.Err() != nil {
```
codyoss, line 54:
```suggestion
	if err := status.Err(); err != nil {
```

codyoss (changes requested):
LGTM, besides the suggestion.

codyoss (approved):

bigquery/snippets/querying/bigquery_query_dry_run.go
shollyman: done

PR #1966, opened 2021-02-25
fix: added execution permissions to script

craiglabenz:


tbpg (approved):
Thanks!

tbpg:
Pushing through with merging because this is fixing an issue with CI.

PR #1970, opened 2021-03-01
samples(spanner): PITR samples backup fix

thiagotnunes:
Instead of using the earliest version time of the database, uses the current time (from spanner). If we used the earliest version time of the database instead we would be creating an empty backup, since the database we are backing up is not 1 hour old (default version retention period).


spanner/spanner_snippets/spanner/spanner_create_backup.go
```diff
@@ -40,9 +41,24 @@ func createBackup(w io.Writer, db, backupID string) error {
...
 	defer adminClient.Close()
-	dbMetadata, err := adminClient.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: db})
+	client, err := spanner.NewClient(ctx, db)
 	if err != nil {
-		return fmt.Errorf("createBackup.GetDatabase: %v", err)
+		return fmt.Errorf("createBackup.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	stmt := spanner.Statement{
```
olavloite, line 50: Would it be better (and at all possible) to move this code away from the sample and into the test code. I worry a little bit that this will confuse users a little, as we are showing an example of something that is not really best practice. This is creating a backup using the current timestamp, which is basically the same as creating a backup without setting a version time. Setting a version time is something that a user should normally only do if they want a specific version to be backed up.
In the Java sample we keep the version time selection in the test code and pass that into the test.

Another possibility could be do show an example where we execute a read/write transaction and then backup the database at the commit timestamp of the read/write transaction. That is a somewhat more reasonable use case, as it shows users how to create a backup containing everything up to a specific commit.

thiagotnunes: That is fair, we are looking to change the samples at the moment to something more realistic than the `earliest_version_time`, since the database is not 1 hour old. We will need to revisit this samples after launch, but for now this change would be better than the current version time that we are setting.

skuruppu: I agree with @olavloite. I don't think we should be putting this logic in the sample. If instead, like in the C# sample, we can compute the version time in the test and pass it in, that would be ideal.

thiagotnunes: Ok, changed the sample to have the version time parameterised.

skuruppu (approved):

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -90,6 +91,24 @@ func initBackupTest(t *testing.T, id, dbName string) (restoreDBName, backupID, c
 		t.Logf("database %s exists in state %s. delete result: %v", db.GetName(), db.GetState().String(),
 			adminClient.DropDatabase(ctx, &adminpb.DropDatabaseRequest{Database: restoreDBName}))
 	}
+	client, err := spanner.NewClient(ctx, restoreDatabaseID)
```
olavloite, line 94: I'm not sure if I'm missing something here, or if this just coincidentally goes well (sometimes): We are creating a client to read from `restoreDatabaseID` here. On the lines above we check whether that database exists, and if it does, we drop it:
```go
	if db, err := adminClient.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: restoreDBName}); err == nil {
		t.Logf("database %s exists in state %s. delete result: %v", db.GetName(), db.GetState().String(),
			adminClient.DropDatabase(ctx, &adminpb.DropDatabaseRequest{Database: restoreDBName}))
	}
```

I would therefore expect this client to return an error that the database does not exist.

thiagotnunes: I think you are correct, it is intriguing that the integration tests are passing.

olavloite (approved):

spanner/spanner_snippets/spanner/spanner_create_backup.go
```diff
@@ -28,7 +28,7 @@ import (
 	adminpb "google.golang.org/genproto/googleapis/spanner/admin/database/v1"
 )
 
-func createBackup(w io.Writer, db, backupID string) error {
+func createBackup(w io.Writer, db, backupID string, versionTime time.Time) error {
```
tbpg, line 31: Will you add sample values for these arguments? See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#function-arguments-for-snippets.

thiagotnunes: Added a sample value for the `versionTime`.

thiagotnunes:
@tbpg hey there, is there anything else that you think we should fix in here? Thanks!

tbpg (approved):

PR #1979, opened 2021-03-04
feat(datacatalog): add a quickstart for datacatalog

shollyman:
Demonstrates:
* creating a tag template
* looking up a resource from another service (bigquery)
* applying a tag based on the template to the resource


datacatalog/datacatalog_quickstart/main.go
```diff
@@ -0,0 +1,195 @@
...
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	datacatalog "cloud.google.com/go/datacatalog/apiv1"
```
tbpg, line 30: Is datacatalog part of BigQuery? Should it be in the `bigquery` directory? Or does it make sense as a separate directory?

datacatalog/datacatalog_quickstart/main.go
```diff
@@ -0,0 +1,195 @@
...
+	datacatalogpb "google.golang.org/genproto/googleapis/cloud/datacatalog/v1"
+)
+
+func main() {
+
```
tbpg, line 35: Delete.

datacatalog/datacatalog_quickstart/main.go
```diff
@@ -0,0 +1,195 @@
...
+	var (
```
tbpg, line 36: Any particular reason to use a `var` group instead of `:=`?

datacatalog/datacatalog_quickstart/main.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return client.CreateTag(ctx, req)
+}
+
+// convertBigQueryResourceRepresentation converts a table identifier in standard sql form
+// (project.datadata.table) into the representation used within data catalog.
+func convertBigQueryResourceRepresentation(table string) (string, error) {
+	parts := strings.Split(table, ".")
+	if len(parts) != 3 {
+		return "", fmt.Errorf("specified table string is not in expected project.dataset.table format")
```
tbpg, line 190: Include the bad string?

tbpg (changes requested):
A few minor comments, overall looks good.

datacatalog/datacatalog_quickstart/main.go
shollyman: It's not part of BQ.  It's a generalized metadata service, but BQ has some features that rely on it.

datacatalog/datacatalog_quickstart/main.go
shollyman: Mostly a carryover from idioms when they're globals.  Removed the var group.

datacatalog/datacatalog_quickstart/main.go
shollyman: done

datacatalog/datacatalog_quickstart/main.go
shollyman: done

tbpg (approved):

PR #1983, opened 2021-03-08
Fixes error assignment

krisrybak:
err is undeclared at this stage, thus proper assignment is required


krisrybak:
@googlebot I signed it!

tbpg (changes requested):
`err` is declared on line 30, so I don't think there is an issue here?

tbpg:
Thank you for the PR! Let us know if you find anything else we can improve in the future.

PR #1986, opened 2021-03-08
feat(eventarc): add pubsub unmarshal library

grant:
Adds the Pub/Sub unmarshal library to the Eventarc sample.

- Adds `pubsub "github.com/googleapis/google-cloudevents-go/cloud/pubsub/v1"`
- Keeps tests passing
- Fixes tests such that the tests always send a Pub/Sub structured JSON object (not just `{}`)

This is pretty similar to the example here:

https://github.com/googleapis/google-cloudevents-go#example-usage


eventarc/pubsub/main.go
```diff
@@ -18,33 +18,37 @@
...
-// See the documentation for more details:
-// https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
-type PubSubMessage struct {
-	Message struct {
-		Data []byte `json:"data,omitempty"`
-		ID   string `json:"id"`
-	} `json:"message"`
-	Subscription string `json:"subscription"`
-}
+	pubsub "github.com/googleapis/google-cloudevents-go/cloud/pubsub/v1"
```
tbpg, line 28: This is a surprising import path. `vN` has a special meaning in Go. It's usually only used for `v2`+. But, this is not a `v1` module -- it's `v0`.

If there is ever a `pubsub/v2`, it will probably need to be a `v2` module. But, there should only be one module at the root of the repo...

eventarc/pubsub/main.go
```diff
@@ -18,33 +18,37 @@
...
-	var e PubSubMessage
-	if err := json.NewDecoder(r.Body).Decode(&e); err != nil {
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
 		http.Error(w, "Bad HTTP Request", http.StatusBadRequest)
 		log.Printf("Bad HTTP Request: %v", http.StatusBadRequest)
 		return
 	}
-	name := string(e.Message.Data)
+	e, err := pubsub.UnmarshalMessagePublishedData(body)
```
tbpg, line 39: This is surprising, too. I'd expect to only use the type and use the `json` package directly.

eventarc/pubsub/main.go
```diff
@@ -18,33 +18,37 @@
...
+	if err != nil {
+		http.Error(w, "Bad Pub/Sub Request", http.StatusBadRequest)
+		log.Printf("Bad Pub/Sub Request: %v", http.StatusBadRequest)
+		return
+	}
+	nameBytes, err := base64.URLEncoding.DecodeString(*e.Message.Data)
```
tbpg, line 45: The data uses URL encoding? The test uses `StdEncoding`.

grant:
We're going to make a few changes to the library first, and then re-visit this PR.

Closing for now, thanks!

PR #1987, opened 2021-03-08
refactor(automl): use Go specific operation helpers

codyoss:
Unlike other languages Go returns Operation types, similar to
Iterator types, to make working with LROs easier. Our samples
should use these helpers were possible instead of the embeded
LRO client.

### DIFF
--- a/automl/operation_status_get.go
+++ b/automl/operation_status_get.go
@@ -22,14 +22,15 @@ import (
 	"io"
 
 	automl "cloud.google.com/go/automl/apiv1"
-	"google.golang.org/genproto/googleapis/longrunning"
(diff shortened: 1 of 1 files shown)


codyoss:
I would have liked to refactor the List sample as well, but there is no easier way of doing this today. LROs are specific to product endpoints so they need to use the same endpoint as the client they are polling operations for. Because of this I have left that sample as is for now but it could be updated to use https://pkg.go.dev/cloud.google.com/go/longrunning/autogen directly. I think we may want to consider adding function to the LRO manual eventually that accepts an endpoint explicitly for these types of use cases where one wants to see all ongoing operations.

tbpg (approved):
Thanks!

PR #2001, opened 2021-03-17
feat: add samples for CMEK with Spanner

olavloite:
Adds samples for Customer Managed Encryption Keys with Spanner:
- Create database using a specific key
- Create backup using a specific key
- Restore backup using a specific key


skuruppu (commented):
Looks good @olavloite. Do we need to add tests? From memory, tests for backup samples are run as part of a nightly build. But I don't know where they live or how they're configured.

spanner/spanner_snippets/spanner/spanner_create_backup_with_encryption_key.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	// Get the name, create time, backup size and encryption key from the backup.
+	backupCreateTime := time.Unix(backup.CreateTime.Seconds, int64(backup.CreateTime.Nanos))
+	fmt.Fprintf(w,
+		"Backup %s of size %d bytes was created at %s using encryption key %s\n",
+		backup.Name,
+		backup.SizeBytes,
+		backupCreateTime.Format(time.RFC3339),
+		backup.EncryptionInfo.KmsKeyVersion)
```
larkee, line 74: Because `EncryptionInfo` will likely not yet be set (see [here](https://github.com/googleapis/googleapis/blob/master/google/spanner/admin/database/v1/spanner_database_admin.proto#L399-L400)), please use `EncryptionConfig.KmsKeyName` instead.

spanner/spanner_snippets/spanner/spanner_create_database_with_encryption_key.go
```diff
@@ -0,0 +1,75 @@
...
+		EncryptionConfig: &adminpb.EncryptionConfig{KmsKeyName: kmsKeyName},
+	}
+	op, err := adminClient.CreateDatabase(ctx, &req)
+	if err != nil {
+		return fmt.Errorf("createDatabaseWithCustomerManagedEncryptionKey.CreateDatabase: %v", err)
+	}
+	if _, err := op.Wait(ctx); err != nil {
+		return fmt.Errorf("createDatabaseWithCustomerManagedEncryptionKey.Wait: %v", err)
+	}
+	fmt.Fprintf(w, "Created database [%s] using encryption key %q\n", db, kmsKeyName)
```
larkee, line 71: Is there any value in using the values from the returned database to show how to access them? I mean, using protos is kind of a staple of Go so I assume not. Does this match the style of the other Go samples?

larkee:
> Looks good @olavloite. Do we need to add tests? From memory, tests for backup samples are run as part of a nightly build. But I don't know where they live or how they're configured.

The tests are defined under `spanner/spanner_snippets/spanner/integration_test.go`

spanner/spanner_snippets/spanner/spanner_create_backup_with_encryption_key.go
```diff
@@ -0,0 +1,78 @@
...
+	expireTime := time.Now().AddDate(0, 0, 14)
+	// Create a backup for a database using a Customer Managed Encryption Key
+	req := adminpb.CreateBackupRequest{
+		Parent:   instanceName,
+		BackupId: backupID,
+		Backup: &adminpb.Backup{
+			Database:   db,
+			ExpireTime: &pbt.Timestamp{Seconds: expireTime.Unix(), Nanos: int32(expireTime.Nanosecond())},
+		},
+		EncryptionConfig: &adminpb.CreateBackupEncryptionConfig{KmsKeyName: kmsKeyName},
```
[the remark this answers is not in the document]
olavloite: Done.

spanner/spanner_snippets/spanner/spanner_create_backup_with_encryption_key.go
olavloite: Hmm... There's seems to be something strange going on with the protos for Go, as there is no `EncryptionConfig` field for the `Backup` message in the protos used by Go: https://github.com/googleapis/googleapis/blob/15b444a98762c31e863087eb688635f2c24feec5/google/spanner/admin/database/v1/backup.proto#L124

Do these need to be re-generated / updated?

spanner/spanner_snippets/spanner/spanner_create_database_with_encryption_key.go
olavloite: Yes, the database metadata is returned by the operation. I've changed it to use the information from that.

olavloite:
> > Looks good @olavloite. Do we need to add tests? From memory, tests for backup samples are run as part of a nightly build. But I don't know where they live or how they're configured.
> 
> The tests are defined under `spanner/spanner_snippets/spanner/integration_test.go`

The integration tests are executed for each PR, and not only for nightly builds, so if we add integration tests for these operations, we will be increasing the build time quite extensively. We also need to set up access permissions to an encryption key that can be used. I've added a test case for `CreateDatabase` to see if the default keys are already accessible for account that is used by the Go build. The `CreateDatabase` operation is normally also the quickest one of the three.

skuruppu:
> > > Looks good @olavloite. Do we need to add tests? From memory, tests for backup samples are run as part of a nightly build. But I don't know where they live or how they're configured.
> > 
> > 
> > The tests are defined under `spanner/spanner_snippets/spanner/integration_test.go`
> 
> The integration tests are executed for each PR, and not only for nightly builds, so if we add integration tests for these operations, we will be increasing the build time quite extensively. We also need to set up access permissions to an encryption key that can be used. I've added a test case for `CreateDatabase` to see if the default keys are already accessible for account that is used by the Go build. The `CreateDatabase` operation is normally also the quickest one of the three.
> 
> EDIT: The default key names (taken from the Java repository) do not seem to be accessible for the account that is used for the integration tests for Go.

@tbpg would you be able add the following permissions for the service account used for running these tests:
- Cloud KMS Admin
- Cloud KMS CryptoKey Encrypter/Decrypter

spanner/spanner_snippets/spanner/spanner_create_backup_with_encryption_key.go
olavloite: I just checked with the Java implementation, and this field is also not present there on the protos, so we can only print the key name that was given to the method, and nothing from the `Backup` proto.

tbpg:
> @tbpg would you be able add the following permissions for the service account used for running these tests:

Done. I gave the SA those permissions on all of the test projects. However, still getting an error saying the key doesn't exist:

```
    integration_test.go:429: failed to create database with customer managed encryption key: 
createDatabaseWithCustomerManagedEncryptionKey.CreateDatabase: rpc error: code = FailedPrecondition desc = KMS Key 
provided is not usable: ServiceAccount: service-176007320581@gcp-sa-spanner.iam.gserviceaccount.com KmsKeyName: 
projects/golang-samples-tests-3/locations/us-central1/keyRings/spanner-test-keyring/cryptoKeys/spanner-test-key Error: 
generic::not_found: CryptoKey projects/golang-samples-tests-3/locations/us-central1/keyRings/spanner-test-keyring/cryptoKeys/spanner-test-key not found. 
[google.rpc.error_details_ext] { message: "CryptoKey projects/golang-samples-tests-3/locations/us-central1/keyRings/spanner-test-keyring/cryptoKeys/spanner-test-key not found." }.
```

skuruppu:
Thanks a lot @tbpg.

@olavloite, I think you need to create a key manually as a once off, maybe through a script, and then this error would hopefully go away.

olavloite:
@skuruppu and @larkee 
I've added integration tests for the CMEK `CreateDatabase`, `CreateBackup` and `RestoreDatabase` samples, including a retry mechanism for restore operations that may fail as there might be multiple tests running against the same Spanner instance at the same time, and only one restore operation is allowed at any time. PTAL.

spanner/spanner_snippets/spanner/spanner_create_backup_with_encryption_key.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	// Get the name, create time, backup size and encryption key from the backup.
+	backupCreateTime := time.Unix(backup.CreateTime.Seconds, int64(backup.CreateTime.Nanos))
+	fmt.Fprintf(w,
+		"Backup %s of size %d bytes was created at %s using encryption key %s\n",
+		backup.Name,
+		backup.SizeBytes,
+		backupCreateTime.Format(time.RFC3339),
+		kmsKeyName)
```
skuruppu, line 77: For the key name, you should be able to get it from the backup object to make sure that what's used for encryption is the same as what you specified as the key.

olavloite: I can't see a field on the `Backup` object that contains a field for that. It does contain an `EncryptionInfo` [struct](https://github.com/googleapis/go-genproto/blob/93e7006c17a68b79787d42654b179335e7337138/googleapis/spanner/admin/database/v1/backup.pb.go#L224), which again [contains](https://github.com/googleapis/go-genproto/blob/93e7006c17a68b79787d42654b179335e7337138/googleapis/spanner/admin/database/v1/common.pb.go#L238) a `KmsKeyVersion` field. We could use that, but at least for `Database` objects it [might take a little while](https://github.com/googleapis/googleapis/blob/5da78db589d423db6bcd6a35637c923fe1a7665c/google/spanner/admin/database/v1/spanner_database_admin.proto#L399) before it is is populated, and I'm not sure whether that is also true for backups.

We could also use the key name from the `CreateBackupRequest` object, but that is not really different from the `kmsKeyName` variable.

Or am I missing something else here?

larkee: Okay, this confusion is my fault. My previous comment on this PR was incorrect. I hadn't realized that `Backup` did not have `encryption_config`.

For `Database`, we should use `encryption_config` as `encryption_info` is lazily instantiated.

For `Backup`, there is no `encryption_config`. The `encryption_info` field on `Backup` does not mention being lazily instantiated and does appear to be set every time. However, using `encryption_info` means the string will include the key version which is inconsistent with the other sample messages.

The two options are:
1) using the `kmsKeyName` i.e. no change
2) using the `EncryptionInfo.kmsKeyVersion`
I need to follow up to see if it's possible to use `EncryptionInfo.kmsKeyVersion` with the database samples for consistency. I'll get back to you in the next day or so :+1:

olavloite: Thanks for the clarification :+1: . I just changed it to using `EncryptionInfo.KmsKeyVersion` after having a quick look at the NodeJS sample. It seems that it is always populated (at least the tests succeed there). Changing it back to using the `kmsKeyName` is simple, so I'll leave it like this until we know which one we prefer.

skuruppu (approved):

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -412,6 +415,109 @@ func TestCreateDatabaseWithRetentionPeriodSample(t *testing.T) {
...
+	dbName, cleanup := initTest(t, randomID())
+	defer cleanup()
+	adminClient, err := database.NewDatabaseAdminClient(context.Background())
+	if err != nil {
+		t.Errorf("failed to create admin client: %v", err)
+	}
+
+	var b bytes.Buffer
+
+	projectId := os.Getenv("GOLANG_SAMPLES_PROJECT_ID")
```
tbpg, line 429: You can get this from:

```
tc := testutil.SystemTest(t)
// Use tc.ProjectID
```

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -412,6 +415,109 @@ func TestCreateDatabaseWithRetentionPeriodSample(t *testing.T) {
...
+	assertContains(t, out, fmt.Sprintf("Created database [%s] using encryption key %q", dbName, kmsKeyName))
+
+	// Try to create a backup of the encrypted database and delete it after the test.
+	backupId := fmt.Sprintf("enc-backup-%s", randomID())
+	defer func() {
+		_ = adminClient.DeleteBackup(context.Background(), &adminpb.DeleteBackupRequest{
+			Name: fmt.Sprintf("%s/backups/%s", instanceName, backupId),
+		})
+	}()
+	b = bytes.Buffer{}
```
tbpg, line 461:
```suggestion
	bytes.Reset()
```

spanner/spanner_snippets/spanner/spanner_create_backup_with_encryption_key.go
```diff
@@ -0,0 +1,81 @@
...
+	"regexp"
+	"time"
+
+	database "cloud.google.com/go/spanner/admin/database/apiv1"
+	pbt "github.com/golang/protobuf/ptypes/timestamp"
+	adminpb "google.golang.org/genproto/googleapis/spanner/admin/database/v1"
+)
+
+func createBackupWithCustomerManagedEncryptionKey(w io.Writer, db, backupID, kmsKeyName string) error {
+	// kmsKeyName = `projects/<project>/locations/<location>/keyRings/<key_ring>/cryptoKeys/<kms_key_name>`
```
tbpg, line 32: Include `db` and `backupID`?

tbpg (commented):
A few small comments.

spanner/spanner_snippets/spanner/integration_test.go
olavloite: Done

spanner/spanner_snippets/spanner/integration_test.go
olavloite: Done

spanner/spanner_snippets/spanner/spanner_create_backup_with_encryption_key.go
olavloite: Good point, done here and in the other samples

larkee:
LGTM :+1: 

As a note, some of the other languages use a static KMS key rather than creating a key to avoid an unnecessary dependency on the KMS client. Do you think relying on the key existing is unreasonable?

olavloite:
> LGTM 👍
> 
> As a note, some of the other languages use a static KMS key rather than creating a key to avoid an unnecessary dependency on the KMS client. Do you think relying on the key existing is unreasonable?

I personally prefer tests that (are able to) create the resources they need as much as possible, except for resources that are very expensive to create. That makes them easier and quicker to try out locally if there's a problem. But I can see the point about not wanting to add an extra dependency on the KMS client, so I'm also perfectly fine with working with a static key.

skuruppu:
@olavloite I think we will leave the tests as is and have the dependency on the KMS client. Most of the time, a key is not created so the dependency is irrelevant. Very rarely, when all the contents of the project gets nuked, then we won't get flaky tests as the client is able to create the key if it's not there. This is pretty important for Go because the test project is not owned by the spanner team so we may not be able to create the static key manually.

skuruppu:
@tbpg would you please be able to merge this for us?

tbpg (approved):

skuruppu:
@tbpg I think you have to manually press the button. It seems that automerge doesn't work :(

PR #2006, opened 2021-03-18
refactor: use testutil.SystemTest for getting project ids

hongalex:
There were some lingering references of getting the test project via `os.Getenv`. This switches the remaining non-main tests to use `testutil.SystemTest(t).ProjectID` instead.


tbpg:
This is intentional -- that way, we can avoid having golang-samples as a dependency in the `go.mod` for the sample functions and apps.

PR #2007, opened 2021-03-18
fix(dataproc): fixing GCE bug associated with new default dataproc image

bradmiro:
Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/2002 and fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/2004.


tbpg (approved):
Thanks!

PR #2008, opened 2021-03-18
fix(pubsublite): improved publish error handling

tmdiep:
Clarified comments, print errors for all publish results and return the publisher client's error.


pubsublite/publish/publish_batch.go
```diff
@@ -63,19 +63,24 @@ func publishWithBatchSettings(w io.Writer, projectID, zone, topicID string, mess
...
 		// Get blocks until the result is ready.
 		id, err := r.Get(ctx)
 		if err != nil {
-			// NOTE: The publisher will terminate upon first error. Create a new
-			// publisher to republish failed messages.
-			return fmt.Errorf("publish error: %v", err)
+			// NOTE: A failed PublishResult indicates that the publisher client
+			// encountered a fatal error and has permanently terminated. After the
+			// fatal error has been resolved, a new publisher client instance must be
+			// created to republish failed messages.
```
tbpg, line 74: Should we show how to collect the failed messages that need to be resent?

tmdiep: Good idea. I updated the quickstart_publisher sample to demonstrate this. But left the other snippets more minimal.

hongalex (approved):

pubsublite/quickstart_publisher/main.go
```diff
@@ -48,33 +54,45 @@ func main() {
...
 	}
 
-	// Print publish results.
-	for _, r := range results {
+	// Print publish results and collect any messages that need to be republished.
+	var toRepublish []*pubsub.Message
+	for _, pm := range published {
 		// Get blocks until the result is ready.
-		id, err := r.Get(ctx)
+		id, err := pm.result.Get(ctx)
```
tbpg, line 70: I think if I were writing this, I would use https://pkg.go.dev/golang.org/x/sync/errgroup. But, that's probably too complex for a sample, and I don't think we've done it before.

In the same vein, an alternative would be to do `publisher.Publish(...).Get()` in a goroutine and put results/errors into a channel/map. Then you wouldn't need the `publishedMessage` type.

I'm not sure that's better. But, I wanted to through a couple alternatives out there to see what you thought. I'm OK to move forward as-is.

pubsublite/quickstart_publisher/main.go
```diff
@@ -48,33 +54,45 @@ func main() {
...
-		})
-		results = append(results, r)
+		}
+		result := publisher.Publish(ctx, msg)
+		published = append(published, &publishedMessage{msg, result})
 	}
 
-	// Print publish results.
-	for _, r := range results {
+	// Print publish results and collect any messages that need to be republished.
```
tbpg, line 66: Should we say how to republish (create a new publisher and try again)?

pubsublite/quickstart_publisher/main.go
tmdiep: I took a look at some of pubsub's samples, and there are some that use `sync.WaitGroup` (e.g. [publish_ordering.go](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/pubsub/topics/publish_ordering.go)), while others loop a slice of PublishResults (e.g. [publish_settings.go](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/pubsub/topics/publish_settings.go)).

I prefer `errgroup.Group` over `sync.WaitGroup`, and have amended this quickstart so that we have at least one realistic example of error handling.

tbpg (approved):

tmdiep:
Tests seem to be flaky today. The tests for pubsublite are passing, but not sure if others are genuine failures.

PR #2018, opened 2021-03-22
feat(monitoring): add Opencensus sample

yuriatgoogle:
feat: provides code for documentation.


monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+	"contrib.go.opencensus.io/exporter/stackdriver"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+	"golang.org/x/exp/rand"
+)
+
+// [END monitoring_sli_metrics_opencensus_setup]
+// [START monitoring_sli_metrics_opencensus_measure]
+// set up metrics
```
tbpg, line 36: Please make comments complete sentences (with periods at the end).

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+var (
+	requestCount       = stats.Int64("oc_request_count", "total request count", "requests")
+	failedRequestCount = stats.Int64("oc_failed_request_count", "count of failed requests", "requests")
+	responseLatency    = stats.Float64("oc_latency_distribution", "distribution of response latencies", "s")
+)
+
+// [END monitoring_sli_metrics_opencensus_measure]
+
+// set up views
```
tbpg, line 45: Include this (complete sentence) comment in the region?

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+		Name:        "oc_failed_request_count",
+		Measure:     failedRequestCount,
+		Description: "count of failed requests",
+		Aggregation: view.Count(),
+	}
+	responseLatencyView = &view.View{
+		Name:        "oc_response_latency",
+		Measure:     responseLatency,
+		Description: "The distribution of the latencies",
+		// bucket definitions must be explicit
```
tbpg, line 64: What does explicit mean here?

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// set up project ID
+	projectID := flag.String("project_id", "", "Cloud Project ID")
+	flag.Parse()
+
+	// register the views
+	if err := view.Register(requestCountView, failedRequestCountView, responseLatencyView); err != nil {
+		log.Fatalf("Failed to register the views: %v", err)
+	}
+	// [START monitoring_sli_metrics_opencensus_exporter]
```
tbpg, line 81: In general, we try to make sample code work if it were copied and pasted directly from the site into an editor. However, this wouldn't work because it has to be inside `func main`.

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+// [END monitoring_sli_metrics_opencensus_view]
+
+func main() {
+
+	// set up project ID
+	projectID := flag.String("project_id", "", "Cloud Project ID")
+	flag.Parse()
+
+	// register the views
+	if err := view.Register(requestCountView, failedRequestCountView, responseLatencyView); err != nil {
```
tbpg, line 78: It doesn't look like there is a region that covers the `view.Register` call? Should there be?

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+	sd.StartMetricsExporter()
+	defer sd.StopMetricsExporter()
+	// [END monitoring_sli_metrics_opencensus_exporter]
+	http.HandleFunc("/", handle)
+	log.Fatal(http.ListenAndServe(":8080", nil))
+}
+
+func handle(w http.ResponseWriter, r *http.Request) {
+	// create context
+	ctx, _ := tag.New(context.Background())
```
tbpg, line 102: Use `r.Context` to pick up anything coming though the request.

```suggestion
	ctx, _ := tag.New(r.Context())
```

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+	requestReceived := time.Now()
+	// [START monitoring_sli_metrics_opencensus_counts]
+	// count the request
+	stats.Record(ctx, requestCount.M(1))
+
+	// fail 10% of the time
+	if rand.Intn(100) > 90 {
```
tbpg, line 109: Nit: `rand.Intn` returns a number in `[0, n)`. So, this would only run for 91, 92, .., 99, which is only 9/100, or 9%. Change to `>= 90` and we should be good.

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+		// count the failed request
+		stats.Record(ctx, failedRequestCount.M(1))
+		// [END monitoring_sli_metrics_opencensus_counts]
+		fmt.Fprintf(w, "intentional error!")
+		// record latency for failure
+		// [START monitoring_sli_metrics_opencensus_latency]
+		stats.Record(ctx, responseLatency.M(time.Since(requestReceived).Seconds()))
+		// [END monitoring_sli_metrics_opencensus_latency]
+		return
+	} else {
```
tbpg, line 119: No need for the `else` because the `if` above `return`s.

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+		delay := time.Duration(rand.Intn(1000)) * time.Millisecond
+		time.Sleep(delay)
+		fmt.Fprintf(w, "Succeeded after "+strconv.Itoa(int(delay)/1000000)+" ms")
```
tbpg, line 122: `fmt.Fprintf` can print using C-style placeholders. You should be able to do:

```suggestion
		fmt.Fprintf(w, "Succeeded after %v", delay)
```

See https://play.golang.org/p/QHPpWNOZqz2.

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// fail 10% of the time
+	if rand.Intn(100) > 90 {
+		// count the failed request
+		stats.Record(ctx, failedRequestCount.M(1))
+		// [END monitoring_sli_metrics_opencensus_counts]
+		fmt.Fprintf(w, "intentional error!")
+		// record latency for failure
+		// [START monitoring_sli_metrics_opencensus_latency]
+		stats.Record(ctx, responseLatency.M(time.Since(requestReceived).Seconds()))
```
tbpg, line 116: Perhaps this should just be in a `defer` call?

monitoring/opencensus/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+	handler.ServeHTTP(rr, req)
+
+	// test response status
+	gotStatus := rr.Code
+	wantStatus := http.StatusOK
+	fmt.Print("status was: " + strconv.Itoa(gotStatus) + "\n")
+
+	if gotStatus != wantStatus {
+		t.Fatalf("Returned wrong status code: got %v want %v",
+			gotStatus, wantStatus)
```
tbpg, line 53: Go doesn't have a line length limit and discourages breaking in the middle of function calls. Since this is not embedded as a sample, we don't have to worry about horizontal scrolling.

https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#line-length

monitoring/opencensus/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"net/http/httptest"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestStartingServer(t *testing.T) {
+	// tc := testutil.SystemTest(t)
```
tbpg, line 29: Delete.

monitoring/opencensus/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// test main endpoint
+	req, err := http.NewRequest("GET", "/", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rr := httptest.NewRecorder()
+	handler := http.HandlerFunc(handle)
+	handler.ServeHTTP(rr, req)
```
tbpg, line 44: This should run the binary built by `BuildMain` directly -- the purpose of that is to test `package main` samples fully.

tbpg (changes requested):

monitoring/opencensus/main.go
yuriatgoogle: Should be done - thank you.

monitoring/opencensus/main.go
yuriatgoogle: Done, thank you.

monitoring/opencensus/main.go
yuriatgoogle: OpenCensus requires you to explicitly specify the buckets.  This is unlike, e.g. OpenTelemetry where you do not need to provide this.

monitoring/opencensus/main.go
yuriatgoogle: Do you have a recommendation for how I might proceed...?

monitoring/opencensus/main.go
yuriatgoogle: Yes, thank you - I expanded the relevant region to include this.

monitoring/opencensus/main.go
yuriatgoogle: Fixed, thank you!

monitoring/opencensus/main.go
yuriatgoogle: Not based on the OpenCensus samples on their website.... but I'm open to guidance!

monitoring/opencensus/main_test.go
yuriatgoogle: Fixed, thank you.

monitoring/opencensus/main_test.go
yuriatgoogle: Thank you, fixed.

monitoring/opencensus/main_test.go
yuriatgoogle: I think I might need your help - how would I do that?

monitoring/opencensus/main.go
tbpg: I guess it depends on how these samples will be shown in the documentation. This fits somewhere in between a quickstart and a snippet, as described by https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#style. So, in that sense, there isn't a good guideline to follow.

This works as-is, but I'm concerned about the selective inclusion/exclusion of code -- users won't see everything they need to be able to run the code.

If they take `monitoring_sli_metrics_opencensus_exporter` and copy it into a `.go` file, they will get a compilation error.

monitoring/opencensus/main.go
tbpg: Looks like this needs an update.

monitoring/opencensus/main.go
tbpg: Also needs an update.

monitoring/opencensus/main.go
tbpg: Check this out: https://play.golang.org/p/0jjjn38MYxD

The `defer` call means "run this function when the current function returns."

So, if you want to track the time from start to the function returning, a `defer` is the standard way to do it. That way, no matter where you return, you don't forget to record the latency.

monitoring/opencensus/main_test.go
tbpg: Thinking about it more, it's a little tricky to run a binary that listens for HTTP requests.

Instead, there is no need for `handler` and `handler.ServeHTTP`. You can just call the handler directly, like this:

https://github.com/GoogleCloudPlatform/golang-samples/blob/ac1f006385cdadc3bccbbfb3fa69b12e3990e077/functions/helloworld/hello_http_test.go#L35-L43

tbpg (changes requested):

monitoring/opencensus/main.go
yuriatgoogle: OK - I get that.  Would adding a comment that this needs to be inside a `func main()` block address your concern? Or would you rather see this perhaps pulled out into a dedicated function? As it stands, it follows the [example](https://opencensus.io/quickstart/go/metrics/#create-the-exporter) almost exactly.

monitoring/opencensus/main.go
yuriatgoogle: Got it - thank you.  Moved recording latency into a `defer` statement.

monitoring/opencensus/main_test.go
yuriatgoogle: Got it - thank you, changed.

monitoring/opencensus/main.go
tbpg: No, I don't think that comment is needed. Just something to think about as you're writing the sample and embedding it on a page -- it should be clear to users how to use the code.

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,125 @@
...
+	// [END monitoring_sli_metrics_opencensus_exporter]
+	http.HandleFunc("/", handle)
+	log.Fatal(http.ListenAndServe(":8080", nil))
+}
+
+func handle(w http.ResponseWriter, r *http.Request) {
+	ctx, _ := tag.New(context.Background())
+	requestReceived := time.Now()
+	// Records latency for failure OR success.
+	// [START monitoring_sli_metrics_opencensus_latency]
```
tbpg, line 103: The `requestReceived` declaration should be within the region.

monitoring/opencensus/main.go
yuriatgoogle: > No, I don't think that comment is needed. Just something to think about as you're writing the sample and embedding it on a page -- it should be clear to users how to use the code.

Yes, the page (which is [live](https://cloud.google.com/stackdriver/docs/solutions/slo-monitoring/sli-metrics/overview?hl=en) now!) has detailed instructions on how to use the code.

> monitoring_sli_metrics_opencensus_view covers two different scopes (top-level and inside func main). I think that should be avoided as it will look like those pieces of code are right next to each other when embedded.

Yes, that's right.  It's actually fine to have those pieces of code next to each other in a general context.  The views COULD be declared within `func main()`, I just didn't think it would make for elegant code.

monitoring/opencensus/main.go
yuriatgoogle: Fixed, thank you!

monitoring/opencensus/main.go
yuriatgoogle: Ah, thank you for explaining that.  Fixed.

monitoring/opencensus/main.go
yuriatgoogle: Right, got it - thank you! Fixed.

monitoring/opencensus/main.go
```diff
@@ -0,0 +1,125 @@
...
+	// [START monitoring_sli_metrics_opencensus_counts]
+	// Counts the request.
+	stats.Record(ctx, requestCount.M(1))
+
+	// Randomly fails 10% of the time.
+	if rand.Intn(100) > 90 {
+		// Counts the error.
+		stats.Record(ctx, failedRequestCount.M(1))
+		// [END monitoring_sli_metrics_opencensus_counts]
+		fmt.Fprintf(w, "intentional error!")
```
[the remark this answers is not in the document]
yuriatgoogle: That's how I had it originally.  However, that then requires that the test expects a 500 some of the time, which could obfuscate, for example, breaking changes, since it would just treat crashes as expected.  So, I wrote the test to validate the body response and return a 200 in either case.

monitoring/opencensus/main.go
yuriatgoogle: Yes, thank you - fixed!

monitoring/opencensus/main.go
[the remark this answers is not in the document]
tbpg: I suppose it doesn't really matter because it isn't part of the sample. But, the only way there would be a 500 is if it gets set in your handler. Any errors or panics or anything would show up in a different way. So, you should be confident that's the only place the 500 comes from.

tbpg (approved):
One minor optional comment. Overall, LGTM.

yuriatgoogle:
> One minor optional comment. Overall, LGTM.

Thank you so much - I'm going to leave the response code as is.  Would you mind resolving the conflicts and merging?

PR #2022, opened 2021-03-25
feat(pubsub): add schema samples

hongalex:
Adds admin samples as well as publishing/subscribing samples with avro and protobuf


pubsub/schemas/create_proto_schema.go
```diff
@@ -0,0 +1,55 @@
...
+	"io"
+	"io/ioutil"
+
+	"cloud.google.com/go/pubsub"
+)
+
+func createProtoSchema(w io.Writer, projectID, schemaID, protoFile string) error {
+	// projectID := "my-project-id"
+	// schemaID := "my-schema"
+	// protoFile = "path/to/a/proto/schema/file(.proto)/formatted/in/protocol/buffers"
```
shollyman, line 30: Could we document this in a bit more detail, or link to some docs that's more specific about what's allowed?

e.g. what happens if I provide a proto file with multiple message definitions, or it contains rpc definitions, etc.

pubsub/schemas/create_topic_with_schema.go
```diff
@@ -0,0 +1,51 @@
...
+	"io"
+
+	"cloud.google.com/go/pubsub"
+)
+
+func createTopicWithSchema(w io.Writer, projectID, topicID, schemaID string, encoding pubsub.SchemaEncoding) error {
+	// projectID := "my-project-id"
+	// topicID := "my-topic"
+	// schemaID := "my-schema-id"
+	// encoding := pubsub.EncodingJSON // or pubsub.EncodingBinary
```
shollyman, line 30: mid-line slashes seem like formatting error

pubsub/schemas/publish_proto_messages.go
```diff
@@ -0,0 +1,78 @@
...
+	// Get the topic encoding type.
+	t := client.Topic(topicID)
+	cfg, err := t.Config(ctx)
+	if err != nil {
+		return fmt.Errorf("topic.Config err: %v", err)
+	}
+	encoding := cfg.SchemaSettings.Encoding
+
+	var msg []byte
+	switch encoding {
```
shollyman, line 52: Do you need to worry about ascii/textproto as well, or only binary and json?

pubsub/schemas/get_schema.go
```diff
@@ -0,0 +1,46 @@
...
+	// projectID := "my-project-id"
+	// schemaID := "my-schema"
+	ctx := context.Background()
+	client, err := pubsub.NewSchemaClient(ctx, projectID)
+	if err != nil {
+		return fmt.Errorf("pubsub.NewSchemaClient: %v", err)
+	}
+	defer client.Close()
+
+	// Retrieve the full schema view. If you don't want to retrive the
```
shollyman, line 36: should this getSchema function have a bool argument  like getFull?  it's not clear what the difference in projection is from the comment.

pubsub/schemas/subscribe_proto_messages.go
```diff
@@ -0,0 +1,73 @@
...
+	// projectID := "my-project-id"
+	// subID := "my-sub"
+	// protoFile = "path/to/a/proto/schema/file(.proto)/formatted/in/protocol/buffers"
+	ctx := context.Background()
+	client, err := pubsub.NewClient(ctx, projectID)
+	if err != nil {
+		return fmt.Errorf("pubsub.NewClient: %v", err)
+	}
+
+	state := &statepb.State{}
```
shollyman, line 41: maybe a comment here explaining that this is creating an instance of the message to be decoded?  First time I read this I thought this was related to pubsub state, not a message about a US state.

pubsub/schemas/schemas_test.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	// Cleanup resources from the previous tests.
+	// This includes schemas, topics, and subscriptions.
+	once.Do(func() {
+		scs, err := listSchemas(io.Discard, tc.ProjectID)
+		if err != nil {
+			t.Fatalf("failed to list schemas: %v", err)
+		}
+		for _, sc := range scs {
```
shollyman, line 66: Do you only want to delete resources that have the appropriate test prefixes (applies to topics and subs as well)?

pubsub/schemas/schemas_test.go
```diff
@@ -0,0 +1,349 @@
...
+			schemaName := strings.Split(sc.Name, "/")
+			deleteSchema(io.Discard, tc.ProjectID, schemaName[len(schemaName)-1])
+		}
+
+		go func() {
```
shollyman, line 71: this seems like it could run late and kill resources under test due to late execution?  Should these goroutines be in a waitgroup or some other control structure?  once.Do() ensures they run only once, but it doesn't ensure they complete before returning.

pubsub/schemas/publish_proto_messages.go
anguillanneuf: Only binary and Json are supported now, but others may get added later.

pubsub/schemas/get_schema.go
hongalex: I think we should keep this as is for flexibility (and is what is specified in the proto). Added a clarifying comment about differences between SchemaViewFull vs Basic.

pubsub/schemas/create_proto_schema.go
hongalex: I linked to the docs for now, but it doesn't provide a lot of context on how the proto definition works. I'll bring this up with the docs team at the next meetings.

pubsub/schemas/create_avro_schema.go
```diff
@@ -28,6 +28,7 @@ import (
 func createAvroSchema(w io.Writer, projectID, schemaID, avscFile string) error {
 	// projectID := "my-project-id"
 	// schemaID := "my-schema"
+	// For more info about the avro schemas see: https://cloud.google.com/pubsub/docs/schemas#avro
```
anguillanneuf, line 31: This link doesn't take me to a specific section on this page, where should it point it?

Since the sample will appear on https://cloud.google.com/pubsub/docs/schemas, is it necessary to add this comment?

hongalex: Hm, that's fair. I added this because of Seth's comment about protos (how do you specify a proto schema?). Do we have a guide to writing avro/proto schemas and if not, I'll probably remove this comment.

hongalex: Per discussion on how-to guide, we'll probably create a blogpost/tutorial at some point in the future but for now, this feature is targeted at the majority of users who are already using schemas so this should not be new.

shollyman (approved):

PR #2034, opened 2021-04-02
fix(dataproc): increase retry count and delay for workflow templates, add UUID to resource names

bradmiro:
https://github.com/GoogleCloudPlatform/golang-samples/issues/2032 and similar are popping up due to race conditions. 
Retries were added to workflow sample, and UUIDs added to resource names for the rest of the samples.


dataproc/dataproc_test.go
```diff
@@ -63,7 +63,7 @@ func TestDataproc(t *testing.T) {
 
 	buf := new(bytes.Buffer)
 
-	testutil.Retry(t, 3, 30*time.Second, func(r *testutil.R) {
```
tbpg, line 66: Retries aren't working well. So, let's give UUIDs a try.

tbpg (changes requested):
Thank you!

dataproc/dataproc_test.go
bradmiro: Done. I did add retries to the workflow sample, as I can't easily slot UUID into the sample without modifying it, which I would like to avoid if possible.

dataproc/instantiate_inline_workflow_template_test.go
```diff
@@ -29,13 +30,14 @@ func TestInstantiateInlineWorkflowTemplate(t *testing.T) {
...
-		t.Fatalf("instantiateInlineWorkflowTemplate got err: %v", err)
-	}
-
-	got := buf.String()
-	if want := fmt.Sprintf("successfully"); !strings.Contains(got, want) {
-		t.Fatalf("instantiateInlineWorkflowTemplate got %q, want %q", got, want)
-	}
+	testutil.Retry(t, 5, 60*time.Second, func(r *testutil.R) {
+		if err := instantiateInlineWorkflowTemplate(buf, tc.ProjectID, region); err != nil {
+			t.Fatalf("instantiateInlineWorkflowTemplate got err: %v", err)
```
tbpg, line 35:
```suggestion
			r.Errorf("instantiateInlineWorkflowTemplate got err: %v", err)
			return
```

dataproc/instantiate_inline_workflow_template_test.go
```diff
@@ -29,13 +30,14 @@ func TestInstantiateInlineWorkflowTemplate(t *testing.T) {
...
+		}
+
+		got := buf.String()
+		if want := fmt.Sprintf("successfully"); !strings.Contains(got, want) {
+			t.Fatalf("instantiateInlineWorkflowTemplate got %q, want %q", got, want)
```
tbpg, line 40: Same comment.

dataproc/instantiate_inline_workflow_template_test.go
```diff
@@ -29,13 +30,14 @@ func TestInstantiateInlineWorkflowTemplate(t *testing.T) {
 	region := "us-central1"
 
 	buf := new(bytes.Buffer)
```
tbpg, line 32: Initialize `buf` inside the retry so it gets reset each time. Basically, just move this one line down.

bradmiro:
@tbpg all feedback addressed

dataproc/instantiate_inline_workflow_template_test.go
```diff
@@ -28,14 +29,16 @@ func TestInstantiateInlineWorkflowTemplate(t *testing.T) {
...
 
-	buf := new(bytes.Buffer)
+	testutil.Retry(t, 5, 60*time.Second, func(r *testutil.R) {
+		buf := new(bytes.Buffer)
 
-	if err := instantiateInlineWorkflowTemplate(buf, tc.ProjectID, region); err != nil {
-		t.Fatalf("instantiateInlineWorkflowTemplate got err: %v", err)
-	}
+		if err := instantiateInlineWorkflowTemplate(buf, tc.ProjectID, region); err != nil {
+			t.Errorf("instantiateInlineWorkflowTemplate got err: %v", err)
```
tbpg, line 36: Looks like this is still an issue. Should be `r.Errorf` followed by `return` rather than `t.Errorf`.

Using `r` makes it so the retry works.

dataproc/instantiate_inline_workflow_template_test.go
```diff
@@ -28,14 +29,16 @@ func TestInstantiateInlineWorkflowTemplate(t *testing.T) {
...
+		}
 
-	got := buf.String()
-	if want := fmt.Sprintf("successfully"); !strings.Contains(got, want) {
-		t.Fatalf("instantiateInlineWorkflowTemplate got %q, want %q", got, want)
-	}
+		got := buf.String()
+		if want := fmt.Sprintf("successfully"); !strings.Contains(got, want) {
+			t.Errorf("instantiateInlineWorkflowTemplate got %q, want %q", got, want)
```
tbpg, line 41: Same comment.

tbpg (changes requested):

dataproc/instantiate_inline_workflow_template_test.go
bradmiro: Fixed!

dataproc/instantiate_inline_workflow_template_test.go
bradmiro: Fixed!

tbpg (approved):
Thanks!

PR #2043, opened 2021-04-09
feat: add blunderbuss config to cloud iot label

gcseh:


.github/blunderbuss.yml
```diff
@@ -39,3 +39,13 @@ assign_issues_by:
   - 'api: spanner'
   to:
   - skuruppu
+- labels:
+  - 'api: cloudiot'
+  to:
+  - attilagargyan
+
+assign_prs_by:
```
tbpg, line 47: If you add to `CODEOWNERS`, you can skip this. But, it doesn't hurt to have it here, too.

gcseh: thanks!

tbpg (approved):

Issue #1645, opened 2020-08-13
eventarc: use event library struct


grant:
In the Cloud Run Events Pub/Sub sample, we want to use the struct provided by the event library, not an inlined struct:

## Code
https://github.com/GoogleCloudPlatform/golang-samples/blob/292e8c9e4b0dd6c34805431d1a894c2bc751dd7f/run/events_pubsub/main.go#L28-L35

## Event type library
https://github.com/googleapis/google-cloudevents-go/blob/c543fb3cb993d15758f84f43a07e03a9ff62e718/src/google/events/cloud/pubsub/v1/data.pb.go#L42-L54

---

I'm not sure if there are more steps needed to be able to import this file.

My Go knowledge for understanding modules is lacking. Can you give me any tips for next steps needed (if any) for using `google-cloudevents-go` @grayside @tbpg?

grant:
The sample is here now:

https://github.com/GoogleCloudPlatform/golang-samples/tree/master/eventarc/pubsub

The event type is here now:

https://github.com/googleapis/google-cloudevents-go/blob/d7141e1f03c0bb45b9c2606553aca380dd23f2ae/cloud/pubsub/v1/MessagePublishedData.go

Assigning to @michaelawyu who is working on this language.

grant:
Pinging @michaelawyu, can you look into this issue? Ping me if there are questions.

grayside:
To be clear, the suggestion here is that we update the non-Eventarc Cloud Run + Pub/Sub sample to use the event type library? Is the Pub/Sub payload identical between Pub/Sub and Eventarc?

grant:
> To be clear, the suggestion here is that we update the non-Eventarc Cloud Run + Pub/Sub sample to use the event type library? Is the Pub/Sub payload identical between Pub/Sub and Eventarc?

Happy to clarify.

The GitHub issue was create on August 12, 2020 – which is significant because at the time, the API was part of Cloud Run and the product was called "Cloud Run Events".

Now the product is called "Eventarc", and there are separate APIs and samples. I've changed the title respectively.

> To be clear, the suggestion here is that we update the non-Eventarc Cloud Run + Pub/Sub sample to use the event type library?

FR is for Eventarc to use the events library for the eventarc sample. I added the sample to the README at one point:

https://github.com/googleapis/google-cloudevents-go#example-usage

>  Is the Pub/Sub payload identical between Pub/Sub and Eventarc?

[...]

PR #2049, opened 2021-04-12
feat(eventarc): use eventarc library for pubsub sample

grant:
Uses the eventarc Golang library for our Eventarc Pub/Sub type.

Automatic base64 decoding for the pub/sub data is a FR, not included here.

fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/1645

### DIFF
--- a/eventarc/pubsub/go.mod
+++ b/eventarc/pubsub/go.mod
@@ -1,5 +1,0 @@
-module github.com/GoogleCloudPlatform/golang-samples/eventarc/pubsub
-
-go 1.13
-
-require github.com/googleapis/google-cloudevents-go v0.0.0-20210412200750-de059319bac5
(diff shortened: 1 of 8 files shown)


grant:
If we get https://github.com/googleapis/google-cloudevents-go/pull/58 in, then we could include the pubsub data base64 decoding here as well.

grant:
@grayside PTAL at this PR with the updated CE library.

grayside (changes requested):
Unstable event type libraries shouldn't be used in samples.

grant:
Per sync with @grayside, we'll create a GitHub issue to clarify the stability of each library and re-visit this PR with the stability evidence backing this PR up.

grant:
Waiting for Go library to update per @grayside request before re-requesting review for this PR.

PR #2059, opened 2021-04-20
docs(appengine): add comment about authenticated push JWT token validation

fayssalmartanigcp:
This snippet is used in the Cloud Pub/Sub docs (https://cloud.google.com/pubsub/docs/push#validating_tokens) and many users are not aware that signature verification of the token is not enough, the claim needs to be validated also.

See similar change in the nodejs repo: GoogleCloudPlatform/nodejs-docs-samples#2222.


appengine/go11x/pubsub/authenicated_push/main.go
```diff
@@ -112,6 +112,12 @@ func (a *app) receiveMessagesHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// IMPORTANT: you should validate claim details not covered by signature
+	// and audience verification above, including:
+	//   - Ensure that `payload.Claims["email"]` is equal to the expected service
+	//     account set up in the push subscription settings.
+	//   - Ensure that `payload.Claims["email_verified"]` is set to true.
```
tbpg, line 119: Should we check this directly in the code, rather than only in the comment?

tbpg (commented):
Thanks!

appengine/go11x/pubsub/authenicated_push/main.go
fayssalmartanigcp: I don't believe we have the service account email here since it's set up by the user to run these tests.

tbpg: I meant the `email_verified` check, not the `email` itself. Unless I'm missing something?

tbpg: We could also use a placeholder email, like we do above:

```
	// Please change http://example.com to match with the value you are
	// providing while creating the subscription.
```

fayssalmartanigcp: Updated, PTAL.

hongalex (approved):

tbpg (approved):

PR #2073, opened 2021-04-29
chore: bump grpc message send limit to 128MB in sample

shollyman:
grpc-go sets 4MB as the default client receive size, so we still can't just remove this option entirely in the face of large row response messages.

### DIFF
--- a/bigquery/bigquery_storage_quickstart/main.go
+++ b/bigquery/bigquery_storage_quickstart/main.go
@@ -42,10 +42,10 @@ import (
 )
 
 // rpcOpts is used to configure the underlying gRPC client to accept large
-// messages.  The BigQuery Storage API may send message blocks up to 10MB
+// messages.  The BigQuery Storage API may send message blocks up to 128MB
 // in size.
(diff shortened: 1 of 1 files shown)


codyoss (approved):
LGTM, is this a common to have to set RPC settings with working with various RPCs? If so I wonder if we should design a hook for per call options to be manually configured. Similar to how we a [hook for per-client options](https://github.com/googleapis/google-cloud-go/wiki/Customizing-constructors).

shollyman:
> LGTM, is this a common to have to set RPC settings with working with various RPCs? If so I wonder if we should design a hook for per call options to be manually configured. Similar to how we a [hook for per-client options](https://github.com/googleapis/google-cloud-go/wiki/Customizing-constructors).

This one in particular needs to be on a streaming call rather than a unary rpc, and I don't necessarily like rightsizing message sizes in this fashion at all.  We're subject to user data shape here, so almost no one will actually use messages of this size as it means they have a  single row > 100 MB.

PR #2080, opened 2021-05-07
fix(all): always call client.Close when creating a client

### LINKED ISSUE #2078: all: many samples don't call client.Close(), leading to memory leaks
The code samples for Secret Manager have "client, err := secretmanager.NewClient(ctx)" but no corresponding "defer client.Close()"  which resulted in my application, which used the code in the samples as a template, having a memory leak.

It'd be great if the sample code could make it clear in the code or comments that .Close() should be called.

Thanks!

tbpg:
Generated with:

```
$ grep -L Close $(grep -El 'New[^(]*Client' **/*.go) | grep -v test | xargs sed -i '/New[^(]*Client/,/}/s/}/}\ndefer client.Close()/'
$ goimports -w .
```

There were ~30 places where `defer client.Close()` was not correct (e.g. when you want to keep a client around in a Cloud Function), so I didn't update those.

I verified `goimports` and `go vet` are happy:
```
$ find . -name go.mod -execdir go vet ./... \;
```

Fixes #2078.

### DIFF
--- a/appengine_flexible/datastore/datastore.go
+++ b/appengine_flexible/datastore/datastore.go
@@ -42,6 +42,7 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+	defer client.Close()
 
 	http.HandleFunc("/", handle)
 	appengine.Main()
--- a/appengine_flexible/datastore/datastore.go
+++ b/appengine_flexible/datastore/datastore.go
@@ -42,7 +42,6 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	defer client.Close()
 
 	http.HandleFunc("/", handle)
 	appengine.Main()
--- a/bigtable/garbagecollection/create_family_gc_intersect.go
+++ b/bigtable/garbagecollection/create_family_gc_intersect.go
@@ -34,7 +34,7 @@ func createFamilyGCIntersect(w io.Writer, projectID, instanceID string, tableNam
 	if err != nil {
 		return fmt.Errorf("bigtable.NewAdminClient: %v", err)
 	}
-	defer client.Close()
+	defer adminClient.Close()
 
 	columnFamilyName := "cf4"
 	if err := adminClient.CreateColumnFamily(ctx, tableName, columnFamilyName); err != nil {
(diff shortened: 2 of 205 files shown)


appengine_flexible/datastore/datastore.go
```diff
@@ -42,6 +42,7 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+	defer client.Close()
```
codyoss, line 45: Looks like this should be datastoreClient. Might be other cases like this.

bigtable/garbagecollection/create_family_gc_intersect.go
```diff
@@ -34,6 +34,7 @@ func createFamilyGCIntersect(w io.Writer, projectID, instanceID string, tableNam
 	if err != nil {
 		return fmt.Errorf("bigtable.NewAdminClient: %v", err)
 	}
+	defer client.Close()
```
codyoss, line 37: another example of above comment.

functions/spanner/spanner.go
```diff
@@ -47,6 +47,7 @@ func HelloSpanner(w http.ResponseWriter, r *http.Request) {
 			log.Printf("spanner.NewClient: %v", err)
 			return
 		}
+		defer client.Close()
```
codyoss, line 50: This would close close too soon as is.

codyoss (changes requested):
Only spot checked a couple. We will want to double check all of the serverless examples as those are written a little different than "standard" examples

functions/spanner/spanner.go
tbpg: Done. Great catch.

appengine_flexible/datastore/datastore.go
tbpg: Done.

bigtable/garbagecollection/create_family_gc_intersect.go
tbpg: Done.

tbpg (commented):
Thanks! Please take another look.

codyoss (approved):
Thanks, this is great!

logging/simplelog/simplelog.go
```diff
@@ -48,11 +48,13 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to create logging client: %v", err)
 	}
+	defer client.Close()
 
 	adminClient, err := logadmin.NewClient(ctx, projID)
 	if err != nil {
 		log.Fatalf("Failed to create logadmin client: %v", err)
 	}
+	defer client.Close()
```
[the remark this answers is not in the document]
nicoleczhu: updated with a commit

nicoleczhu (approved):

tbpg:
Great catch! Thank you!

PR #2092, opened 2021-05-26
docs(datastore): improved cursor sample in docs

craiglabenz:
New content of region tags include origin and iteration of `cursorStr` variable

### DIFF
--- a/datastore/snippets/snippet_test.go
+++ b/datastore/snippets/snippet_test.go
@@ -537,7 +537,7 @@ func SnippetIterator_Cursor() {
 	// When run within a loop, this yields the cursor to continue reading results
 	cursorStr = nextCursor.String()
 	// [END datastore_cursor_paging]
-	_ = err        // Check the error.
+	_ = err // Check the error.
 }
 
 func SnippetQuery_EventualConsistency() {
--- a/datastore/snippets/snippet_test.go
+++ b/datastore/snippets/snippet_test.go
@@ -508,36 +508,42 @@ func SnippetIterator_Cursor() {
 	client, _ := datastore.NewClient(ctx, "my-proj")
 	defer client.Close()
 	// [START datastore_cursor_paging]
-	cursorStr := ""
-	const pageSize = 5
-	query := datastore.NewQuery("Tasks").Limit(pageSize)
-	if cursorStr != "" {
-		cursor, err := datastore.DecodeCursor(cursorStr)
-		if err != nil {
-			log.Fatalf("Bad cursor %q: %v", cursorStr, err)
+	for {
+		cursorStr := ""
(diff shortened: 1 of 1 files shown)


datastore/snippets/snippet_test.go
```diff
@@ -534,9 +534,10 @@ func SnippetIterator_Cursor() {
 
 	// Get the cursor for the next page of results.
 	nextCursor, err := it.Cursor()
+	// When run within a loop, this yields the cursor to continue reading results
```
tbpg, line 537: I don't understand this sample. There is no loop, but there should be? Should we add one?

I'm confused -- `cursorStr` is set to `""`, then immediately checked `if cursorStr != ""`.

tbpg (changes requested):

datastore/snippets/snippet_test.go
craiglabenz: My thinking for the sample was simply to show all variable initialization, but it could make sense to actually set up a loop. I'll have to further examine the implications for tests, however.

craiglabenz: I've updated this sample to exist within a loop, but I'm not actually certain this samples repo is set up to test everything like it claims.

datastore/snippets/snippet_test.go
```diff
@@ -507,36 +507,43 @@ func SnippetIterator_Cursor() {
...
-	query := datastore.NewQuery("Tasks").Limit(pageSize)
-	if cursorStr != "" {
-		cursor, err := datastore.DecodeCursor(cursorStr)
-		if err != nil {
-			log.Fatalf("Bad cursor %q: %v", cursorStr, err)
+	for {
+		cursorStr := ""
+		const pageSize = 5
+		query := datastore.NewQuery("Tasks").Limit(pageSize)
+		if cursorStr != "" {
```
tbpg, line 515: I think at this point, `cursorStr` will always be `""`. Should `cursorStr` be initialized outside the loop?

datastore/snippets/snippet_test.go
```diff
@@ -507,36 +507,43 @@ func SnippetIterator_Cursor() {
...
-		_, err = it.Next(&task)
-	}
-	if err != iterator.Done {
-		log.Fatalf("Failed fetching results: %v", err)
-	}
+		// Read the tasks.
+		var tasks []Task
+		var task Task
+		it := client.Run(ctx, query)
+		_, err := it.Next(&task)
```
tbpg, line 527: Could this iterator have many entries, not just the one that gets read?

tbpg (changes requested):

craiglabenz:
Closed in favor of #2173

PR #2093, opened 2021-06-01
opentelemetry: export traces in batch rather than synchronously

punya:
Synchronous export carries a very heavy performance penalty (added
latency for each request), whereas batching with default options only
adds a 5s delay before a given trace is exported.

With this change, we guide users towards the right starting point. If
they have more specialized needs that call for synchronous export, they
can learn more from the upstream OTel docs.


punya:
@ahmetb @dashpole @jsuereth PTAL

tbpg (approved):

PR #2094, opened 2021-06-09
feat(spanner): update samples for optimizer statistics package

hengfengli:


spanner/spanner_snippets/spanner/spanner_create_client_with_query_options.go
```diff
@@ -30,7 +30,10 @@ import (
 func createClientWithQueryOptions(w io.Writer, database string) error {
 	ctx := context.Background()
 	queryOptions := spanner.QueryOptions{
-		Options: &sppb.ExecuteSqlRequest_QueryOptions{OptimizerVersion: "1"},
+		Options: &sppb.ExecuteSqlRequest_QueryOptions{
+			OptimizerVersion:           "1",
+			OptimizerStatisticsPackage: "auto_20191128_14_47_22UTC",
```
olavloite, line 35: Will this value always be accepted by the backend? I agree that it would be best to show an actual value here, instead of only showing how to set something like `latest`, but I don't know if this will fail during testing on (some) databases.

hengfengli: I think it will always be valid. I replicated this from https://github.com/googleapis/java-spanner/pull/1239.

olavloite: Good point, it worked there, so it should work here as well.

olavloite (approved):
I'm happy with the samples. I'm not sure whether the changes to the non-Spanner `go.mod` files should be included in this PR.

hengfengli:
> I'm happy with the samples. I'm not sure whether the changes to the non-Spanner `go.mod` files should be included in this PR.

The changes have to be in this PR. Otherwise, `Go / Build and Lint` check would fail.

hengfengli:
@tbpg Can you please help us to merge this PR? Thanks!

hengfengli:
@JustinBeckwith The failed test is not related to our change: 

```
--- FAIL: TestCreateBucketClassLocation (11.72s)
    buckets_test.go:55: deleteBucket: Bucket("golang-samples-tests-storage-buckets-tests-attrs").Delete: context deadline exceeded
```

codyoss (approved):

hengfengli:
@codyoss Can you help me to merge it? I don't have the permission to do it. Thanks!

spanner/spanner_snippets/spanner/spanner_create_client_with_query_options.go
olavloite: Hmmm.... It seems that this (sometimes) goes wrong: https://github.com/googleapis/java-spanner/issues/1273#issuecomment-879742347

hengfengli: Interesting. It's supposed to always have it.

PR #2095, opened 2021-06-09
fix(opentelemetry): use the right flush method

ahmetb:
According to internal chats, Shutdown() is not meant to be called by the user,
ForceFlush() is the right thing to do here.

Context: http://doc/1R82XXlE8dVt1C-kLYRH2gFLVCzSgKiRna0yroZTsjhs/edit?resourcekey=0-iYHnmYFHYuh8ltcMbPFxhw&disco=AAAAMyjSPL0

cc: @aabmass @punya

### DIFF
--- a/opentelemetry/trace/main.go
+++ b/opentelemetry/trace/main.go
@@ -37,7 +37,6 @@ func main() {
 	if err != nil {
 		log.Fatalf("texporter.NewExporter: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


punya (approved):
Thanks for catching this @ahmetb.

aabmass (approved):
LGTM as well, but needs approval from @tbpg or another owner

codyoss (approved):

PR #2098, opened 2021-06-17
fix(functions): use golang image for CI/CD example

### LINKED ISSUE #1848: functions: CI/CD example
[The example for cloud functions CI/CD](https://github.com/GoogleCloudPlatform/golang-samples/blob/a12e2fd198c022aa062c6897713347e1816f737d/functions/ci_cd/cloudbuild.yaml#L17) doesn't feel right. I'm just starting to learn how to use the service and golang so I might be completely wrong, but:
1. It pulls golang image directly from mirror.gcr.io. According to [mirror documentation](https://cloud.google.com/container-registry/docs/pulling-cached-images):
- _Google Cloud services such as Cloud Build and Google Kubernetes Engine automatically check for cached images before attempting to pull an image from Docker Hub._
- _Only obtain cached images on mirror.gcr.io by configuring the Docker daemon. A request to pull directly from mirror.gcr.io will fail if a cached copy of the image does not exist._
2. It will pull latest (currently 1.15.5) golang image. Cloud functions support 1.11.6 and 1.13.8 runtimes ([docs](https://cloud.google.com/functions/docs/concepts/go-runtime#execution_environment)) using custom Ubuntu 18.04 based images.

There are a few options for making it better:
1. Just change the image to golang:1.13.8 (example uses go113 runtime) and pull it from docker hub (using the mirror transparently).
2. Extend the example by creating a custom Ubuntu based image with golang 1.13.8. This adds complexity, especially if we don't want to rebuild it on every CI trigger, but it's a step closer to production environment than official debian based images from docker hub.
3. This would be a feature request, but IMO the best solution from CI best practices perspective: publish go113 and go111 runtimes that are used in the production environment (or new images synced with the production if they contain any secret sauce).

tbpg:
Fixes #1848.

### DIFF
--- a/functions/ci_cd/cloudbuild.yaml
+++ b/functions/ci_cd/cloudbuild.yaml
@@ -14,9 +14,9 @@
 
 # [START functions_ci_cd_cloud_build]
 steps:
-- name: 'mirror.gcr.io/library/golang'
+- name: golang
   args: ['go', 'version']
-- name: 'mirror.gcr.io/library/golang'
+- name: golang
   args: ['go', 'test']
   dir: 'function/dir/from/repo/root'
 - name: 'gcr.io/cloud-builders/gcloud'


codyoss (approved):
LGTM

PR #2102, opened 2021-06-18
chore(.github): split build and lint into multiple jobs

### LINKED ISSUE #2101: all: consider splitting GitHub Actions to speed up processing
See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/.github/workflows/go.yaml. We could split the `build` job into more than one job so the steps can run in parallel, thereby running more quickly.

tbpg:
This reduces the Actions runtime from ~6.5 minutes to ~5 minutes. Not ideal, but the slowest one is `vet`, which can't really be sped up any more, from what I can tell (`vet` each module in parallel?)

Fixes #2101.


codyoss (approved):
LGTM

tbpg:
I updated the repo settings to require the new set of Actions checks.

PR #2104, opened 2021-06-18
fix(deps): revert version contrib.go.opencensus.io/exporter/stackdriver

shollyman:
https://github.com/GoogleCloudPlatform/golang-samples/pull/2083 bumped
this, but didn't test older go versions which contain conflict related
to the time package.

Subsequent changes to address tested versions handled in other PRs.

PR #2105, opened 2021-06-18
fix(workflows): test with both go 1.16 and 1.11

shollyman:
Switches github actions to matrix builds so we can test multiple go versions.

Follows on from https://github.com/GoogleCloudPlatform/golang-samples/pull/2104 to ensure we're testing more broadly across go versions, as we also invoke these tests as part of google-cloud-go.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -54,6 +54,9 @@ jobs:
   test:
     name: Root tests ${{ matrix.go }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go: [ '1.16', '1.11' ]
     steps:
       - name: Setup Go
         uses: actions/setup-go@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yaml
```diff
@@ -8,13 +8,16 @@ on:
       - master
 jobs:
   build:
-    name: Build and Lint
+    name: Build and Lint ${{ matrix.go }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go: [ '1.16' ]
```
tbpg, line 15: I don't think this has the intended effect. The root tests basically never change and shouldn't be a problem with 1.11.

However, some modules in this repo use a Go version > 1.11. We shouldn't try to build those with 1.11, obviously.

So, I think this needs to be done in Kokoro. Why didn't the upgrade PR break when we brought in a dependency that depended on Go > 1.11? It did for a different dependency (x/net).

tbpg (changes requested):

.github/workflows/go.yaml
shollyman: I'll go ahead and drop this PR.

PR #2117, opened 2021-06-24
docs(functions): updated gcp logo

iamrajiv:
docs(functions): updated gcp logo


averikitsch:
Feel free to reopen, but it seems like this change is unnecessary.

PR #2118, opened 2021-06-24
fix: add retries to run test

averikitsch:
Fixes issue #2106


averikitsch:
Help needed to fix lint

run/testing/h2c.e2e_test.go
```diff
@@ -49,7 +50,10 @@ func TestHTTP2Server(t *testing.T) {
 		t.Fatalf("Service.ParsedURL: %v", err)
 	}
 
-	h2Client := &http.Client{Transport: &http2.Transport{}}
+	retryClient := retryablehttp.NewClient()
```
tbpg, line 53: Don't think we want to add another dependency. We can use `testutil.Retry` for retries.

PR #2119, opened 2021-06-25
feat(spanner): add samples for low-cost instances

hengfengli:

PR #2122, opened 2021-06-25
test(spanner): use a separate instance for each test

olavloite:
Creates a separate Cloud Spanner instance for each test. The instance is deleted after the test is executed, and otherwise cleaned up by a later test run if the test fails to delete the instance. All tests are marked as parallel to improve test execution speed. Having each test use a separate instance also reduces the probability that the backup tests slow each other down, as backup operations are limited to 1 per instance. With each test having its own instance, the backup tests should no longer need to wait for each other.


hengfengli:
@olavloite How many instances do we need? My concern is that we may have a limit for the number of instances that we can create in a project.

olavloite:
> @olavloite How many instances do we need? My concern is that we may have a limit for the number of instances that we can create in a project.

The current setup will create as many instances as that there are Spanner sample tests that may run in parallel. As the current integration test contains 5 test functions, it means that it could create up to 5 instances at the same time.

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -604,6 +541,102 @@ func mustRunSample(t *testing.T, f sampleFuncWithContext, dbName, errMsg string)
 	return b.String()
 }
 
+func createTestInstance(t *testing.T) (instanceName string, cleanup func()) {
```
hengfengli, line 544: Shouldn't we directly pass `projectID` into this function? The current way looks confusing that we parse the projectID from the instanceName and then we construct a new instanceName.

hengfengli: What I mean is to move `parseInstanceName` outside of this function to get the projectID first and then call this function.

olavloite: Yes, this is a little confusing / strange in order to keep the tests behave the same as now using the same environment variables. I've moved the instance name parsing outside the `createTestInstance` function to make it a little less confusing (albeit with a little code duplication).

hengfengli:
Also, another strange point is that although we specify an instance in `GOLANG_SAMPLES_SPANNER`, but we will never use it. Instead, we always create new instances under the project.

hengfengli:
Okay, it's used somewhere else. Can we leave a comment to explain why we are doing this?

hengfengli (approved):

olavloite:
> Okay, it's used somewhere else. Can we leave a comment to explain why we are doing this? I mean we are not using the instance in `GOLANG_SAMPLES_SPANNER`.

I've added a comment for it, but I also changed it a little bit. The environment variable is used for other samples, but only the project id is used in these tests, so it does not make sense to have a method that returns the entire instance name and let the individual test methods extract the project id from it. Instead I've changed the method to only return the project id and documented why we are doing that.

olavloite:
@tbpg This PR has the same problem with the linting as [this PR](https://github.com/GoogleCloudPlatform/golang-samples/pull/2119#issuecomment-868157817). What is the best way to deal with this?

codyoss:
@olavloite This should be fixed if you update the branch

codyoss (approved):

Issue #2125, opened 2021-06-28
bug: functions/security/idtoken.go


bweston92:
The following doesn't seem to work:

```
https://github.com/GoogleCloudPlatform/golang-samples/blob/HEAD/functions/security/idtoken.go
```

by default using a service account file with it I get the following:

```
neither scopes or audience are available for the self-signed JWT
```

I can get it to make a request but have to use `option.WithAudiences(targetURL)` in the call to `idtoken.NewClient` and it works fine.

Is this desired behaviour?
If so can improve these documents?
If not, where should I list a bug?

codyoss:
This is being tracked on [google-api-go-client](https://github.com/googleapis/google-api-go-client/issues/1092) and has a [fix out for review](https://github.com/googleapis/google-api-go-client/pull/1096). For now please use google.golang.org/api@v0.48.0, sorry for the inconvenience.

PR #2127, opened 2021-06-30
fix: update google.golang.org/api to v0.50.0

codyoss:
Only updated modules that used idtoken and could have experienced this issue.

Fixes: #2125


tbpg (changes requested):
Looks like there are some empty files? Also, are we sure we need to add require lines where they weren't there before?

tbpg (approved):

PR #2130, opened 2021-07-01
feat(secretmanager): add examples for using etags

gwhitehawk:


secretmanager/delete_secret_with_etag.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"context"
+	"fmt"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
+)
+
+// deleteSecretWithEtag deletes the secret with the given name and all of its versions.
+func deleteSecretWithEtag(name string, etag string) error {
```
codyoss, line 27: nit: A similar thing applys to all samples

```suggestion
func deleteSecretWithEtag(name, etag string) error {
```

secretmanager/delete_secret_with_etag.go
```diff
@@ -0,0 +1,51 @@
...
+	// name := "projects/my-project/secrets/my-secret"
```
codyoss, line 28: nit: can you also please provide a sample etag value.

codyoss (commented):
Just a couple nits or else it LGTM

secretmanager/delete_secret_with_etag.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
+)
+
+// deleteSecretWithEtag deletes the secret with the given name and all of its versions.
+func deleteSecretWithEtag(name, etag string) error {
+	// name := "projects/my-project/secrets/my-secret"
+	// etag := "\"123\""
```
codyoss, line 29: Should this value actually be a quoted string? If so maybe use the a ` character quotation to avoid having to escape.

gwhitehawk: Yes, it is supposed to be quoted as per https://google.aip.dev/154. I'll fix the notation. Let me also squash the commits, not sure why the ops from commandline don't show up in the ui.

codyoss: We always squash on merge so don't worry about the history 😄

gwhitehawk: Fixed now.

codyoss (approved):
LGTM, thank you!

sethvargo (approved):

PR #2146, opened 2021-07-12
chore(asset): try to reduce flakiness

codyoss:
Double the amount of reties and add the force flag to overwrite
values if present.

Fixes: #2142

### DIFF
--- a/asset/quickstart/export-assets-bigquery/main.go
+++ b/asset/quickstart/export-assets-bigquery/main.go
@@ -45,6 +45,7 @@ func main() {
 				BigqueryDestination: &assetpb.BigQueryDestination{
 					Dataset: dataset,
 					Table:   "test",
+					Force:   true,
 				},
 			},
 		},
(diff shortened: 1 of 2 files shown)


asset/quickstart/export-assets-bigquery/main.go
```diff
@@ -45,6 +45,7 @@ func main() {
 				BigqueryDestination: &assetpb.BigQueryDestination{
 					Dataset: dataset,
 					Table:   "test",
+					Force:   true,
```
tbpg, line 48: Are we sure we want this in the sample code? Do we need a comment?

tbpg (changes requested):

asset/quickstart/export-assets-bigquery/main.go
codyoss: It is in the sample for the other two languages that have a sample for this (Java and Python). I can add a comment, but it would just be restating what the docs say for this field.

tbpg: Sounds good. Thanks!

tbpg (approved):

Issue #2139, opened 2021-07-08
Tokens expire after 1 hour - use Tokensource instead.


jankrynauw:
https://github.com/GoogleCloudPlatform/golang-samples/blob/42a8597d5beb089c313eb618ac1363af5e404be0/run/grpc-ping/connection.go#L30

The current example manually adds a token to each client request.  For some longer running services, this becomes an issue since the client is usually instantiated when launching the service, requests made after 1 hour of the initial client instantiation will fail since the token will have expired by then.

Rather use the TokenSource methodology where the underlying token is validated and only refreshed when needed.

Here is a revised `NewConn` method illustrating a potential solution to the above:
```go
type grpcTokenSource struct {
	oauth.TokenSource
}

// NewConn creates a new gRPC connection.
// host should be of the form domain:port, e.g., example.com:443
func NewConn(ctx context.Context, host string, insecure bool) (*grpc.ClientConn, error) {
	var opts []grpc.DialOption
	if host != "" {
		opts = append(opts, grpc.WithAuthority(host))
	}

	if insecure {
		opts = append(opts, grpc.WithInsecure())
	} else {
		systemRoots, err := x509.SystemCertPool()
		if err != nil {
			return nil, err
		}
		cred := credentials.NewTLS(&tls.Config{
			RootCAs: systemRoots,
		})
		opts = append(opts, grpc.WithTransportCredentials(cred))
	}

	// use a tokenSource to automatically inject tokens with each underlying client request
	audience := "https://" + strings.Split(host, ":")[0]
	tokenSource, err := idtoken.NewTokenSource(ctx, audience, option.WithAudiences(audience))
	if err != nil {
		return nil, status.Errorf(
			codes.Unauthenticated,
			"NewTokenSource: %s", err,
		)
	}
	opts = append(opts, grpc.WithPerRPCCredentials(grpcTokenSource{
		TokenSource: oauth.TokenSource{
			tokenSource,
		},
	}))

	return grpc.Dial(host, opts...)
}
```

grayside:
Hello jankrynauw,

Thanks for the issue and coding up a proposed improvement, that is very generous with your time.

To ensure I understand what you're saying, you think this code sample should 
properly handle token refreshing, a limitation of the current code pointed out in:

https://github.com/GoogleCloudPlatform/golang-samples/blob/42a8597d5beb089c313eb618ac1363af5e404be0/run/grpc-ping/request_auth.go#L39-L42

[...]

PR #2153, opened 2021-07-15
docs(run): Tokens expire after 1 hour - use Tokensource instead.

jankrynauw:
The current example manually adds a token to each client request. For some longer running services, this becomes an issue since the client is usually instantiated when launching the service, requests made after 1 hour of the initial client instantiation will fail since the token will have expired by then.

Rather use the TokenSource methodology where the underlying token is validated and only refreshed when needed.

### DIFF
--- a/run/grpc-ping/connection.go
+++ b/run/grpc-ping/connection.go
@@ -18,8 +18,15 @@ package main
 // [START run_grpc_conn]
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"google.golang.org/api/idtoken"
+	"google.golang.org/api/option"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/oauth"
+	"google.golang.org/grpc/status"
+	"strings"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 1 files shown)


jankrynauw:
@grayside you mentioned that you would like to see the changes made to the `request_auth.go` file, and not inside the `NewConn()` function.   Not sure whether this is possible, without changing the signature of the `pingRequestWithAuth` method - which receives `conn *grpc.ClientConn` parameter.  The Tokensource is added via the options with the `grpc.Dial(host, opts...)` method which returns a connection.

grayside (changes requested):
My understanding of the current implementation is a new token is being created for every request, rather than reusing the tokens for their 1 hour lifespan.

With that in mind, I think we need to expand the token source implementation in request_auth.go so the TokenSource can be reused between requests.

I'd like to constraint the significant changes to request_auth.go because that keeps the sample aligned with the documentation: https://cloud.google.com/run/docs/triggering/grpc#request-auth

I appreciate your willingness to contribute, and acknowledge that coding for the carefully scoped spaces in our documentation can be an awkward requirement. If this exceeds your time, I will carry forward the plan.

muncus:
In light of @grayside 's comments above, it seems like we want to address this in a different way, so i'll reopen your initial issue #2139 and close this PR. All updates (and PRs) will be linked to your issue, so you can stay up to date on our progress.
Thanks for your patience!

PR #1893, opened 2021-01-13
fix(endpoints): update gke deployment apiVersion to apps/v1

duluong:
Deprecated APIs was removed from 1.16

PR #2157, opened 2021-07-21
fix(endpoints): update sample deployment apiVersion to apps/v1

lescoggi:
Deprecated APIs was removed from 1.16.

Similar changes made in #1893 need to be applied to this sample.


lescoggi:
Apologies, not sure if I need to do anything for the Kokoro CI check. Let me know.

Issue #2162, opened 2021-07-22
storage: TestPostPolicyV4 failed


codyoss:
See full log [here](https://source.cloud.google.com/results/invocations/8110b3b8-edbf-468a-ba7b-750304ce2736/targets).

Failure message:

```
=== RUN   TestPostPolicyV4
    objects_test.go:488: status code in response mismatch: got 404 want 204
        Body: HTTP/2.0 404 Not Found
        Content-Length: 188
        Content-Type: application/xml; charset=UTF-8
        Date: Thu, 22 Jul 2021 16:08:07 GMT
        Server: UploadServer
        Vary: Origin
        X-Guploader-Uploadid: XXXXXXXXX

        <?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist.</Message><Details>The specified bucket does not exist.</Details></Error>
    objects_test.go:501: Failed to retrieve attributes: storage: object doesn't exist
--- FAIL: TestPostPolicyV4 (4.24s)
```

codyoss:
This failed on the presubmit for a recent PR I made: #2161. Changes were unrelated so I manually merged that PR and filed this issue. I agree that this seems like a permissions thing.

BrennaEpp:
Thanks for the response @codyoss. An issue for this has been filed internally. As we have been unable to reproduce this error locally, for the moment, this issue will be closed with #2168 which improves logging in case this error happens again. Please let us know if it does.

PR #2168, opened 2021-07-27
test(storage): Improve logging for objects test

BrennaEpp:
Adds request logging for `TestPostPolicyV4` for #2162


storage/objects/objects_test.go
```diff
@@ -479,14 +485,22 @@ func TestPostPolicyV4(t *testing.T) {
...
+	}
+
 	res, err := http.DefaultClient.Do(req)
 	if err != nil {
 		t.Fatalf("client.Do: %v", err)
 	}
 	if g, w := res.StatusCode, 204; g != w {
 		blob, _ := httputil.DumpResponse(res, true)
-		t.Errorf("status code in response mismatch: got %d want %d\nBody: %s", g, w, blob)
+		t.Errorf("status code in response mismatch: got %d want %d\nBody: %s\nattempted request: %s", g, w, blob, string(requestDump))
```
tritone, line 501: I would change this to put the request first, then the response, something like:

`...Request: %v\n\nResponse: %v\n", requestDump, blob)`

storage/objects/objects_test.go
```diff
@@ -479,14 +485,22 @@ func TestPostPolicyV4(t *testing.T) {
...
+	if err != nil {
+		t.Logf("error dumping request for logging: %v", err)
+	}
+
 	res, err := http.DefaultClient.Do(req)
 	if err != nil {
 		t.Fatalf("client.Do: %v", err)
 	}
 	if g, w := res.StatusCode, 204; g != w {
 		blob, _ := httputil.DumpResponse(res, true)
```
tritone, line 500: Let's call this responseDump

storage/objects/objects_test.go
```diff
@@ -422,6 +425,9 @@ func TestPostPolicyV4(t *testing.T) {
 	bucketName := tc.ProjectID + "-post-policy-bucket-name"
 	objectName := "foo.txt"
 	serviceAccount := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
+	if serviceAccount == "" {
```
tritone, line 428: I'm kind of dubious about this actually-- if something changes in how we do creds the test will be silently skipped. @codyoss thoughts?

tritone (commented):
Thanks for looking into this!

storage/objects/objects_test.go
codyoss: I think this is fine, but I don't think it is needed since this will always be set in the case of system tests. https://github.com/GoogleCloudPlatform/golang-samples/blob/60e4b3700f89663d46ddd80fdca3a70c6326fa7d/testing/kokoro/system_tests.sh#L106

If this changed I think a lot would break.

storage/objects/objects_test.go
BrennaEpp: Shouldn't it be `requestDump`?

storage/objects/objects_test.go
BrennaEpp: I think the check is useful for local testing/debugging. I changed it to output an error rather than skip the test. Is that okay or should I just remove the check? Note that the `TestV4SignedURL()` in this file also does this check and skips if not set.

storage/objects/objects_test.go
tritone: No, `res` is the HTTP response. We want to make sure to log BOTH the request we are sending and the response from the server.

storage/objects/objects_test.go
tritone: Either way is fine with me.

storage/objects/objects_test.go
BrennaEpp: I don't quite understand - the lines here are dumping the request body into a variable before we send it so we can log it later if needed. The response gets dumped later on (line 500) if there is an unexpected result.

tritone (approved):

PR #2160, opened 2021-07-21
feat(compute): add compute samples

FrodoTheTrue:

PR #2174, opened 2021-08-04
fix(compute): iterator to match new return values

codyoss:
This is to unblock CI in google-cloud-go. Note #2160 will also need to now be refactored before merge.


compute/list_instances.go
```diff
@@ -41,17 +42,18 @@ func listInstances(w io.Writer, projectID string, zone string) error {
 		Zone:    zone,
 	}
 
-	resp, err := client.List(ctx, req)
-	if err != nil {
-		return fmt.Errorf("List instances request: %v", err)
-	}
-
+	it := client.List(ctx, req)
```
tbpg, line 45: This is as breaking as they come. Is there no way around this?

tbpg (changes requested):

tbpg (approved):

codyoss:
The CI failures here are unrelated. Force merging this to unblock google-cloud-go CI.

PR #2178, opened 2021-08-05
feat(spanner): add samples for setting default leader

asthamohta:


spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -440,6 +440,78 @@ func TestCustomerManagedEncryptionKeys(t *testing.T) {
...
+	assertContains(t, out, fmt.Sprintf("Database DDL is as follows"))
+
+	//Try to Query Information Schema Database Options
+	b.Reset()
+	if err := queryInformationSchemaDatabaseOptions(&b, dbName); err != nil {
+		t.Errorf("failed to query information schema database options: %v", err)
+	}
+	out = b.String()
+	assertContains(t, out, fmt.Sprintf("The result of the query to get"))
+}
```
hengfengli, line 514: I think we should have a space between functions.

spanner/spanner_snippets/spanner/spanner_list_databases.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return err
+	}
+	defer adminClient.Close()
+
+	iter := adminClient.ListDatabases(ctx, &adminpb.ListDatabasesRequest{
+		Parent: instanceId,
+	})
+
+	printDatabases := func(iter *database.DatabaseIterator) error {
```
hengfengli, line 39: do we need to define this function? why can't we put the loop iteration outside?

spanner/spanner_snippets/spanner/spanner_query_information_schema_database_options.go
```diff
@@ -0,0 +1,62 @@
...
+	"regexp"
+
+	"cloud.google.com/go/spanner"
+	"google.golang.org/api/iterator"
+)
+
+func queryInformationSchemaDatabaseOptions(w io.Writer, db string) error {
+	matches := regexp.MustCompile("^(.+)/databases/(.+)$").FindStringSubmatch(db)
+	if matches == nil || len(matches) != 3 {
+		return fmt.Errorf("createDatabaseWithCustomerManagedEncryptionKey: invalid database id %q", db)
```
hengfengli, line 31: `createDatabaseWithCustomerManagedEncryptionKey`  does not match this sample's name.

spanner/spanner_snippets/spanner/spanner_query_information_schema_database_options.go
```diff
@@ -0,0 +1,62 @@
...
+
+	for {
+		row, err := iter.Next()
+		if err == iterator.Done {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		var OPTION_NAME, OPTION_VALUE string
```
hengfengli, line 54: these two variables should not be capitalized.

spanner/spanner_snippets/spanner/spanner_query_information_schema_database_options.go
```diff
@@ -0,0 +1,62 @@
...
+	databaseId := matches[2]
+
+	ctx := context.Background()
+	client, err := spanner.NewClient(ctx, db)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	stmt := spanner.Statement{SQL: `SELECT SingerId, AlbumId, MarketingBudget FROM Albums`}
```
hengfengli, line 42: The query is incorrect. This should be a query to the info schema.

spanner/spanner_snippets/spanner/spanner_get_database_ddl.go
```diff
@@ -0,0 +1,54 @@
...
+
+	op, err := adminClient.GetDatabaseDdl(ctx, &adminpb.GetDatabaseDdlRequest{
+		Database: db,
+	})
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(w, "Database DDL is as follows: \n [%s]", op.GetStatements())
```
hengfengli, line 48: is `%v` better?

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -440,6 +440,78 @@ func TestCustomerManagedEncryptionKeys(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
+	defer cancel()
+
+	// Try to get Instance Configs
+	config := fmt.Sprintf("projects/%s/instanceConfigs/%s", projectID, "nam6")
+	if err := getInstanceConfig(ctx, &b, config); err != nil {
+		t.Errorf("failed to create get instance configs: %v", err)
+	}
+	out := b.String()
+	assertContains(t, out, fmt.Sprintf("Available leader options for instance config"))
```
hengfengli, line 462: `fmt.Sprintf(` is not needed. Same for below similar cases.

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -440,6 +440,78 @@ func TestCustomerManagedEncryptionKeys(t *testing.T) {
 	assertContains(t, out, fmt.Sprintf("using encryption key %s", kmsKeyName))
 }
 
+func TestCreateDatabaseWithDefaultLeaderSample(t *testing.T) {
+	_ = testutil.SystemTest(t)
+	t.Parallel()
+
+	instName, dbName, cleanup := initTest(t, randomID())
```
hengfengli, line 447: `dbName` is never used.

hengfengli: My bad. My editor gives me a wrong info.

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -440,6 +440,78 @@ func TestCustomerManagedEncryptionKeys(t *testing.T) {
...
+
+	// Try to get Database DDL
+	b.Reset()
+	if err := getDatabaseDdl(ctx, &b, dbName); err != nil {
+		t.Errorf("failed to get Database DDL: %v", err)
+	}
+	out = b.String()
+	assertContains(t, out, fmt.Sprintf("Database DDL is as follows"))
+
+	//Try to Query Information Schema Database Options
```
hengfengli, line 507: Add a space before `Try`.

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -440,6 +440,78 @@ func TestCustomerManagedEncryptionKeys(t *testing.T) {
...
+
+	// Try to get list of Databases
+	b.Reset()
+	if err := listDatabases(ctx, &b, instName); err != nil {
+		t.Errorf("failed to get list of Databases: %v", err)
+	}
+	out = b.String()
+	assertContains(t, out, fmt.Sprintf("Databases for instance"))
+
+	t.SkipNow()
```
hengfengli, line 480: Why t.SkipNow()? We should remove this.

hengfengli (changes requested):
Thanks for making these samples so quick. There are some places need to be fixed. I don't think the current test instance would work since it is a regional instance. 

We need to change the `createTestInstance` method to make `InstanceConfig` as a parameter: 

https://github.com/GoogleCloudPlatform/golang-samples/blob/43677838ad7c8772d3b86b50b9146f824cea644f/spanner/spanner_snippets/spanner/integration_test.go#L674

If not given, use `regional-us-central1` as default.

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -52,7 +52,16 @@ var (
 
 func initTest(t *testing.T, id string) (instName, dbName string, cleanup func()) {
 	projectID := getSampleProjectId(t)
-	instName, cleanup = createTestInstance(t, projectID)
+	instName, cleanup = createTestInstance(t, projectID, false)
```
hengfengli, line 55: It's better to pass the instance config name as a param instead of a boolean value.

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -52,7 +52,16 @@ var (
...
+	dbID := validLength(fmt.Sprintf("smpl-%s", id), t)
+	dbName = fmt.Sprintf("%s/databases/%s", instName, dbID)
+
+	return
+}
+
+func initTestWithConfig(t *testing.T, id string) (instName, dbName string, cleanup func()) {
```
hengfengli, line 62: We should allow the caller to set the instance config: 

```go
func initTestWithConfig(t *testing.T, id, instanceConfigName string) (instName, dbName string, cleanup func()) {

instName, cleanup = createTestInstance(t, projectID, instanceConfigName)
```

spanner/spanner_snippets/spanner/spanner_query_information_schema_database_options.go
```diff
@@ -51,11 +53,11 @@ func queryInformationSchemaDatabaseOptions(w io.Writer, db string) error {
...
 			return err
 		}
-		var OPTION_NAME, OPTION_VALUE string
-		if err := row.Columns(&OPTION_NAME, &OPTION_VALUE); err != nil {
+		var option_name, option_value string
+		if err := row.Columns(&option_name, &option_value); err != nil {
 			return err
 		}
-		fmt.Fprintf(w, "The result of the query to get %s for %s is %s", OPTION_NAME, databaseId, OPTION_VALUE)
+		fmt.Fprintf(w, "The result of the query to get %s for %s is %s", option_name, databaseId, option_value)
```
hengfengli, line 60: nits: for Go's conventions, it should be `databaseID`. I guess you got this from the existing codebase. But better to fix it.

hengfengli (approved):
Thanks a lot for working on this. LGTM.

hengfengli:
@codyoss Can we get an approval from you? Thanks.

spanner/spanner_snippets/spanner/spanner_query_information_schema_database_options.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"io"
+	"regexp"
+
+	"cloud.google.com/go/spanner"
+	"google.golang.org/api/iterator"
+)
+
+func queryInformationSchemaDatabaseOptions(w io.Writer, db string) error {
+	matches := regexp.MustCompile("^(.+)/databases/(.+)$").FindStringSubmatch(db)
```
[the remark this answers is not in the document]
hengfengli: @codyoss Do you suggest we change to: 

```go
func queryInformationSchemaDatabaseOptions(w io.Writer, projectID, instanceID, databaseID string) error {
```

But the thing is that when we create the client we still need to construct the full path: 

```go
        // db = `projects/<project>/instances/<instance-id>/database/<database-id>`
	client, err := spanner.NewClient(ctx, db)
```

Another thing is that in most of the existing samples, we pass the full path as a parameter. We probably would like to follow the same convention.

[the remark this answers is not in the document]
codyoss: Unless that values are used individually I would say no. Just one db string /w an example should be fine.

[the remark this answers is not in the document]
hengfengli: Does this mean that this comment is resolved? Sorry I got a bit confused.

hengfengli:
@codyoss A strange thing is that the four CI tests are getting stuck since last night. Do you know what it is the reason?

codyoss (approved):

PR #2179, opened 2021-08-09
feat(compute): add pagination samples

FrodoTheTrue:


compute/list_all_instances.go
```diff
@@ -35,12 +36,17 @@ func listAllInstances(w io.Writer, projectID string) error {
 	}
 	defer instancesClient.Close()
 
+	// Use the `MaxResults` parameter to limit the number of results that the API returns per response page.
```
codyoss, line 39: Nothing wrong with changes here, but would you update L31. Looks like an incorrect sample value.

codyoss (commented):
Just one small thing or else looks good, thanks!

compute/list_all_instances.go
FrodoTheTrue: updated

codyoss (approved):
Thanks!

PR #2183, opened 2021-08-10
fix(asset): fix tests

codyoss:
There was recently an intentional breaking change in the way
asset LROs respond. Updaing the test to just check for a successful
completion.

Fixes: #2176
Fixes: #2182


tbpg (approved):
:grimacing:

PR #2191, opened 2021-08-12
feat(kms): add samples for new rng and hmac APIs

sethvargo:
This adds samples for new KMS APIs.


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/GoogleCloudPlatform/golang-samples
 go 1.11
 
 require (
-	cloud.google.com/go v0.90.0
```
sethvargo, line 6: The mod changes are from updating the kms and kmspb packages + `go mod tidy`

sethvargo:
I'm not entirely sure what CI is complaining about with respect to the go.mod. Any ideas

codyoss:
@sethvargo Hope you don't mind a pushed a commit to see if it is all just tidy related. Because this is a multi-module project where some of the sub-modules depend on the root module with replace statements -- if you modify the root go.mod you have to tidy any with a replace statement. I have an bash function to help me with this: 

```
tidyall() {
    go mod tidy
    wd="$PWD"
    for i in $(find $wd -name go.mod); do
        pushd $(dirname $i)
        go mod tidy
        popd
    done
    cd $wd
}
```

codyoss (approved):
LGTM

tbpg:
One liner (mentioned in the error message):
```
find . -name go.mod -execdir go mod tidy \;
```

sethvargo:
Sweet thank you both!

sethvargo:
Hmm - is the go111 failure related?

codyoss:
Let me try to push one more commit to see if we can get this working. It's looks like with the changes in the go.mod a transitive dep now requires Go 1.13, so 1.11 will fail.

sethvargo:
Thank you @codyoss - let me know how I can help.

codyoss:
Forgot to come back to this yesterday sorry. It was the use of `%w` in go1.11 test that was causing the test failures. Moved usage to `%v`.

sethvargo:
Oh my b. Thank you for catching that and thank you for merging!

PR #2197, opened 2021-08-18
fix(kms): asymmetric public key ec error message

hasheddan:
Fixes error message in asymmetric public key check for ECDSA signature
verification example.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>


codyoss (approved):
Looks good to me, thank you for your contribution!

sethvargo (approved):
Yea...... I totally definitely didn't copy and paste. You can prove nothing 😄

PR #2212, opened 2021-08-31
fix(compute): fix compute samples after taking breaking change

codyoss:


tbpg (approved):

codyoss:
Failure unrelated to PR. Merging to unblock google-cloud-go

PR #2219, opened 2021-09-09
feat(cloudsql): add support for SSL certificates

enocom:
This commit adds support for configuring SSL certificates for Postgres
and MySQL.

### DIFF
--- a/cloudsql/mysql/database-sql/cloudsql.go
+++ b/cloudsql/mysql/database-sql/cloudsql.go
@@ -19,16 +19,20 @@
 package main
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"database/sql"
+	"errors"
 	"fmt"
 	"html/template"
+	"io/ioutil"
 	"log"
 	"math"
 	"net/http"
 	"os"
 	"strconv"
 
-	_ "github.com/go-sql-driver/mysql"
+	"github.com/go-sql-driver/mysql"
 )
 
 // vote struct contains a single row from the votes table in the database.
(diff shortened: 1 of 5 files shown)


cloudsql/mysql/database-sql/cloudsql.go
```diff
@@ -277,6 +281,41 @@ func initTCPConnectionPool() (*sql.DB, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		mysql.RegisterTLSConfig("cloudsql", &tls.Config{
+			RootCAs:               pool,
+			Certificates:          []tls.Certificate{cert},
+			InsecureSkipVerify:    true,
+			VerifyPeerCertificate: verifyPeerCertFunc(pool),
+		})
+		dbURI += "&tls=cloudsql"
```
codyoss, line 314: This variable is not declared in the region tag being added, should it be? Also, not edited in this sample but it looks like the declaration and initialization of this variable could be brought together with `:=`

codyoss (approved):
LGTM, just one small question.

cloudsql/mysql/database-sql/cloudsql.go
enocom: Good point. I'll update the sample to use `:=` and send a separate PR. Because the `dbURI` needs to be modified only if a user wants to configure SSL certs, we're stuck with this approach of appending to the existing variable. In our upcoming quickstart work, we'll show this section immediately after the section which introduces `dbURI`.

codyoss: SGTM

enocom:
FYI I've had to try a few different approaches on copying certificates that may or may not exist. The Go code is unchanged.

Now, I've added a `certs` directory to avoid any complex Dockerfile commands. This seems like the cleanest approach.

enocom:
@codyoss I've requested another review. There's no nice way to copy files that may or may not be present (e.g., optionally configured certificates) in Docker. I've opted for a simple approach of making a directory with a `.gitkeep` file, but that's required an update to the bad files test. PTAL.

codyoss (approved):
SGTM

PR #2222, opened 2021-09-14
feat(firestore): add partition queries samples

crwilcox:


firestore/firestore_snippets/collection_group_query.go
```diff
@@ -51,3 +51,66 @@ func collectionGroupQuery(w io.Writer, projectID string) error {
...
+
+func partitionQuery(ctx context.Context, client *firestore.Client) error {
+	// [START firestore_partition_query]
+	cities := client.CollectionGroup("cities")
+
+	// Get a partioned query for the cities collection group, with a maximum
+	// partition count of 10
+	partitionedQueries, err := cities.GetPartitionedQueries(ctx, 10)
+	if err != nil {
+		fmt.Println(err)
```
kolea2, line 63: do we want to just print here? Should we throw instead?

crwilcox: I think I borrowed this from elsewhere. As the sample region doesn't include the func, it is odd we are returning from, essentially, nowhere. Though I saw other places that write to writer (use `Fprintln`) which could have testing advantages. Though I just noticed on ln 77 I do return err making this even odder. @tbpg I'd love your thoughts here if you want to weigh in. Then I can make this all consistent.

crwilcox: Perfect. Thank you for your input here. You are correct, those samples are all pretty light. That may be something to address, but it seems a more holistic thing than starting with this one sample.

firestore/firestore_snippets/collection_group_query.go
```diff
@@ -51,3 +51,66 @@ func collectionGroupQuery(w io.Writer, projectID string) error {
...
+
+func partitionQuery(ctx context.Context, client *firestore.Client) error {
+	// [START firestore_partition_query]
+	cities := client.CollectionGroup("cities")
+
+	// Get a partioned query for the cities collection group, with a maximum
+	// partition count of 10
+	partitionedQueries, err := cities.GetPartitionedQueries(ctx, 10)
+	if err != nil {
+		return err
```
tbpg, line 63: Might want to use `fmt.Errorf` to add more context to the error when returning it. See https://github.com/GoogleCloudPlatform/golang-samples/blob/master/CONTRIBUTING.md#return-errors.

tbpg (approved):

kolea2 (approved):

PR #2223, opened 2021-09-15
feat(compute): add firewall samples

FrodoTheTrue:


compute/firewall/create_firewall_rule.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+// creates a simple firewall rule allowing for incoming HTTP and HTTPS access from the entire Internet.
```
codyoss, line 28: nit: start doc comments with the name of the method: "// createFirewallRule ..." (Applies for other samples too)
nit: remove the word simple

compute/firewall/firewall_test.go
```diff
@@ -0,0 +1,112 @@
...
+	compute "cloud.google.com/go/compute/apiv1"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+)
+
+func TestFirewallSnippets(t *testing.T) {
+	ctx := context.Background()
+	firewallsClient, err := compute.NewFirewallsRESTClient(ctx)
+	if err != nil {
+		t.Errorf("NewInstancesRESTClient: %v", err)
```
codyoss, line 35: I think at least these first couple errors should actually be t.Fatalf. If they don't succeed the remainder of the test will not either.

codyoss (changes requested):
Just a couple of small things, else it LGTM.

codyoss (approved):
LGTM

PR #2224, opened 2021-09-15
fix(healthcare): add request field to bundle

noerog:
Adds "request" field to bundle body and updates the FHIR store to use the latest version (R4).

Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/2218


codyoss (approved):
Thanks!

PR #2231, opened 2021-09-21
chore(cloudsql): declare and assign with :=

enocom:
Follow up from https://github.com/GoogleCloudPlatform/golang-samples/pull/2219#discussion_r711112356.


codyoss (approved):
Thank you!

PR #2234, opened 2021-09-22
test(storage): Add bucket setup/teardown to bucket tests

### LINKED ISSUE #2228: storage/buckets: TestLifecycleManagement failed
Note: #2196 was also for this test, but it was closed more than 10 days ago. So, I didn't mark it flaky.

----

commit: 104b239a400fee1e0735672c23b4a9433be4f622
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/cfe7c7a5-cd04-42b8-9925-31ece7200b77), [Sponge](http://sponge2/cfe7c7a5-cd04-42b8-9925-31ece7200b77)
status: failed
<details><summary>Test output</summary><br><pre>    buckets_test.go:438: enableBucketLifecycleManagement: Bucket("golang-samples-tests-storage-buckets-tests").Update: googleapi: Error 429: The project exceeded the rate limit for bucket operations (buckets creating, updating and deleting)., rateLimitExceeded</pre></details>

BrennaEpp:
This should hopefully resolve the gcs rate limit issue for these tests. 
Fixes #2228


storage/buckets/buckets_test.go
```diff
@@ -31,42 +32,77 @@ import (
...
+	os.Exit(exit)
+}
+
 func TestCreate(t *testing.T) {
 	tc := testutil.SystemTest(t)
-	bucketName := tc.ProjectID + "-storage-buckets-tests"
+	bucketName := testutil.UniqueBucketName(testPrefix)
+	ctx := context.Background()
+
+	client, err := storage.NewClient(ctx)
```
tritone, line 57: Could we just make this client a global variable for this file and initialize it in `TestMain`?

storage/buckets/buckets_test.go
```diff
@@ -31,42 +32,77 @@ import (
...
+	if err != nil {
+		t.Fatalf("storage.NewClient: %v", err)
+	}
+	t.Cleanup(func() {
```
tritone, line 61: `t.Cleanup` is new in Go 1.14-- do we have test runs for older versions in this repo? @codyoss

storage/buckets/buckets_test.go
```diff
@@ -31,42 +32,77 @@ import (
 	iampb "google.golang.org/genproto/googleapis/iam/v1"
 )
 
+const (
+	testPrefix      = "storage-buckets-test"
```
tritone, line 36: Let's include the projectID in this as well if possible?

tritone (commented):
Looks like a good approach! A few small things.

storage/buckets/buckets_test.go
codyoss: We test on 1.11 and latest Go versions. In order to use this feature though you would need to add a go.mod file somewhere in storage directory tree that declares a Go version of at least 1.14. Our test runners trigger off of the language version specified in the go.mod.

storage/buckets/buckets_test.go
BrennaEpp: We could include it here, but it looks like the individual tests should still call `testutil.SystemTest(t)` to skip them if the env variable for the project isn't set, and also gives us that project id. So adding a global projectID here wouldn't be very useful unless we want to refactor that part as well

storage/buckets/buckets_test.go
BrennaEpp: yes, great point!!

codyoss (approved):

PR #2235, opened 2021-09-23
feat: Adding cloud run jobs sample

dinagraves:


dinagraves:
@grayside  I can't get this test to pass -- main.go doesn't seem to be using the environment variables I'm setting at all.

dinagraves:
If I move the os.Getenv to inside the `main` function, then the tests pass.  Why?

grayside:
When you run the test, it compiles a test binary and then executes the tests.

The global-scoped `os.Getenv` calls are running before the tests. Then, when your individual unit tests override the environment variables, there's nothing in `main` updating the variable with the new value.

The mindset difference is that your test is not including and executing main.go at runtime, it's already got that file and now you are specifically calling main() and nothing outside of it.

dinagraves:
Thank you!  Is the solution to have an init() in the test file?  Or to just move all the environment variables inside the function?

run/jobs/main.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"fmt"
+	"log"
+	"math/rand"
+	"os"
+	"strconv"
+	"time"
+)
+
+func getEnvVars() (string, string, string, string) {
```
dinagraves, line 26: @grayside is this an ugly/awful solution?

grayside:
Yes, initialize inside the function.

If it weren't `main`, we could consider whether the function should be reading environment variables or if something outside the function is responsible for bridging environment to parameters. That's a general design space around handling globals and managing side effects.

run/jobs/main.go
grayside: Try thinking about initializing a strongly typed config or app struct. Something you can createFromEnv or initialize by manually in tests. I can show you a couple examples of that pattern if helpful.

dinagraves: How closely should it look like the canonical sample?  https://github.com/averikitsch/nodejs-docs-samples/blob/jobs/run/jobs/index.js

averikitsch: Should we strongly type for the Job's defined variables, ie TASK_NUM and ATTEMPT_NUM? I lean on the side of idiomatic Go than matching the canonical.

grayside: Go has some flexibility on this issue, and it depends.

In general, I prefer to have a layer that captures environment and "production" preferences such as: https://github.com/GoogleCloudPlatform/golang-samples/blob/main/getting-started/bookshelf/bookshelf.go#L56

A couple reasons for a central app struct:

1. If the app is organized "just so", it becomes easier for unit & integration tests to inject different config/behaviors without dealing with global state or adding checks deeper in the app.
2. Strongly typing/validating these values improve explicitness, local development, and portability. Problems can be detected on startup instead of deeper into the app's processing, and the requirements to run the app in a different environment than Jobs would be more clear.

Thanks for asking if we should follow canonical practice! +1 to "how to start the app follows language practice", but with new things how to draw that line can be be unclear.

run/jobs/main.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Retrieve User-defined env vars
+	var SLEEP_MS = os.Getenv("SLEEP_MS")
+	var FAIL_RATE = os.Getenv("FAIL_RATE")
+	return TASK_NUM, ATTEMPT_NUM, SLEEP_MS, FAIL_RATE
+}
+
+func main() {
+	TASK_NUM, ATTEMPT_NUM, SLEEP_MS, FAIL_RATE := getEnvVars()
+	log.Printf("Starting Task #%s, Attempt #%s ...", TASK_NUM, ATTEMPT_NUM)
```
grayside, line 39: These log statements seem like a strong candidate to be deprecated by product improvements. Really like that we're making sure they happen! Maybe we should confirm if they'll be needed for awhile?

run/jobs/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"bytes"
+	"log"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestSuccessfulJob(t *testing.T) {
```
grayside, line 25: Should there be TestJobSuccess with two cases: around FAIL_RATE empty/0 and FAIL_RATE 1? We're not testing the fault injection is honored.

run/jobs/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+	os.Setenv("SLEEP_MS", "0")
+	os.Setenv("TASK_NUM", "1")
+	os.Setenv("ATTEMPT_NUM", "1")
+	os.Setenv("FAIL_RATE", "")
```
grayside, line 29: Just following up my comment about central struct vs. global dependency, these variables are now overridden, which means if we don't revert them after this test case, we've now changed the environment for other tests.

run/jobs/main.go
```diff
@@ -0,0 +1,74 @@
...
+	"log"
+	"math/rand"
+	"os"
+	"strconv"
+	"time"
+)
+
+func getEnvVars() (string, string, string, string) {
+	// Retrieve Job-defined env vars
+	var TASK_NUM = os.Getenv("TASK_NUM")
```
grayside, line 28: Unexported constants should be lowerCamelCase.
https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

dinagraves:
@grayside I tried using the struct. LMK if I'm on the right path.

run/jobs/main.go
```diff
@@ -0,0 +1,84 @@
...
+	taskNum    string
+	attemptNum string
+
+	// User-defined env vars
+	sleepMs  int64
+	failRate float64
+}
+
+func stringToInt(s string) int64 {
+	num, _ := strconv.ParseInt(s, 10, 64)
```
grayside, line 37: suggestion: Suppressing the parsing errors is not idiomatic. I suggest making these more specific about the variable they support and adding a log.Fatal, or returning the error and handling it in main(). This comes close to blocker but I think falls just short.

In my opinion, I would have a `configFromEnv() (Config, error)` function and fatal the error in main. Config in this case is a replacement name for `type EnvVars`, what's important about them is not that they are environment, but that they are the config from the app that happens to come from the environment.

run/jobs/main.go
```diff
@@ -0,0 +1,84 @@
...
+	return num
+}
+
+func stringToFloat(s string) float64 {
+	num, _ := strconv.ParseFloat(s, 64)
+	return num
+}
+
+func main() {
```
grayside, line 46: If this were a "real" app, I'd have main() just assemble the config values and log the task/attempt.

Collapsing the "core work" into main probably helps overall readability, and since this is a more "getting started" sample I'm in favor of the approach here.

dinagraves: Ack.

run/jobs/main_test.go
dinagraves: If FAIL_RATE is 1, it won't succeed.  Should I test that?  I test the actual failure function below with fail_rate of 1.

grayside: I think testing it in the more complete way is nice but not required.

run/jobs/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	main()
+	log.SetOutput(os.Stdout)
+	output := buf.String()
+
+	start := "Starting Task #1, Attempt #1 ..."
+	finish := "Completed Task #1, Attempt #1"
+
+	if !(strings.Contains(output, start) && strings.Contains(output, finish)) {
+		t.Fatalf("\nExpected string to contain:\n%s\n%s\n\nGot:\n%s", start, finish, output)
```
grayside, line 42: nit: Seems like this should be a `t.Errorf`, and use the more typical want/got syntax. I'd also split into separate asserts.

grayside (approved):
LGTM, left a couple more really minor points but non-blocking.

PR #2249, opened 2021-09-28
feat(bigquery): add migration quickstart

shollyman:


bigquery/bigquery_migration_quickstart/main.go
```diff
@@ -0,0 +1,137 @@
...
+	reportWorkflowStatus(workflow)
+}
+
+// executeTranslationWorkflow constructs a migration workflow that performs some offline SQL translation.
+func executeTranslationWorkflow(ctx context.Context, client *migration.Client, projectID, location string) (*migrationpb.MigrationWorkflow, error) {
+
+	// Tasks are extensible; the translation task is defined by the BigQuery Migration API.
+
+	translationDetails := &translationtaskpb.TranslationTaskDetails{
+		// TODO: we need to stage some suitable translation inputs and complete this definition.
```
tswast, line 70: Is this a GCS path? If so, maybe we could stick something in our sample data bucket?

bigquery/bigquery_migration_quickstart/main.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Finally, construct the workflow creation request.
+	req := &migrationpb.CreateMigrationWorkflowRequest{
+		Parent: fmt.Sprintf("projects/%s/locations/%s", projectID, location),
+		MigrationWorkflow: &migrationpb.MigrationWorkflow{
+			DisplayName: "example SQL conversion",
+			Tasks: map[string]*migrationpb.MigrationTask{
+				"example_conversion": {
+					Type:    "Translation_Teradata",
+					Details: anyDetails,
```
tswast, line 90: I'm a bit confused why the API doesn't use a oneof here. Presumably it only supports a limited number of types + detail messages.

bigquery/bigquery_migration_quickstart/main.go
```diff
@@ -0,0 +1,137 @@
...
+	// This is an asyncronous process, so we now poll the workflow
+	// until completion or a suitable timeout has elapsed.
+	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+	for {
+		select {
+		case <-timeoutCtx.Done():
+			return nil, fmt.Errorf("task %s didn't complete due to context expiring", workflow.GetName())
+		default:
+			polledWorkflow, err := client.GetMigrationWorkflow(timeoutCtx, &migrationpb.GetMigrationWorkflowRequest{
```
tswast, line 111: Oh dear. No long running operation API?

bigquery/bigquery_migration_quickstart/main.go
shollyman: It's because the intent is to allow user extension; so you end up with an arbitrary set of task definitions without a central registry.

bigquery/bigquery_migration_quickstart/main.go
shollyman: Not at alpha at least.

shollyman:
Referencing this here since it's not tracked as part of the PR, but it core to the quickstart:

> gsutil cat gs://cloud-samples-data/bigquery/migration/translation/input/queries.txt
```
DATABASE ExampleDatabase;
SELECT * FROM teratable1;
SELECT invalidcolumn, fname FROM teratable2;
INSERT INTO teratable3 (foo, bar, baz) VALUES (1,2,3);
```

> gsutil cat gs://cloud-samples-data/bigquery/migration/translation/schema/ddl.txt
```
CREATE TABLE teratable1 (code char, examplejson JSON(1000));
CREATE TABLE teratable2 (idx INT, username VARCHAR(64), fname
VARCHAR(100)) UNIQUE PRIMARY INDEX (idx);
```

shollyman:
Sample output to GCS:

> gsutil cat gs://`output prefix omitted`/queries.txt
```
<Translation Failure for query at original line 1, column 1: DATABASE ExampleDatabase;>

-- #Translation Warning# Could not fully verify query, please provide table information for: teratable1
SELECT * FROM teratable1;

-- #Translation Warning# Below Verification Error encountered for translation output. The corresponding input query is at original line 3, column 1.
-- #Translation Warning# Unrecognized name: invalidcolumn [at 1:8]
SELECT invalidcolumn, fname FROM teratable2;

-- #Translation Warning# Could not fully verify query, please provide table information for: teratable3
INSERT INTO teratable3 (foo, bar, baz) VALUES (1, 2, 3);
```

> gsutil cat gs:/`output prefix omitted`/queries_errors.csv
```
Input_Query,Line,Column,Category,Message,Type
"DATABASE ExampleDatabase;
",1,1,SqlParseException,Non-query expression encountered in illegal context,ERROR
```

Issue #2289, opened 2021-11-01
dialogflow: README example commands do not work


chrbsg:
golang-samples/dialogflow/README.md states:

```
### Detect intent

(Text, audio, streaming audio)

[Source code](detect_intent/detect_intent.go)

**Usage:** `go run detect_intent/detect_intent.go -help`

### Intent management

[Source code](intent_management/intent_management.go)

**Usage:** `go run intent_management/intent_management.go -help`

### Entity type management

[Source code](entity_type_management/entity_type_management.go)

**Usage:** `go run entity_type_management/entity_type_management.go -help`

### Entity management

[Source code](entity_management/entity_management.go)

**Usage:** `go run entity_management/entity_management.go -help`

### Session entity type management

[Source code](session_entity_type_management/session_entity_type_management.go)

**Usage:** `go run session_entity_type_management/session_entity_type_management.go -help`
```

But these commands do not work, because there is no main package, or main functions:

```
$ go run detect_intent/detect_intent.go -help
go run: cannot run non-main package
$ grep -r 'package main'
$ 
```

codyoss:
Thanks for filing this issues, I believe these used to be runnable examples, but are no longer so. I will get out a PR to update the README to reflect this.

PR #2290, opened 2021-11-01
chore(dialogflow): update readme

codyoss:
Theses samples seem like they used to be runable, but they are no
longer so. Simplify README to reflect this.

Fixes: #2289


tbpg (approved):
This is why I don't like putting manually-maintained info in docs. :smile:

PR #2296, opened 2021-11-08
feat(storagetransfer): add quickstart for storage transfer service

JesseLovelace:
Adds a quickstart for storage transfer service, along with a test

I have never written Go code before (ever), so apologies for any mistakes!


storagetransfer/quickstart_test.go
```diff
@@ -0,0 +1,111 @@
...
+		t.Errorf("quickstart: %#v", err)
+	}
+
+	got := buf.String()
+	if want := "transferJobs/"; !strings.Contains(got, want) {
+		t.Errorf("got %q, want %q", got, want)
+	}
+
+	re := regexp.MustCompile("transferJobs/\\d+")
+	name := re.FindString(got)
```
tritone, line 66: I don't love doing this with a regex from output-- could you have the quickstart function return the job name (or something containing it) instead?

storagetransfer/quickstart_test.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		t.Fatalf("Bucket creation failed: %v", err)
+	}
+	defer testutil.DeleteBucketIfExists(ctx, str, sinkBucketName)
+
+	sourceBucketName, err := testutil.CreateTestBucket(ctx, t, str, tc.ProjectID, "sts-go-source")
+	if err != nil {
+		t.Fatalf("Bucket creation failed: %v", err)
+	}
+	defer testutil.DeleteBucketIfExists(ctx, str, sourceBucketName)
```
tritone, line 50: If I understand correctly, both of these buckets are empty. Does that pose any problems for testing the transfer job, or does it not matter?

storagetransfer/quickstart_test.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func grantStsPermissions(bucketName string, projectID string) {
+	ctx := context.Background()
+
+	req := &storagetransferpb.GetGoogleServiceAccountRequest{
+		ProjectId: projectID,
+	}
+
+	resp, err := sts.GetGoogleServiceAccount(ctx, req)
```
tritone, line 86: Possibly totally off-base but-- should this code really be in a sample as well, rather than just in the test? Do the docs make it clear how to give these permissions to the STS service account?

tritone (commented):
Welcome to Go! Few comments, some of which are because I don't know anything about STS.

Also, to fix the build, you first need to run `go get cloud.google.com/go/storagetransfer` and probably `go mod tidy` as well. Then we can see if there are other things to fix.

storagetransfer/quickstart.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"fmt"
+	"io"
+	"log"
+
+	storagetransfer "cloud.google.com/go/storagetransfer/apiv1"
+	storagetransferpb "google.golang.org/genproto/googleapis/storagetransfer/v1"
+)
+
+// quickstart creates and runs a transfer job between two GCS buckets.
```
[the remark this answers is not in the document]
JesseLovelace: Probably not super typical, but a valid use case and the simplest thing we could think of for a get-off-the-ground quickstart

storagetransfer/quickstart.go
```diff
@@ -0,0 +1,58 @@
...
+					GcsDataSink: &storagetransferpb.GcsData{BucketName: sinkGCSBucket}},
+			},
+			Status: storagetransferpb.TransferJob_ENABLED,
+		},
+	}
+	resp, err := client.CreateTransferJob(ctx, req)
+	if err != nil {
+		log.Fatalf("Failed to create transfer job: %v", err)
+	}
+	_, err = client.RunTransferJob(ctx, &storagetransferpb.RunTransferJobRequest{
```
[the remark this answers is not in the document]
JesseLovelace: The customer would only need JobName to monitor the job, which they already have

storagetransfer/quickstart_test.go
JesseLovelace: It doesn't matter, the job creates and runs successfully anyway, doing nothing

storagetransfer/quickstart_test.go
JesseLovelace: The docs explain how to do it in the console, but not with code. I agree that it should be a sample and it's something I want to add and will propose at some point, just not sure how to go about it yet since it's done entirely with the GCS client and not with the STS client, so there has to be some explainer on how to depend on the GCS client as well, I'm not sure if there's other examples of docs in one client having samples for a different client. Either way, it's not going to be part of the this stage

JesseLovelace:
Addressed your comments, however running `go get cloud.google.com/go/storagetransfer` didn't seem to do anything

storagetransfer/quickstart.go
```diff
@@ -2,17 +2,15 @@ package storagetransfer
...
 import (
+	storagetransfer "cloud.google.com/go/storagetransfer/apiv1"
 	"context"
 	"fmt"
-	"io"
-	"log"
-
-	storagetransfer "cloud.google.com/go/storagetransfer/apiv1"
 	storagetransferpb "google.golang.org/genproto/googleapis/storagetransfer/v1"
+	"io"
```
tritone, line 9: Core language dependencies should come first; run goimports to fix (intellij can do this for you).

storagetransfer/quickstart_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestQuickstart(t *testing.T) {
+	tc := testutil.SystemTest(t)
+	ctx := context.Background()
+
+	str, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("storage.NewClient: %v", err)
```
tritone, line 25: This and L31 should be `t.Fatalf` instead of log.

tritone (commented):
Two more little things, otherwise LGTM once we fix the build.

storagetransfer/quickstart_test.go
```diff
@@ -0,0 +1,105 @@
...
+func grantSTSPermissions(bucketName string, projectID string, sts *storagetransfer.Client, str *storage.Client) {
+	ctx := context.Background()
+
+	req := &storagetransferpb.GetGoogleServiceAccountRequest{
+		ProjectId: projectID,
+	}
+
+	resp, err := sts.GetGoogleServiceAccount(ctx, req)
+	if err != nil {
+		fmt.Print("Error getting service account")
```
tritone, line 82: This also needs to cause an error rather than proceeding. You should either make the function return an error to the caller, or pass in the `testing.T` as an arg so that you can cause the test failure here.

tritone (approved):
LGTM pending presubmits passing. @codyoss we need a review from you as well because we don't have ownership over this subdir.

codyoss (approved):

JesseLovelace:
Seems like this is just pending https://github.com/GoogleCloudPlatform/golang-samples/issues/2300

tritone:
Yeah, seems to be consistently failing. I made #2308 to fix.

PR #2297, opened 2021-11-09
feat(pubsublite): add regional topic samples

hongalex:
Based on Python https://github.com/googleapis/python-pubsublite/pull/266 and Java https://github.com/googleapis/java-pubsublite/pull/929


codyoss (approved):
LGTM

pubsublite/admin/update_topic.go
```diff
@@ -36,7 +36,12 @@ func updateTopic(w io.Writer, projectID, region, zone, topicID string) error {
...
 
-	topicPath := fmt.Sprintf("projects/%s/locations/%s/topics/%s", projectID, zone, topicID)
+	var topicPath string
+	if regional {
+		topicPath = fmt.Sprintf("projects/%s/locations/%s/topics/%s", projectID, region, topicID)
+	} else {
+		topicPath = fmt.Sprintf("projects/%s/locations/%s/topics/%s", projectID, zone, topicID)
+	}
 	// For ranges of fields in TopicConfigToUpdate, see https://pkg.go.dev/cloud.google.com/go/pubsublite/#TopicConfigToUpdate
 	config := pubsublite.TopicConfigToUpdate{
```
anguillanneuf, line 46: Can we also show ThroughputReservation as one of the field to be updated in this update topic example? Sorry I didn't think of this one in the first pass.

anguillanneuf (changes requested):
Generally LGTM. One request for the update topic example and one request to add a comment in the list topics example. Listing topics in a region will not return zonal topics in that region even if they use the same reservation.

pubsublite/admin/update_topic.go
hongalex: Why is the `regional` bool missing from the Python/Java samples for update topic? I also noticed it is missing for get topic as well.

anguillanneuf (approved):

pubsublite/admin/update_topic.go
anguillanneuf: Update topic Java/Python samples have been updated to use a bool for regional too. Thanks for pointing that out!

PR #2298, opened 2021-11-09
fix(cloudsql) database max connection lifetime

icecrime:
Google Cloud recommends limiting a database connection's lifetime to 1800 seconds, but samples use a 1800 value for `time.Duration` which evaluates to 1800 nanoseconds.

Express the arguments to `SetConnMaxLifetime` as `time.Second`.

### DIFF
--- a/cloudsql/mysql/database-sql/cloudsql.go
+++ b/cloudsql/mysql/database-sql/cloudsql.go
@@ -31,6 +31,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"time"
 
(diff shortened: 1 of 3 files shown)


icecrime:
@googlebot I fixed it.

icecrime:
By the way the same problem can be found in [this documentation page](https://cloud.google.com/sql/docs/postgres/manage-connections#duration).

codyoss (approved):
LGTM, thanks for the fix!

codyoss:
@icecrime The sample you edited will auto-update that page too 😄  Thanks again!

PR #2304, opened 2021-11-16
feat(functionsv2): Add sample for Audit Logs

muncus:
Sample function prints fields to the logs, and has been manually tested with GCS write audit logs.

This is the Go version of this canonical sample: https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/2991f86146742d00725c4dd73bf3719bd25e809b/functions/v2/index.js#L65


functions/functionsv2/helloauditlog/hello_auditlog.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/cloudevents/sdk-go/v2/event"
+	auditevents "github.com/googleapis/google-cloudevents-go/cloud/audit/v1"
+)
+
+// HelloAuditLog receives a Cloud Audit Log event, and logs a few fields.
```
grayside, line 29: suggestion: process -- when tracking against a canonical sample it can help to link the canonical in the PR description. https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/2991f86146742d00725c4dd73bf3719bd25e809b/functions/v2/index.js#L65

functions/functionsv2/helloauditlog/hello_auditlog.go
```diff
@@ -0,0 +1,52 @@
...
+// Package helloworld provides a set of Cloud Functions samples.
+package helloworld
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/cloudevents/sdk-go/v2/event"
+	auditevents "github.com/googleapis/google-cloudevents-go/cloud/audit/v1"
```
grayside, line 26: blocking: Unfortunately this library is not GA and we shouldn't use it yet. I'm curious how you found your way to this library, and what we might do to prevent others for deciding to use it?

functions/functionsv2/helloauditlog/hello_auditlog.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// HelloAuditLog receives a Cloud Audit Log event, and logs a few fields.
+func HelloAuditLog(ctx context.Context, e event.Event) error {
+	log.Printf("Event Type: %s", e.Type())
+	log.Printf("Subject: %s", e.Subject())
```
grayside, line 32: non-blocking: canonical sample includes "Method" property in output, and it looks like output order is different. Perfect alignment is not required, but I think it's useful to help polyglot developers contrast the code, and sometimes reveals unintended, subtle variations in the developer experience.

functions/functionsv2/helloauditlog/hello_auditlog_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+func TestHelloAuditLog(t *testing.T) {
+
+	tests := []struct {
+		name         string
+		subject      string
+		payload      auditevents.ProtoPayload
+		expectedLogs []string
+	}{
+		{"sample-output",
```
grayside, line 56: suggestion: name declarations for longer struct definitions improve readability. I went looking for whether we have a standard on this, and I couldn't find one so it may just be opinion ;)

Since I'm curious if I'm missing something, @codyoss do you know?
I skimmed through https://golang.org/doc/effective_go and https://github.com/golang/go/wiki/CodeReviewComments

grayside (changes requested):
This is a very good sample PR, thank you. I can tell a lot of care was taken to track existing practices & patterns.

Unfortunately the google-cloudevents-go library is not recommended for use, we need to try another way. You can see in https://github.com/GoogleCloudPlatform/golang-samples/blob/main/functions/functionsv2/hellopubsub/hello_pubsub_test.go we created an ad hoc struct. We hope to switch to that library in the future.

functions/functionsv2/helloauditlog/hello_auditlog.go
muncus: I'll remove the use of this library, and let you know when i'm ready for another review.

There is a documentation page that appears to conflate the two things called "cloud event": https://cloud.google.com/eventarc/docs/cloudevents#open-source
there are additional references in the Common Events section that link to protobufs, and the google-cloudevents repo docs.

functions/functionsv2/helloauditlog/hello_auditlog.go
muncus: Done. ive added the link to the PR description. Thanks for providing the link!

functions/functionsv2/helloauditlog/hello_auditlog.go
muncus: This may warrant a followup issue for the cloudevent library, because the method name property is not present in the [event.Event object](https://pkg.go.dev/github.com/cloudevents/sdk-go/v2@v2.6.1/event#Event).
How do you suggest we proceed?

functions/functionsv2/helloauditlog/hello_auditlog_test.go
muncus: convention or not, i'm happy to add that. it seems like a reasonable practice, even if it is not standard. (will leave comment open for cody's input).

muncus (commented):
I think i've addressed your comments, and this is ready for another round of review.
(some comments left open for your input, feel free to resolve if you're satisfied).

functions/functionsv2/helloauditlog/hello_auditlog.go
muncus: The inconsistency i mentioned here has now been corrected in the canonical sample. this is now consistent.

functions/functionsv2/helloauditlog/hello_auditlog.go
```diff
@@ -0,0 +1,68 @@
...
+	// Print out details from the CloudEvent itself
+	// See https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#subject
+	// for details on the Subject field
+	log.Printf("Event Type: %s", e.Type())
+	log.Printf("Subject: %s", e.Subject())
+
+	// Decode the Cloud Audit Logging message embedded in the CloudEvent
+	logentry := &AuditLogEntry{}
+	if err := e.DataAs(logentry); err != nil {
+		ferr := fmt.Errorf("event.DataAs: %v", err)
```
grayside, line 53: nit: Maybe this should wrap the error with %w?

grayside (approved):

functions/functionsv2/helloauditlog/hello_auditlog.go
muncus: Done!

ace-n:
@tbpg can you rubber-stamp this?

functions/functionsv2/helloauditlog/hello_auditlog.go
```diff
@@ -0,0 +1,68 @@
...
+
+// AuditLogEntry represents a LogEntry as described at
+// https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry
+type AuditLogEntry struct {
+	ProtoPayload *AuditLogProtoPayload `json:"protoPayload"`
+}
+
+// AuditLogProtoPayload represents AuditLog within the LogEntry.protoPayload
+// See https://cloud.google.com/logging/docs/reference/audit/auditlog/rest/Shared.Types/AuditLog
+type AuditLogProtoPayload struct {
```
codyoss, line 36: Should we me making use of generated proto stubs? https://github.com/googleapis/go-genproto/blob/4c6863e31247658eee07a5b0075340c04399d1d4/googleapis/cloud/audit/audit_log.pb.go#L43

Or does this even lib not work with those types?

grayside: Hey Cody, we're following a medium-term pattern of inlining the data structure. We're doing this both as a bridge to the future library dependency we plan to take (from https://github.com/googleapis/google-cloudevents libraries) and to cover a gap in documentation that's still being resolved.

I'm not strongly opposed to go using that reference, if you feel it's a reasonable thing for developers to take direct dependency. It would be a difference for this sample vs. the other GCF v2 samples.

codyoss: The inline struct is fine by me, I just wanted to make sure this choice was intentional. Anyone that uses our client libraries would already have this dep in their tree.

codyoss (approved):

PR #2307, opened 2021-11-18
docs: update CONTRIBUTING with go.sum info

tritone:
It's easy to miss this step and will cause hard-to-debug
build failures if you do.


codyoss (approved):
Thank you!

PR #2312, opened 2021-12-02
feat(cloudsql/mysql): update to V2 sample

enocom:
* Add connector example in addition to TCP and UNIX socket example

* Add support for Cloud Functions
    - Create a cmd/app directory for the main package
    - Rename the top-level package "cloudsql"
    - Move template into Go file for simple deployment
    - use "/Votes" as endpoint for casting a vote to match default Cloud
      Functions naming scheme

* Simplify application code

* Add new region tags while leaving old tags in place

This commit leaves a few open TODOs:
- [x] ~Migrate to the new region tags~ Pin region tags in docs until these changes are launched
- [x] Update the README
- [x] Fix the integration tests


cloudsql/postgres/database-sql/cloudsql_test.go
```diff
@@ -12,14 +12,16 @@
...
+package cloudsql
 
 import (
 	"bytes"
 	"net/http/httptest"
 	"os"
 	"strings"
 	"testing"
+
+	cloudsql "github.com/GoogleCloudPlatform/golang-samples/cloudsql/postgres/database-sql"
```
kurtisvg, line 24: hmm isn't this package the package we are in? Why do we need an import here?

cloudsql/postgres/database-sql/cmd/app/.gcloudignore
```diff
@@ -0,0 +1,25 @@
+# This file specifies files that are *not* uploaded to Google Cloud
```
kurtisvg, line 1: Do we need a .gcloudignore here? Doesn't it use .gitignore by default?

kurtisvg (changes requested):
We probably should have split this into 3 prs (one for each db engine) to help limit the size.

enocom:
I'll split this into two more PRs. We'll start with MySQL here (since that's where the majority of comments are).

cloudsql/postgres/database-sql/cmd/app/.gcloudignore
enocom: This is a mistaken addition. I'll remove it when I send the postgres PR.

cloudsql/postgres/database-sql/cloudsql_test.go
enocom: Another bogus addition. Will fix in the postgres PR.

cloudsql/mysql/database-sql/cloudsql.go
```diff
@@ -12,361 +12,329 @@
...
-	// [END cloud_sql_mysql_databasesql_limit]
-
-	// [START cloud_sql_mysql_databasesql_lifetime]
-
-	// Set Maximum time (in seconds) that a connection can remain open.
-	dbPool.SetConnMaxLifetime(1800 * time.Second)
-
-	// [END cloud_sql_mysql_databasesql_lifetime]
-}
+var indexHTML = `
```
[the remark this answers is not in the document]
enocom: A separate file is much nicer, but getting it to work in Cloud Functions requires the code look in a special place to find the template. So rather than introduce runtime-specific concerns into the app, I chose to imbed the template for simplicity.

cloudsql/mysql/database-sql/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
+# Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
enocom: Fixed.

cloudsql/mysql/database-sql/connect_connector.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		return v
+	}
+
+	var (
+		dbUser                 = mustGetenv("DB_USER")                  // e.g. 'my-db-user'
+		dbPwd                  = mustGetenv("DB_PASS")                  // e.g. 'my-db-password'
+		dbName                 = mustGetenv("DB_NAME")                  // e.g. 'my-database'
+		instanceConnectionName = mustGetenv("INSTANCE_CONNECTION_NAME") // e.g. 'project:region:instance'
+	)
```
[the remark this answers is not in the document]
enocom: Definitely. I'll do that.

cloudsql/mysql/database-sql/README.md
```diff
@@ -72,16 +72,10 @@ sudo mkdir ./cloudsql
...
-
-```bash
-export DB_SOCKET_DIR=./cloudsql
-```
-
 Use these terminal commands to initialize environment variables:
 ```bash
 export GOOGLE_APPLICATION_CREDENTIALS=/path/to/service/account/key.json
-export INSTANCE_CONNECTION_NAME='<MY-PROJECT>:<INSTANCE-REGION>:<INSTANCE-NAME>'
+export UNIX_SOCKET_PATH='./cloudsql/<MY-PROJECT>:<INSTANCE-REGION>:<INSTANCE-NAME>'
```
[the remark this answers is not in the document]
enocom: I like it. Will do.

cloudsql/mysql/database-sql/connect_connector.go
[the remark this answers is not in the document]
enocom: Actually, I forgot that we're purposely showing people in the region tags where these values come from. So, yes, it does end up as boilerplate, but has the nice quality of being perfectly clear about where these values come from.

enocom:
Adding @jackwotherspoon and @jsimonweb as reviewers. Here's what I think the reference implementation for our connectivity samples should look like. Note the README needs to be rewritten or at least revised, but I'm leaving the work aside for the moment.

cloudsql/mysql/database-sql/cmd/app/main.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func main() {
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = "8080"
+	}
+	log.Printf("Listening on port %s", port)
+
+	http.HandleFunc("/", cloudsql.Votes)
```
kurtisvg, line 32: I'm not sure I understand why we put this in a different package. Wouldn't it make sense to all be in one?

kurtisvg (approved):

cloudsql/mysql/database-sql/cmd/app/main.go
enocom: This is necessary to make functions work. Cloud Functions wants a non-main package. Everything else wants a main package.

cloudsql/mysql/database-sql/cloudsql_test.go
```diff
@@ -22,138 +22,183 @@ import (
...
+	switch ct {
+	case useTCP:
+		d.unixPath = ""
+		d.instConnName = ""
+	case useUnix:
+		d.host = ""
+		d.instConnName = ""
+	default: // connector
+		d.host = ""
+		d.host = ""
```
jsimonweb, line 80: Duplicated line. Should this be?
d.unixPath = ""

jsimonweb:
Manually testing this PR by deploying the MySQL sample app to Cloud Run using a MySQL instance:

Works: MySQL instance with Public IP
Fails: MySQL instance with Private IP (SSL enabled) and SSL server/client certs included in container
Error: 
> unable to create table: Config error: instance does not have IP of type "PUBLIC" (connection name = "[project:region:instance]")

Performing the same test by deploying the MySQL sample app to Cloud Run without the changes in this PR:
Works: MySQL instance with Public IP
Works: MySQL instance with Private IP (SSL enabled) and SSL server/client certs included in container

enocom:
@jsimonweb Did you use the connector for private IP in your test?

enocom:
The quickstarts rely on main and will break if we merge this. We need to pin the quickstarts before merging this.

cloudsql/mysql/database-sql/Dockerfile
```diff
@@ -42,9 +42,6 @@ RUN set -x && apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -
 
 # Copy the binary to the production image from the builder stage.
 COPY --from=builder /app/server /app/server
-COPY ./templates /app/templates
-# Copy any certificates IFF present.
-COPY ./certs /app/certs
```
jsimonweb, line 45:
```suggestion

# Copy any certificates IF present.
COPY ./certs /app/certs
```

jsimonweb (approved):

PR #2318, opened 2021-12-07
feat(storage): update SignedURL samples to handle auth for user 

### LINKED ISSUE #2079: storage: please improve SignedURL example
The [current situation for generating SignedURLs using the GCP SDK is awkward](https://github.com/googleapis/google-cloud-go/issues/1130#issuecomment-668863950), and that is not the fault of this repository. 

However, I still find the current sample for SignedURLs [generate_v4_get_object_signed_url.go](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/storage/objects/generate_v4_get_object_signed_url.go) to be problematic. It only works for a service account, but if you run it in GKE as Workload Identity or as a GCP User using default credentials (`authorized_user`) it will fail without any real guidance on how to succeed. Documentation about alternatives is very hard to come by at present.

I find `SignBytes` with `SignBlobRequest` to be a better mechanism than `PrivateKey` with the Service Account secret key for the `SignedURLOptions`, as it is at least possible for the code to succeed using Workload Identity or `authorized_user` default credentials if the IAM permissions are set up properly. This is a working example (which could use still further improvement):
```
package main

import (
	"context"
	"flag"
	"fmt"
	"io"
	"log"
	"os"
	"strings"
	"time"

	credentials "cloud.google.com/go/iam/credentials/apiv1"
	credentialspb "google.golang.org/genproto/googleapis/iam/credentials/v1"

	"cloud.google.com/go/storage"
	"golang.org/x/oauth2/google"
	"google.golang.org/api/option"
)

const serviceAccountEmailOrId = "SERVICEACCOUNT_ID@GCPPROJECT.iam.gserviceaccount.com" //nolint:lll // duh

func main() {
	ctx := context.Background()

	gcpCredentials, err := google.FindDefaultCredentials(ctx, storage.ScopeFullControl)
	if err != nil {
		log.Fatal(err)
	}
	client, err := storage.NewClient(ctx, option.WithCredentials(gcpCredentials))
	if err != nil {
		log.Fatal(err)
	}

	var o string
	flag.StringVar(&o, "o", "", "source object; in the format of <bucket:object>")
	flag.Parse()

	names := strings.Split(o, ":")
	if len(names) < 2 {
		usage("missing -o flag")
	}
	bucket, object := names[0], names[1]

	if len(os.Args) < 3 {
		usage("missing subcommand")
	}

	switch os.Args[2] {
	case "write":
		if err := write(client, bucket, object); err != nil {
			log.Fatalf("Cannot write object: %v", err)
		}
	case "sign":
		value, signErr := generateV4GetObjectSignedURL(os.Stdout, bucket, object, gcpCredentials)
		if signErr != nil {
			log.Fatalf("Cannot sign object: %+v", signErr)
		}
		log.Println(value)
	}
}

const helptext = `usage: objects -o=bucket:name [subcommand] <args...>

subcommands:
	- write
	- sign
`

func usage(msg string) {
	if msg != "" {
		fmt.Fprintln(os.Stderr, msg)
	}
	fmt.Fprint(os.Stderr, helptext)
	os.Exit(2)
}

func write(client *storage.Client, bucket, object string) error {
	ctx := context.Background()
	f, err := os.Open(object)
	if err != nil {
		return err
	}
	defer f.Close()

	ctx, cancel := context.WithTimeout(ctx, time.Second*50)
	defer cancel()
	wc := client.Bucket(bucket).Object(object).NewWriter(ctx)
	if _, err = io.Copy(wc, f); err != nil {
		return err
	}
	if err := wc.Close(); err != nil {
		return err
	}
	return nil
}

// generateV4GetObjectSignedURL generates object signed URL with GET method.
// w is stdout for logging
//
func generateV4GetObjectSignedURL(
	w io.Writer,
	bucket,
	object string,
	gcpCredentials *google.Credentials,
) (string, error) {
	fmt.Fprintln(w, "Got here!")
	ctx := context.Background()

	var opts *storage.SignedURLOptions

	credentialsClient, err := credentials.NewIamCredentialsClient(
		ctx, option.WithCredentials(gcpCredentials))
	if err != nil {
		return "", err
	}

	defer credentialsClient.Close()
	// service account will work fine, but gcp authorized_user type credentials
	// requires Service Account Token Creator and Service Account User IAM roles
	// on the service Account (districts team should have this)
	opts = &storage.SignedURLOptions{
		Method:         "GET",
		GoogleAccessID: serviceAccountEmailOrId,
		SignBytes: func(b []byte) ([]byte, error) {
			req := &credentialspb.SignBlobRequest{
				Payload: b,
				Name:    serviceAccountEmailOrId,
			}
			resp, signBlobErr := credentialsClient.SignBlob(ctx, req)
			if signBlobErr != nil {
				return nil, signBlobErr
			}
			return resp.SignedBlob, signBlobErr
		},
		Expires: time.Now().Add(15 * time.Minute),
	}

	u, signErr := storage.SignedURL(bucket, object, opts)
	if signErr != nil {
		return "", fmt.Errorf("unable to generate SignedURL: %+v", signErr)
	}

	fmt.Fprintln(w, "You can use this signed URL with any user agent, for example:")
	fmt.Fprintf(w, "curl %q\n", u)
	return u, nil
}

```

BrennaEpp:
Fixes #2079

Similar change to PostPolicy samples should be made once Storage 19.0 is released.

### DIFF
--- a/storage/objects/generate_v4_get_object_signed_url.go
+++ b/storage/objects/generate_v4_get_object_signed_url.go
@@ -39,7 +39,7 @@ func generateV4GetObjectSignedURL(w io.Writer, bucket, object string) (string, e
 	// Signing a URL requires credentials authorized to sign a URL. You can pass
 	// these in through SignedURLOptions with one of the following options:
 	// 		a. a Google service account private key, obtainable from the Google Developers Console
-	// 		b. a Google Access ID with iam.serviceAccounts.signBlob permissions
+	// or	b. a Google Access ID with iam.serviceAccounts.signBlob permissions
 	// or  	c. a SignBytes function implementing custom signing.
 	// In this example, none of these options are used, which means the following
 	// code only works when the credentials are defined via the environment variable
--- a/storage/objects/generate_v4_get_object_signed_url.go
+++ b/storage/objects/generate_v4_get_object_signed_url.go
@@ -16,38 +16,41 @@ package objects
 
 // [START storage_generate_signed_url_v4]
 import (
+	"context"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"time"
 
 	"cloud.google.com/go/storage"
-	"golang.org/x/oauth2/google"
 )
 
 // generateV4GetObjectSignedURL generates object signed URL with GET method.
-func generateV4GetObjectSignedURL(w io.Writer, bucket, object, serviceAccount string) (string, error) {
+// Signing a URL requires credentials authorized to sign a URL. You can pass
+// these in as a Google service account private key (obtainable from the Google
+// Developers Console) or using a SignBytes function implementing custom signing.
+// In this example, neither of these options are used, which means the following
+// code only works when the credentials are defined via the environment variable
+// GOOGLE_APPLICATION_CREDENTIALS.
+func generateV4GetObjectSignedURL(w io.Writer, bucket, object string) (string, error) {
 	// bucket := "bucket-name"
 	// object := "object-name"
-	// serviceAccount := "service_account.json"
-	jsonKey, err := ioutil.ReadFile(serviceAccount)
+
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
 	if err != nil {
-		return "", fmt.Errorf("ioutil.ReadFile: %v", err)
+		return "", fmt.Errorf("storage.NewClient: %v", err)
 	}
-	conf, err := google.JWTConfigFromJSON(jsonKey)
-	if err != nil {
-		return "", fmt.Errorf("google.JWTConfigFromJSON: %v", err)
-	}
+	defer client.Close()
+
 	opts := &storage.SignedURLOptions{
-		Scheme:         storage.SigningSchemeV4,
-		Method:         "GET",
-		GoogleAccessID: conf.Email,
-		PrivateKey:     conf.PrivateKey,
-		Expires:        time.Now().Add(15 * time.Minute),
+		Scheme:  storage.SigningSchemeV4,
+		Method:  "GET",
+		Expires: time.Now().Add(15 * time.Minute),
 	}
-	u, err := storage.SignedURL(bucket, object, opts)
+
+	u, err := client.Bucket(bucket).SignedURL(object, opts)
 	if err != nil {
-		return "", fmt.Errorf("storage.SignedURL: %v", err)
+		return "", fmt.Errorf("Bucket(%q).SignedURL: %v", bucket, err)
 	}
 
 	fmt.Fprintln(w, "Generated GET signed URL:")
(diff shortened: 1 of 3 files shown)


storage/objects/generate_v4_get_object_signed_url.go
```diff
@@ -16,38 +16,41 @@ package objects
...
 )
 
 // generateV4GetObjectSignedURL generates object signed URL with GET method.
-func generateV4GetObjectSignedURL(w io.Writer, bucket, object, serviceAccount string) (string, error) {
+// Signing a URL requires credentials authorized to sign a URL. You can pass
+// these in as a Google service account private key (obtainable from the Google
+// Developers Console) or using a SignBytes function implementing custom signing.
+// In this example, neither of these options are used, which means the following
+// code only works when the credentials are defined via the environment variable
+// GOOGLE_APPLICATION_CREDENTIALS.
```
tritone, line 33: This is not quite accurate, correct? It could also populate signBlob.

I would also maybe move this comment down to where the creds could be passed in via the SignedURLOptions and make it explicit that that's where you would do it.

storage/objects/generate_v4_put_object_signed_url.go
```diff
@@ -16,42 +16,46 @@ package objects
...
 
-// generateV4PutObjectSignedURL generates object signed URL with PUT method.
-func generateV4PutObjectSignedURL(w io.Writer, bucket, object, serviceAccount string) (string, error) {
+// generateV4GetObjectSignedURL generates object signed URL with PUT method.
+// Signing a URL requires credentials authorized to sign a URL. You can pass
+// these in as a Google service account private key (obtainable from the Google
+// Developers Console) or using a SignBytes function implementing custom signing.
+// In this example, neither of these options are used, which means the following
+// code only works when the credentials are defined via the environment variable
+// GOOGLE_APPLICATION_CREDENTIALS.
```
tritone, line 33: Same as above

storage/objects/objects_test.go
```diff
@@ -360,14 +360,14 @@ func TestV4SignedURL(t *testing.T) {
 
 	bucketName := tc.ProjectID + "-signed-url-bucket-name"
 	objectName := "foo.txt"
-	serviceAccount := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
-	if serviceAccount == "" {
+
+	if os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") == "" {
```
tritone, line 364: I think we can assume that users need to have credentials in order to run these tests.

tritone (commented):
Some docs questions; otherwise very nice cleanup!

storage/objects/objects_test.go
BrennaEpp: You're right, I'm deleting the check. It will error with an appropriate message if the credentials are not set.

tritone (approved):
One formatting nit, otherwise looks good!

PR #2320, opened 2021-12-09
feat(functions): New v2 sample for GCE labeling on new instance creation

muncus:
- feat(functions): gcfv2 sample for GCE audit logs.
- add a check for malformed subject.


functions/functionsv2/label_gce_instance/go.mod
```diff
@@ -0,0 +1,29 @@
...
+	cloud.google.com/go v0.97.0
+	github.com/googleapis/gax-go/v2 v2.1.0
+	github.com/json-iterator/go v1.1.10 // indirect
+	github.com/modern-go/concurrent v0.0.0-20180228061459-e0a39a4cb421 // indirect
+	github.com/modern-go/reflect2 v0.0.0-20180701023420-4b7aa43c6742 // indirect
+	google.golang.org/genproto v0.0.0-20210924002016-3dee208752a0
+	google.golang.org/protobuf v1.27.1
+)
+
+require (
```
grayside, line 17: Not familiar with this syntax using multiple require blocks. Perhaps a `go mod tidy` is in order?

functions/functionsv2/label_gce_instance/label_gce_instance.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"context"
+	"fmt"
+	"log"
+	"regexp"
+	"strings"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/googleapis/gax-go/v2"
```
grayside, line 29: Very rare to see gax-go as a sample dependency. Curious to see where it's used as I continue the review.

functions/functionsv2/label_gce_instance/label_gce_instance.go
```diff
@@ -0,0 +1,136 @@
...
+	logentry := &AuditLogEntry{}
+	if err := e.DataAs(logentry); err != nil {
+		werr := fmt.Errorf("event.DataAs() : %w", err)
+		log.Printf("Error parsing proto payload: %s", werr)
+		return werr
+	}
+	payload := logentry.ProtoPayload
+	creator, ok := payload.AuthenticationInfo["principalEmail"]
+	if !ok {
+		e := fmt.Errorf("principalEmail not found in cloud event payload: %v", payload)
```
grayside, line 78: suggestion: Given how common it is to use `e` as a secondary name for `err`, maybe we should change the variable representing event to `ce` or `ev`?

functions/functionsv2/label_gce_instance/label_gce_instance.go
```diff
@@ -0,0 +1,136 @@
...
+	if len(paths) < 6 {
+		return fmt.Errorf("invalid event subject: %s", e.Subject())
+	}
+	project := paths[2]
+	zone := paths[4]
+	instance := paths[6]
+
+	// Format the `creator` label value to match GCE label requirements
+	// See https://cloud.google.com/compute/docs/labeling-resources#requirements
+	labelsan := regexp.MustCompile("[^a-z0-9_-]+")
```
grayside, line 96: e.g., "label sanitization"?

functions/functionsv2/label_gce_instance/label_gce_instance_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"testing"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/googleapis/gax-go/v2"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+)
+
+// FakeInstancesClient is a GceInstancesClient that does not call GCE, and returns stub values.
+type FakeInstancesClient struct {
```
grayside, line 30: Thanks for putting together a thorough unit test!

Generally, we prefer integration tests (or extra credit, [e2e tests](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/functions/helloworld/hello_http_system_test.go)). These lead to the test doubling as a check on overall sample reliability.

Identifying how to consistently meet that standard in CloudEvent functions samples is forthcoming, so this is not a blocking concern but I wanted to flag as a specific case since this testing approach is affecting the sample UX.

grayside (approved):
Looking great, left a few points of feedback but nothing blocking. We may want to request GCE DPE product review before merging.

Process Note: Title the PR using conventionalcommits.org, amongst other things this helps the automation identify which product to use. New samples generally `feat(product-short-name): Short sample explanation`

functions/functionsv2/label_gce_instance/go.mod
muncus: This was not an intentional choice, and tidy does not merge the two blocks.
I've merged them manually and everything seems fine.

functions/functionsv2/label_gce_instance/label_gce_instance.go
muncus: good call. fixed.

functions/functionsv2/label_gce_instance/label_gce_instance_test.go
muncus: I agree that this approach to testing is not great for the sample readability and UX (though it is one of the [recommended approaches](https://github.com/googleapis/google-cloud-go/blob/main/testing.md#testing-using-mocks). )

I see at least the following options, and am open to others:
We could leave the tests as-is, or ( to prioritize the sample UX over the testing) switch to just testing the function's return value.
How do you suggest we proceed?

codyoss: I think it may be cleaner to actually use a fake here instead of a mock for the sake of the sample, see the first section in the doc you linked.  Then we don't need to define an interface in the sample at all and can just work with the real client.

muncus: I discussed this with Cody, and it turns out the GCE services require an [http api approach](https://github.com/googleapis/google-api-go-client/blob/main/testing.md#testing-http-services-using-fakes), which is different from the grpc guidance I initially linked.
I've updated the sample tests to use a fake server.

functions/functionsv2/label_gce_instance/label_gce_instance.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+// Cloud Function that receives GCE instance creation Audit Logs, and adds a
+// `creator` label to the instance.
+func LabelGceInstance(ctx context.Context, ev event.Event) error {
+	// Extract parameters from the Cloud Event and Cloud Audit Log data
+	logentry := &AuditLogEntry{}
+	if err := ev.DataAs(logentry); err != nil {
+		werr := fmt.Errorf("event.DataAs() : %w", err)
```
codyoss, line 64: nit: perhaps just shadow err here.

functions/functionsv2/label_gce_instance/label_gce_instance.go
```diff
@@ -0,0 +1,129 @@
...
+		log.Printf("Error parsing proto payload: %s", werr)
```
codyoss, line 65: nit: I am not sure we need these extra logs where errors are returned. Seem duplicative.

functions/functionsv2/label_gce_instance/label_gce_instance.go
```diff
@@ -0,0 +1,129 @@
...
+		return werr
+	}
+	payload := logentry.ProtoPayload
+	creator, ok := payload.AuthenticationInfo["principalEmail"]
+	if !ok {
```
codyoss, line 70: nit: on line below consider err for naming(github wont let me comment on the line below right now...)

functions/functionsv2/label_gce_instance/label_gce_instance_test.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func NewFakeInstancesServer(i *computepb.Instance) *FakeInstancesServer {
+	fake := &FakeInstancesServer{
+		instance:      i,
+		changedLabels: make(map[string]string),
+	}
+	fake.HandleFunc("/compute/v1/projects/PROJECT/zones/ZONE/instances/INSTANCE",
+		func(w http.ResponseWriter, r *http.Request) {
+			rbytes, err := json.Marshal(fake.instance)
```
codyoss, line 44: Since we are working with generated proto stubs we should use the protojson library: https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson

codyoss (changes requested):
Just a couple small things, or else it LGTM. Thanks for refactoring the testing!

functions/functionsv2/label_gce_instance/label_gce_instance.go
muncus: Done here, and in a similar case below on line 99

functions/functionsv2/label_gce_instance/label_gce_instance.go
muncus: That was my initial though as well, but in my experience, just returning the error gets you a log message that the response was a 500, but does not include the string representation of the returned error.
I think the added logging is worthwhile, since it provides necessary debugging info for users of the sample.

functions/functionsv2/label_gce_instance/label_gce_instance_test.go
muncus: Thanks, i did not know about the protojson library. done.

functions/functionsv2/label_gce_instance/label_gce_instance.go
codyoss: Ack, we can do it here if you feel strongly, but this is not something we have been doing in general with our samples.

codyoss (approved):

PR #2325, opened 2021-12-09
samples(storage): streaming file upload and download

shaffeeullah:
For this page: https://cloud.google.com/storage/docs/streaming#storage-stream-upload-object-go


storage/objects/objects_test.go
```diff
@@ -67,7 +67,7 @@ func TestObjects(t *testing.T) {
 		t.Fatalf("uploadFile(%q): %v", object2, err)
 	}
 
-	if err := uploadFile(ioutil.Discard, bucketVersioning, object1); err != nil {
+	if err := streamFileUpload(ioutil.Discard, bucketVersioning, object1); err != nil {
 		t.Fatalf("uploadFile(%q): %v", object1, err)
```
BrennaEpp, line 71: This line should be updated to print out the different function

`t.Fatalf("streamFileUpload(%q) ...`

storage/objects/stream_file_upload.go
```diff
@@ -0,0 +1,61 @@
...
+	wc.ChunkSize = 0 // note retries are not supported for chunk size 0
+	if _, err = io.Copy(wc, buf); err != nil {
+		return fmt.Errorf("io.Copy: %v", err)
+	}
+	// Data can continue to be added to the file until the writer is closed
+	if err := wc.Close(); err != nil {
+		return fmt.Errorf("Writer.Close: %v", err)
+	}
+	fmt.Fprintf(w, "%v uploaded to %v.\n", object, bucket)
+	//object created (match node)
```
BrennaEpp, line 56: Forgot to delete this comment?

storage/objects/stream_file_upload.go
```diff
@@ -0,0 +1,61 @@
...
+	ctx, cancel := context.WithTimeout(ctx, time.Second*50)
+	defer cancel()
+
+	// Upload an object with storage.Writer.
+	wc := client.Bucket(bucket).Object(object).NewWriter(ctx)
+	wc.ChunkSize = 0 // note retries are not supported for chunk size 0
+	if _, err = io.Copy(wc, buf); err != nil {
+		return fmt.Errorf("io.Copy: %v", err)
+	}
+	// Data can continue to be added to the file until the writer is closed
```
BrennaEpp, line 51: Should we add an example of this? Perhaps copy in another buffer?

also nit: comment should end with a period to be consistent with the rest of the samples

BrennaEpp (commented):
Thanks for working on this! Just a couple comments

storage/objects/stream_file_upload.go
shaffeeullah: Spoke with @tritone regarding this and we decided that it might over-complicate the sample to add an example. Go users would presumably know how to do this since it is not a storage operation. What do you think? I'm okay either way.

storage/objects/stream_file_upload.go
shaffeeullah: Good callout! Fixed.

storage/objects/objects_test.go
shaffeeullah: Good catch

storage/objects/stream_file_upload.go
BrennaEpp: Yeah, that's true. The comment should probably be enough

BrennaEpp (approved):
LGTM

storage/objects/stream_file_upload.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2020 Google LLC
```
ryan-dayrit, line 1: Should this be `Copyright 2021 Google LLC`?

shaffeeullah: good catch!

ryan-dayrit (approved):
lgtm

PR #2331, opened 2021-12-14
docs(bigquery): add samples for materialized views

shollyman:


tswast (approved):
LGTM. I'm not sure why we decided to use a slightly different query in Python sample.

PR #2335, opened 2021-12-15
samples(storage): download byte range

shaffeeullah:


storage/objects/download_byte_range.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"cloud.google.com/go/storage"
+)
+
+// downloadByteRange downloads a specific byte range of an object to a file.
+func downloadByteRange(w io.Writer, bucket, object string, startByte int64, length int64, destFileName string) error {
+	// bucket := "bucket-name"
+	// object := "object-name"
+	// startByte := 0
+	// length := 20
```
tritone, line 33: I see in the Node sample you have start byte and end byte rather than length. Maybe it's good to do the same here and show how you would subtract to get length in the sample, for consistency? I just want it to be clear that the required params are different in Go.

storage/objects/objects_test.go
```diff
@@ -204,6 +204,26 @@ func TestObjects(t *testing.T) {
 		}
 	})
 
+	t.Run("downloadByteRange", func(t *testing.T) {
+		dir, err := ioutil.TempDir("", "downloadByteRangeTestTempDir")
```
tritone, line 208: You could potentially de-duplicate this with your other sample test and just create one tempdir at the beginning of TestObjects to use for both subtests-- especially if you're going to add additional samples that will require a local dir.

storage/objects/download_byte_range.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*50)
+	defer cancel()
+
+	f, err := os.Create(destFileName)
+	if err != nil {
+		return fmt.Errorf("os.Create: %v", err)
+	}
```
tritone, line 48: The file should be closed at the end of the sample as well. In this case, you shouldn't use a defer statement because we need to check for errors in writing the file. Instead just do it manually after L58.

I may have forgotten about this in the previous sample, could you double-check that one as well and fix if necessary?

tritone (commented):
Thanks!

storage/objects/download_byte_range.go
```diff
@@ -0,0 +1,72 @@
...
+		return fmt.Errorf("Object(%q).NewReader: %v", object, err)
+	}
+	defer rc.Close()
+
+	if _, err := io.Copy(f, rc); err != nil {
+		return fmt.Errorf("io.Copy: %v", err)
+	}
+
+	err = f.Close()
+	if err != nil {
```
tritone, line 62: nit: can inline this error as well.

storage/objects/download_file.go
```diff
@@ -54,6 +54,12 @@ func downloadFile(w io.Writer, bucket, object string, destFileName string) error
 	if _, err := io.Copy(f, rc); err != nil {
 		return fmt.Errorf("io.Copy: %v", err)
 	}
+
+	err = f.Close()
+	if err != nil {
```
tritone, line 59: And same here

tritone (approved):
One nit, otherwise LGTM

storage/objects/download_file.go
shaffeeullah: Awesome, thanks Chris!

PR #2337, opened 2021-12-16
test(spanner): fix backup tests

rahul2393:
TestCustomerManagedEncryptionKeys need encryption key in same region as test instance, and ListBackupOperation needs metadata.@type as first filter option similar issue in C++ https://github.com/googleapis/google-cloud-cpp/issues/7741.

Fixes: https://github.com/GoogleCloudPlatform/golang-samples/issues/2333 https://github.com/GoogleCloudPlatform/golang-samples/issues/2334


hengfengli (approved):
LGTM. Nice catch!

PR #2353, opened 2021-12-20
branch changes from sample app for code review

davidianstyle:


vision/vision_quickstart/main.go
```diff
@@ -20,44 +20,153 @@ package main
...
-		log.Fatalf("Failed to detect labels: %v", err)
+	// Declare a SafeSearch values to include
+	safeSearchValuesToInclude := map[string]bool {
+		"UNKOWN": true,
+		"POSSIBLE": true,
+		"LIKELY": true,
+		"VERY_LIKELY": true,
+	}
+
+	var names []string
```
rmhofmann, line 85: i like to explicitly initialize as an empty slice, like `names := []string{}`

vision/vision_quickstart/main.go
```diff
@@ -20,44 +20,153 @@ package main
...
+
+// [START storage_copy_file]
+
+// copyFile copies an object into specified bucket.
+func copyFile(w io.Writer, dstBucket, srcBucket, srcObject string, categorization string, confidence string) error {
+	// dstBucket := "bucket-1"
+	// srcBucket := "bucket-2"
+	// srcObject := "object"
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
```
rmhofmann, line 152: rather than redefine ctx and client, could pass them thru as args to copyFile.  Most of the Go code i write takes (ctx, c) as the first two arguments, for this reason

davidianstyle: Ah, good point. I copied this from `https://github.com/GoogleCloudPlatform/golang-samples/blob/main/storage/objects/copy_file.go`. Is there a better way I should be importing that function rather than copy/paste?

Mstandastote (changes requested):
#2353

PR #2356, opened 2021-12-23
sample(storage): add retry config sample

tritone:
Adds a sample for a user-configurable retry strategy using the
new surface in the storage library. Sample will go here:
https://cloud.google.com/storage/docs/retry-strategy#tools

Depends on library changes being release and dependency on
storage bumped in this repo.


BrennaEpp (approved):
👍

storage/retry/configure_retries.go
```diff
@@ -0,0 +1,71 @@
...
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"github.com/googleapis/gax-go/v2"
+)
+
+// configureRetries configures a custom retry strategy for an API call.
```
cbonnie, line 28: // Configures a custom retry strategy for a single API call.

storage/retry/configure_retries.go
```diff
@@ -0,0 +1,71 @@
...
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("storage.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	// Configure retries for all operations using this ObjectHandle. Retries may
+	// also be configured on the BucketHandle or Client types.
+	o := client.Bucket(bucket).Object(object).Retryer(
+		// Use WithBackoff to control the timing of the exponential backoff. As
```
cbonnie, line 42: "// Use WithBackoff to control the timing of the exponential backoff."

storage/retry/configure_retries.go
```diff
@@ -0,0 +1,71 @@
...
+		// configured here, there will be an initial delay between retries of up to
+		// 2s, a maximum delay of 60s, and the pause length will be multiplied by 3
+		// in each iteration. The actual length of pauses between retries is subject
+		// to random jitter.
+		storage.WithBackoff(gax.Backoff{
```
cbonnie, line 47: I would do this as: 

// Set the initial retry delay to a maximum of 2 seconds. The length of pauses between retries is subject to random jitter.
Initial: 2 * time.Second, 
// Set the maximum retry delay to 60 seconds.
Max: 60 * time.Second,
// Set the backoff multiplier to 3.0.
...

This reflects the language described in the docs and also matches the language in other client samples

cbonnie: Can we also show how to customize the max number of retry attempts?

storage/retry/configure_retries.go
```diff
@@ -0,0 +1,71 @@
...
+			Initial:    2 * time.Second,
+			Max:        60 * time.Second,
+			Multiplier: 3,
+		}),
+		// Use WithPolicy to configure the idempotency policy. By default, only
```
cbonnie, line 52: "// Use WithPolicy to customize retry so that all requests are retried even if they are non-idempotent."

storage/retry/configure_retries.go
```diff
@@ -0,0 +1,71 @@
...
+		// even if it is non-idempotent.
+		storage.WithPolicy(storage.RetryAlways),
+	)
+
+	// Use context timeouts to set an overall deadline on the call, including all
+	// potential retries.
+	ctx, cancel := context.WithTimeout(ctx, 500*time.Second)
+	defer cancel()
+
+	// Delete an object using the specified policy.
```
cbonnie, line 63: // Delete an object using the specified retry policy"

storage/retry/configure_retries.go
tritone: Thanks for the suggestions-- I took these. As for max attempts, we do not provide that lever in Go; users can provide an overall timeout instead (as on L60).

cbonnie (approved):

PR #2363, opened 2021-12-29
samples(compute): add create instances compute samples

FrodoTheTrue:


compute/instances/create-start-instance/utils_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+)
+
+func waitZoneOp(projectId, zone string, op compute.Operation) error {
```
codyoss, line 25: I think these helper methods should take the context as an argument.

compute/instances/create-start-instance/utils_test.go
```diff
@@ -0,0 +1,73 @@
...
+	ctx := context.Background()
+	zoneOperationsClient, err := compute.NewZoneOperationsRESTClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewZoneOperationsRESTClient: %v", err)
```
codyoss, line 29: no need to wrap the errors in this file as they seem to be wrapped again in calling code. You can just return the error for these helpers.

compute/instances/create-start-instance/create_instance_with_subnet.go
```diff
@@ -0,0 +1,116 @@
...
+	ctx := context.Background()
+	instancesClient, err := compute.NewInstancesRESTClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewInstancesRESTClient: %v", err)
+	}
+	imagesClient, err := compute.NewImagesRESTClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewImagesRESTClient: %v", err)
+	}
+	defer instancesClient.Close()
```
codyoss, line 45: nit: I would move this defer up under the error check. If the second client errored the client would be left open.

codyoss: This happens in other files as well.

compute/instances/create-start-instance/create_instance_with_snapshotted_data_disk.go
```diff
@@ -0,0 +1,124 @@
...
+			Project:   projectID,
+			Zone:      zone,
+		}
+		zoneOp, err := zoneOperationsClient.Wait(ctx, waitReq)
+		if err != nil {
+			return fmt.Errorf("unable to wait for the operation: %v", err)
+		}
+
+		if *zoneOp.Status.Enum() == computepb.Operation_DONE {
+			fmt.Fprintf(w, "Instance created\n")
```
codyoss, line 116: Since there is no formatting used here you could use fmt.Fprintln instead and omit the newline.

codyoss (commented):
Just a few nits

codyoss (approved):

PR #2376, opened 2022-01-07
chore: split out into more modules

codyoss:
Some builds times were quite long when change was made in the root
modules as many sub-directories were apart of this module. This commit
adds more modules so that our change detector can be more selective about
what tests are run. This should also hopefully help with some quota issues
as tests will be run less frequently.


tritone (approved):

hongalex (approved):
Pub/Sub stuff looks good to me. At some point, we might want to bump the Pub/Sub version of other modules too.

codyoss:
There will be a follow up PR to this one that removes a lot of the replace statements. it can't be done here without some hackery so I will wait and do it as a followup.

PR #2379, opened 2022-01-14
docs(bigquery): add storage write pending stream example

shollyman:
This is in the vein of the python bare API sample, but goes a bit further in exploring the surface.  Given the differences, it may be appropriate to use a different region tag.


bigquery/snippets/managedwriter/bigquerystorage_append_rows_pending.go
```diff
@@ -0,0 +1,205 @@
...
+
+		// DATETIME uses the literal format.
+		DatetimeCol: proto.String("2022-01-01 12:13:14.000000"),
+
+		// GEOGRAPHY uses Well-Known-Text (WKT) format.
+		GeographyCol: proto.String("POINT(-122.350220 47.649154)"),
+
+		// NUMERIC and BIGNUMERIC can be passed as string, or more efficiently
+		// using a packed byte representation.
+		NumericCol:    proto.String("99999999999999999999999999999.999999999"),
```
tswast, line 67: We intentionally used a rather small schema in `bigquerystorage_append_rows_pending`. 

https://github.com/googleapis/python-bigquery-storage/blob/main/samples/snippets/append_rows_pending.py

`bigquerystorage_append_rows_raw_proto2` has a more thorough schema

https://github.com/googleapis/python-bigquery-storage/blob/main/samples/snippets/append_rows_proto2.py

but indeed perhaps a different name is desired for Go since it's less "raw" than Python.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_pending.go
```diff
@@ -0,0 +1,205 @@
...
+	// Advance our current offset.
+	curOffset = curOffset + 1
+
+	// This time, we'll append three more rows in a single request.
+	result, err = managedStream.AppendRows(ctx, [][]byte{rowBytes, rowBytes, rowBytes}, managedwriter.WithOffset(curOffset))
+	if err != nil {
+		return fmt.Errorf("AppendRows second call error: %v", err)
+	}
+
+	// Block and wait for the result from the backend.
```
tswast, line 168: Perhaps pretty clear from the term "block", but perhaps we call out that typically that this would be done in a separate thread/goroutine?

bigquery/snippets/managedwriter/bigquerystorage_append_rows_pending.go
```diff
@@ -0,0 +1,205 @@
...
+	// we'll keep track of the current offset in the stream with curOffset.
+	var curOffset int64
+
+	result, err := managedStream.AppendRows(ctx, [][]byte{rowBytes}, managedwriter.WithOffset(0))
+	if err != nil {
+		return fmt.Errorf("AppendRows first call error: %v", err)
+	}
+
+	// GetResult blocks until we receive a response from the API.
+	recvOffset, err := result.GetResult(ctx)
```
tswast, line 153: Maybe best to add all the `GetResult` calls to the end to show off that multiple appends can get queued at a time?

bigquery/snippets/managedwriter/exampleproto/sample_data.proto
```diff
@@ -0,0 +1,61 @@
...
+
+
+
+// Define a message type representing the rows in your table.
+message SampleData {
+
+  // Define a nested message to encode STRUCT column values.
+  message SampleStruct {
+    optional int64 sub_int_col = 1;
+  }
```
tswast, line 31: Do we want to unnest this for the purpose of exercising the normalize? Or call out that if the messages are nested like this, the normalization step is optional?

bigquery/snippets/managedwriter/integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+	if err := client.Dataset(testDatasetID).Table(testTableID).Create(ctx, &bigquery.TableMetadata{
+		Schema: pendingSchema,
+	}); err != nil {
+		t.Fatalf("failed to create destination table(%q %q): %v", testDatasetID, testTableID, err)
+	}
+
+	if err := appendToPendingStream(log.Writer(), tc.ProjectID, testDatasetID, testTableID); err != nil {
+		t.Errorf("appendToPendingStream(%q %q): %v", testDatasetID, testTableID, err)
+	}
+
```
tswast, line 90: Might want to query the table afterwards to double-check the rows were committed? I think it has to be a query though and not a tabledata.list call, right?

shollyman: I do that level of validation in the library itself.  That might make sense for a longer-form sample app though.

bigquery/snippets/managedwriter/exampleproto/sample_data.proto
shollyman: Refactored the proto to have two distinct messages that require normalizing, and updated the comment.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_pending.go
shollyman: Switched to a slice of appends which get checked after the appends are all sent.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_pending.go
shollyman: Updated the comments around this to make it clearer.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_pending.go
shollyman: Switched to region tag bigquerystorage_write_pending_complexschema

tswast (approved):
Looks great, thanks

PR #2388, opened 2022-01-21
feat: add code samples and tests for jobs with captions

irataxy:
Fixes b:210892114

https://cloud.google.com/transcoder/images/caption.srt needs to be uploaded to the sample project bucket (cloud-samples-data) under the media/ directory.


media/transcoder/create_job_with_embedded_captions.go
```diff
@@ -0,0 +1,157 @@
...
+	defer client.Close()
+
+	req := &transcoderpb.CreateJobRequest{
+		Parent: fmt.Sprintf("projects/%s/locations/%s", projectID, location),
+		Job: &transcoderpb.Job{
+			OutputUri: outputURI,
+			JobConfig: &transcoderpb.Job_Config{
+				Config: &transcoderpb.JobConfig{
+					Inputs: []*transcoderpb.Input{
+						&transcoderpb.Input{
```
codyoss, line 51: When instantiating a slice item you can omit the types of the individual items. So here you can remove the `&transcoderpb.Input` part. There are about a dozen of these in this file.

codyoss: Also in other files.

codyoss (changes requested):
Just a couple of style nits

media/transcoder/create_job_with_embedded_captions.go
irataxy: Done in these files. Filed b/218557754 to clean up the rest of the files.

codyoss (approved):
Thanks!

PR #2390, opened 2022-01-22
Create README.md for App Engine Flex samples

shreejad:
Since gcloud 367.0.0 introduces a regression in `gcloud app deploy` for Flex code samples, add a README to guide users on how to preserve old functionality.


codyoss:
In general we try to avoid having READMEs in this repo. If documentation is needed it should be put on cloud.google.com as that is how most people engage with this content.

codyoss (changes requested):
See note above.

codyoss:
Closing as I have not heard back a response. This can be re-opened if still relevant.

PR #2392, opened 2022-01-26
chore(samples): add tail logs sample using api2

minherz:
Adds a sample of the entries.tail API use with GAPIC primitives.
Adds a test to validate the execution. The test success is to complete within a designated time.
Given entries.tail API is in Beta, the test introduces artificial delay between establishing a stream for tailing logs entries and first log ingestion.


logging/taillogs/taillogs.go
```diff
@@ -0,0 +1,62 @@
...
+	"context"
+	"fmt"
+	"io"
+	"log"
+
+	logging "cloud.google.com/go/logging/apiv2"
+	loggingpb "google.golang.org/genproto/googleapis/logging/v2"
+)
+
+func tailLogs(projectID string) {
```
savijatv, line 28: Can we add a comment about this function?

minherz: Added. Are there guidelines or template for code snippets in Go?

logging/taillogs/taillogs.go
```diff
@@ -0,0 +1,65 @@
...
+	logging "cloud.google.com/go/logging/apiv2"
+	loggingpb "google.golang.org/genproto/googleapis/logging/v2"
+)
+
+// tailLogs creates a channel to stream log entries that were recently ingested for a project
+func tailLogs(projectID string) {
+	// projectID := "your_project_id"
+
+	ctx := context.Background()
+	client, err := logging.NewClient(ctx)
```
codyoss, line 33: After checking the error we should defer a client.Close()

logging/taillogs/taillogs.go
```diff
@@ -0,0 +1,65 @@
...
+	req := &loggingpb.TailLogEntriesRequest{
+		ResourceNames: []string{
+			"projects/" + projectID,
+		},
+	}
+	if err := stream.Send(req); err != nil {
+		log.Fatalf("stream.Send error: %v", err)
+	}
+
+	defer stream.CloseSend()
```
codyoss, line 50: This defer should be moved up after the error check where is was created.

logging/taillogs/taillogs.go
```diff
@@ -0,0 +1,65 @@
...
+	// read and print two streamed log entries
+	for counter := 0; counter < 2; counter++ {
+		resp, err := stream.Recv()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			log.Fatalf("stream.Recv error: %v", err)
+		}
+		fmt.Printf("received:\n%v\n", resp)
```
codyoss, line 61: Up to you, but another way things like this are commonly tested is that the sample accepts an io.Writer and we verify the results in the test.

logging/taillogs/taillogs.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// tailLogs creates a channel to stream log entries that were recently ingested for a project
+func tailLogs(projectID string) {
+	// projectID := "your_project_id"
+
+	ctx := context.Background()
+	client, err := logging.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("NewClient error: %v", err)
```
codyoss, line 35: log.Fatal is basically a panic. We should instead return an error from this function.

logging/taillogs/taillogs_test.go
```diff
@@ -0,0 +1,87 @@
...
+	uuid, err := uuid.NewRandom()
+	logID := fmt.Sprintf("tail-sample-log-%s", uuid.String()[:8])
+
+	go func() {
+		// 10 seconds is a recommended time to wait till streaming channel is established
+		// when entries.tail reach PRD the sleep should be removed
+		time.Sleep(10 * time.Second)
+
+		logger := client.Logger(logID)
+		defer logger.Flush()
```
codyoss, line 57: I think this can just be moved below the two log lines and not defered. The func has no error points, so we can be explicit.

codyoss (changes requested):
A couple comments. I think the select approach is good in this case.

logging/taillogs/taillogs.go
minherz: :+1:

logging/taillogs/taillogs.go
minherz: :+1:

logging/taillogs/taillogs.go
minherz: Replaced all `log.Fatalf` with `return fmt.Errorf`

logging/taillogs/taillogs_test.go
minherz: :+1:

logging/taillogs/taillogs_test.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	defer client.Close()
+
+	adminClient, err := logadmin.NewClient(ctx, projectID)
+	if err != nil {
+		t.Fatalf("Failed to create logadmin client: %v", err)
+	}
+	defer adminClient.Close()
+
+	uuid, err := uuid.NewRandom()
```
codyoss, line 48: nit: please name this something else as this collides with package name uuid. Although it compiles this shadows and could lead to strange effects if the test is ever refactored down the road.

codyoss: Also, please check the error

codyoss (changes requested):
Just a couple of more small things.

logging/taillogs/taillogs_test.go
minherz: Sure

codyoss (approved):

logging/taillogs/taillogs_test.go
```diff
@@ -0,0 +1,89 @@
...
+	// cannot use t.Cleanup() due to go111 support
+	defer func() {
+		adminClient.DeleteLog(ctx, logID)
+	}()
+
+	// following implements custom 2 min timeout for running tailLogs() sample
+	success := make(chan int, 1)
+	go func() {
+		// ingest a couple of logs to finish the test
+		tailLogs(projectID)
```
codyoss, line 79: Oops, missed this after the refactor. Need to check this error as well.

codyoss (changes requested):

logging/taillogs/taillogs_test.go
minherz: Damn it. You are right (again)

codyoss (approved):

PR #2394, opened 2022-01-27
refactor(pubsub): rename region tags

hongalex:
Region tag used was incorrect, update to the correct one.


anguillanneuf (approved):

anguillanneuf:
Hm, why is the test so flaky?

codyoss:
It looks like the gRPC dep accidentally got bumped, will restore in a different CL to unblock

PR #2401, opened 2022-01-28
docs(iam-samples): extract PrivateKeyData from ServiceAccountkey

Sita04:
For more context: b/216208252

Fix: Extracting the PrivateKeyData which contains the json key credential in base64 format.


codyoss:
Please run the command listed in the failed test. It looks like there is some extra whitespace in the changes.

codyoss (approved):

PR #2402, opened 2022-01-29
fix(opencensus): trace exemplar example to show how to use attachments

jsuereth:
The documentation for this is around trace exemplars, which requires a SpanContext attachment.

This is a common source of confusion when using Exemplars (the need to attach one of the documented types as an `Any`).

### DIFF
--- a/opencensus/trace_exemplar.go
+++ b/opencensus/trace_exemplar.go
@@ -19,15 +19,33 @@ package opencensus
 
 // [START monitoring_opencensus_configure_trace_exemplar]
 import (
+	"fmt"
 	"time"
 
 	googlepb "github.com/golang/protobuf/ptypes/timestamp"
 	distributionpb "google.golang.org/genproto/googleapis/api/distribution"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
+	"google.golang.org/protobuf/types/known/anypb"
 )
 
-func createDataPointWithExemplar() *monitoringpb.Point {
+// Generates metric TimeSeries points containing Exemplars with attached tracing span.
+func createDataPointWithExemplar(projectID string) (*monitoringpb.Point, error) {
+	// projectID := "my-cloud-project-id"
 	end := time.Now().Unix()
+	traceId := "0000000000000001"
+	spanId := "00000001"
+	spanCtx, err := anypb.New(&monitoringpb.SpanContext{
+		SpanName: fmt.Sprintf("projects/%s/traces/%s/spans/%s", projectID, traceId, spanId),
+	})
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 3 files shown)


dashpole (approved):

opencensus/trace_exemplar.go
```diff
@@ -19,15 +19,25 @@ package opencensus
...
 	"time"
 
 	googlepb "github.com/golang/protobuf/ptypes/timestamp"
 	distributionpb "google.golang.org/genproto/googleapis/api/distribution"
 	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
+	"google.golang.org/protobuf/types/known/anypb"
 )
 
-func createDataPointWithExemplar() *monitoringpb.Point {
+func createDataPointWithExemplar(projectID string) *monitoringpb.Point {
```
codyoss, line 31: Please add a comment to this function: https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#comment-functions-and-packages

codyoss: Please supply example value for projectID: https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#function-arguments-for-snippets

opencensus/trace_exemplar.go
```diff
@@ -19,15 +19,25 @@ package opencensus
...
 	end := time.Now().Unix()
+	traceId := "0000000000000001"
+	spanId := "00000001"
+	spanCtx, _ := anypb.New(&monitoringpb.SpanContext{
```
codyoss, line 35: Please check the errors, can update function to return an error.

opencensus/trace_exemplar_test.go
```diff
@@ -27,7 +27,7 @@ import (
 
 func writeTimeSeriesData(projectID string) error {
 	ctx := context.Background()
-	dataPoint := createDataPointWithExemplar()
+	dataPoint := createDataPointWithExemplar(projectID)
```
codyoss, line 30: The client used below should defer a client.Close()

codyoss (changes requested):
A couple small things.

opencensus/trace_exemplar.go
jsuereth: Fixed.

opencensus/trace_exemplar.go
jsuereth: Fixed

opencensus/trace_exemplar_test.go
jsuereth: Fixed

codyoss (approved):
LGTM, thanks!

PR #2405, opened 2022-01-31
Update bigquery_create_table.go

leoncozzi:
Creating the function to be exported by capitalizing it.


codyoss:
This function is meant to function as a snippet that could be copy and pasted into code. This code is not meant to be directly imported. Closing as by design.

PR #2411, opened 2022-02-03
fix(compute): update compute and operation usage

noahdietz:
Removes all direct uses of the operation clients for polling, and replaces them with use of the Operation wrapper `Wait` method instead. Also updates `compute` to v.1.2.0 that contains these changes.

cc: @FrodoTheTrue @rsamborski


codyoss (approved):
LGTM, this makes the code so much easier to read!

FrodoTheTrue (approved):
LGTM 

One questions: does it make sense to leave the wait_for_operation example in current form: https://github.com/GoogleCloudPlatform/golang-samples/blob/main/compute/wait_for_operation.go ?
Or it is better not to use while loop at all?

rsamborski:
@FrodoTheTrue I believe that with full LRO we shouldn't need to while loop anymore. Nevertheless I would still keep an example showing how to check the LRO status in our docs with separate region tag. It can be part of another sample, PTAL at PHP which has full LRO since GA: https://github.com/GoogleCloudPlatform/php-docs-samples/blob/931263a19cbf9933c08adba9a217933374a77aac/compute/cloud-client/instances/src/create_instance.php#L91

noahdietz:
> One questions: does it make sense to leave the wait_for_operation example in current form:

I had the same question, but I wasn't sure. If one only has the operation name (and project/zone) then it would be a useful snippet to be able to reference. Generally folks should use the Operation.Wait helper, but there could be an instance where that sample would be useful. WDYT?

noahdietz:
@rsamborski beat me to it :) let's keep it!

codyoss:
Helping push through as tests seem to be passing.

PR #2412, opened 2022-02-04
chore: remove replace statements

codyoss:
This is a follow-up to #2376 which should make it easier to get
things merged into this repo. Without the replace statments an
update in the root module will no longer have cascading go.sum
changes for every other module. This also does mean though that if
a helper method needs to be updated in the future, that resides in
root/internal, that update will have to be made over two PRs so it
can be imported into the sub-modules. This change will also make it
so less tests are run with every PR should should speed up the feedback
loop for contributors to this repo.


codyoss:
cc @hongalex after this change lands I think some of the issues you brought up in the last meeting should be resolved.

tritone (approved):

hongalex (approved):

codyoss:
Failed do to a test flake, pushing this through.

PR #2421, opened 2022-02-08
feature(functions): add HTTP sample for functionsv2

anniefu:


functions/functionsv2/helloworld/hello_world.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/GoogleCloudPlatform/functions-framework-go/functions"
+)
+
+func init() {
+	functions.HTTP("HelloGet", helloGet)
+}
+
+// helloGet is an HTTP Cloud Function.
+func helloGet(w http.ResponseWriter, r *http.Request) {
```
ace-n, line 32: Should the first `h` be capitalized?

i.e. `HelloGet`?

anniefu: With declarative function signatures, you no longer need to export your function for them to work (the name of the function is based off the first string parameter in the `functions.HTTP()` call) Given that, it's probably best to keep it package local to avoid exposing it to other packages that might try and take a dependency on the function that the author doesn't want.

ace-n (approved):
LGTM, assuming this works on **both** Gen 1 and Gen 2 GCF.

codyoss:
@ace-n @anniefu I think these new snippets should have a new region tag. As is this region tag already exists in "v1" samples

PR #2317, opened 2021-12-06
refactor(functions): update HTTP sample to use declarative function

anniefu:
signatures

As described by GoogleCloudPlatform/functions-framework-go#99

PR #2422, opened 2022-02-08
fix(functions): revert to non-declarative HTTP signatures

anniefu:
Declarative and non-declarative signatures shouldn't be mixed within a
package, so leave the original samples as non-declarative.

Effectively reverts https://github.com/GoogleCloudPlatform/golang-samples/pull/2317


ace-n (approved):

ace-n:
(To clarify, we're moving the "declarative" HTTP sample to `functions/functionsv2` in #2421.)

Issue #2423, opened 2022-02-08
remove old asset main samples


codyoss:
Once docs are migrated, clean these up.

- analyze-iam-policy-longrunning-gcs
- export-assets-bigquery

PR #2424, opened 2022-02-08
refactor(asset): refactor a couple of samples

codyoss:
These samples are called quickstarts but looking at other samples
in different langs they are presented as functions to call. The main
wraping around these tests make them hard to debug when their are
failures. Made them more of a traditional sample. Will delete old
code once docs point to new sample.

Fixes: #2418
Fixes: #2414
Updates: #2423


tbpg (commented):
There is still `quickstart` in the path. Also, these are still using `package main`. Is that intentional?

PR #2438, opened 2022-02-15
feat(storage): add preconditions to stream file upload sample

BrennaEpp:


storage/objects/stream_file_upload.go
```diff
@@ -43,8 +43,20 @@ func streamFileUpload(w io.Writer, bucket, object string) error {
...
+
+	// Add preconditions so that the upload may safely retry on transient errors.
+	// In this case, we assume the object does not yet exist.
+	// To add preconditions for an existing object use the following instead:
+	// attrs, err := o.Attrs(ctx)
+	// if err != nil {
+	// 	return fmt.Errorf("object.Attrs: %v", err)
+	// }
+	// o = o.If(storage.Conditions{GenerationMatch: attrs.Generation})
+	o = o.If(storage.Conditions{DoesNotExist: true})
```
frankyn, line 56: Suggestion:
```go
	// Add preconditions so that the upload may safely retry on transient errors.
	// In this case, we assume the object does not yet exist.
	o = o.If(storage.Conditions{DoesNotExist: true})

	// To add preconditions for an existing object use the following instead:
	// attrs, err := o.Attrs(ctx)
	// if err != nil {
	// 	return fmt.Errorf("object.Attrs: %v", err)
	// }
	// o = o.If(storage.Conditions{GenerationMatch: attrs.Generation})

```

storage/objects/stream_file_upload.go
```diff
@@ -43,8 +43,20 @@ func streamFileUpload(w io.Writer, bucket, object string) error {
 	ctx, cancel := context.WithTimeout(ctx, time.Second*50)
 	defer cancel()
 
+	o := client.Bucket(bucket).Object(object)
+
+	// Add preconditions so that the upload may safely retry on transient errors.
+	// In this case, we assume the object does not yet exist.
```
cbonnie, line 49: I would change the line "In this case, we assume the object does not yet exist" to "The object does not yet exist in this code sample."

cbonnie, line 49: What's meant by "in this case"?

BrennaEpp: For this code sample. I can change this line to "In this code sample, we assume..."

storage/objects/stream_file_upload.go
BrennaEpp: Done!

storage/objects/stream_file_upload.go
```diff
@@ -43,8 +43,21 @@ func streamFileUpload(w io.Writer, bucket, object string) error {
 	ctx, cancel := context.WithTimeout(ctx, time.Second*50)
 	defer cancel()
 
+	o := client.Bucket(bucket).Object(object)
+
+	// Add preconditions so that the upload may safely retry on transient errors.
+	// In this code sample, we assume the object does not yet exist.
```
cbonnie, line 49: Suggestion:

Replace "In this code sample..." with "The following code sample sets a `Match` precondition to 0. The request to upload an object is aborted if the live object already exists in your storage bucket."

storage/objects/stream_file_upload.go
```diff
@@ -43,8 +43,21 @@ func streamFileUpload(w io.Writer, bucket, object string) error {
...
+	o = o.If(storage.Conditions{DoesNotExist: true})
+
+	// To add preconditions for an existing object use the following instead:
```
cbonnie, line 52: Replace "To add preconditions..." with "Sets a generation-match precondition for an existing object. The request to upload is successful if the object's generation number matches your precondition criteria."

storage/objects/stream_file_upload.go
```diff
@@ -43,8 +43,21 @@ func streamFileUpload(w io.Writer, bucket, object string) error {
 	ctx, cancel := context.WithTimeout(ctx, time.Second*50)
 	defer cancel()
 
+	o := client.Bucket(bucket).Object(object)
+
+	// Add preconditions so that the upload may safely retry on transient errors.
```
cbonnie, line 48: After some thought, I think we should add a link to the Go ref docs here so the user can understand what "generation" means. So:

"Add preconditions so that the upload may safely retry on transient errors. See pkg.go.dev/cloud.google.com/go/storage#hdr-Conditions for information on preconditions."

wdyt?

storage/objects/stream_file_upload.go
tritone: I think this doesn't make it clear that you would have to un-comment the line below to make this work, and that you have to choose one of the two options.

storage/objects/stream_file_upload.go
tritone, line 48: IMO the main reason to add a precondition is to avoid race conditions and data corruption as described here https://cloud.google.com/storage/docs/request-preconditions#examples_of_race_conditions_and_data_corruption . Safe retries are a side effect. Maybe a subtle distinction but I think we should encourage using a precondition on its own terms rather than emphasizing the connection with retries here.

tritone: We're going to need to include this comment in every sample with a precondition, and there are a lot of them. Would it make more sense to include these links in one place (probably a new section of the generations & preconditions docs) rather than putting them in a comment in each sample?

storage/objects/stream_file_upload.go
```diff
@@ -43,8 +43,21 @@ func streamFileUpload(w io.Writer, bucket, object string) error {
...
+	// attrs, err := o.Attrs(ctx)
+	// if err != nil {
+	// 	return fmt.Errorf("object.Attrs: %v", err)
+	// }
+	// o = o.If(storage.Conditions{GenerationMatch: attrs.Generation})
+
 	// Upload an object with storage.Writer.
-	wc := client.Bucket(bucket).Object(object).NewWriter(ctx)
+	wc := o.NewWriter(ctx)
 	wc.ChunkSize = 0 // note retries are not supported for chunk size 0.
```
tritone, line 61: This will actually nullify the effect of adding the precondition for this sample-- there is no way that a Writer with ChunkSize=0 can retry, even if preconditions are present. The reason is that the data will not be present in a buffer and so a new request body for the retried request cannot be constructed.

BrennaEpp, line 61: I just saw this part of the sample - this means we probably shouldn't include the preconditions for this particular sample, but I'd like it reviewed anyway to get a template for other samples that require preconditions.

BrennaEpp: yeah I just saw this - regardless I'd like this reviewed for other samples

storage/objects/stream_file_upload.go
BrennaEpp: I changed the wording. I'm not entirely happy with it so PTAL. I agree with Chris that we should make it clear that you have to choose one of the two options.

storage/objects/stream_file_upload.go
BrennaEpp: I'm concerned about this wording since it may be confusing to users why we use this condition instead of GenerationMatch and/or may make them try to use `GenerationMatch: 0` instead (which does nothing, as we do say in the go docs). On the other hand, this is accurate and may help users understand that the `DoesNotExist` precondition is equivalent to setting a generation match but for non-existing objects.

storage/objects/stream_file_upload.go
```diff
@@ -43,8 +43,24 @@ func streamFileUpload(w io.Writer, bucket, object string) error {
...
+	o := client.Bucket(bucket).Object(object)
+
+	// Set a generation-match precondition. The request to upload is aborted
+	// if the object's generation number does not match your precondition
+	// criteria. This avoids race conditions and data corruption.
+	// For an object that does not yet exist, set the DoesNotExist precondition.
+	o = o.If(storage.Conditions{DoesNotExist: true})
+
+	// If the live object already exists in your bucket, uncomment the following
+	// lines instead of using the above line.
```
tritone, line 55: I would say something like:

> 	// If the live object already exists in your bucket, instead set a generation match condition using the live object's generation number.

tritone (commented):
One comment, otherwise, I like this wording better in general! Maybe we should add the proposed wording on the doc.

storage/objects/stream_file_upload.go
cbonnie: @BrennaEpp Do you have a proposal for what wording you think would be best?

storage/objects/stream_file_upload.go
cbonnie: What do you propose as a change to the existing wording?

storage/objects/stream_file_upload.go
BrennaEpp: @cbonnie I added above the following, which applies to both setting `DoesNotExist` and `GenerationMatch: ...` code: "Set a generation-match precondition. The request to upload is aborted if the object's generation number does not match your precondition criteria. This avoids race conditions and data corruption." 
Then for this line:
"If the live object already exists in your bucket, instead set a generation match condition using the live object's generation number as follows."

storage/objects/stream_file_upload.go
```diff
@@ -43,8 +43,24 @@ func streamFileUpload(w io.Writer, bucket, object string) error {
...
+	o := client.Bucket(bucket).Object(object)
+
+	// Set a generation-match precondition. The request to upload is aborted
+	// if the object's generation number does not match your precondition
+	// criteria. This avoids race conditions and data corruption.
+	// For an object that does not yet exist, set the DoesNotExist precondition.
+	o = o.If(storage.Conditions{DoesNotExist: true})
+
+	// If the live object already exists in your bucket, instead set a generation
+	// match condition using the live object's generation number as follows.
```
cbonnie, line 55: "generation-match precondition" instead of "generation match condition"

cbonnie, line 55: "as follows" is not needed (makes the comment here consistent with the one above)

cbonnie (approved):
Left two nits, otherwise LGTM!

BrennaEpp:
Closing this, will open other PRs changing the samples per the final wording from this discussion

PR #2443, opened 2022-02-17
feat(storage): add preconditions to bucket samples

BrennaEpp:
Adds preconditions to all bucket samples that require them as conditionally idempotent ops.


tritone (commented):
Looks good, same question as for the object samples-- should we explicitly say that adding a precondition is optional? Would be good to have consensus on this.

cbonnie (approved):

enocom (approved):
The Go use looks good to me.

Is there a reason why this snippet is repeated across multiple files, as opposed to being in a single sample?

BrennaEpp:
Closing this as we determined to not include preconditions in the samples for bucket metadata operations - they clutter the sample and data races etc. are not a big concern with these samples (because people tend to not make many conflicting metadata changes).

PR #2445, opened 2022-02-17
feat(storage): add etag to HMAC key samples

BrennaEpp:
Adds etag conditions to the ops that require them to be idempotent.


tritone (approved):

BrennaEpp:
Closing this as we determined to not include preconditions in the samples for hmac key ops.

PR #2453, opened 2022-03-01
chore(storage): delete unused samples

tritone:
All of the samples deleted here are in the old format and are
no longer used in any docs (english or translated).

Remaining samples in buckets/main.go and objects/main.go are
used in English and/or translated pages. Most have already been
duplicated and updated elsewhere in this repo. We just need
to fix internal docs links before removing.


codyoss (approved):
I love negative commits!

PR #2455, opened 2022-03-02
samples(compute): add preventing accidental vm deletion samples

FrodoTheTrue:


enocom:
Adding @muncus who will make a pass on this and then I'll provide a review as well.

compute/instances/preventing-accidental-vm-deletion/preventing_accidental_vm_deletion_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	expectedResult = fmt.Sprintf("Instance %s has DeleteProtection value: %v", instanceName, false)
+	if got := buf.String(); !strings.Contains(got, expectedResult) {
+		t.Errorf("getDeleteProtection got %q, want %q", got, expectedResult)
+	}
+
+	buf.Reset()
+
+	deleteReq := &computepb.DeleteInstanceRequest{
```
muncus, line 91: This should probably move to a Cleanup() function, so that we get more reliable deletion.
Especially since you've used Fatalf() above, which halts test execution and will prevent vm deletion in the event of a test failure.

compute/instances/preventing-accidental-vm-deletion/set_delete_protection.go
```diff
@@ -0,0 +1,63 @@
...
+		Resource:           instanceName,
+		DeletionProtection: proto.Bool(deleteProtection),
+	}
+
+	op, err := instancesClient.SetDeletionProtection(ctx, req)
+	if err != nil {
+		return fmt.Errorf("unable to set deletion protection: %v", err)
+	}
+
+	if err = op.Wait(ctx); err != nil {
```
muncus, line 54: as in my earlier comment, we should use a Context object with a deadline here.

muncus (changes requested):
Thanks for submitting this PR! 
Note that you may also receive comments from @enocom on this PR.

compute/instances/preventing-accidental-vm-deletion/create_instance.go
```diff
@@ -0,0 +1,85 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+// createInstance sends an instance creation request to the Compute Engine API and wait for it to complete.
```
enocom, line 28: s/wait/waits/

compute/instances/preventing-accidental-vm-deletion/create_instance.go
```diff
@@ -0,0 +1,85 @@
...
+func createInstance(w io.Writer, projectID, zone, instanceName string, deleteProtection bool) error {
+	// projectID := "your_project_id"
```
enocom, line 30: These shouldn't be commented out and include a TODO to the developer to update them.

``` golang
// TODO(developer): Replace these variables before running the sample.
projectID := "your_project_id"
```

enocom: Same with the occurrences below.

compute/instances/preventing-accidental-vm-deletion/create_instance.go
```diff
@@ -0,0 +1,85 @@
...
+	// zone := "europe-central2-b"
+	// instanceName := "your_instance_name"
+	// deleteProtection := false
+
+	ctx := context.Background()
+	instancesClient, err := compute.NewInstancesRESTClient(ctx)
```
enocom, line 36: `c`, `cl`, or even `client` is more idiomatic as a variable name.

compute/instances/preventing-accidental-vm-deletion/create_instance.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"io"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+// createInstance sends an instance creation request to the Compute Engine API and wait for it to complete.
+func createInstance(w io.Writer, projectID, zone, instanceName string, deleteProtection bool) error {
```
enocom, line 29: Why use an `io.Writer` here? Would just a simple `fmt.Println` or string return value work to report success?

compute/instances/preventing-accidental-vm-deletion/preventing_accidental_vm_deletion_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+func TestPreventingAccidentalVMDeletionSnippets(t *testing.T) {
+	ctx := context.Background()
+	instancesClient, err := compute.NewInstancesRESTClient(ctx)
+	if err != nil {
+		t.Fatalf("NewInstancesRESTClient: %v", err)
+	}
+	defer instancesClient.Close()
+
+	var seededRand *rand.Rand = rand.New(
```
enocom, line 39: `seededRand` could be shorted to just `r`, but could probably be removed entirely in favor of a timestamp as suggested in [the comment below](https://github.com/GoogleCloudPlatform/golang-samples/pull/2455/files#r818171889).

compute/instances/preventing-accidental-vm-deletion/preventing_accidental_vm_deletion_test.go
```diff
@@ -0,0 +1,105 @@
...
+	zone := "europe-central2-b"
+	instanceName := "test-instance-" + fmt.Sprint(seededRand.Int())
+
+	buf := &bytes.Buffer{}
+
+	if err := createInstance(buf, tc.ProjectID, zone, instanceName, true); err != nil {
+		t.Fatalf("createInstance got err: %v", err)
+	}
+
+	expectedResult := "Instance create"
```
enocom, line 51: The string matching here is probably unnecessary. A nil error return is adequate to ensure the operation worked. Otherwise, consider returning a value that could be checked instead.

enocom: Also, prefer `got` instead of `expectedResult`.

compute/instances/preventing-accidental-vm-deletion/create_instance.go
enocom: I do see that we're using `instancesClient` in the other samples, though, so consistency is probably more important for now.

enocom (changes requested):

compute/instances/preventing-accidental-vm-deletion/preventing_accidental_vm_deletion_test.go
codyoss: If we wanted a different name here I think `want` would be the correct shorter name.

compute/instances/preventing-accidental-vm-deletion/create_instance.go
enocom: Oh I see. There's a mismatch with the style guide here: https://googlecloudplatform.github.io/samples-style-guide/#method-structure. We'll have to sort this out separately.

compute/instances/preventing-accidental-vm-deletion/create_instance.go
```diff
@@ -0,0 +1,85 @@
...
+			},
+		},
+	}
+
+	op, err := instancesClient.Insert(ctx, req)
+	if err != nil {
+		return fmt.Errorf("unable to create instance: %v", err)
+	}
+
+	if err = op.Wait(ctx); err != nil {
```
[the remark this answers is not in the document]
codyoss: If we want to use timeout context's I think we should be consistent and use them everywhere in the sample base. I don't think we are using them in samples I have reviewed. Are other langs providing timeout when calling LROs?

compute/instances/preventing-accidental-vm-deletion/preventing_accidental_vm_deletion_test.go
enocom: :man_facepalming: yes, `want`. Woops.

compute/instances/preventing-accidental-vm-deletion/create_instance.go
enocom: Oh I see. There's a mismatch with the style guide here: https://googlecloudplatform.github.io/samples-style-guide/#method-structure. We'll have to sort this out separately.

compute/instances/preventing-accidental-vm-deletion/create_instance.go
```diff
@@ -0,0 +1,85 @@
...
+		return fmt.Errorf("unable to wait for the operation: %v", err)
+	}
+
+	fmt.Fprintf(w, "Instance created\n")
```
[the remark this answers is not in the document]
FrodoTheTrue: The point is good, but Let's left it as is to be consistent

compute/instances/preventing-accidental-vm-deletion/preventing_accidental_vm_deletion_test.go
FrodoTheTrue: Switched to errorf, so now we will execute deleteInstance in any case (as in another samples)

compute/instances/preventing-accidental-vm-deletion/preventing_accidental_vm_deletion_test.go
FrodoTheTrue: switched to `r`

FrodoTheTrue:
@enocom @muncus added fixes for the comments

enocom (approved):

compute/instances/preventing-accidental-vm-deletion/create_instance.go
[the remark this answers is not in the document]
muncus: I'm fine with omitting them here for consistency, and I'll find a different way to advocate for their usage more generally :)

muncus (approved):

PR #2472, opened 2022-03-09
fix(serverless): clean up ID token example

ace-n:
~**DO NOT MERGE** until tests pass.~

The system tests are failing on `main` on my local machine, so we should confirm they're working before merging this.

### DIFF
--- a/functions/security/idtoken.go
+++ b/functions/security/idtoken.go
@@ -56,5 +56,6 @@ func makeGetRequest(w io.Writer, targetURL string, audience string) error {
 
 	return nil
 }
+
 // [END cloudrun_service_to_service_auth]
 // [END functions_bearer_token]
--- a/functions/security/idtoken.go
+++ b/functions/security/idtoken.go
@@ -28,14 +28,13 @@ import (
 
 // `makeGetRequest` makes a request to the provided `targetURL`
 // with an authenticated client using audience `audience`.
-//
-// [END functions_bearer_token]
-// Example `audience` value (Cloud Run): https://my-cloud-run-service.run.app/
-// [END cloudrun_service_to_service_auth]
-// [START functions_bearer_token]
-// Example `audience` value (Cloud Functions): https://project-region-projectid.cloudfunctions.net/myFunction
-// [START cloudrun_service_to_service_auth]
 func makeGetRequest(w io.Writer, targetURL string, audience string) error {
+	// [END functions_bearer_token]
+	// Example `audience` value (Cloud Run): https://my-cloud-run-service.run.app/
+	// [END cloudrun_service_to_service_auth]
+	// [START functions_bearer_token]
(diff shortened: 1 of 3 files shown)


functions/security/idtoken.go
```diff
@@ -27,14 +26,21 @@ import (
 	"google.golang.org/api/idtoken"
 )
 
-// makeGetRequest makes a request to the provided targetURL with an authenticated client.
-func makeGetRequest(w io.Writer, targetURL string) error {
-	// functionURL := "https://TARGET_URL"
+// `makeGetRequest` makes a request to the provided `targetURL`
```
codyoss, line 29: The func docs should be the thing directly on-top-of the function. Region tags should either be before the docs or after the method.

It looks like this method does not follow our conventions for Go samples. Examples should be in the function body like: https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#function-arguments-for-snippets

codyoss (changes requested):
Just a couple of small things

functions/security/idtoken.go
```diff
@@ -27,14 +26,21 @@ import (
...
+// with an authenticated client using audience `audience`.
+//
+// [END functions_bearer_token]
+// Example `audience` value (Cloud Run): https://my-cloud-run-service.run.app/
+// [END cloudrun_service_to_service_auth]
+// [START functions_bearer_token]
+// Example `audience` value (Cloud Functions): https://project-region-projectid.cloudfunctions.net/myFunction
+// [START cloudrun_service_to_service_auth]
+func makeGetRequest(w io.Writer, targetURL string, audience string) error {
```
muncus, line 38: This context may be in the accompanying docs, but in all uses I see here, targetURL and audience are identical. I'd suggest including an explanation of when these args might differ, for other readers without context.

functions/security/idtoken_cloudfunction.go
```diff
@@ -23,13 +23,14 @@ import (
 
 // MakeGetRequestCloudFunction wraps makeGetRequest in an HTTP Cloud Function for system test deployment.
 func MakeGetRequestCloudFunction(w http.ResponseWriter, r *http.Request) {
-	if os.Getenv("TARGET_URL") == "" {
+	url := os.Getenv("TARGET_URL")
```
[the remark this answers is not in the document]
ace-n: Fixed.

functions/security/idtoken.go
ace-n: Fixed.

functions/security/idtoken.go
ace-n: Fixed.

muncus (approved):
I discussed the region_tag shenanigans with Ace offline, and while it is unusual, we do not have a better story for re-use of sample code between samples.

codyoss (approved):

PR #2476, opened 2022-03-10
feat(run/grpc-ping): reuse TokenSource as suggested in #2139

muncus:
Adam, as per our discussion earlier today, here's a solution that reuses the TokenSource between multiple calls. Its only really good for a single `audience`, but i think it helps improve the sample, in line with the request in #2139

### DIFF
--- a/run/grpc-ping/request_auth.go
+++ b/run/grpc-ping/request_auth.go
@@ -22,12 +22,17 @@ import (
 	"fmt"
 	"time"
 
+	"golang.org/x/oauth2"
 	"google.golang.org/api/idtoken"
 	"google.golang.org/grpc"
 	grpcMetadata "google.golang.org/grpc/metadata"
 
 	pb "github.com/GoogleCloudPlatform/golang-samples/run/grpc-ping/pkg/api/v1"
 )
+
+// Store the TokenSource at the package level so it persists across multiple calls.
+// This allows for token reuse (provided the audience is the same).
+var tokenSource oauth2.TokenSource
 
 // pingRequestWithAuth mints a new Identity Token for each request.
 // This token has a 1 hour expiry and should be reused.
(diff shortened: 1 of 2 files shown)


run/grpc-ping/request_auth.go
```diff
@@ -22,27 +22,36 @@ import (
...
 	"google.golang.org/api/idtoken"
 	"google.golang.org/grpc"
 	grpcMetadata "google.golang.org/grpc/metadata"
 
 	pb "github.com/GoogleCloudPlatform/golang-samples/run/grpc-ping/pkg/api/v1"
 )
 
+// Store the TokenSource at the package level so it persists across multiple calls.
+// This allows for token reuse (provided the audience is the same).
+var tokenSource oauth2.TokenSource
```
codyoss, line 35: What would you think about:

- Wrapping this in a struct
- Making the function a method
- Initializing the struct if nil

It is often an anti-pattern to have global state, so it would be nice to avoid that in a sample if possible.

grayside: Is the idea that the call would look like:

```
pingRequestWithAuth{conn *grpc.ClientConn, p *pb.Request, audience string}.Execute()
```

I'm a bit conflicted between "more right" and "more direct".

I think this idea works but we'll need to be thoughtful with the comments, we're adding more complexity and breaking up some of the side-by-side comparability in the docs between no-auth and auth.

Here's where this fits: https://cloud.google.com/run/docs/triggering/grpc

codyoss:
```go
type app struct {
  ts oauth2.TokenSource
}

func (a *app) (conn *grpc.ClientConn, p *pb.Request) (*pb.Response, error) { ... }
```

Seeing that this is really a full app with a main I think solidifies to me that we should use this pattern. The struct can be initialized in  main to set the tokensource

codyoss: You could still contain this all in one block. Like I mentioned above you could have a nil check in the method:

```go
func (a *app) makeRequest(...) ... {
  if a == nil {
    a.init()
  }
}

func (a *app) init() {
  a := // ...
}
```

This adds a bit of complexity but I would argue the code is more correct/robust. Thoughts?

meredithslota:
Closing out as "stale" — feel free to reopen if there is intention to finish this. Thanks!

PR #2480, opened 2022-03-14
feat(internal/cloudrunci): add support for Cloud Run Jobs for running tests

muncus:
- feat(internal/cloudrunci): CI capabilities for Cloud Run Jobs.
- Avoid Platform objects in RunJobs

### DIFF
--- a/internal/cloudrunci/cloudrunjobs.go
+++ b/internal/cloudrunci/cloudrunjobs.go
@@ -1,0 +1,326 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cloudrunci facilitates end-to-end testing against the production Cloud Run.
+//
+// This is a specialized tool that could be used in addition to unit tests. It
+// calls the `gcloud beta run` command directly.
+//
+// gcloud (https://cloud.google.com/sdk) must be installed. You must be authorized via
+// the gcloud command-line tool (`gcloud auth login`).
+//
+// You may specify the location of gcloud via the GCLOUD_BIN environment variable.
+package cloudrunci
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os/exec"
+	"strings"
+	"time"
+
+	"cloud.google.com/go/logging/logadmin"
+	"google.golang.org/api/iterator"
+)
+
+// RunJob describes a Cloud Run Job
+// The typical usage flow of a RunJob is to call the following methods, which
+// call the corresponding "gcloud run jobs" commands:
+// Build(), Create(), Run().
+type RunJob struct {
+	// Name is an ID, used for logging and to generate a unique version to this run.
+	Name string
+
+	// The directory containing the job's code.
+	Dir string
+
+	// The container image name to deploy. If left blank the container will be built
+	// and pushed to gcr.io/[ProjectID]/[Name]:[Revision]
+	Image string
+
+	// The project to deploy to.
+	ProjectID string
+
+	// The platform to deploy to.
+	Platform Platform
+
+	// Additional runtime environment variable overrides for the app.
+	Env EnvVars
+
+	// Build this Image as a BuildPack, without using a Dockerfile
+	AsBuildpack bool
+
+	built   bool // True if container image has been built.
+	created bool // True if job has been created.
+	started bool // true if the Job has been started.
+}
+
+// NewJob creates a new RunJob to be run with Cloud Run Jobs.
+// It will default to the ManagedPlatform in region us-central1,
+// and build a container image as needed for deployment.
+func NewJob(name, projectID string) *RunJob {
+	return &RunJob{
+		Name:      name,
+		ProjectID: projectID,
+		Platform:  ManagedPlatform{Region: "us-central1"},
+	}
+}
+
+// validate confirms all required service properties are present.
+func (j *RunJob) validate() error {
+	if j.ProjectID == "" {
+		return errors.New("Project ID missing")
+	}
+	if j.Platform == nil {
+		return errors.New("Platform configuration missing")
+	}
+	if err := j.Platform.Validate(); err != nil {
+		return err
+	}
+	if err := j.Env.Validate(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


internal/cloudrunci/cloudrunjobs.go
```diff
@@ -0,0 +1,324 @@
...
+		return errors.New("Project ID missing")
+	}
+	if err := j.Env.Validate(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// version returns the revision that the service will be deployed to.
```
averikitsch, line 103: nit: use execution instead of revision

run/testing/jobs.e2e_test.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2019 Google LLC
```
averikitsch, line 1: 2019 --> 2022

run/testing/jobs.e2e_test.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	if err := crj.Create(); err != nil {
+		t.Fatalf("Create %q: %v", crj.Name, err)
+	}
+	if err := crj.Run(); err != nil {
+		t.Errorf("Run(%s): %s", crj.Name, err)
+	}
+	defer crj.Clean()
+
```
averikitsch, line 42: I would love to see a check for the logs that the job creates (a log should include "Task").

internal/cloudrunci/cloudrunjobs.go
```diff
@@ -0,0 +1,324 @@
...
+
+func (j *RunJob) CommonGCloudFlags() []string {
+	return []string{
+		"--region", j.Region,
+		"--project", j.ProjectID,
+	}
+
+}
+
+// validate confirms all required service properties are present.
```
averikitsch, line 91:
```suggestion
// validate confirms all required job properties are present.
```

internal/cloudrunci/cloudrunjobs.go
```diff
@@ -0,0 +1,324 @@
...
+	return nil
+}
+
+// version returns the revision that the service will be deployed to.
+// This identifier is also used to locate relevant log messages.
+func (j *RunJob) version() string {
+	return j.Name + "-" + runID
+}
+
+// Create deploys the job to Cloud Run Jobs, but does not start it.
```
averikitsch, line 109:
```suggestion
// Creates the Cloud Run job, but does not start it.
```

internal/cloudrunci/cloudrunjobs.go
```diff
@@ -0,0 +1,324 @@
...
+		j.Image,
+		"--force-delete-tags",
+	}
+
+	cmd := exec.Command(gcloudBin, args...)
+	cmd.Dir = j.Dir
+	return cmd
+}
+
+func (j *RunJob) deleteRunJobCmd() *exec.Cmd {
```
averikitsch, line 272: I believe but need to check that you can't delete a job if an execution is currently running; therefore we may need to  delete the execution first

internal/cloudrunci/cloudrunjobs.go
```diff
@@ -0,0 +1,324 @@
...
+		if err != nil {
+			return fmt.Errorf("gcloud: %v: %q", j.version(), err)
+		}
+		j.built = false
+	}
+
+	return nil
+}
+
+func (j *RunJob) createCmd() *exec.Cmd {
```
averikitsch, line 198: In the future, we may need to support flags, such as tasks, parallelism and retries

averikitsch (commented):
Just a few nits, but overall the functionality that is needed is provided. Thanks for taking this on Marc!

internal/cloudrunci/cloudrunjobs.go
muncus: Currently, Run() waits for the job to complete before returning. Should we (a) Not wait or (b) Make Wait optional, and probably default?

run/testing/jobs.e2e_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestCloudRunJobs(t *testing.T) {
+	tc := testutil.EndToEndTest(t)
+
+	crj := &cloudrunci.Job{
+		Name:        "runjobs",
+		ProjectID:   tc.ProjectID,
+		Dir:         "../jobs",
```
grayside, line 30: out-of-scope: I missed before that the Cloud Run Jobs quickstart is directly in run/jobs. It should be in run/jobs/quickstart or similar, one more directory down, so we have clear separation in place to add more sample jobs.

run/testing/jobs.e2e_test.go
grayside: Maybe the `Completed Task` check was added since this comment? Note that under the hood, LogEntries is already filtering to logs associated with this Job as a resource.

grayside (approved):
Aside from the question about Job vs. Execution log filtering, this LGTM.

Since every PR test pipeline is assigned its own project, ensuring PRs to the same Cloud Run Jobs sample can't taint each other, this is not a blocking concern.

run/testing/jobs.e2e_test.go
muncus: Yes, this check was added in response, and i forgot to close out the conversation.

run/testing/jobs.e2e_test.go
```diff
@@ -0,0 +1,56 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cloudruntests
+
+import (
+	"testing"
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/cloudrunci"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
```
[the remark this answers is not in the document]
muncus: IMO, the simplest outcome is the replace directive. I think this leads to the principle of least surprise (i.e. the code in the local directory is the code that is compiled).
I'm also fine with reverting the `replace` directive after this PR if desired.

internal/cloudrunci/cloudrunjobs.go
```diff
@@ -0,0 +1,327 @@
...
+
+func (j *Job) LogEntries(filter string, find string, maxAttempts int) (bool, error) {
+	ctx := context.Background()
+	client, err := logadmin.NewClient(ctx, j.ProjectID)
+	if err != nil {
+		return false, fmt.Errorf("logadmin.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	preparedFilter := fmt.Sprintf(`resource.type="cloud_run_revision" resource.labels.service_name="%s" %s`, j.version(), filter)
```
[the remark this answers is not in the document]
muncus: I agree that we *should*, but the execution id is not available to us, given the way we create it.
`resource.labels.revision_name` field contains the execution id, but there's no way for us to know it without some significant revision to Job creation.
Given that the test object creates a new job, and cleans it up, I expect the value of execution-level filtering to be relatively minor and probably not worth the necessary changes. Would you agree?

I will add a note to the Job docstring about LogEntries() not working if Run() is called more than once.

PR #2483, opened 2022-03-16
feat(storage): add setting client endpoint sample

### LINKED ISSUE #2463: storage: add endpoint samples
We need samples added that describe how to set the endpoint for our storage clients.
Such as setting for localhost/emulator, or private/vpc, or for the soon to be added locational endpoints feature.

Sample tracker & Request:

https://docs.google.com/document/d/1aeL7jgs789s7xg9khTxGzGWkb4k-t6LUk-uqhhs1Vuw/edit#

cojenco:
Adds sample for changing request endpoint on storage client

Fixes #2463

Thanks @BrennaEpp for helping fix the local goimports issue🎉 
Cleans up and replaces #2464

### DIFF
--- a/storage/service_account/client_endpoint/client_endpoint_test.go
+++ b/storage/service_account/client_endpoint/client_endpoint_test.go
@@ -16,19 +16,21 @@ package clientendpoint
 
 import (
 	"bytes"
+	"os"
 	"strings"
 	"testing"
-
-	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
 )
 
 func TestSetClientEndpoint(t *testing.T) {
-	tc := testutil.SystemTest(t)
+	if os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") == "" {
+		t.Skip("GOOGLE_APPLICATION_CREDENTIALS not set")
+	}
+
 	customEndpoint := "http://localhost:8080/storage/v1/"
 
 	var buf bytes.Buffer
 	if err := setClientEndpoint(&buf, customEndpoint); err != nil {
-		t.Errorf("setClientEndpoint in project %s: %s", tc.ProjectID, err)
+		t.Errorf("setClientEndpoint: %s", err)
 	}
 
 	if got, want := buf.String(), "request endpoint set for the client"; !strings.Contains(got, want) {
--- a/storage/service_account/client_endpoint/client_endpoint_test.go
+++ b/storage/service_account/client_endpoint/client_endpoint_test.go
@@ -1,0 +1,37 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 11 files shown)


storage/service_account/client_endpoint/client_endpoint_test.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+func TestSetClientEndpoint(t *testing.T) {
+	tc := testutil.SystemTest(t)
```
muncus, line 26: I don't think we need this line, because the ProjectID does not get used in the `setClientEndpoint()` method, making this more of a unit test than a "system test".

muncus (commented):
Thanks for this nice, concise sample snippet!

storage/service_account/client_endpoint/client_endpoint_test.go
cojenco: @muncus I used testutil based on the [guide](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#use-testutil-for-tests). In this case, do you recommend to remove the testutil and skip tests instead?

storage/service_account/client_endpoint/set_client_endpoint.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/option"
+)
+
+// setClientEndpoint sets the request endpoint.
+func setClientEndpoint(w io.Writer, customEndpoint string) error {
+	// customEndpoint := "custom-request-endpoint"
```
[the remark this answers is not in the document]
cojenco: great point! updated the comment, thanks!

storage/service_account/client_endpoint/client_endpoint_test.go
muncus: I hesitate to suggest less testing 😅 but my understanding here is that testutil is not needed, because the test code does not actually call the Storage service, and thus no Project ID is actually needed.
That's my perspective, but I'll tag in @enocom for a second opinion.

storage/service_account/client_endpoint/client_endpoint_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"bytes"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestSetClientEndpoint(t *testing.T) {
+	if os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") == "" {
```
enocom, line 25: This seems reasonable to me. Although using the other approach would be fine as well IMHO.

enocom (approved):

storage/service_account/client_endpoint/client_endpoint_test.go
cojenco: Sounds good! Indeed no project ID is needed for this sample, though authentication is not avoided for `setClientEndpoint()`. So added a test skip to ensure the test passes under the right conditions.

PR #2493, opened 2022-03-30
feat(spanner): add PG samples for getting-started

rahul2393:
This PR contains the equivalent samples for PG dialect to be linked from https://cloud.google.com/spanner/docs/getting-started/go


enocom:
Adding @BrennaEpp for a first pass and then I'll take a look.

spanner/spanner_snippets/snippet.go
```diff
@@ -477,6 +483,234 @@ func writeWithTransactionUsingDML(ctx context.Context, w io.Writer, client *span
...
+	_, err := client.ReadWriteTransaction(ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
+		// getBudget returns the budget for a record with a given albumId and singerId.
+		getBudget := func(albumID, singerID int64) (int64, error) {
+			key := spanner.Key{albumID, singerID}
+			row, err := txn.ReadRow(ctx, "Albums", key, []string{"MarketingBudget"})
+			if err != nil {
+				return 0, err
+			}
+			var budget int64
+			if err := row.Column(0, &budget); err != nil {
```
BrennaEpp, line 663: Could this just be returned directly instead of checking the error case? Or perhaps (here and elsewhere), we should just add some information to the errors returned? See https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#return-errors

BrennaEpp (commented):
Just a couple comments. Is it possible to add some tests for the new samples as well?

spanner/spanner_snippets/snippet.go
```diff
@@ -477,6 +483,234 @@ func writeWithTransactionUsingDML(ctx context.Context, w io.Writer, client *span
 
 // [END spanner_dml_getting_started_update]
 
+// [START spanner_postgresql_create_database]
+
+func pgCreateDatabase(ctx context.Context, w io.Writer, adminClient *database.DatabaseAdminClient, db string) error {
+	matches := regexp.MustCompile("^(.*)/databases/(.*)$").FindStringSubmatch(db)
+	if matches == nil || len(matches) != 3 {
+		return fmt.Errorf("invalid database id %s", db)
```
enocom, line 491: %v is generally preferred

spanner/spanner_snippets/snippet.go
```diff
@@ -477,6 +483,234 @@ func writeWithTransactionUsingDML(ctx context.Context, w io.Writer, client *span
...
+		// includes database names in the CREATE DATABASE statement.
+		CreateStatement: `CREATE DATABASE "` + matches[2] + `"`,
+	})
+	if err != nil {
+		return err
+	}
+	if _, err := op.Wait(ctx); err != nil {
+		return err
+	}
+	// Databases that are created with PostgreSQL dialect do not support extra DDL statements in the `CreateDatabase` call.
```
enocom, line 506: Please wrap this comment to under 100 characters, or even shorter (80). This will be hard to read on cloud.google.com. See https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#line-length.

spanner/spanner_snippets/snippet.go
```diff
@@ -477,6 +483,234 @@ func writeWithTransactionUsingDML(ctx context.Context, w io.Writer, client *span
...
+		},
+	}
+	opUpdate, err := adminClient.UpdateDatabaseDdl(ctx, updateReq)
+	if err != nil {
+		return err
+	}
+	if err := opUpdate.Wait(ctx); err != nil {
+		return err
+	}
+	fmt.Fprintf(w, "Created database [%s]\n", db)
```
enocom, line 532: %v here too

enocom (changes requested):

spanner/spanner_snippets/snippet.go
```diff
@@ -477,6 +483,234 @@ func writeWithTransactionUsingDML(ctx context.Context, w io.Writer, client *span
 
 // [END spanner_dml_getting_started_update]
 
+// [START spanner_postgresql_create_database]
```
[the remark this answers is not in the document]
enocom: Looks like a lot of this file didn't follow this rule, but I agree we should split these out into individual files so we can show the imports.

rahul2393:
Thanks @enocom and @BrennaEpp, the aim of the PR is to update the [getting-started/go](https://cloud.google.com/spanner/docs/getting-started/go) page to write samples for GSQL and PG using tabs like [this](https://cloud.devsite.corp.google.com/spanner/docs/storing-numeric-data#performing_approximate_aggregations_sorting_and_calculations), for that I have added samples for PG here in separate files and added integration tests, the samples are copied to snippets.go because we the getting-started page has one liners for customers to run the operations but the actual samples in the page will point to individual files and not the samples. 
I have tried to follow the contribution guideline as per suggestion and updated the PR, please help review again
Thanks

enocom (changes requested):
The code LGTM.

We should have only one set of region tags, though. Since this is adding to an existing sample, we might as well keep the pattern consistent.

enocom (approved):

enocom:
This is blocked on https://github.com/GoogleCloudPlatform/golang-samples/pull/2502.

PR #2499, opened 2022-04-04
chore(cloudsql): add note about secret manager to all

enocom:


cloudsql/mysql/database-sql/connect_connector.go
```diff
@@ -35,7 +35,10 @@ func connectWithConnector() (*sql.DB, error) {
 		}
 		return v
 	}
-
+	// Note: Saving credentials in environment variables is convenient, but not
```
codyoss, line 38: I assume all of the samples in each lang is doing something similar here? If so should this text just be migrated up into the surrounding docs on cloud site?

codyoss (approved):
LGTM, just one question.

cloudsql/mysql/database-sql/connect_connector.go
enocom: Since we embed this code into the docs, we want to make call out the option to use secret manager right next to the calls to os.Getenv, just to make it explicit. It's possible also warn users in the docs as well.

enocom:
Blocked by https://github.com/GoogleCloudPlatform/golang-samples/pull/2502.

PR #2507, opened 2022-04-11
test(iam): Add Retry to TestServiceAccounts

quartzmo:
Attempt to fix occasional test flakes apparently caused by eventual consistency of recently created Service Account with retry of `renameServiceAccount`.

closes: #2486

### DIFF
--- a/iam/snippets/service_accounts_test.go
+++ b/iam/snippets/service_accounts_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
 	"github.com/gofrs/uuid"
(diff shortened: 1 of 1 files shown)


iam/snippets/service_accounts_test.go
```diff
@@ -41,14 +42,17 @@ func TestServiceAccounts(t *testing.T) {
...
-	}
-	wantDispName := "Updated Test"
-	if wantDispName != account.DisplayName {
-		t.Fatalf("renameServiceAccount: account.DisplayName is %q, wanted %q", account.Name, wantDispName)
-	}
+
+	testutil.Retry(t, 5, 5*time.Second, func(r *testutil.R) {
+		account, err = renameServiceAccount(buf, account.Email, "Updated Test")
+		if err != nil {
+			r.Errorf("renameServiceAccount: %v", err)
```
tritone, line 49: This should have a return after it to avoid a panic on L52 I believe.

quartzmo: Fixed, thank you! Let's see what the real problem is now...

Issue #2169, opened 2021-07-29
mailgun.go example not working


aitchpea:
https://github.com/GoogleCloudPlatform/golang-samples/blob/master/docs/appengine/mail/mailgun/mailgun.go

Does not work with the latest AppEngine as it does not include the changes https://cloud.google.com/appengine/docs/standard/go/go-differences  of request.Context()

Also, I live in Europe and once I had updated the above changes I found the example still did not work as I needed the below extra line of code (can it be included to avoid others needing to do lots of searching to understand the issue? Clearly it would be commented out for the general case)

message := mg.NewMessage(sender, subject, body, recipient)
/// The below is needed if the code is running in Europe
mg.SetAPIBase(mailgun.APIBaseEU)

Can you show me the documentation on what the difference is between code in golang-samples/blob/master/appengine vs golang-samples/blob/master/docs ? The structure does not make logical sense to me so I wondered if I am missing something.

broady:
Hi, yes, this sample is not supported on Go 1.12. Was that indicated anywhere?

aitchpea:
Apologies - I do not understand your question as it could mean "statement: we do not support this sample in Go 1.12, have we stated anywhere it would be" or "I agree with your point and can see that it would not be supported on Go 1.12"  etc ?

Just to understand your view - do you think Google should not support the update of samples to the newer versions of Go? I just looked again at the Readme and could not see any definition of what ongoing support these examples would have (apart from saying that Pull requests and raising Issues would be acceptable). Is there some other channel / request I should make?

markxp:
The [app.yaml](https://raw.githubusercontent.com/GoogleCloudPlatform/golang-samples/master/docs/appengine/mail/app.yaml) shows it must be 1st generation standard environment, because it uses `inbound-services: mail` (not availalbe for 2nd generation, only `warmup`, even 1st generation in go111).

I was confused about these flacky stuff for a long time. Unfortunely it is usually not well-commented in sample codes.

aitchpea:
Thanks - your points make sense and show that this is indeed old code, my point is partly a question of what is the aim for Google representatives leaving this code and not keeping it updated. It in the docs directory and I would not have looked in here myself when I was looking for mailgun examples - but a quick Google search found them, so if I can find it then many others can too. If it was clearly documented that this is old code / unsupported etc it would save time for anyone else stumbling in here. 

[...]

PR #2509, opened 2022-04-12
fix(docs): update mail example to use appengine/v2

codyoss:
Fixes: #2169


enocom (approved):
CI needs a minor goimports fix.

PR #2510, opened 2022-04-12
fix(healthcare): increase FHIR store create test retry

noerog:
Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/2417.


codyoss (approved):
Thanks!

codyoss:
@noerog Seems like there are some permission errors here.

noerog:
It looks like a separate issue is occurring, which is that TestHL7V2Store is now failing in the same way that TestFHIRStore was. TestHL7V2Store has a retry where it's failing, so I'll need to see if we need to increase the retry or try something else.

PR #2514, opened 2022-04-14
chore: bump compute deps to 1.6.0

codyoss:
There was a breaking change in genproto related to compute. This should fix current test failures that pull in a modern genproto version.


tritone (approved):

codyoss:
Force merging to fix builds that are blocked

PR #2516, opened 2022-04-15
fix(bigquery/storage): handle retryable RESOURCE_EXHAUSTED errors in bigquery storage sample

esert-g:
* Adds special handling for retryable RESOURCE_EXHAUSTED errors returned from bigquery storage api, so they aren't treated as regular errors.
* Fixes retries being reset every time an error is observed, so at most 3 back-to-back errors are retried.

### DIFF
--- a/bigquery/bigquery_storage_quickstart/main.go
+++ b/bigquery/bigquery_storage_quickstart/main.go
@@ -38,7 +38,10 @@ import (
 	gax "github.com/googleapis/gax-go/v2"
 	goavro "github.com/linkedin/goavro/v2"
 	bqStoragepb "google.golang.org/genproto/googleapis/cloud/bigquery/storage/v1"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // rpcOpts is used to configure the underlying gRPC client to accept large
@@ -207,8 +210,8 @@ func processStream(ctx context.Context, client *bqStorage.BigQueryReadClient, st
 	// stream, implement a retry that resets once progress is made.
 	retryLimit := 3
 
+	retries := 0
 	for {
-		retries := 0
 		// Send the initiating request to start streaming row blocks.
 		rowStream, err := client.ReadRows(ctx, &bqStoragepb.ReadRowsRequest{
 			ReadStream: st,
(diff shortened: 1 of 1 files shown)


bigquery/bigquery_storage_quickstart/main.go
```diff
@@ -224,10 +227,31 @@ func processStream(ctx context.Context, client *bqStorage.BigQueryReadClient, st
 			if err == io.EOF {
 				return nil
 			}
-			if err != nil {
-				retries++
-				if retries >= retryLimit {
-					return fmt.Errorf("processStream retries exhausted: %v", err)
+			if err == nil {
```
shollyman, line 230: Can you reorder this so the `err==nil` case is last?  As it stands, the logic checks for io.EOF, no error, then generic error (where we inspect for the exhaustion/retryinfo).

bigquery/bigquery_storage_quickstart/main.go
```diff
@@ -224,10 +227,31 @@ func processStream(ctx context.Context, client *bqStorage.BigQueryReadClient, st
...
+				// Reset retries after a successful response.
+				retries = 0
+			} else {
+				// If there is an error, check whether it is a retryable error with a retry delay and sleep instead of increasing retries count.
+				var retryDelayDuration time.Duration
+				if errorStatus, ok := status.FromError(err); ok && errorStatus.Code() == codes.ResourceExhausted {
```
shollyman, line 236: Do we ever anticipate the backend adding RetryInfo to responses other than ResourceExhausted?  If so, perhaps we don't special case the `ResourceExhausted`.

shollyman (commented):
Thanks for the PR, a couple minor items then we should be able to get this merged.

bigquery/bigquery_storage_quickstart/main.go
esert-g: Done.

bigquery/bigquery_storage_quickstart/main.go
esert-g: BigQuery storage backend will only return a RetryInfo with ResourceExhausted.

shollyman (approved):

bigquery/bigquery_storage_quickstart/main.go
```diff
@@ -223,15 +226,35 @@ func processStream(ctx context.Context, client *bqStorage.BigQueryReadClient, st
 			r, err := rowStream.Recv()
 			if err == io.EOF {
 				return nil
-			}
-			if err != nil {
-				retries++
-				if retries >= retryLimit {
-					return fmt.Errorf("processStream retries exhausted: %v", err)
+			} else if err != nil {
```
enocom, line 229: The else here is superfluous with the return above.

bigquery/bigquery_storage_quickstart/main.go
```diff
@@ -223,15 +226,35 @@ func processStream(ctx context.Context, client *bqStorage.BigQueryReadClient, st
...
+				// If there is an error, check whether it is a retryable error with a retry delay and sleep instead of increasing retries count.
```
enocom, line 230: Please wrap this comment to under 100 characters.

enocom (changes requested):
A couple minor things

bigquery/bigquery_storage_quickstart/main.go
esert-g: Done

bigquery/bigquery_storage_quickstart/main.go
esert-g: Done

enocom (approved):

PR #2524, opened 2022-04-20
feat(compute): add custom-machine-type helper class

FrodoTheTrue:


FrodoTheTrue:
Helper class from the PR: https://github.com/GoogleCloudPlatform/golang-samples/pull/2482

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+type CustomMachineType struct {
+	zone, cpuSeries     string
+	memoryMb, coreCount int
+	typeLimit           TypeLimit
+}
+
+func (t CustomMachineType) check() error {
+	// Check whether the requested parameters are allowed. Find more information about limitations of custom machine
```
enocom, line 59: This is a helpful comment that should be declared above the `check` function. Also, let's make sure to wrap this under 100 or even better 80 characters.

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,126 @@
...
+		return fmt.Errorf("requested memory is too large.. Maximum memory allowed for %v is %v MB per core", t.cpuSeries, t.typeLimit.maxMemPerCore)
+	}
+	if t.memoryMb > t.typeLimit.extraMemoryLimit && t.typeLimit.allowExtraMemory {
+		return fmt.Errorf("requested memory is too large.. Maximum memory allowed for %v is %v MB", t.cpuSeries, t.typeLimit.extraMemoryLimit)
+	}
+
+	return nil
+}
+
+func (t CustomMachineType) IsExtraMemoryUsed() bool {
```
enocom, line 88: Don't export this function.

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	if t.IsExtraMemoryUsed() {
+		return fmt.Sprintf("zones/%v/machineTypes/%v-%v-%v-ext", t.zone, t.cpuSeries, t.coreCount, t.memoryMb)
+	}
+
+	return fmt.Sprintf("zones/%v/machineTypes/%v-%v-%v", t.zone, t.cpuSeries, t.coreCount, t.memoryMb)
+}
+
+func (t CustomMachineType) ShortString() string {
```
enocom, line 106: Since this function returns the machine type, should it be called `machineType` instead?

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,126 @@
...
+	// Return machine type in a format without the zone. For example, n2-custom-0-10240.
```
enocom, line 107: Let's move this to above the function to make it a proper Godoc string.

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,126 @@
...
+	CPUSeriesN1Limit       = TypeLimit{append([]int{1}, MakeRange(2, 97, 2)...), 922, 6656, true, 624 << 10}
+)
+
+type CustomMachineType struct {
+	zone, cpuSeries     string
+	memoryMb, coreCount int
+	typeLimit           TypeLimit
+}
+
+func (t CustomMachineType) check() error {
```
enocom, line 58: I wonder if there's a better name than check. Maybe validate? Also, would this be better as a function rather than a method?

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,126 @@
...
+	ss := strings.Split(t.String(), "/")
+	return ss[len(ss)-1]
+}
+
+func createCustomMachineType(zone, cpuSeries string, memoryMb, coreCount int, tl TypeLimit) (*CustomMachineType, error) {
+	if containsString([]string{E2Small, E2Micro, E2Medium}, cpuSeries) {
+		coreCount = 2
+	}
+	cmt := CustomMachineType{zone, cpuSeries, memoryMb, coreCount, tl}
+	check := cmt.check()
```
enocom, line 118: the variable here should be `err` since it's an error. Also, consider using this terser syntax:

``` golang
if err := cmt.check(); err != nil {
  // ... 
}
```

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if check != nil {
+		return &CustomMachineType{}, check
+	}
+	return &cmt, nil
```
enocom, line 123: Taking the address of `cmt` here is fine, but to my eye it's better to move the `&` to where the value is declared, i.e.,

``` golang
cmt := &CustomMachineType{zone, cpuSeries, memoryMb, coreCount, tl}
```

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,126 @@
...
+	// This format is used to create instance templates.
+	ss := strings.Split(t.String(), "/")
+	return ss[len(ss)-1]
+}
+
+func createCustomMachineType(zone, cpuSeries string, memoryMb, coreCount int, tl TypeLimit) (*CustomMachineType, error) {
+	if containsString([]string{E2Small, E2Micro, E2Medium}, cpuSeries) {
+		coreCount = 2
+	}
+	cmt := CustomMachineType{zone, cpuSeries, memoryMb, coreCount, tl}
```
enocom, line 117: Specify the fields and don't rely on positional arguments here.

compute/instances/custom-machine-type/helper_class_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+var customMachineTypeTests = []struct {
+	cpuSeries string
+	memory    int
+	cpu       int
+	cpuLimit  TypeLimit
+	out       string
+	outShort  string
+}{
+	{N1, 8192, 8, CPUSeriesN1Limit, "zones/europe-central2-b/machineTypes/custom-8-8192", "custom-8-8192"},
```
enocom, line 30: Let's use declare the fields explicitly here, e.g.,

``` golang
{
  cpuSeries: n1,
  memory: 8192,
  // ... etc
}
```

compute/instances/custom-machine-type/helper_class_test.go
```diff
@@ -0,0 +1,74 @@
...
+		})
+	}
+}
+
+func TestCustomMachineTypeErrorsSnippets(t *testing.T) {
+	zone := "europe-central2-b"
+
+	// bad memory 256
+	_, err := createCustomMachineType(zone, N1, 8194, 8, CPUSeriesN1Limit)
+	expectedResult := "requested memory must be a multiple of 256 MB"
```
enocom, line 63: Typically `want` is used for an expected result.

compute/instances/custom-machine-type/helper_class_test.go
```diff
@@ -0,0 +1,74 @@
...
+	if fmt.Sprint(err) != expectedResult {
```
enocom, line 64: `err.Error()`

compute/instances/custom-machine-type/helper_class_test.go
```diff
@@ -0,0 +1,74 @@
...
+		t.Errorf("createCustomMachineType should return error: %s %v", expectedResult, err)
```
enocom, line 65: `%v` is almost always the thing to use.

enocom (changes requested):
Thanks for splitting the big PR up.

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+)
+
+type customMachineType struct {
+	zone, cpuSeries     string
+	memoryMb, coreCount int
+	typeLimit
+}
+
+// Validates whether the requested parameters are allowed.
```
enocom, line 90: "Validates" should be "validate"

compute/instances/custom-machine-type/utils.go
```diff
@@ -0,0 +1,25 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package snippets
+
+func containsString(s []string, str string) bool {
```
enocom, line 17: Can this be moved to where it's used?

FrodoTheTrue: I am going to use the function in other samples in the same folder. What's the best way to do the same logic in different files in one package (I guess to rename the function in every file is super wired weird)

compute/instances/custom-machine-type/helper_class.go
```diff
@@ -0,0 +1,168 @@
...
+		return fmt.Errorf("requested memory is too large. Maximum memory allowed for %v is %v MB per core", cmt.cpuSeries, cmt.typeLimit.maxMemPerCore)
+	}
+	if cmt.memoryMb > cmt.typeLimit.extraMemoryLimit && cmt.typeLimit.allowExtraMemory {
+		return fmt.Errorf("requested memory is too large. Maximum memory allowed for %v is %v MB", cmt.cpuSeries, cmt.typeLimit.extraMemoryLimit)
+	}
+
+	return nil
+}
+
+// Returns the custom machine type in form of a string acceptable by Compute Engine API.
```
enocom, line 128: This should be `String returns ...`

compute/instances/custom-machine-type/utils.go
enocom: Since it's such a simple function, I think duplicating it makes sense, probably by declaring it inline where necessary. That way, sample reads see the entire code in a single region tag.

For example:

``` golang
func (t customMachineType) String() string {
        containsString := func(s []string, str string) bool {
            // ...
        }
	if containsString([]string{e2Small, e2Micro, e2Medium}, t.cpuSeries) {
		return fmt.Sprintf("zones/%v/machineTypes/%v-%v", t.zone, t.cpuSeries, t.memoryMb)
	}
        // ...
}
```

PR #2525, opened 2022-04-20
feat(compute): add windows os image sample

FrodoTheTrue:


enocom:
I'll make a pass on this after @BrennaEpp takes a look.

enocom (changes requested):
Just a few minor presentation things.

compute/instances/windows/create-os-image/create_windows_os_image.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		instance, err := instancesClient.Get(ctx, instanceReq)
+		if err != nil {
+			return fmt.Errorf("unable to get instance: %v", err)
+		}
+
+		if instance.GetStatus() != "TERMINATED" && instance.GetStatus() != "STOPPED" {
+			if !forceCreate {
+				return fmt.Errorf(
+					`instance %s should be stopped.
```
enocom, line 93: Depending on whether people will be running this actual sample or not, we might get away from this long line problem by making this message into a comment. Alternatively, consider removing the tabs in this string literal to produce something like:

``` golang
if !forceCreate {
    return fmt.Errorf(`one long error message goes
 here and will render nicely in the terminal`) // intentionally not re-indenting this line
 }
 ```

enocom (approved):
One minor error formatting problem, otherwise this looks good.

PR #2528, opened 2022-04-21
fix: Change desired translated text

engelke:
It appears that the API now translates beef to "Boeuf" instead of "Bœuf"


ace-n (approved):

engelke:
Done elsewhere with a direct commit to main.

PR #2532, opened 2022-04-21
chore(functions/ocr/app): Relax test expectation to prevent flakes

quartzmo:
Sometimes the Vision API returns "Filets de Bœuf" as expected by the test, other times it returns "Filets de Boeuf", causing the failure. This PR relaxes the expectation.

closes: #2523


codyoss (approved):
LGTM

PR #2534, opened 2022-04-22
Use storage.SignedURL in generate_v4_get_object_signed_url example

aleksandrzhiliaev:
This PR updates example for object signed urls

### DIFF
--- a/storage/objects/generate_v4_get_object_signed_url.go
+++ b/storage/objects/generate_v4_get_object_signed_url.go
@@ -51,7 +51,7 @@ func generateV4GetObjectSignedURL(w io.Writer, bucket, object string) (string, e
 		Expires: time.Now().Add(15 * time.Minute),
 	}
 
-	u, err := client.Bucket(bucket).SignedURL(object, opts)
+	u, err := storage.SignedURL(bucket, object, opts)
 	if err != nil {
 		return "", fmt.Errorf("Bucket(%q).SignedURL: %v", bucket, err)
 	}


storage/objects/generate_v4_get_object_signed_url.go
```diff
@@ -51,7 +51,7 @@ func generateV4GetObjectSignedURL(w io.Writer, bucket, object string) (string, e
 		Expires: time.Now().Add(15 * time.Minute),
 	}
 
-	u, err := client.Bucket(bucket).SignedURL(object, opts)
+	u, err := storage.SignedURL(bucket, object, opts)
```
tritone, line 54: We specifically updated this sample to use `BucketHandle.SignedURL` because it is much more user-friendly for auth, so we think it should be the preferred approach for new users. See #2318 

There is a short example for storage.SignedURL here: https://pkg.go.dev/cloud.google.com/go/storage#hdr-Signed_URLs

tritone (commented):
Can you speak a little more as to the reason for this PR? Feel free to file an issue in the repo.

storage/objects/generate_v4_get_object_signed_url.go
aleksandrzhiliaev: Hey, sorry just found that my local version of the storage library was an old one. Yea makes sense for me too. Closing PR.

tritone: No worries! Hope the new version works better for you.

PR #2542, opened 2022-04-26
feat: add code samples and tests for Live Stream

irataxy:
b: 219762506


enocom (approved):
Normally, we prefer PRs of ~500 changes. But this looks solid. The Live Stream API needs to be enabled on our testing project, which I'll handle separately.

PR #2545, opened 2022-04-27
feat(compute): add compute_machine_type samples 2

FrodoTheTrue:


compute/instances/custom-machine-type/create_shared_with_helper.go
```diff
@@ -0,0 +1,223 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+func customMachineTypeSharedCoreHelper(zone, cpuSeries string, memory int) (error, string) {
```
enocom, line 28: The error should be the last return value.

compute/instances/custom-machine-type/create_with_helper.go
```diff
@@ -0,0 +1,222 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+func customMachineTypeHelper(zone, cpuSeries string, coreCount, memory int) (error, string) {
```
enocom, line 28: Same here. Error should be the last return value.

enocom: What's a better name for this function? Helper isn't descriptive. Perhaps `instanceURI`? Or whatever the appropriate term is for "zones/%v/machineTypes/%v-%v-%v"?

enocom (changes requested):

enocom (approved):

PR #2546, opened 2022-04-29
chore(bigquery): switch bigquery query snippets to direct iterator

shollyman:
This updates the existing query snippets to demonstrate directly accessing the iterator.  Previously, these samples used the pattern of managing/monitoring the job directly via Run -> Wait -> Read.

This does two things:  simplifies the query samples, and potentially enables faster results access.

Samples that don't access query results were left unmodified (e.g. dry run, DML queries, etc).

### DIFF
--- a/bigquery/snippets/querying/bigquery_query.go
+++ b/bigquery/snippets/querying/bigquery_query.go
@@ -43,7 +43,7 @@ func queryBasic(w io.Writer, projectID string) error {
 	// Run the query and process the returned row iterator.
 	it, err := q.Read(ctx)
 	if err != nil {
-		return fmt.Errorf("query.Read(: %v", err)
+		return fmt.Errorf("query.Read(): %v", err)
 	}
 	for {
 		var row []bigquery.Value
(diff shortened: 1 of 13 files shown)


bigquery/snippets/querying/bigquery_query.go
```diff
@@ -40,19 +40,11 @@ func queryBasic(w io.Writer, projectID string) error {
 			"LIMIT 100")
 	// Location must match that of the dataset(s) referenced in the query.
 	q.Location = "US"
-	// Run the query and print results when the query job is completed.
-	job, err := q.Run(ctx)
+	// Run the query and process the returned row iterator.
+	it, err := q.Read(ctx)
 	if err != nil {
-		return err
+		return fmt.Errorf("query.Read(: %v", err)
```
tritone, line 46: assuming these should all be `"query.Read(): %v` (closing the parens)?

tritone (approved):

bigquery/snippets/querying/bigquery_query.go
shollyman: There is a saying among my people to express my feelings here:

"D'oh!"

PR #2549, opened 2022-05-02
feat(compute): custom vm type samples 3

FrodoTheTrue:


enocom (approved):

compute/instances/custom-machine-type/custom_machine_type_test.go
```diff
@@ -166,4 +166,92 @@ func TestComputeCreateInstanceWithCustomMachineTypeSnippets(t *testing.T) {
 	if err != nil {
 		t.Errorf("deleteInstance got err: %v", err)
 	}
+
+	buf.Reset()
+
+	want = "Instance updated"
+
+	if err := createInstanceWithCustomMachineType(buf, tc.ProjectID, zone, instanceName, customMT); err != nil {
+		t.Errorf("createInstanceWithCustomMachineType got err: %v", err)
```
codyoss, line 175: for these test consider if the test is still valid if a precondition fails. For example if the instance is never created in this step you are unable to modify or get the instance so this should fail the test with a t.Fatalf.

enocom:
@FrodoTheTrue There are some test errors that look pretty simple to fix. The sample itself looks good.

compute/instances/custom-machine-type/update_memory.go
```diff
@@ -0,0 +1,120 @@
...
+	"io"
+	"strings"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+// modifyInstanceWithExtendedMemory sends an instance creation request
+// to the Compute Engine API and wait for it to complete.
```
codyoss, line 30: nit: waits

codyoss (changes requested):
Just a couple small things, thanks for the PR.

codyoss (approved):

PR #2554, opened 2022-05-03
fix(healthcare): Wait for dataset operation to complete before continuing with tests

nikklassen:
This is basically the same as the fix in the Python tests GoogleCloudPlatform/python-docs-samples#7501

### DIFF
--- a/healthcare/dataset_create.go
+++ b/healthcare/dataset_create.go
@@ -19,13 +19,16 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"time"
 
 	healthcare "google.golang.org/api/healthcare/v1"
 )
 
 // createDataset creates a dataset.
 func createDataset(w io.Writer, projectID, location, datasetID string) error {
-	ctx := context.Background()
+	// Set a deadline for the dataset to become initialized.
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
 
 	healthcareService, err := healthcare.NewService(ctx)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


healthcare/dicom_test.go
```diff
@@ -58,9 +59,13 @@ func TestDICOMStore(t *testing.T) {
 		})
 	}
 
-	if err := createDataset(ioutil.Discard, tc.ProjectID, location, datasetID); err != nil {
-		t.Skipf("Unable to create test dataset: %v", err)
-		return
+	{
```
enocom, line 62: Ditto above.

healthcare/fhir_test.go
```diff
@@ -50,9 +50,13 @@ func TestFHIRStore(t *testing.T) {
 		})
 	}
 
-	if err := createDataset(ioutil.Discard, tc.ProjectID, location, datasetID); err != nil {
-		t.Skipf("Unable to create test dataset: %v", err)
-		return
+	{
```
enocom, line 53: And here.

healthcare/hl7v2_test.go
```diff
@@ -44,11 +45,14 @@ func TestHL7V2Store(t *testing.T) {
 		})
 	}
 
-	testutil.Retry(t, 10, time.Second, func(r *testutil.R) {
-		if err := createDataset(ioutil.Discard, tc.ProjectID, location, datasetID); err != nil {
-			r.Errorf("Unable to create test dataset: %v", err)
+	{
```
enocom, line 48: And here.

enocom (changes requested):

healthcare/dataset_test.go
```diff
@@ -50,9 +51,15 @@ func TestDataset(t *testing.T) {
 		})
 	}
 
-	if err := createDataset(buf, tc.ProjectID, location, datasetID); err != nil {
-		t.Fatalf("createDataset got err: %v", err)
+	{
```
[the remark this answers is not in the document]
tritone: +1, better to use a named context.

healthcare/dataset_test.go
```diff
@@ -50,9 +51,15 @@ func TestDataset(t *testing.T) {
...
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+		if err := createDataset(ctx, buf, tc.ProjectID, location, datasetID); err != nil {
+			cancel()
```
tritone, line 57: I don't believe cancelling explicitly is necessary when you use a timeout... @enocom can you sanity check me on that?

healthcare/dataset_test.go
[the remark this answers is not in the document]
enocom: A new variable name is more conventional in my opinion and reads a little better as not everyone knows this trick.

healthcare/dataset_create.go
```diff
@@ -19,14 +19,13 @@ import (
...
 
 	healthcare "google.golang.org/api/healthcare/v1"
 )
 
 // createDataset creates a dataset.
-func createDataset(w io.Writer, projectID, location, datasetID string) error {
-	ctx := context.Background()
-
+func createDataset(ctx context.Context, w io.Writer, projectID, location, datasetID string) error {
 	healthcareService, err := healthcare.NewService(ctx)
```
tritone, line 29: IMO it's better to create the context in the sample and set a timeout explicitly. This also is required by https://googlecloudplatform.github.io/samples-style-guide/#minimal-arguments Users will have to do this themselves in any case.

See example here: https://github.com/GoogleCloudPlatform/golang-samples/blob/HEAD/storage/buckets/create_bucket_class_location.go#L39

tritone (approved):

healthcare/dataset_create.go
nikklassen: thanks for pointing that out, I'll prefer following the style guide

enocom:
I'll leave this to @tritone to approve.

PR #2556, opened 2022-05-05
Adds STS samples

JesseLovelace:
Adds STS samples to bring Go up to speed with Java and Python. Also updates the test package to be set up for multiple tests.


storagetransfer/transfer_from_aws.go
```diff
@@ -0,0 +1,95 @@
...
+	// jobDescription = "Transfers objects from an AWS bucket to a GCS bucket"
+
+	// The name of the Aws bucket to transfer objects from
+	// awsSourceBucket := "my-source-bucket"
+
+	// The name of the GCS bucket to transfer objects to
+	// gcsSinkBucket := "my-sink-bucket"
+
+	// The time to start the transfer
+	// startTime := time.Now()
```
tritone, line 30: For jobDescription, startTime and the creds, it probably makes more sense to declare them inside the function rather than passing them in via arguments. Doesn't look necessary for testing in this case. Same comment applies to other samples.

https://googlecloudplatform.github.io/samples-style-guide/#minimal-arguments

storagetransfer/transfer_from_aws.go
```diff
@@ -0,0 +1,95 @@
...
+	"google.golang.org/genproto/googleapis/type/date"
+	"google.golang.org/genproto/googleapis/type/timeofday"
+	"io"
+	"time"
+
+	storagetransfer "cloud.google.com/go/storagetransfer/apiv1"
+	storagetransferpb "google.golang.org/genproto/googleapis/storagetransfer/v1"
+)
+
+func transfer_from_aws(w io.Writer, projectID string, jobDescription string, awsSourceBucket string, gcsSinkBucket string, startTime time.Time, awsAccessKeyID string, awsSecretKey string) (*storagetransferpb.TransferJob, error) {
```
tritone, line 16: Use camelCase for function names (here and elsewhere).

storagetransfer/transfer_from_aws.go
```diff
@@ -0,0 +1,95 @@
...
+	// gcsSinkBucket := "my-sink-bucket"
+
+	// The time to start the transfer
+	// startTime := time.Now()
+
+	// The AWS access key credential, should be accessed via environment variable for security
+	// awsAccessKeyID := os.Getenv("AWS_ACCESS_KEY_ID")
+
+	// The AWS secret key credential, should be accessed via environment variable for security
+	// awsAccessKeyID := os.Getenv("AWS_SECRET_KEY")
```
tritone, line 36: Should be awsSecretKey I believe.

storagetransfer/check_latest_transfer_operation.go
```diff
@@ -0,0 +1,54 @@
...
+		return nil, fmt.Errorf("storagetransfer.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	job, err := client.GetTransferJob(ctx, &storagetransferpb.GetTransferJobRequest{
+		JobName:   jobName,
+		ProjectId: projectID,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get transfer job: %v", err)
```
tritone, line 32: Error strings shouldn't be capitalized (here and elsewhere): https://github.com/golang/go/wiki/CodeReviewComments#error-strings

storagetransfer/storagetransfer_test.go
```diff
@@ -0,0 +1,266 @@
...
+		t.Errorf("check_latest_transfer_operation: Operation returned didn't have a valid operation name: %q", op.Name)
+	}
+
+	got := buf.String()
+	if want := op.Name; !strings.Contains(got, want) {
+		t.Errorf("check_latest_transfer_operation: got %q, want %q", got, want)
+	}
+	fmt.Println(got)
+
+	tj := &storagetransferpb.TransferJob{
```
tritone, line 223: Could you factor out this cleanup logic into a separate function? It could even be called at the beginning of each test with `defer`

tritone (commented):
Generally looks fine, a few Go style things.

storagetransfer/transfer_from_aws.go
JesseLovelace: Thanks for pointing me there!

As an aside, this style guide also says to avoid return types whenever possible, which is what Java does, every GCS java sample is a `void`. 

I have return types here because (I believe) that's what the GCS Go samples do. Is there some reason for that, or is it just historical?

tritone: Good question! Returning an `error` is just idiomatic error handling for Go, so we decided to make that an exception to the rule (after a lot of debate). The canonical snippet samples for Go have an error returned from all methods.

However, you're right that returning values other than errors is discouraged-- so, probably better to avoid it unless it's absolutely necessary for testing.

storagetransfer/storagetransfer_test.go
```diff
@@ -16,21 +16,18 @@ package storagetransfer
 
 import (
 	"bytes"
+	"cloud.google.com/go/iam"
```
tritone, line 19: Imports should be separated by language builtin vs. other dependencies (which will have a full import URL). Intellij should handle this correctly for you I think.

storagetransfer/transfer_from_aws.go
```diff
@@ -7,41 +7,42 @@ import (
...
-	// awsAccessKeyID := os.Getenv("AWS_SECRET_KEY")
-
 	ctx := context.Background()
 	client, err := storagetransfer.NewClient(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("storagetransfer.NewClient: %v", err)
 	}
 	defer client.Close()
 
+	// A description of this job
```
tritone, line 34: Comments here and following seem unnecessary, but fine to leave them if they are the same across languages.

tritone (commented):
nice clean up; a couple minor nits. Also LMK if you need help resolving the merge conflict.

storagetransfer/transfer_from_aws.go
JesseLovelace: Yeah the comments are just copied from the argument list, I'm going to leave them to call attention to the "these are the variables you can change for your on use case!" idea, but open to changing it in the future

tritone (approved):

storagetransfer/storagetransfer_test.go
codyoss: `goimports -w`

PR #2558, opened 2022-05-06
feat(compute): add preemtible samples

FrodoTheTrue:


compute/instances/preemptible/create_preemtible.go
```diff
@@ -0,0 +1,108 @@
...
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+// createPreemtibleInstance creates a new preemptible VM instance
+// with Debian 10 operating system.
+func createPreemtibleInstance(
+	w io.Writer,
+	projectID, zone, instanceName string,
+	preemtible bool,
```
codyoss, line 33: nit: I am not sure if this param should be an input. The way the description is worded for this sample I think that value is expected to always be true, right? If so please remove.

compute/instances/preemptible/preemptible_test.go
```diff
@@ -0,0 +1,100 @@
...
+	ctx := context.Background()
+	var r *rand.Rand = rand.New(
+		rand.NewSource(time.Now().UnixNano()))
+	tc := testutil.SystemTest(t)
+	zone := "europe-central2-b"
+	instanceName := fmt.Sprintf("test-vm-%v-%v", time.Now().Format("01-02-2006"), r.Int())
+
+	buf := &bytes.Buffer{}
+
+	err := createPreemtibleInstance(buf, tc.ProjectID, zone, instanceName, true)
```
codyoss, line 60: nit: anytime there is just an error returned that will be checked right away use the syntax: `if err := foo(); err != nil {...}`

compute/instances/preemptible/preemptible_test.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		t.Errorf("createPreemtibleInstance got err: %v", err)
```
codyoss, line 62: nit: make these first two asserts fatalf. Middle ones can remain errorf so that we always get the opportunity to delete the instance at the end.

codyoss (changes requested):
just a couple of small things.

PR #2565, opened 2022-05-13
feat: add code samples and tests for Video Stitcher

irataxy:
b: 202972969

### DIFF
--- a/media/stitcher/README.md
+++ b/media/stitcher/README.md
@@ -1,0 +1,5 @@
+# Video Stitcher API Go Samples
+
+This directory contains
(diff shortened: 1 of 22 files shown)


iennae:
Could you break this up into smaller PRs? This is a lot to review in one batch. By developing in small batches this makes it easier for folks to assess the code since we have asynchronous code reviews.

irataxy:
> Could you break this up into smaller PRs? This is a lot to review in one batch. By developing in small batches this makes it easier for folks to assess the code since we have asynchronous code reviews.

Done in https://github.com/GoogleCloudPlatform/golang-samples/pull/2572.

PR #2573, opened 2022-05-20
fix: Add some resilience to the grpc tests

muncus:
These tests occasionally flake, and adding a couple retries here should
(hopefully) resolve the flakes.


run/testing/grpc_server_streaming.e2e_test.go
```diff
@@ -71,12 +72,19 @@ func TestGRPCServerStreamingService(t *testing.T) {
...
 	recvMsgs := 0
+	recvFailures := 0
 	for {
 		_, err := resp.Recv()
 		if err == io.EOF {
 			break
 		} else if err != nil {
-			t.Fatalf("rpc StreamTime.Recv: %v", err)
+			recvFailures++
+			if recvFailures < 5 {
```
ace-n, line 82: Nit: move `5` to a predefined constant? ([Context](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants))

ace-n (approved):

run/testing/grpc_server_streaming.e2e_test.go
```diff
@@ -71,15 +72,22 @@ func TestGRPCServerStreamingService(t *testing.T) {
 	}
 
 	recvMsgs := 0
+	recvFailures := 0
```
enocom, line 75: Let's change this to just `var recvFailures int` and do the same above while we're at it.

run/testing/grpc_server_streaming.e2e_test.go
```diff
@@ -71,15 +72,22 @@ func TestGRPCServerStreamingService(t *testing.T) {
...
 	for {
 		_, err := resp.Recv()
-		if err == io.EOF {
+		if err == nil {
+			recvMsgs++
+		} else if err == io.EOF {
```
enocom, line 80: Let's check for io.EOF first.

run/testing/grpc_server_streaming.e2e_test.go
```diff
@@ -71,15 +72,22 @@ func TestGRPCServerStreamingService(t *testing.T) {
 	}
 
 	recvMsgs := 0
+	recvFailures := 0
 	for {
 		_, err := resp.Recv()
-		if err == io.EOF {
+		if err == nil {
```
enocom, line 78: It's preferred to indent the error flow. So this would become (with the io.EOF check first):

``` golang
_, err := resp.Recv()
if err == io.EOF {
  break
}
if err != nil {
  recvFailures++
  if recvFailures < 5 {
    // ...
  }
  continue
}

recvMsgs++

enocom (changes requested):

enocom (approved):

PR #2583, opened 2022-06-02
feat(secretmanager): create update_secret_with_alias.go

nataliesalvati:
Update samples for Access By Alias Launch


secretmanager/update_secret_with_alias.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2019 Google LLC
```
enocom, line 1: This needs to be 2022.

secretmanager/update_secret_with_alias.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"io"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
+	"google.golang.org/genproto/protobuf/field_mask"
+)
+
+// updateSecret updates the metadata about an existing secret.
+func updateSecret(w io.Writer, name string) error {
```
enocom, line 29: This will need to change to prevent colliding with the existing `updateSecret`. Maybe `updateSecretWithAlias`?

enocom (changes requested):
Some minor details. Also, there should probably be a test for this as well.

secretmanager/update_secret_with_alias.go
nataliesalvati: Updated name, date, and test

secretmanager/secretmanager_test.go
```diff
@@ -577,6 +577,34 @@ func TestUpdateSecret(t *testing.T) {
...
+	secret := testSecret(t, tc.ProjectID)
+	defer testCleanupSecret(t, secret.Name)
+
+	var b bytes.Buffer
+	if err := updateSecretWithAlias(&b, secret.Name); err != nil {
+		t.Fatal(err)
+	}
+
+	if got, want := b.String(), "Updated secret"; !strings.Contains(got, want) {
+		t.Errorf("updateSecret: expected %q to contain %q", got, want)
```
enocom, line 592: updateSecretWithAlias

secretmanager/update_secret_with_alias.go
```diff
@@ -0,0 +1,62 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
+	"google.golang.org/genproto/protobuf/field_mask"
+)
+
+// updateSecret updates the metadata about an existing secret.
```
enocom, line 28: updateSecretWithAlias

enocom: This doc string should probably say something about updating a secret with aliases instead of metadata?

enocom (changes requested):
Some minor things.

secretmanager/update_secret_with_alias.go
nataliesalvati: Fixed

secretmanager/secretmanager_test.go
nataliesalvati: This would be inconsistent with the etags case, it's still a call to updatesecret

secretmanager/update_secret_with_alias.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create secretmanager client: %v", err)
+	}
+	defer client.Close()
+
+	// Build the request.
+	req := &secretmanagerpb.UpdateSecretRequest{
+		Secret: &secretmanagerpb.Secret{
+			Name: name,
+			versionAliases: map[string]string{
```
enocom, line 44: This needs to be `VersionAliases` or whatever the field is called.

enocom (changes requested):
Missed a build failure.

secretmanager/update_secret_with_alias.go
nataliesalvati: Fixed, my b

secretmanager/update_secret_with_alias.go
```diff
@@ -0,0 +1,62 @@
...
+		return fmt.Errorf("failed to create secretmanager client: %v", err)
+	}
+	defer client.Close()
+
+	// Build the request.
+	req := &secretmanagerpb.UpdateSecretRequest{
+		Secret: &secretmanagerpb.Secret{
+			Name: name,
+			VersionAliases: map[string]string{
+				"test": "1",
```
enocom, line 45: `"1"` needs to be just `1`.

enocom (commented):
If you run go get ./... within the secret manager directory, the go tool will fix the go.mod issue.

secretmanager/update_secret_with_alias.go
nataliesalvati: Set, sorry it took me so long on the go package updates

enocom:
This is still failing with:

```
secretmanager_test.go:163: failed to update secret: rpc error: code = InvalidArgument desc = Aliases cannot be assigned to versions that don't exist
```


Looks like there's a test helper to create a secret version that we'll want to use here:

https://github.com/GoogleCloudPlatform/golang-samples/blob/main/secretmanager/secretmanager_test.go#L121-L128

secretmanager/secretmanager_test.go
```diff
@@ -604,3 +604,33 @@ func TestUpdateSecretWithEtag(t *testing.T) {
...
+
+	client, ctx := testClient(t)
+	s, err := client.GetSecret(ctx, &secretmanagerpb.GetSecretRequest{
+		Name: secret.Name,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if got, want := s.Labels, map[string]int64{"test": 1}; !reflect.DeepEqual(got, want) {
```
enocom, line 633: Should this should be `s.VersionAliases`?

enocom (approved):

PR #2585, opened 2022-06-06
feat: add Video Stitcher live session samples and tests

irataxy:
b:202972969

### DIFF
--- a/media/videostitcher/create_live_session.go
+++ b/media/videostitcher/create_live_session.go
@@ -33,8 +33,10 @@ func createLiveSession(w io.Writer, projectID, sourceURI, slateID string) error 
 	// manifest (.mpd) file or an M3U playlist manifest (.m3u8) file.
 	// sourceURI := "https://storage.googleapis.com/my-bucket/main.mpd"
 	// slateID := "my-slate"
-
-	// See Single Inline Linear (https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/tags)
+	// See https://cloud.google.com/video-stitcher/docs/concepts for information
+	// on ad tags and ad metadata. This sample uses an ad tag URL that displays
+	// a Single Inline Linear ad
+	// (https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/tags).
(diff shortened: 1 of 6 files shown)


media/videostitcher/create_live_session.go
```diff
@@ -0,0 +1,70 @@
...
+// createLiveSession creates a livestream session in which to insert ads.
+// Live sessions are ephemeral resources that expire after a few minutes.
+func createLiveSession(w io.Writer, projectID, sourceURI, adTagURI, slateID string) error {
+	// projectID := "my-project-id"
+
+	// Uri of the media to stitch; this URI must reference either an MPEG-DASH
+	// manifest (.mpd) file or an M3U playlist manifest (.m3u8) file.
+	// sourceURI := "https://storage.googleapis.com/my-bucket/main.mpd"
+
+	// See Single Inline Linear (https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/tags)
```
tritone, line 36: Are there actual docs describing the API that could be linked to rather than the list of examples? Seems confusing. Also, this could probably be a local variable rather than an argument; I don't think it needs to be subbed in for testing.

media/videostitcher/stitcher_test.go
```diff
@@ -404,3 +412,141 @@ func TestVodSessions(t *testing.T) {
...
+	adTagDetailsID := strSlice[len(strSlice)-1]
+	adTagDetailsID = strings.TrimRight(adTagDetailsID, "\n")
+	adTagDetailsName := fmt.Sprintf("projects/%s/locations/%s/liveSessions/%s/liveAdTagDetails/%s", projectNumber, location, sessionID, adTagDetailsID)
+	if !strings.Contains(got, adTagDetailsName) {
+		t.Errorf("listLiveAdTagDetails got\n----\n%v\n----\nWant to contain:\n----\n%v\n----\n", got, adTagDetailsName)
+	}
+	buf.Reset()
+
+	// Get the specified ad tag detail for a given live session.
+	testutil.Retry(t, 1, 2*time.Second, func(r *testutil.R) {
```
tritone, line 543: I think setting maxAttempts to 1 means that it will not be retried. Consider using more generous limits to avoid test flakes.

tritone (commented):
couple minor comments, in general LGTM!

media/videostitcher/stitcher_test.go
irataxy: Changed to 3

media/videostitcher/create_live_session.go
irataxy: Added a link to some conceptual documents. Made the ad tag URL a local variable.

PR #2598, opened 2022-06-14
fix(gcfv2/imagemagick): Catch storage Close() errors, and include zombie image.

muncus:


tritone (approved):
Good catch-- definitely worth checking if there are other instances of `Writer.Close` not being checked floating around.

PR #2600, opened 2022-06-14
chore: try to get renovate working

quartzmo:
Similar to googleapis/google-cloud-go#6050

> Switching the group name should make changes come from a new branch on the renovate fork I believe. This may fix the issue that we are no longer getting PRs from our current branch.


codyoss (approved):

codyoss:
And now we wait!

PR #2615, opened 2022-06-28
feat: add sample to disable billingAccount

tonnytg:
On doc https://cloud.google.com/billing/docs/how-to/notify#cap_disable_billing_to_stop_usage
don't have any Go code, this can be used to disable sandbox when pub/sub triggered by the budget alert.

### DIFF
--- a/billing/disableBillingAccount/go.mod
+++ b/billing/disableBillingAccount/go.mod
@@ -1,0 +1,23 @@
+module github.com/GoogleCloudPlatform/golang-samples/billing
+
+go 1.18
+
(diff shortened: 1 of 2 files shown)


tritone (commented):
Hi @tonnytg , thanks for the PR. Unfortunately, as it says in the repo README, we can't accept new samples from external contributors in this repo.

I'll pass this along to the billing tech writing team and hopefully they can address the gap that you noticed in the docs at https://cloud.google.com/billing/docs/how-to/notify#cap_disable_billing_to_stop_usage .

PR #2618, opened 2022-06-29
chore(functions): clarify equality of endpoint and audience

ace-n:
Partial fix for b/237501078


muncus (approved):
Thanks for clarifying this, Ace!

PR #2623, opened 2022-07-05
samples(compute): add moving instances samples [do not merge]

FrodoTheTrue:


compute/get_instance.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	defer instancesClient.Close()
+
+	reqInstance := &computepb.GetInstanceRequest{
+		Project:  projectID,
+		Zone:     zone,
+		Instance: instanceName,
+	}
+
+	return instancesClient.Get(ctx, reqInstance)
```
codyoss, line 46: Samples should only return errors, please use the passed in writer to print something and verify.

compute/disks/create_disk_from_snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+	w io.Writer,
+	projectID, zone, diskName, diskType, snapshotLink string,
+	diskSizeGb int64,
+) error {
+	// projectID := "your_project_id"
+	// zone := "europe-central2-b"
+	// diskName := "your_disk_name"
+	// diskType := "zones/us-west3-b/diskTypes/pd-ssd"
+	// snapshotLink := "projects/project_name/global/snapshots/snapshot_name"
+	// diskSizeGb := 120
```
codyoss, line 39: nit: hard code this value, does not need to be variable.

codyoss: And in other samples too.

compute/disks/set_disk_autodelete.go
```diff
@@ -0,0 +1,96 @@
...
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+)
+
+// setDiskAutodelete sets the autodelete flag of a disk to given value.
+func setDiskAutodelete(
+	w io.Writer,
+	projectID, zone, instanceName, diskName string,
+	autoDelete bool,
```
codyoss, line 31: I would not pass this value. The default is false so I think we should hard code this to true for this sample.

compute/disks/disks_test.go
```diff
@@ -0,0 +1,270 @@
...
+		t.Fatalf("NewInstancesRESTClient: %v", err)
+	}
+
+	defer instancesClient.Close()
+
+	buf := &bytes.Buffer{}
+
+	err = createDisk(ctx, tc.ProjectID, zone, diskName, sourceImage)
+	if err != nil {
+		t.Errorf("createDisk got err: %v", err)
```
codyoss, line 160: Many of these should be a t.Fatalf. Errorf allows the test to keep running after failure, but that is not what is desired here.

compute/disks/disks_test.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	err = createDiskSnapshot(ctx, tc.ProjectID, zone, diskName, snapshotName)
+	if err != nil {
+		t.Errorf("createDiskSnapshot got err: %v", err)
+	}
+
+	diskSnapshotLink := fmt.Sprintf("projects/%s/global/snapshots/%s", tc.ProjectID, snapshotName)
```
codyoss, line 168: define this above with other vars

compute/disks/disks_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	op, err := instancesClient.Delete(ctx, req)
+	if err != nil {
+		return err
+	}
+
+	return op.Wait(ctx)
+}
+
+func TestComputeDisksSnippets(t *testing.T) {
```
codyoss, line 135: Can this be split up into individual tests with a TestMain for setup? Or use t.Run to help break this up. Hard to follow what is being tested as is.

compute/instances/create-start-instance/create_instance_with_existing_disks.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		disks = append(disks, disk)
+	}
+
+	attachedDisks := [](*computepb.AttachedDisk){}
+
+	for _, disk := range disks {
+		attachedDisk := &computepb.AttachedDisk{
+			Source: proto.String(disk.GetSelfLink()),
```
codyoss, line 74: Is SelfLink just a resource path? Can this be constructed instead of fetched?

compute/snapshots/create_snapshot.go
```diff
@@ -0,0 +1,126 @@
...
+// createSnapshot creates a snapshot of a disk.
+func createSnapshot(
+	w io.Writer,
+	projectID, diskName, snapshotName, zone, region, location, diskProjectID string,
+) error {
+	// projectID := "your_project_id"
+	// diskName := "your_disk_name"
+	// snapshotName := "your_snapshot_name"
+	// zone := "europe-central2-b"
+	// region := ""
```
codyoss, line 37: provide a value

compute/snapshots/create_snapshot.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"io"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+// createSnapshot creates a snapshot of a disk.
+func createSnapshot(
```
codyoss, line 29: This should be two seperate samples as there is branching paths for two different sets of input and validation logic.

codyoss (changes requested):
This PR contains a few different areas of work and over 1000 LoC. In the future please try to split PRs like this up into multiple.

compute/instances/create-start-instance/create_instance_with_existing_disks.go
FrodoTheTrue: Here I use the same logic as we use in the canonical python sample: https://github.com/googleapis/python-compute/blob/main/samples/snippets/instances/create_start_instance/create_with_existing_disks.py#L230

FrodoTheTrue:
Sure, this PR is just a WIP. I have splitted this PR into 3 parts and fixed your comments. We could jump into them for more detailed comments, if neded:
1) Disks samples: https://github.com/GoogleCloudPlatform/golang-samples/pull/2628
2) Snapshots samples: https://github.com/GoogleCloudPlatform/golang-samples/pull/2629
3) Samples for moving instances doc: https://github.com/GoogleCloudPlatform/golang-samples/pull/2630

codyoss:
@FrodoTheTrue Can we close this?

FrodoTheTrue:
yep, sure

PR #2624, opened 2022-07-06
feat(storagetransfer): add samples for posix transfers and manifest

JesseLovelace:
Tests likely to fail initially because some pub/sub permissions will need to be set up--working on it! (Edit: permissions added, thanks cody!)


storagetransfer/transfer_between_posix.go
```diff
@@ -0,0 +1,88 @@
...
+
+	storagetransfer "cloud.google.com/go/storagetransfer/apiv1"
+	storagetransferpb "google.golang.org/genproto/googleapis/storagetransfer/v1"
+)
+
+func transferBetweenPosix(w io.Writer, projectID string, sourceAgentPoolName string, sinkAgentPoolName string, rootDirectory string, destinationDirectory string, bucketName string) (*storagetransferpb.TransferJob, error) {
+	// Your project id
+	// projectId := "my-project-id"
+
+	// The agent pool associated with the POSIX data source. If not provided, defaults to the default agent
```
tritone, line 32: Here and elsewhere-- there's no way to "not provide" a function arg in Go. How would the user get the default values if they want them? Is it enough to pass an empty string, or does something different need to happen in how the request is constructed?

storagetransfer/storagetransfer_test.go
```diff
@@ -188,6 +189,120 @@ func TestGetLatestTransferOperation(t *testing.T) {
...
 }
 
+func TestDownloadToPosix(t *testing.T) {
+	tc := testutil.SystemTest(t)
+
+	buf := new(bytes.Buffer)
+
+	rootDirectory, err := ioutil.TempDir("", "download-to-posix-test")
+	if err != nil {
+		t.Errorf("download_to_posix: %#v", err)
```
tritone, line 199: This should probably be t.Fatalf since the test can't proceed if this fails. Same for L210 and elsewhere in other samples

storagetransfer/storagetransfer_test.go
```diff
@@ -188,6 +189,120 @@ func TestGetLatestTransferOperation(t *testing.T) {
...
+	rootDirectory, err := ioutil.TempDir("", "transfer-using-manifest-test")
+	if err != nil {
+		t.Errorf("transfer_using_manifest: %#v", err)
+	}
+	defer os.RemoveAll(rootDirectory)
+
+	sourceAgentPoolName := "" //use default agent pool
+	object := sc.Bucket(gcsSourceBucket).Object("manifest.csv")
+	defer object.Delete(context.Background())
+	manifestLocation := "gs://" + gcsSourceBucket + "/manifest.csv"
```
tritone, line 291: Maybe better to pass gcsSourceBucket as an arg and construct this string in-line in the sample? Seems important to be clear that users must use the gs:// prefix.

tritone (commented):
Few small things, overall looks good

storagetransfer/transfer_between_posix.go
JesseLovelace: It would just be passing in an empty string, or even just not specifying the parameter, either works

tritone: Gotcha, makes sense. I guess since this is docs inside the function rather than a function godoc, it's fine to phrase it this way.

tritone (approved):

PR #2626, opened 2022-07-07
fix: set the default mongo port to be the actual mongo port

tpryan:
Currently pointing to 80 for historical reasons in the base tutorial.  Let's change that.


enocom (approved):
Should this be something a user could change?

PR #2630, opened 2022-07-08
samples(compute): add compute samples for moving instances doc

FrodoTheTrue:


compute/instances/create-start-instance/create_instance_test.go
```diff
@@ -100,4 +146,43 @@ func TestComputeCreateInstanceSnippets(t *testing.T) {
 	if err != nil {
 		t.Errorf("deleteInstance got err: %v", err)
 	}
+
+	buf.Reset()
+
+	err = createDisk(ctx, tc.ProjectID, zone, diskName2, *newestDebian.SelfLink)
+	if err != nil {
+		t.Errorf("createDisk got err: %v", err)
```
codyoss, line 154: prereqs and creation tasks shoul be t.Fatalf, the following 3

codyoss (changes requested):
Just one small thing.

compute/instances/create-start-instance/create_instance_test.go
FrodoTheTrue: Why we should modify in the following parts? From what I see 154 line makes sense to make fatalf, but we will not delete this disk if we will make next calls as fatalf? Or I am missing smth

codyoss: I think as resources are created perhaps we should defer deleting them, thoughts?

FrodoTheTrue: Definitely agree, but adding Fatalf will not help from what I see
In Node.js we do this trick: https://github.com/googleapis/nodejs-compute/blob/main/samples/test/disks.test.js#L131 (deleting resources with the same name structure)

codyoss: If you want to add a general cleanup call like that here I am okay with that.

codyoss (approved):

Issue #2637, opened 2022-07-15
Weird error message


timograf:
Hello

Looks like you got a copy/paste error all over the file, since you are not calling "bigtable.NewAdminClient" but "bigtable.NewClient".

https://github.com/GoogleCloudPlatform/golang-samples/blob/ba5aa117abd4ccd77864dc66e7d142e6e03098db/bigtable/reads/reads.go#L40

codyoss:
Thank you for the thoughtful feedback!

PR #2638, opened 2022-07-15
test(bigtable): fixup a handful of style violations

codyoss:
Fixes: #2637


muncus (approved):
Thanks for fixing these up!

PR #2679, opened 2022-08-09
chore(cloudsql): add consistent region tags

enocom:
This commit adds region tags that match the other languages. A future
commit will remove the old region tags.


cloudsql/postgres/database-sql/connect_tcp.go
```diff
@@ -82,4 +83,5 @@ func connectTCPSocket() (*sql.DB, error) {
 }
 
 // [END cloud_sql_postgres_databasesql_sslcerts]
+// [START cloud_sql_postgres_databasesql_connect_tcp_sslcerts]
```
jackwotherspoon, line 86: should be `END` tag

jackwotherspoon (changes requested):

jackwotherspoon (approved):

Issue #2435, opened 2022-02-15
Fix a typo in the comment for resetInstance


savijatv:
https://github.com/GoogleCloudPlatform/golang-samples/blob/dc44c2939e6320b58ea4b344c0618b8bc96b27c6/compute/reset_instance.go#L27

The comment before reset should say 'Resets a running instance..." instead of 'Resets a stopped instance...'

PR #2683, opened 2022-08-10
fix: Fix typo on function comment

jhonatanlteodoro:
Refs: #2435


codyoss (dismissed):
Thanks for the contributiion!

PR #2688, opened 2022-08-11
feat(pubsub): add exactly once delivery samples

hongalex:


pubsub/subscriptions/pull_exactly_once_delivery.go
```diff
@@ -0,0 +1,76 @@
...
+	sub := client.Subscription(subID)
+
+	// Receive messages for 10 seconds, which simplifies testing.
+	// Comment this out in production, since `Receive` should
+	// be used as a long running operation.
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+	err = sub.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
+		fmt.Fprintf(w, "Got message: %q\n", string(msg.Data))
+		r := msg.AckWithResult()
```
maheshgattani, line 50: Can we add a comment similar to the publish sample: https://github.com/GoogleCloudPlatform/golang-samples/blob/83461c3e2568fd4db5bc5717d9eb31b63ab33170/pubsub/topics/publish.go#L41. Basically referring to the blocking behavior here.

maheshgattani (approved):

enocom (approved):

maheshgattani (approved):

pubsub/subscriptions/pull_exactly_once_delivery.go
```diff
@@ -0,0 +1,83 @@
...
+
+		switch status {
+		case pubsub.AcknowledgeStatusSuccess:
+			fmt.Fprintf(w, "Message successfully acked: %s", msg.ID)
+		case pubsub.AcknowledgeStatusInvalidAckID:
+			fmt.Fprintf(w, "Message failed to ack with response of Invalid. ID: %s", msg.ID)
+		case pubsub.AcknowledgeStatusPermissionDenied:
+			fmt.Fprintf(w, "Message failed to ack with response of Permission Denied. ID: %s", msg.ID)
+		case pubsub.AcknowledgeStatusFailedPrecondition:
+			fmt.Fprintf(w, "Message failed to ack with response of Permission Denied. ID: %s", msg.ID)
```
pradn, line 71: Failed precondition

pubsub/subscriptions/pull_exactly_once_delivery.go
```diff
@@ -0,0 +1,83 @@
...
+		fmt.Fprintf(w, "Got message: %q\n", string(msg.Data))
+		r := msg.AckWithResult()
+		// Block until the result is returned and a pubsub.AcknowledgeStatus
+		// is returned for the acked message.
+		status, err := r.Get(ctx)
+		if err != nil {
+			fmt.Fprintf(w, "MessageID: %s failed when calling result.Get: %v", msg.ID, err)
+		}
+
+		switch status {
```
pradn, line 63: The handling of the error illustrates the error types, but doesn't tell the user what they should do with permanent ack failures. I don't think it's easy to come up with a simple example that helps users understand how to deal with these errors in real life. Just some thoughts.

pradn (changes requested):

pubsub/subscriptions/pull_exactly_once_delivery.go
hongalex: Yeah. I was using the [Java sample as reference](https://github.com/googleapis/java-pubsub/blob/37eaff859422bb215ace202ffd0adf8a651dadb5/samples/snippets/src/main/java/pubsub/SubscribeWithExactlyOnceConsumerWithResponseExample.java). I think we struggled to provide a useful example for ordered key's `ResumePublish` as well.

PR #2692, opened 2022-08-17
feat(dataflow): Add Go Flex Template example

jrmccluskey:
Adds a code snippet and instructions on how to build and run a Dataflow flex template using the Apache Beam Go SDK. This is meant to be the Go equivalent of the [Python](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/main/dataflow/flex-templates/streaming_beam) and [Java](https://github.com/GoogleCloudPlatform/java-docs-samples/tree/main/dataflow/flex-templates/streaming_beam_sql) tutorials, albeit with a simpler wordcount pipeline.


dataflow/flex-templates/wordcount/metadata.json
```diff
@@ -0,0 +1,18 @@
...
+            "isOptional": true
+        },
+        {
+            "name": "output",
+            "label": "Output file.",
+            "helpText": "GCS location to write output to.",
+            "isOptional": false
+        }
+    ]
+}
```
enocom, line 18: nit: newline

dataflow/flex-templates/wordcount/wordcount.go
```diff
@@ -0,0 +1,118 @@
...
+// formatFn is a DoFn that formats a word and its count as a string.
+func formatFn(w string, c int) string {
+	return fmt.Sprintf("%s: %v", w, c)
+}
+
+// CountWords is a composite transform that counts the words of a PCollection
+// of lines. It expects a PCollection of type string and returns a PCollection
+// of type KV<string,int>. The Beam type checker enforces these constraints
+// during pipeline construction.
+func CountWords(s beam.Scope, lines beam.PCollection) beam.PCollection {
```
enocom, line 89: Does this function need to be exported?

dataflow/flex-templates/wordcount/wordcount.go
```diff
@@ -0,0 +1,118 @@
...
+	smallWords      = beam.NewCounter("extract", "smallWords")
+	lineLen         = beam.NewDistribution("extract", "lineLenDistro")
+)
+
+// extractFn is a DoFn that emits the words in a given line and keeps a count for small words.
+type extractFn struct {
+	SmallWordLength int `json:"smallWordLength"`
+}
+
+func (f *extractFn) ProcessElement(ctx context.Context, line string, emit func(string)) {
```
enocom, line 65: I assume the Beam Pipeline requires this to be exported. Is that right?

Also, a GoDoc for the method would be nice.

enocom (changes requested):
Should this have some tests?

dataflow/flex-templates/wordcount/wordcount.go
jrmccluskey: Yes the lifecycle methods for structure DoFns have to be exported, along with any internal fields that need to be serialized (like SmallWordLength.)

dataflow/flex-templates/wordcount/metadata.json
jrmccluskey: Done.

dataflow/flex-templates/wordcount/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM gcr.io/dataflow-templates-base/go-template-launcher-base
+
+ARG WORKDIR=/dataflow/template
+RUN mkdir -p ${WORKDIR}
+
+COPY wordcount ${WORKDIR}/wordcount
+
+ENV FLEX_TEMPLATE_GO_BINARY="${WORKDIR}/wordcount"
+
+ENTRYPOINT ["/opt/google/dataflow/go_template_launcher"]
```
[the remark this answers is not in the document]
jrmccluskey: Done.

dataflow/flex-templates/wordcount/wordcount.go
jrmccluskey: I don't think it has to be, it could be made unexported if you would prefer.

jrmccluskey:
We have [some tests for this that exist already](https://github.com/apache/beam/blob/7fea1ef262cad725ce30651ebb441d56dba0c11c/sdks/go/test/integration/wordcount/wordcount_test.go), although they're somewhat abstract since they leverage the counters and use some hashing to validate the output instead of looking at the PCollection content. I can add them if you'd prefer a test example with the snippet though.

tritone:
> We have [some tests for this that exist already](https://github.com/apache/beam/blob/7fea1ef262cad725ce30651ebb441d56dba0c11c/sdks/go/test/integration/wordcount/wordcount_test.go), although they're somewhat abstract since they leverage the counters and use some hashing to validate the output instead of looking at the PCollection content. I can add them if you'd prefer a test example with the snippet though.

We require integration tests for all samples in this repo -- see https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#testing for guidelines.

dataflow/flex-templates/wordcount/wordcount.go
enocom: In that case, not exporing from package main is preferred.

enocom:
Some kind of test coverage would be good, if only to ensure this sample keeps working for the future.

jrmccluskey:
Added tests

enocom (approved):

PR #2698, opened 2022-08-23
docs(translate): add granular markers to v3 translate api for code sample extraction

SKrupa:
This aligns the code samples with some other language samples such as [NodeJS](https://github.com/googleapis/nodejs-translate/blob/main/samples/v3/translate_translate_text.js) so that code samples can be extracted in documentation.


translate/v3/translate_text.go
```diff
@@ -55,15 +59,16 @@ func translateText(w io.Writer, projectID string, sourceLang string, targetLang
 	if err != nil {
 		return fmt.Errorf("TranslateText: %v", err)
 	}
+	// [END translate_v3_translate_text_2]
 
+	// [START translate_v3_translate_text_2]
```
quartzmo, line 64: Copy/paste error? Should this block be `translate_v3_translate_text_3` as in the Node sample?

quartzmo (changes requested):

translate/v3/translate_text.go
SKrupa: Whoops, thanks!!

quartzmo (approved):

SKrupa:
@codyoss are there any outstanding issues I need to address for merging this PR? We're planning on updating our extension with these updated samples soon, so would appreciate a sign off. Thanks!

enocom (approved):

PR #2704, opened 2022-08-26
docs: clears up difference of go from other languages

iennae:
in sample browser, other languages have an explicit base64 decode. This adds some documentation so that folks don't file bugs on "missing a decode step" 

b/238760874

### DIFF
--- a/run/pubsub/main.go
+++ b/run/pubsub/main.go
@@ -50,7 +50,6 @@ func main() {
 // PubSubMessage is the payload of a Pub/Sub event.
 // See the documentation for more details:
 // https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
-// byte slice unmarshalling handles base64 decoding
 type PubSubMessage struct {
 	Message struct {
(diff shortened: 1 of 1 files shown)


run/pubsub/main.go
```diff
@@ -50,6 +50,7 @@ func main() {
 // PubSubMessage is the payload of a Pub/Sub event.
 // See the documentation for more details:
 // https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
+// byte slice unmarshalling handles base64 decoding
```
averikitsch, line 53: I understand why the comment is here but should it be on line 71?

iennae: maybe? I contemplated that, but I didn't want to break up the code either.

averikitsch (approved):

run/pubsub/main.go
tritone, line 53: Nit: capitalize/use a period for a sentence comment. go/go-style/decisions#comment-sentences

tritone (approved):

PR #2706, opened 2022-08-30
feat(auth): add auth samples

FrodoTheTrue:


quartzmo (changes requested):
Please rename the following files to use snake-case:

* `auth/idTokenFromImpersonatedCredentials.go`
* `auth/idTokenFromMetadataServer.go`
* `auth/idTokenFromServiceAccount.go`
* `auth/verifyGoogleIdToken.go`

auth/idTokenFromImpersonatedCredentials.go
```diff
@@ -0,0 +1,73 @@
...
+	// working environment.
+	credentials, err := google.FindDefaultCredentials(ctx, scope)
+	if err != nil {
+		return fmt.Errorf("failed to generate default credentials: %v", err)
+	}
+
+	// delegates: The chained list of delegates required to grant the final accessToken.
+	// For more information, see:
+	// https://cloud.google.com/iam/docs/create-short-lived-credentials-direct#sa-credentials-permissions
+	// Delegate is NOT USED here.
```
quartzmo, line 48: "Delegate is NOT USED here."

What does this mean?

auth/idTokenFromMetadataServer.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create NewTokenSource: %v", err)
+	}
+
+	// Get the ID token.
+	// Once you've obtained the ID token, you can use it to make an authenticated call
+	// to the target audience.
+	_, err = ts.Token()
+	if err != nil {
+		return fmt.Errorf("enable to receive token: %v", err)
```
quartzmo, line 52: "enable"? Maybe meant: "failed" ?

auth/idTokenFromServiceAccount.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create NewTokenSource: %v", err)
+	}
+
+	// Get the ID token.
+	// Once you've obtained the ID token, you can use it to make an authenticated call
+	// to the target audience.
+	_, err = ts.Token()
+	if err != nil {
+		return fmt.Errorf("enable to receive token: %v", err)
```
quartzmo, line 59: "enable"? Maybe meant: "failed" ?

auth/idTokenFromImpersonatedCredentials.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return fmt.Errorf("IDTokenSource error: %v", err)
+	}
+
+	// Get the ID token.
+	// Once you've obtained the ID token, you can use it to make an authenticated call
+	// to the target audience.
+	_, err = ts.Token()
+	if err != nil {
+		return fmt.Errorf("enable to receive token: %v", err)
```
quartzmo, line 66: "enable"? Maybe meant: "failed" ?

auth/authenticate_implicit_with_adc.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+)
+
+// authenticateImplicitWithAdc auto-detects credentials
+// and prints storage buckets by authenticating with ADC.
```
quartzmo, line 28: Maybe state the full-length term Application Default Credentials, or else provide a link?

auth/authenticate_explicit_with_adc.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/storage"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+)
+
+// authenticateExplicitWithAdc prints storage buckets by authenticating with ADC.
```
quartzmo, line 29: Maybe state the full-length term Application Default Credentials, or else provide a link?

auth/auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	ctx := context.Background()
+	tc := testutil.SystemTest(t)
+	audience := "iap.googleapis.com"
+
+	want := "Listed all storage buckets."
+
+	buf := &bytes.Buffer{}
+
+	if err := authenticateExplicitWithAdc(buf); err != nil {
+		t.Errorf("authenticateExplicitWithAdc got err: %v", err)
```
quartzmo, line 39: Use `t.Fatalf` here because subsequent operation(s) depend on this one succeeding?

auth/auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if got := buf.String(); !strings.Contains(got, want) {
+		t.Errorf("authenticateExplicitWithAdc got %q, want %q", got, want)
+	}
+
+	buf.Reset()
+
+	if err := authenticateImplicitWithAdc(buf, tc.ProjectID); err != nil {
+		t.Errorf("authenticateImplicitWithAdc got err: %v", err)
```
quartzmo, line 48: Use `t.Fatalf` here because subsequent operation(s) depend on this one succeeding?

auth/auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if got := buf.String(); !strings.Contains(got, want) {
+		t.Errorf("authenticateImplicitWithAdc got %q, want %q", got, want)
+	}
+
+	want = "Generated ID token."
+	buf.Reset()
+
+	if err := getIdTokenFromMetadataServer(buf, audience); err != nil {
+		t.Errorf("getIdTokenFromMetadataServer got err: %v", err)
```
quartzmo, line 58: Use `t.Fatalf` here because subsequent operation(s) depend on this one succeeding?

auth/auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if got := buf.String(); !strings.Contains(got, want) {
+		t.Errorf("getIdTokenFromMetadataServer got %q, want %q", got, want)
+	}
+
+	buf.Reset()
+
+	if err := getIdTokenFromServiceAccount(buf, os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"), audience); err != nil {
+		t.Errorf("getIdTokenFromServiceAccount got err: %v", err)
```
quartzmo, line 67: Use `t.Fatalf` here because subsequent operation(s) depend on this one succeeding?

auth/auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	if got := buf.String(); !strings.Contains(got, want) {
+		t.Errorf("getIdTokenFromServiceAccount got %q, want %q", got, want)
+	}
+
+	buf.Reset()
+	want = "ID token verified."
+
+	credentials, err := google.FindDefaultCredentials(ctx)
+	if err != nil {
+		t.Errorf("failed to generate default credentials: %v", err)
```
quartzmo, line 78: Use `t.Fatalf` here because subsequent operation(s) depend on this one succeeding?

auth/auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	ts, err := idtoken.NewTokenSource(ctx, audience, option.WithCredentials(credentials))
+	if err != nil {
+		t.Errorf("failed to create NewTokenSource: %v", err)
```
quartzmo, line 83: Use `t.Fatalf` here because subsequent operation(s) depend on this one succeeding?

auth/auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	token, err := ts.Token()
+	if err != nil {
+		t.Errorf("failed to get ID token: %v", err)
```
quartzmo, line 88: Use `t.Fatalf` here because subsequent operation(s) depend on this one succeeding?

auth/auth_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if err := verifyGoogleIdToken(buf, token.AccessToken, audience); err != nil {
+		t.Errorf("verifyGoogleIdToken got err: %v", err)
```
quartzmo, line 92: Use `t.Fatalf` here because subsequent operation(s) depend on this one succeeding?

quartzmo (changes requested):

auth/authenticate_explicit_with_adc.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to generate default credentials: %v", err)
+	}
+	// If you are authenticating to a Cloud API, you can let the library include the default scope,
+	// https://www.googleapis.com/auth/cloud-platform, because IAM is used to provide fine-grained
+	// permissions for Cloud.
+	// For more information on scopes to use,
+	// see: https://developers.google.com/identity/protocols/oauth2/scopes
+
+	// onstruct the Storage client.
```
piaxc, line 47: typo, should be "Construct"

auth/authenticate_implicit_with_adc.go
piaxc: +1, maybe just
authenticateImplicitWithAdc uses Application Default Credentials
to automatically find credentials and authenticate.

auth/authenticate_implicit_with_adc.go
```diff
@@ -0,0 +1,61 @@
...
+func authenticateImplicitWithAdc(w io.Writer, projectId string) error {
+	// projectId := "your_project_id"
+
+	ctx := context.Background()
+
+	// This snippet demonstrates how to list buckets.
```
piaxc, line 34: Please remove, while strictly true, it's not really the point of the snippet.

piaxc (changes requested):
Thanks Fedor!

auth/idTokenFromImpersonatedCredentials.go
FrodoTheTrue: Moved to the request params

FrodoTheTrue:
@quartzmo could you please take a loot once more?
Also, I am receiving stange lint error ...

```
To fix this check, run "goimports -w . && find . -name go.mod -execdir go mod tidy \;"
HEAD detached at pull/2706/merge
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   auth/auth_test.go
	modified:   auth/id_token_from_impersonated_credentials.go

no changes added to commit (use "git add" and/or "git commit -a")
Error: Process completed with exit code 1.
```

Not sure what to do with that

FrodoTheTrue:
@codyoss could you please help me with an error above?

codyoss:
Looks like you need to run goimports to format the files.

FrodoTheTrue:
@quartzmo could you please take a look once more?

quartzmo (approved):

auth/id_token_from_impersonated_credentials.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/impersonate"
+	"google.golang.org/api/option"
+)
+
+// getIdTokenFromImpersonatedCredentials uses a service account (SA1) to impersonate as
+// another service account (SA2) and obtain id token for the impersonated account.
+// To obtain token for SA2, SA1 should have the "roles/iam.serviceAccountTokenCreator" permission on SA2.
+func getIdTokenFromImpersonatedCredentials(w io.Writer, scope, targetAudience, impersonatedServiceAccount string) error {
```
codyoss, line 31: Can we add a test for this?

auth/id_token_from_impersonated_credentials.go
```diff
@@ -0,0 +1,71 @@
...
+	// scope := "https://www.googleapis.com/auth/cloud-platform"
+	// targetAudience := "http://www.example.com"
+	// impersonatedServiceAccount := "name@project.service.gserviceaccount.com"
+
+	ctx := context.Background()
+
+	// Construct the GoogleCredentials object which obtains the default configuration from your
+	// working environment.
+	credentials, err := google.FindDefaultCredentials(ctx, scope)
```
codyoss, line 40: Not creating this type and passing this has the same affect, can you remove this from the example?

auth/id_token_from_metadata_server.go
```diff
@@ -0,0 +1,59 @@
...
+// getIdTokenFromMetadataServer uses the Google Cloud metadata server environment
+// to create an identity token and add it to the HTTP request as part of an Authorization header.
+func getIdTokenFromMetadataServer(w io.Writer, url string) error {
+	// url := "http://www.example.com"
+
+	ctx := context.Background()
+
+	// Construct the GoogleCredentials object which obtains the default configuration from your
+	// working environment.
+	credentials, err := google.FindDefaultCredentials(ctx)
```
codyoss, line 37: Same as previous file, I think this should just be removed as it is not needed to make the code work.

codyoss (changes requested):

auth/id_token_from_impersonated_credentials.go
FrodoTheTrue: It will work, but the goal of the samples is to show how to explicitly set up the credentials
Examples from other languages: https://github.com/googleapis/google-auth-library-python/blob/main/samples/cloud-client/snippets/idtoken_from_impersonated_credentials.py#L47

auth/id_token_from_impersonated_credentials.go
FrodoTheTrue: It's not clear how to test impersonated token creation, so the team decided for now to skip this in all languages.
Ex: https://github.com/googleapis/google-auth-library-python/blob/main/samples/cloud-client/snippets/snippets_test.py

auth/id_token_from_impersonated_credentials.go
codyoss: For these two can we add to the comment to say that this would be done implicitly or something to that nature? I think it is fine to be explicit. But if we don't say we are being verbose users are going to follow this pattern without knowing that there is a simpler way.

FrodoTheTrue: I guess we have "authenticate explicit" example and it should be enough from my view.

Again, same for other languages (there are additional info): https://github.com/googleapis/google-auth-library-python/pull/1102/files#diff-96271f4091235622616d8e8fa7d45ddf6da62cd552544721dc10214a3562d57dR47

codyoss: That is true there is a sample like that but think about it from a user point of view. I don't think that sample and this one will be shown on the same page, right? If there are not they don't have the context of the other sample. I get that this is how things were done in other langs too, but I wonder if it is right to do so there as well.

Whenever possible I think we should have users rely on our auto-detection logic and only in exceptional cases be explicit, when required. I will approve this, but might be worth talking to the team about.

codyoss (approved):

Issue #1946, opened 2021-02-09
bigquery: demonstrate use of arrow serialization with bigquerystorage reading


shollyman:
Copied from https://github.com/googleapis/google-cloud-go/issues/3682, filed by https://github.com/mrgleeco.

## Request:

A golang analog to the Java implementation as described at https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1

## What we need:

Example showing how to use https://pkg.go.dev/github.com/apache/arrow/go/arrow and relevant subpackages for reading individual row batches, and at least some schema handling code.

shollyman:
https://arrow.apache.org/docs/status.html#implementation-status tracks the deltas by language.  Looks like we're still missing some necessary types for go (null and decima256), and it looks like there's still a requirement that the flatbuffer to be decoded has the schema embedded (the storage API uses an external schema message).  Need to either work around this via constructing/merging schema into a flatbuffer or go fix arrow upstream.

shollyman:
started contributions in apache arrow

alvarowolfx:
@shollyman looks like implementation for Decimal256 was recently added to Go ( see https://github.com/apache/arrow/commit/6d1bc624454b8a83e5c90410b51c2718059c394b ) and the Null data type also appears on the implementation status dashboard. I'll take a look later on the schema part that you mentioned here.

alvarowolfx:
Resolved by https://github.com/GoogleCloudPlatform/golang-samples/pull/2710

PR #2710, opened 2022-09-08
feat(bigquery): add arrow row format example

alvarowolfx:
Improve BigQuery quickstart by adding an example on how to use the Arrow format.


shollyman (approved):
Nice!  It looks like the work on arrow flight is where the necessary changes in the schema handling in arrow/ipc landed.  Thanks for taking a look at this!

alvarowolfx:
Resolves https://github.com/GoogleCloudPlatform/golang-samples/issues/1946

bigquery/bigquery_storage_quickstart/main.go
```diff
@@ -266,7 +314,49 @@ func processStream(ctx context.Context, client *bqStorage.BigQueryReadClient, st
...
+	r, err := ipc.NewReader(buf, ipc.WithAllocator(mem))
+	if err != nil {
+		return err
+	}
+	aschema := r.Schema()
+	for {
+		select {
+		case <-ctx.Done():
+			// Context was cancelled.  Stop.
+			return nil
```
codyoss, line 339: nit: This error should be propagated back to the user.

alvarowolfx: Does returning `ctx.Err()` make sense in this context ? 

```
case <-ctx.Done():
	// Context was cancelled.  Stop.
	return ctx.Err()
```

codyoss (approved):

PR #2713, opened 2022-09-14
feat: Add snippets for Spanner DML with returning clause

c2nes:
Samples are provided for INSERT, DELETE, and UPDATE in both GoogleSQL and PostgreSQL dialects. To provide a more compelling example for the INSERT case, a generated column has been added in the "create_database" example so that the generated value can be returned in the INSERT examples. This changed the number of mutations generated for inserts to the Artists table, and the commitStats test case was updated as a result.

### DIFF
--- a/spanner/spanner_snippets/spanner/pg_spanner_dml_delete_returning.go
+++ b/spanner/spanner_snippets/spanner/pg_spanner_dml_delete_returning.go
@@ -1,0 +1,72 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spanner
+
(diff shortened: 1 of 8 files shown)


spanner/spanner_snippets/spanner/pg_spanner_dml_delete_returning.go
```diff
@@ -0,0 +1,70 @@
...
+		defer iter.Stop()
+		for {
+			row, err := iter.Next()
+			if err == iterator.Done {
+				break
+			}
+			if err != nil {
+				return err
+			}
+			var singerID int64
```
enocom, line 57: nit:

``` golang
var (
  singerID int64
  fullName string
)
```

spanner/spanner_snippets/spanner/spanner_dml_delete_returning.go
```diff
@@ -0,0 +1,70 @@
...
+		defer iter.Stop()
+		for {
+			row, err := iter.Next()
+			if err == iterator.Done {
+				break
+			}
+			if err != nil {
+				return err
+			}
+			var singerID int64
```
enocom, line 57: Same nit as above.

enocom (approved):
One minor nit. Otherwise LGTM

spanner/spanner_snippets/spanner/pg_spanner_dml_delete_returning.go
c2nes: Done.

spanner/spanner_snippets/spanner/spanner_dml_delete_returning.go
c2nes: Done.

PR #2726, opened 2022-09-30
chore: update go version to 1.19

LukeSchlangen:


run/helloworld/go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/GoogleCloudPlatform/golang-samples/run/helloworld
 
-go 1.13
+go 1.19
```
codyoss, line 3: Is there motivation for this besides just to bump it up? This really only affects code being allow to use new language features. Also if this is going to get bumped we should be sure to bump the appropriate Dockerfile to match.

codyoss (approved):

run/helloworld/go.mod
codyoss: After talking offline this seems fine, but I think we should bump the dockerfile too stil.

LukeSchlangen: Sounds good! Added commit to bump the Dockerfile.

PR #2731, opened 2022-10-07
feat(bigquery): add sample to insert data with complex data types

alvarowolfx:
Towards internal b/155095509

### DIFF
--- a/bigquery/snippets/table/bigquery_inserting_data_types.go
+++ b/bigquery/snippets/table/bigquery_inserting_data_types.go
@@ -1,0 +1,106 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package table
+
+// [START bigquery_inserting_data_types]
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"cloud.google.com/go/bigquery"
+	"cloud.google.com/go/civil"
+)
+
+// ComplexType represents a complex row item
+type ComplexType struct {
+	Name         string                 `bigquery:"name"`
+	Age          int                    `bigquery:"age"`
+	School       []byte                 `bigquery:"school"`
+	Location     bigquery.NullGeography `bigquery:"location"`
+	Measurements []float64              `bigquery:"measurements"`
+	DatesTime    DatesTime              `bigquery:"datesTime"`
+}
+
+// DatesTime represents different date/time representation
+type DatesTime struct {
+	Day        civil.Date     `bigquery:"day"`
+	FirstTime  civil.DateTime `bigquery:"firstTime"`
+	SecondTime civil.Time     `bigquery:"secondTime"`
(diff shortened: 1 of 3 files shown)


bigquery/snippets/table/bigquery_inserting_data_types.go
```diff
@@ -0,0 +1,106 @@
...
+	// schema, err := bigquery.InferSchema(ComplexType{})
+
+	table := client.Dataset(datasetID).Table(tableID)
+	err = table.Create(ctx, &bigquery.TableMetadata{
+		Schema: schema,
+	})
+	if err != nil {
+		return fmt.Errorf("table.Create: %w", err)
+	}
+	day, _ := civil.ParseDate("2019-01-12")
```
shollyman, line 81: May want to leave a comment about eliding the error checking for these static values, generally we want to guide users towards good error handling processes.

bigquery/snippets/table/bigquery_inserting_data_types.go
```diff
@@ -0,0 +1,106 @@
...
+			Day:        day,
+			FirstTime:  firstTime,
+			SecondTime: secondTime,
+			ThirdTime:  thirdTime,
+		},
+	}
+	rows := []*ComplexType{row}
+	inserter := table.Inserter()
+	if err := inserter.Put(ctx, rows); err != nil {
+		return err
```
shollyman, line 101: It may be worth showing how to check for a [MultiError](https://pkg.go.dev/cloud.google.com/go/bigquery#MultiError) here, for users who are generating a mix of good and bad rows.

shollyman (approved):
called out two minor items for your consideration

bigquery/snippets/table/bigquery_inserting_data_types.go
alvarowolfx: I came up with two versions for showing error handling but no sure which one to use:

1. Try to insert a list of rows, with one being a bad one, prints the errors, but the function itself return error as nil because we are skipping invalid rows.
```
rows := []*ComplexType{row}
badRow := &ComplexType{
    Name: "John",
    Age:  24,
    // this row is missing required fields
}
rows = append(rows, badRow)

inserter := table.Inserter()
inserter.SkipInvalidRows = true
err = inserter.Put(ctx, rows)
if err != nil {
    if multiErr, ok := err.(bigquery.PutMultiError); ok {
        for _, putErr := range multiErr {
            fmt.Printf("failed to insert row %d with err: %v \n", putErr.RowIndex, putErr.Error())
        }
        return nil
    }
    return err
}
return nil
```

2.  Add a commented `badRow`, keep the default behavior of not skipping invalid rows and return an error if the MultiError is found:

```
rows := []*ComplexType{row}
/* Uncomment to simulate insert errors
badRow := &ComplexType{
    Name: "John",
    Age:  24,
    // this row is missing required fields
}
rows = append(rows, badRow)*/

inserter := table.Inserter()
err = inserter.Put(ctx, rows)
if err != nil {
    if multiErr, ok := err.(bigquery.PutMultiError); ok {
        for _, putErr := range multiErr {
            fmt.Printf("failed to insert row %d with err: %v \n", putErr.RowIndex, putErr.Error())
        }
    }
    return err
}
return nil
```

enocom: Both approaches look good to me. I have a slight preference for the second one, just because the sample works out of the box, and lets a user "break it" if they want to.

Also, nit: if you do go with the second approach use `//` style comments which are somewhat easier to remove in my opinion.

enocom (approved):

bigquery/snippets/table/bigquery_inserting_data_types.go
alvarowolfx: cool, I went with the second approach and just push a change to the comment style

PR #2736, opened 2022-10-10
feat(batch): Initial Batch sample

Shnatsel:
I'm opening this PR to get feedback on style early. I need to write 5 more samples for Batch, so I want to figure it out before I have to edit 5 of these for any changes.

This is meant to be a close translation of https://github.com/googleapis/python-batch/pull/22 and https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2775

I've experimented with initializing entire structs at once instead of using `new(SomeType)` and then assigning fields, but it caused rather unpleasant rightward drift:
https://github.com/GoogleCloudPlatform/golang-samples/blob/205b9e49a31d0594da19eda9b6077519d25445aa/batch/create_with_script_no_mounting.go

TODO:
 - [x] Write the rest of the samples once the style is settled
 - [x] Add tests


enocom:
@Shnatsel The rightward drift actually looks better to me. To handle the long lines, we can pull out child variables for things like `TaskGroup` and `AllocationPolicy`. Using `new` and building a structure piece by piece is actually less common in Go.

Would you mind updating with the suggestion above and we can polish this together?

Shnatsel:
I've split the megastructure construction into a bunch of smaller structures, still constructed by value instead of via field assignment, as you have suggested.

I think this is more readable than either of the previous layouts.

Is there anything else I should change, or am I good to go to I start working on the other 5 samples?

batch/create_with_container_no_mounting.go
```diff
@@ -0,0 +1,121 @@
...
+	ctx := context.Background()
+	batchClient, err := batch.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewClient: %v", err)
+	}
+	defer batchClient.Close()
+
+	container := &batchpb.Runnable_Container{
+		ImageUri:   "gcr.io/google-containers/busybox",
+		Commands:   []string{"-c", "echo Hello world! This is task ${BATCH_TASK_INDEX}. This job has a total of ${BATCH_TASK_COUNT} tasks."},
```
enocom, line 43: To avoid horizontal scroll let's shorten the command to something like:

```
"echo Task = ${BATCH_TASK_INDEX} Total = ${BATCH_TASK_COUNT}"
```

batch/create_with_container_no_mounting.go
```diff
@@ -0,0 +1,121 @@
...
+		Entrypoint: "/bin/sh",
+	}
+
+	// We can specify what resources are requested by each task.
+	resources := &batchpb.ComputeResource{
+		CpuMilli:  2000, // in milliseconds per cpu-second. This means the task requires 2 whole CPUs.
```
enocom, line 49: Let's move the comment above CpiMilli.

batch/create_with_container_no_mounting.go
```diff
@@ -0,0 +1,121 @@
...
+		MemoryMib: 16,
+	}
+
+	maxRunDuration := &durationpb.Duration{
```
enocom, line 53: nit: these two short declarations could just be inlined.

batch/create_with_container_no_mounting.go
enocom: Ditto below.

batch/create_with_container_no_mounting.go
enocom: Ditto below.

enocom (changes requested):
Snippets are looking good. We just need some tests now.

enocom:
@Shnatsel when you're ready for another pass, just re-request a review above.

Shnatsel:
@enocom I'm done adding samples, and I've covered all of them by test. I'd appreciate another look so I could merge this. Thanks!

enocom (approved):

PR #2739, opened 2022-10-13
functions: Update GCF Memorystore sample to use Functions Framework

ericschmidtatwork:
Updates the Memorystore sample for Cloud Functions to use the explicit/declarative Functions Framework syntax


muncus (approved):
Eric,
Thanks for this update, and apologies for the delay.
It looks like you need to run `go mod tidy`, but you should be good to go after that.

ericschmidtatwork:
@muncus Done, thanks!

PR #2743, opened 2022-10-20
feat(batch): Batch task samples

Shnatsel:
Adds Go samples for listing and getting Batch tasks, a sub-object of jobs.

Follow-up to #2736


batch/get_task.go
```diff
@@ -0,0 +1,56 @@
...
+
+	ctx := context.Background()
+	batchClient, err := batch.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewClient: %v", err)
+	}
+	defer batchClient.Close()
+
+	req := &batchpb.GetTaskRequest{
+		Name: fmt.Sprintf("projects/%s/locations/%s/jobs/%s/taskGroups/%s/tasks/%d", projectID, region, jobName, taskGroup, taskNumber),
```
enocom, line 43: Let's put the values starting with projectID on the next line to avoid the long line here.

batch/job_basics_test.go
```diff
@@ -55,6 +55,35 @@ func TestBatchJobCRUD(t *testing.T) {
 
 	buf.Reset()
 
+	// Tasks take a couple of seconds to be created on the server side.
+	// We're going to poll until they're created, or give up if the errors are persistent.
+	var attempts = 0
```
enocom, line 60: `var attempts int`

enocom (changes requested):
Two minor things.

batch/get_task.go
Shnatsel: I'm surprised `go fmt` doesn't do that automatically. Fixed, thanks!

enocom (approved):

PR #2750, opened 2022-11-04
feat(bigquery): add set client endpoint sample

alvarowolfx:
Towards internal b/254876447


shollyman (approved):
Thanks for putting this together!

It was a surprise to me that we need the api prefix and not just the hostname, so this is great that you test api interactions.

quartzmo (approved):
LGTM

PR #2766, opened 2022-11-16
feat(cloudsql): add MySQL Auto IAM AuthN sample

enocom:


enocom:
blocked on https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/309

cloudsql/mysql/database-sql/cloudsql.go
```diff
@@ -156,7 +156,16 @@ func mustConnect() *sql.DB {
 
 	// Use the connector when INSTANCE_CONNECTION_NAME (proj:region:instance) is defined.
 	if os.Getenv("INSTANCE_CONNECTION_NAME") != "" {
-		db, err = connectWithConnector()
+		if os.Getenv("DB_USER") == "" && os.Getenv("DB_IAM_USER") == "" {
+			log.Fatal("Warning: One of DB_USER or DB_IAM_USER must be defined")
```
telpirion, line 160: NIT: Should this log string be lower case?

cloudsql/mysql/database-sql/connect_connector_iam_authn.go
```diff
@@ -0,0 +1,72 @@
+// Copyright 2022 Google LLC
```
telpirion, line 1: Do we need a test for this sample?

telpirion (approved):

cloudsql/mysql/database-sql/cloudsql.go
enocom: Good question. I had to check this. It's just error messages that should start with a lowercase letter.

cloudsql/mysql/database-sql/connect_connector_iam_authn.go
enocom: This is just a minor variation of https://github.com/GoogleCloudPlatform/golang-samples/blob/main/cloudsql/mysql/database-sql/connect_connector.go where the only difference is the addition of a `cloudsqlconn.WithIAMAuthN()` option. For the connector sample, we have a [test in place](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/cloudsql/mysql/database-sql/cloudsql_test.go#L164). So I think we're good actually.

PR #2772, opened 2022-12-02
feat(compute): Disk snapshot samples

Shnatsel:


compute/snapshots/delete_snapshots_by_filter.go
```diff
@@ -0,0 +1,92 @@
...
+		if err != nil {
+			return err
+		}
+		delete(w, ctx, snapshotsClient, projectID, *snapshot.Name)
+	}
+	return nil
+}
+
+// snapshotList returns an iterator over disk snapshots in the given project
+func snapshotList(w io.Writer, ctx context.Context, client *compute.SnapshotsClient, projectID, filter string) (*compute.SnapshotIterator, error) {
```
telpirion, line 62: For a single region_tag / code snippet, we tend to keep all of the code inside one function. I recommend removing this function (and delete), and then adding this code inline in the `deleteByFilter()` method.

compute/snapshots/delete_snapshots_by_filter.go
```diff
@@ -0,0 +1,92 @@
...
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+)
+
+// Deletes ALL disk snapshots in the project that match the filter
+func deleteByFilter(w io.Writer, projectID, filter string) error {
+	// projectID := "your_project_id"
+
+	// Learn more about filters:
+	// https://cloud.google.com/python/docs/reference/compute/latest/google.cloud.compute_v1.types.ListSnapshotsRequest
+	// filter := ""
```
telpirion, line 34: NIT: Since this variable is passed in to the function and there's another developer-provided variable (`projectID`), I would move this comment to be right after `projectID`.

compute/snapshots/list_snapshots.go
```diff
@@ -0,0 +1,61 @@
...
+	"google.golang.org/api/iterator"
+	computepb "google.golang.org/genproto/googleapis/cloud/compute/v1"
+)
+
+// listSnapshots prints a list of disk snapshots in the given project
+func listSnapshots(w io.Writer, projectID, filter string) error {
+	// projectID := "your_project_id"
+	// Filter applied when listing snapshots. Learn more about filters:
+	// https://cloud.google.com/python/docs/reference/compute/latest/google.cloud.compute_v1.types.ListSnapshotsRequest
+	// filter := ""
```
telpirion, line 33: Same as above--I'd move this comment to be right after the other commented-out variable declaration.

compute/snapshots/snapshots_test.go
```diff
@@ -145,6 +145,26 @@ func TestComputeSnapshotsSnippets(t *testing.T) {
 		t.Errorf("createSnapshot got %q, want %q", got, want)
 	}
 
+	buf.Reset()
+	want = snapshotName
+
+	if err := listSnapshots(buf, tc.ProjectID, ""); err != nil {
```
telpirion, line 151: I think the typical practice elsewhere is 1 test case per region tag/sample.

telpirion (changes requested):

compute/snapshots/snapshots_test.go
Shnatsel: There is no single convention, it varies based on the needs of given tests. It would do more harm than good in this case.

There's a lot of initialization that needs to be done up front in this file (see lines 123-132), which creates a lot of boilerplate, which would have to be duplicated if I split this up into separate tests. Besides, some steps rely on previous steps to be complete (creation must come before listing, listing must come before deletion) so it wouldn't be "one testcase = one snippet" anyway.

Splitting up the test function would make sense if we were looking to parallelize the tests (I've done this for another product already), but these tests execute quickly enough that this is not a concern.

compute/snapshots/delete_snapshots_by_filter.go
Shnatsel: I have inlined the functions as you said and I think it's more confusing this way, even with comments. But as you wish.

compute/snapshots/snapshots_test.go
Shnatsel: Nevermind, I take that back. Done.

telpirion:
I'm still not a big fan of how this sample uses `t.Run()`, but I won't stop you from merging.

telpirion (approved):

Shnatsel:
What is a better way to use `t.Run()`?

Shnatsel:
I am basically following the testing pattern already laid out, but if there's a better way I'd be happy to use it for future PRs, and perhaps convert the existing tests to it later if it provides tangible benefits.

PR #2775, opened 2022-12-05
feat(asset): added saved queries sample code

weilin1995:


quartzmo:
@weilin1995 Can you rewrite your commit message (and rebase at the same time if you want) to pass the conventional commits check? Changing it to the current PR title should work.

asset/quickstart/update-saved-query/main.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2019 Google LLC
```
quartzmo, line 1: Should be 2022.

asset/quickstart/delete-saved-query/main.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	defer client.Close()
+ 
+	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
+	cloudresourcemanagerClient, err := cloudresourcemanager.NewService(ctx)
+	if err != nil {
+		log.Fatalf("cloudresourcemanager.NewService: %v", err)
+	}
+ 
+	project, err := cloudresourcemanagerClient.Projects.Get(projectID).Do()
```
quartzmo, line 53: Why not use the GAPIC `resourcemanager` client?

See example: https://github.com/googleapis/google-cloud-go/blob/main/resourcemanager/apiv3/projects_client_example_test.go#L19-L68

PR #2776, opened 2022-12-06
feat(spanner): add samples for Proto columns 

harshachinta:
This PR adds samples to perform insert, read and delete operations on Proto columns - Proto Messages and Proto Enum.


harshachinta:
@enocom @rahul2393 
Please review this PR as the initial PR for supporting samples for Proto Columns.

PR #2777, opened 2022-12-06
feat(spanner): add samples for Proto columns with null values

harshachinta:
This PR adds samples to perform insert and read operations on Proto columns - Proto Messages and Proto Enum that can take null values.


harshachinta:
Please review this PR after reviewing [PR-2776](https://github.com/GoogleCloudPlatform/golang-samples/pull/2776).

PR #2779, opened 2022-12-06
feat(kms): Add docs samples for KMS key import.

bdhess:


kms/kms_test.go
```diff
@@ -484,6 +484,68 @@ func TestIAMRemoveMember(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		if got, want := b.String(), "Created import job"; !strings.Contains(got, want) {
+			t.Errorf("createImportJob: expected %q to contain %q", got, want)
+		}
+	})
+	importJobName := fmt.Sprintf("%s/importJobs/%s", fixture.KeyRingName, importJobID)
+
+	t.Run("checkStateImportJob", func(t *testing.T) {
```
telpirion, line 516: Why is this being run in a separate goroutine?

kms/kms_test.go
```diff
@@ -484,6 +484,68 @@ func TestIAMRemoveMember(t *testing.T) {
...
+		var b bytes.Buffer
+		for !strings.Contains(b.String(), "ACTIVE") {
+			if err := checkStateImportJob(&b, importJobName); err != nil {
+				t.Fatal(err)
+			}
+		}
+	})
+
+	t.Run("importManuallyWrappedKey", func(t *testing.T) {
```
telpirion, line 525: Why is this being run in a separate goroutine?

bdhess: As above - is there some style guide rule that's violated here?

kms/kms_test.go
telpirion: If so, then please create a separate test case for each sample.

kms/kms_test.go
telpirion: No rule per se is being violated. However, embedding  calls to `t.Run()` creates multiple problems:

1. Adds unnecessary concurrency
2. Creates a harder-to-maintain test case

kms/kms_test.go
tritone: IMO using subtests is generally a fine pattern and not overly complex. However, the typical use for them is table-driven tests where cases are independent of one another. In this case, it seems like you are maintaining state between test cases, for example if `createKeyForImport` fails then I think later cases will fail/panic also because they don't have the necessary state, which then creates more noise to debug the output.

I think what would be less confusing would be either:
1. (better) Don't use subtests; write one test and use `t.Fatalf` to fail fast if a sample that creates resources fails.
2. (best) Separate out into separate tests which do their own setup/cleanup calling library code directly, rather than relying on setup/cleanup provided by other samples.

LMK if you have more questions about this.

bdhess: The samples executed in the (single) test are all snippets that are part of a complex multi-step process.  See https://cloud.google.com/kms/docs/importing-a-key which is already done in Python.  These are really only separate samples because we want to insert prose in between them for context in the doc page.

Splitting into a single test case per sample would just end up with a ton of duplicate code, because a test for sample n necessarily needs the results of samples[1:n-1].

bdhess: In any case, I removed the subtests and used approach 1 suggested by @tritone

tritone (approved):

tritone:
BTW, @bdhess, in the future if you could add separate commit(s) with review changes rather than amending your original commit, it makes it easier to see the diffs via github for reviewing. Thanks!

telpirion (approved):

PR #2780, opened 2022-12-07
feat(compute): add disk image samples

Shnatsel:
Adapted from [the canonical samples in Python](https://github.com/googleapis/python-compute/pull/270), and fixed a TOC/TOU issue in the creation samples by making the server process the `forceCreate` flag.

I am also discussing applying a similar change in Python with the author of the Python sample.


compute/disk-images/create_disk_image.go
```diff
@@ -0,0 +1,92 @@
...
+	// sourceDiskName := "your_disk_name"
+	// imageName := "my_image"
+	// // If storageLocations empty, automatically selects the closest one to the source
+	// storageLocations = []string{}
+	// // If forceCreate is set to `true`, proceeds even if the disk is attached to
+	// // a running instance. This may compromise integrity of the image!
+	// forceCreate = false
+
+	ctx := context.Background()
+	disksClient, err := compute.NewDisksRESTClient(ctx)
```
telpirion, line 46: Do you need the REST client? Or is there a gRPC client you can use?

compute/disk-images/delete_disk_image.go
```diff
@@ -0,0 +1,58 @@
...
+// Deletes the specified disk image
+func deleteDiskImage(
+	w io.Writer,
+	projectID, imageName string,
+) error {
+	// projectID := "your_project_id"
+	// imageName := "your_image"
+
+	ctx := context.Background()
+	imagesClient, err := compute.NewImagesRESTClient(ctx)
```
telpirion, line 36: Same as above--can you use a gRPC client here?

compute/disk-images/disk_image_test.go
```diff
@@ -0,0 +1,117 @@
...
+		rand.NewSource(time.Now().UnixNano()))
+	tc := testutil.SystemTest(t)
+	zone := "us-central1-a"
+	imageName := fmt.Sprintf("test-image-go-%v-%v", time.Now().Format("01-02-2006"), r.Int())
+	diskName := fmt.Sprintf("test-disk-go-%v-%v", time.Now().Format("01-02-2006"), r.Int())
+	sourceImage := "projects/debian-cloud/global/images/family/debian-11"
+
+	buf := &bytes.Buffer{}
+
+	err := createDisk(ctx, tc.ProjectID, zone, diskName, sourceImage)
```
telpirion, line 87: If you need to set up resources before running the test, create a `TestMain(m *testing.M)` method that creates the resources.

telpirion (changes requested):

compute/disk-images/create_disk_image.go
Shnatsel: There is no gRPC client for Compute APIs. Compute Engine APIs predate gRPC rollout in the Cloud.

This causes all sorts of issues and additional complexity because most of the tooling expects gRPC (documentation generation, etc) so we generate fake protos from REST APIs to interoperate with the tooling. But that's fortunately irrelevant here.

compute/disk-images/disk_image_test.go
Shnatsel: No, I am *testing the disk creation.* Running this inside the test is fully intentional.

telpirion (approved):

PR #2801, opened 2022-12-29
feat(compute): Regional disk samples

Shnatsel:
To go on this page: https://cloud.google.com/compute/docs/disks/regional-persistent-disk


compute/disks/disks_test.go
```diff
@@ -204,6 +255,41 @@ func TestComputeDisksSnippets(t *testing.T) {
 		}
 	})
 
+	t.Run("Create and resize a regional disk", func(t *testing.T) {
```
muncus, line 258: Do these tests need to be in the same function as all the others, or could this be done in a more focused (and maintainable!)  `TestRegionalDiskSamples()` method?

compute/disks/disks_test.go
```diff
@@ -263,66 +349,113 @@ func TestComputeDisksSnippets(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("deleteInstance got err: %v", err)
+	}
+
+	// disks attached to a VM instance must be deleted after deleting the instance
+
+	// disk 1 is set to auto-delete, no need to delete it explicitly
+
+	if err := deleteRegionalDisk(buf, tc.ProjectID, region, instanceDiskName2); err != nil {
+		t.Fatalf("deleteRegionalDisk got err: %v", err)
```
muncus, line 455: These 2 `t.Fatalf()` should be `t.Errorf()`, so that problems with disk2 deletion do not also leave disk3 lying around.

FYI, you could also use [`t.Cleanup`](https://pkg.go.dev/testing#T.Cleanup) to ensure these resources are cleaned up, as our minimum go version is 1.15.

Shnatsel: Good catch, thanks!

I didn't want to use t.Cleanup because it doesn't report errors correctly, making it difficult to understand what went wrong when it fails:

https://github.com/golang/go/issues/51395
https://github.com/golang/go/issues/29207

compute/disks/disks_test.go
muncus: That's fair. I think these tests would benefit from some refactoring, but that's beyond the scope of what you're doing here.

compute/disks/disks_test.go
muncus: that's a good reason to avoid using t.Cleanup. thanks for the context links!

muncus (approved):

PR #2802, opened 2022-12-29
feat(storagetransfer): add azure sample

JesseLovelace:
Adds a sample for using transfer service to transfer from an Azure container


storagetransfer/storagetransfer_test.go
```diff
@@ -89,6 +91,20 @@ func TestMain(m *testing.M) {
 		log.Fatalf("couldn't create S3 bucket: %v", err)
 	}
 
+	fmt.Println(os.Getenv("AZURE_CONNECTION_STRING"))
```
tritone, line 94: Remove print debugging

storagetransfer/storagetransfer_test.go
```diff
@@ -89,6 +91,20 @@ func TestMain(m *testing.M) {
...
+	connectionString := os.Getenv("AZURE_CONNECTION_STRING") +
+		";" + "AccountName=" + os.Getenv("AZURE_STORAGE_ACCOUNT")
+	azClient, err := azblob.NewClientFromConnectionString(connectionString, nil)
+	if err != nil {
+		log.Fatal("Invalid credentials with error: " + err.Error())
```
tritone, line 99: Maybe "couldn't create Azure client: "?

testing/kokoro/system_tests.sh
```diff
@@ -98,7 +98,13 @@ export GOLANG_SAMPLES_FIRESTORE_PROJECT=golang-samples-fire-0
 
 set -x
 
-source ${KOKORO_GFILE_DIR}/secret_manager/go-aws-secrets
+export STS_AWS_SECRET=`gcloud secrets versions access latest --project cloud-devrel-kokoro-resources --secret=go-storagetransfer-aws`
```
tritone, line 101: Why is this updated for AWS keys here as well?

storagetransfer/transfer_from_azure.go
```diff
@@ -0,0 +1,111 @@
...
+	"time"
+
+	storagetransfer "cloud.google.com/go/storagetransfer/apiv1"
+	"cloud.google.com/go/storagetransfer/apiv1/storagetransferpb"
+	"google.golang.org/genproto/googleapis/type/date"
+	"google.golang.org/genproto/googleapis/type/timeofday"
+)
+
+func transferFromAzure(w io.Writer, projectID string, azureStorageAccountName string, azureSourceContainer string, gcsSinkBucket string) (*storagetransferpb.TransferJob, error) {
+	// Your Google Cloud Project ID
```
tritone, line 32: nit: all full-line comments should end with a period.

storagetransfer/transfer_from_azure.go
```diff
@@ -0,0 +1,111 @@
...
+					Year:  int32(startTime.Year()),
+					Month: int32(startTime.Month()),
+					Day:   int32(startTime.Day()),
+				},
+				ScheduleEndDate: &date.Date{
+					Year:  int32(startTime.Year()),
+					Month: int32(startTime.Month()),
+					Day:   int32(startTime.Day()),
+				},
+				StartTimeOfDay: &timeofday.TimeOfDay{
```
tritone, line 88: Are all these fields required by the proto? Seems a bit clunky.

tritone (commented):
Couple minor things, otherwise LGTM

testing/kokoro/system_tests.sh
JesseLovelace: We actually updated the AWS creds recently, at the same time we created Azure creds.

Basically, before we used AWS creds which were owned by a particular Googler and were just attached to his gcard. Now, they're owned by the data transfer service team, and we're retiring those old creds

storagetransfer/transfer_from_azure.go
JesseLovelace: I can cut it

tritone: Ultimately between you and the TW team, I am fine either way. If you cut it might be good to add a comment that it will start immediately by default.

tritone (approved):

PR #2803, opened 2023-01-03
feat(compute): resize instance sample

Shnatsel:
For https://cloud.google.com/compute/docs/instances/changing-machine-type-of-stopped-instance


compute/change_machine_type.go
```diff
@@ -0,0 +1,111 @@
...
+			if v == str {
+				return true
+			}
+		}
+
+		return false
+	}
+
+	// Make sure that the machine is turned off
+	if !containsString([]string{"TERMINATED", "STOPPED"}, instance.GetStatus()) {
```
telpirion, line 68: I think there's an easier way to do this that makes it clearer what you're trying to do.

```
if !strings.Contains("STOPPED TERMINATED", instance.GetStatus()) {
      // ...
}

```

Shnatsel: Is this a Go idiom? It feels like a clever hack to me, and it took me a while to understand what's actually happening here.

This is copied from another sample too, so it may also be a good idea to keep them the same for consistency.

telpirion: Unfortunately Go doesn't have an easy way to determine whether an element is a member of a slice or array. One way, of course is to write a loop that determines membership (original PR code). Another way is to search for substrings.

My comment is merely a suggestion of an alternative. Feel free to accept or reject as you see fit.

Shnatsel: I see. I'm going to keep the current implementation for consistency with other samples, then.

telpirion (approved):

PR #2807, opened 2023-01-11
fix(testing): Work around kokoro image changes

muncus:
Recent changes to Kokoro have caused this initial `git` command to error, resulting in no tests being run.

While we're skipping the check here, it is still done in `testing/kokoro/system_test.sh`, so I do not anticipate any cost increase over out previous baseline (prior to the breakage).

### DIFF
--- a/testing/kokoro/trampoline.sh
+++ b/testing/kokoro/trampoline.sh
@@ -20,13 +20,13 @@ date
 
 cd github/golang-samples || exit 1
 
-SIGNIFICANT_CHANGES="$(git --no-pager diff --name-only main..HEAD | grep -Ev '(\.md$|^\.github)' || true)"
+# SIGNIFICANT_CHANGES="$(git --no-pager diff --name-only main..HEAD | grep -Ev '(\.md$|^\.github)' || true)"
 
-# If this is a PR with only insignificant changes, don't run any tests.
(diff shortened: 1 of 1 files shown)


codyoss:
Instead of commenting this out, can we instead try to mark this as a safe repo with something like `git config --global --add safe.directory ...`. I think this will allow us to run the commands if I understand correct

muncus:
I don't believe this is a safe.directory issue, because the error message is `not in a git repository` in the failing build, rather than the safe.directory message of `fatal: detected dubious ownership in repository`.
I tested safe.directory fix just to verify, and [got the same error](https://fusion2.corp.google.com/invocations/5ee8f835-c0ab-4884-9224-6e116c9ada18/targets/cloud-devrel%2Fgo%2Fgolang-samples%2Fpresubmit%2Fearliest-version/log)

codyoss (approved):

codyoss:
@muncus Hmm, kk thanks. Can you file an issue to undo this so we make sure this eventually gets restored?

muncus:
@codyoss filed #2811 to track.
The check is still, AFAICT being performed in the `system_test.sh` script that kokoro runs - is there a reason we need it in both places? Feel free to follow up on that bug with any other changes you'd like to see.

PR #2810, opened 2023-01-12
fix(testing): use safe.directory to mark our cloned repo as safe

muncus:


muncus:
this fix does not work. closing.

PR #2813, opened 2023-01-12
docs(storage): fix precondition comments

BrennaEpp:
- fix comments that reference "upload" to reflect the correct operation being performed
- fix comments that mention "generation" instead of "metageneration" where applicable


tritone (approved):
Thanks for the fix!

PR #2815, opened 2023-01-13
chore(functions/http): Update stateless sample to use gen2 signature

ddegraf:
Updated stateless sample using gen2 signature.

Tested:
- go test
- deploy + curl


tritone (approved):

muncus (approved):
LGTM, but the checks show you'll need a `go mod tidy` in here. That should kick off the kokoro tests, too (they look like they've not run for some reason).

PR #2816, opened 2023-01-13
chore(functions/http): Update xml sample to use gen2 signature

ddegraf:
Updated xml sample using gen2 signature.

Tested:
- go test
- deploy + curl


ace-n (approved):
LGTM for `functions` based on #2815 

(We still need a Golang LGTM; I'll ask around.)

muncus (approved):

PR #2834, opened 2023-01-24
docs(documentai): Add Initial Sample for Document AI

holtskinner:
<s> NOTE: The test project will need the following API enabled

- `documentai.googleapis.com`

And a Document OCR Processor will need to be created in the project

https://cloud.google.com/document-ai/docs/create-processor

And the Processor ID should replace the `aaaaaaaaa` in the `process_document_test.go` file </s>

EDIT: Not needed for quickstart only


documentai/snippets/quickstart.go
```diff
@@ -0,0 +1,73 @@
...
+
+	projectID := "PROJECT_ID"
+	locationID := "us"
+	processorID := "aaaaaaaa" // Create a Processor before running sample
+	filePath := "invoice.pdf"
+	mimeType := "application/pdf"
+
+	ctx := context.Background()
+
+	endpoint := fmt.Sprintf("%s-documentai.googleapis.com:443", locationID)
```
telpirion, line 41: Don't switch between `fmt` and `log` in the same sample. Pick one and stick with it.

telpirion (commented):
This sample needs a test.

documentai/snippets/quickstart.go
holtskinner: log is only used for error reporting. All of the other samples I looked at followed this pattern

holtskinner: nvm, found this in the contributing guide https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#return-errors

telpirion:
I believe that the issue in merging here is caused by a dependency error else where. #2853 will hopefully fix.

holtskinner:
Hmm, it seems like the build is still failing in the same spot even after the updates...

holtskinner:
Created new PR #2856

PR #2836, opened 2023-01-26
chore(cloudsql): split out Postgres Auto IAM AuthN sample

jackwotherspoon:
Splitting out Auto IAM AuthN sample for Go Connector into its own file.

### DIFF
--- a/cloudsql/postgres/database-sql/connect_connector.go
+++ b/cloudsql/postgres/database-sql/connect_connector.go
@@ -55,7 +55,6 @@ func connectWithConnector() (*sql.DB, error) {
 	if err != nil {
 		return nil, err
 	}
-	// [START cloud_sql_postgres_databasesql_auto_iam_authn]
 	var opts []cloudsqlconn.Option
 	if usePrivate != "" {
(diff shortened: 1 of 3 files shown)


cloudsql/postgres/database-sql/connect_connector.go
```diff
@@ -61,9 +57,6 @@ func connectWithConnector() (*sql.DB, error) {
 	}
 	// [START cloud_sql_postgres_databasesql_auto_iam_authn]
 	var opts []cloudsqlconn.Option
-	if dbIAMUser != "" {
```
enocom, line 64: One small nit: we should probably leave this in. Cut the docs over. And then remove this.

enocom (approved):

cloudsql/postgres/database-sql/connect_connector.go
jackwotherspoon: Removed the region tag from this file as it is currently not being rendered within the https://cloud.google.com/sql/docs/postgres/iam-logins page due to a bug showing the Java sample for Go.

PR #2851, opened 2023-02-03
feat(spanner): add samples for Drop Database Protection

harshachinta:
This PR adds samples and integration tests for Drop Database Protection feature.

### DIFF
--- a/spanner/spanner_snippets/spanner/integration_test.go
+++ b/spanner/spanner_snippets/spanner/integration_test.go
@@ -35,6 +35,7 @@ import (
 	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
 	"github.com/google/uuid"
 	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
 	adminpb "google.golang.org/genproto/googleapis/spanner/admin/database/v1"
 	instancepb "google.golang.org/genproto/googleapis/spanner/admin/instance/v1"
 	"google.golang.org/grpc/codes"
@@ -582,6 +583,35 @@ func TestCreateDatabaseWithDefaultLeaderSample(t *testing.T) {
 	assertContains(t, out, "The result of the query to get")
 }
 
+func TestDropDatabaseProtectionSample(t *testing.T) {
+	_ = testutil.SystemTest(t)
+	t.Parallel()
+
+	_, dbName, cleanup := initTest(t, randomID())
+	defer cleanup()
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
+	defer cancel()
+
+	var out string
+	mustRunSample(t, createDatabase, dbName, "failed to create a database")
+	out = runSampleWithContext(ctx, t, dropDatabaseProtection, dbName, "failed to enable drop database protection for a database")
+	assertContains(t, out, fmt.Sprintf("Enabled drop database protection to database [%s]\n", dbName))
+
+	databaseAdmin, err := database.NewDatabaseAdminClient(ctx, option.WithEndpoint("staging-wrenchworks.sandbox.googleapis.com:443"))
+	if err != nil {
+		log.Fatalf("cannot create databaseAdmin client: %v", err)
+	}
+
+	database, err := databaseAdmin.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: dbName})
+	if err != nil {
+		log.Fatalf("error during GetDatabase for db %v: %v", dbName, err)
+	}
+	if !database.GetEnableDropProtection() {
+		t.Errorf("got output %t; want it to contain %t", database.GetEnableDropProtection(), true)
+	}
+}
+
 func TestCustomInstanceConfigSample(t *testing.T) {
 	_ = testutil.SystemTest(t)
 	t.Parallel()
@@ -1089,11 +1119,11 @@ func createTestInstance(t *testing.T, projectID string, instanceConfigName strin
 	ctx := context.Background()
 	instanceID := fmt.Sprintf("go-sample-%s", uuid.New().String()[:16])
 	instanceName = fmt.Sprintf("projects/%s/instances/%s", projectID, instanceID)
-	instanceAdmin, err := instance.NewInstanceAdminClient(ctx)
+	instanceAdmin, err := instance.NewInstanceAdminClient(ctx, option.WithEndpoint("staging-wrenchworks.sandbox.googleapis.com:443"))
 	if err != nil {
 		t.Fatalf("failed to create InstanceAdminClient: %v", err)
 	}
-	databaseAdmin, err := database.NewDatabaseAdminClient(ctx)
+	databaseAdmin, err := database.NewDatabaseAdminClient(ctx, option.WithEndpoint("staging-wrenchworks.sandbox.googleapis.com:443"))
 	if err != nil {
 		t.Fatalf("failed to create DatabaseAdminClient: %v", err)
 	}
--- a/spanner/spanner_snippets/spanner/integration_test.go
+++ b/spanner/spanner_snippets/spanner/integration_test.go
@@ -598,6 +598,7 @@ func TestDropDatabaseProtectionSample(t *testing.T) {
 	out = runSampleWithContext(ctx, t, dropDatabaseProtection, dbName, "failed to enable drop database protection for a database")
 	assertContains(t, out, fmt.Sprintf("Enabled drop database protection to database [%s]\n", dbName))
 
+	// TODO(sriharshach): Remove endpoint
 	databaseAdmin, err := database.NewDatabaseAdminClient(ctx, option.WithEndpoint("staging-wrenchworks.sandbox.googleapis.com:443"))
 	if err != nil {
 		log.Fatalf("cannot create databaseAdmin client: %v", err)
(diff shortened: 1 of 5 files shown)


spanner/spanner_snippets/spanner/spanner_drop_database_protection.go
```diff
@@ -0,0 +1,62 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spanner
+
+// [START spanner_drop_database_protection]
```
rajatbhatta, line 17:
```suggestion
// [START spanner_update_database]
```

spanner/spanner_snippets/spanner/spanner_drop_database_protection.go
```diff
@@ -0,0 +1,62 @@
...
+		return err
+	}
+	if _, err := op.Wait(ctx); err != nil {
+		return err
+	}
+	fmt.Fprintf(w, "Enabled drop database protection to database [%s]\n", db)
+	return nil
+}
+
+// [END spanner_drop_database_protection]
```
rajatbhatta, line 62:
```suggestion
// [END spanner_update_database]
```

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -582,6 +583,36 @@ func TestCreateDatabaseWithDefaultLeaderSample(t *testing.T) {
 	assertContains(t, out, "The result of the query to get")
 }
 
+func TestDropDatabaseProtectionSample(t *testing.T) {
```
rajatbhatta, line 586:
```suggestion
func TestUpdateDatabaseSample(t *testing.T) {
```

spanner/spanner_snippets/spanner/spanner_drop_database_protection.go
harshachinta: Done

spanner/spanner_snippets/spanner/spanner_drop_database_protection.go
harshachinta: Done

spanner/spanner_snippets/spanner/integration_test.go
harshachinta: Done

spanner/spanner_snippets/spanner/spanner_drop_database_protection.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2023 Google LLC
```
rajatbhatta, line 1: Change the filename to spanner_update_database.go.

rajatbhatta:
LGTM apart from a requested change in file name and a general comment: Can we remove the TODOs and keep this PR clean (having only the changes which will be merged ultimately)? You can add a comment on the PR or mention as Note in the title saying that tests are expected to fail because the feature is not enabled in /span/cloud. We can have them locally of course, to test the changes.

spanner/spanner_snippets/spanner/spanner_drop_database_protection.go
harshachinta: Done

rajatbhatta:
@harshachinta: Can we rename the PR title, and description to indicate UpdateDatabase as a feature (not Drop Database Protection)? Also, letting the user know where we are in the program flow would be a good user experience. Can we add some logs for that? Similar to [link](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/spanner/spanner_snippets/spanner/spanner_copy_backup.go). You can refer to Node.js [PR](https://github.com/googleapis/nodejs-spanner/pull/1812/files#diff-fad67b2d8d064a52c464d9e2b0f558ed128ff2b5a7eb5c7e744db1ec9b1d602a).

rajatbhatta (approved):

harshachinta:
@telpirion Can you please review this PR on high priority?

spanner/spanner_snippets/spanner/spanner_update_database.go
```diff
@@ -0,0 +1,65 @@
...
+	"io"
+	"regexp"
+
+	database "cloud.google.com/go/spanner/admin/database/apiv1"
+	adminpb "cloud.google.com/go/spanner/admin/database/apiv1/databasepb"
+	"google.golang.org/genproto/protobuf/field_mask"
+)
+
+func updateDatabase(ctx context.Context, w io.Writer, db string) error {
+	matches := regexp.MustCompile("^(.*)/databases/(.*)$").FindStringSubmatch(db)
```
tritone, line 30: I'm not sure this check belongs in the code sample. Typical practice is to put an example of what the variable should be in the function body, e.g.:
```
// db := "projectID/databases/db-name"
```
(or whatever the expected format would be) so that the user can uncomment and sub their own.

spanner/spanner_snippets/spanner/spanner_update_database.go
```diff
@@ -0,0 +1,65 @@
...
+		Database: &adminpb.Database{
+			Name:                 db,
+			EnableDropProtection: true,
+		},
+		UpdateMask: &field_mask.FieldMask{
+			Paths: []string{"enable_drop_protection"},
+		},
+	})
+	if err != nil {
+		return err
```
tritone, line 53: It's useful to wrap each of the return error cases so it's clear where the error is coming from, e.g. `return fmt.Errorf("UpdateDatabase: %w", err)`

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -582,6 +583,53 @@ func TestCreateDatabaseWithDefaultLeaderSample(t *testing.T) {
...
 }
 
+func TestUpdateDatabaseSample(t *testing.T) {
+	_ = testutil.SystemTest(t)
+	t.Parallel()
+
+	_, dbName, cleanup := initTest(t, randomID())
+	defer cleanup()
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Hour)
```
tritone, line 593: This seems too long; do you expect the operation to actually take an hour?

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -582,6 +583,53 @@ func TestCreateDatabaseWithDefaultLeaderSample(t *testing.T) {
...
+	database, err := databaseAdmin.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: dbName})
+	if err != nil {
+		log.Fatalf("error during GetDatabase for db %v: %v", dbName, err)
+	}
+	// Verify if the EnableDropProtection field got enabled
+	if !database.GetEnableDropProtection() {
+		t.Errorf("got output %t; want it to contain %t", database.GetEnableDropProtection(), true)
+	}
+
+	// Disable Drop db protection for the cleanup to delete the databases
```
tritone, line 615: Perhaps the overall test cleanup should do this and retry in case of a transient failure?

spanner/spanner_snippets/spanner/integration_test.go
harshachinta: @tritone 
We have been using 60 minutes as a standard for all the existing tests, so following the same here also.
Ref: https://github.com/GoogleCloudPlatform/golang-samples/blob/main/spanner/spanner_snippets/spanner/integration_test.go#L136

spanner/spanner_snippets/spanner/spanner_update_database.go
harshachinta: Thanks for the suggestion @tritone. This would really help.

spanner/spanner_snippets/spanner/spanner_update_database.go
harshachinta: @tritone 
Sure, I followed the existing samples that had this check https://github.com/GoogleCloudPlatform/golang-samples/blob/main/spanner/spanner_snippets/spanner/spanner_create_database.go#L29.
I agree with you that putting an example is better in samples. Have made relevant change.

spanner/spanner_snippets/spanner/spanner_update_database.go
```diff
@@ -0,0 +1,60 @@
...
+	adminpb "cloud.google.com/go/spanner/admin/database/apiv1/databasepb"
+	"google.golang.org/genproto/protobuf/field_mask"
+)
+
+func updateDatabase(ctx context.Context, w io.Writer, db string) error {
+	// db = `projects/<project>/instances/<instance-id>/database/<database-id>`
+	// Instantiate database admin client.
+	adminClient, err := database.NewDatabaseAdminClient(ctx)
+	if err != nil {
+		return fmt.Errorf("updateDatabase.NewDatabaseAdminClient: %v", err)
```
tritone, line 33: nit: I'd reference the function including the import name rather than the caller function; e.g. `database.NewDatabaseAdminClient`. Same on L48 & L54 below

spanner/spanner_snippets/spanner/integration_test.go
tritone: Okay got it, no problem if this works for you but something to keep an eye on if you start getting test hangs/timeouts.

spanner/spanner_snippets/spanner/integration_test.go
tritone: Thanks for the explanation... two thoughts on this:
1. Maybe the logic to set `EnableDropProtection: false` should go in the cleanup function rather than the test itself?
1. I think it'd be a good idea to ensure that the call to `databaseAdmin.UpdateDatabase` is itself retried in cases of transient errors to make sure you don't end up with stranded resources in the test project.

Both of these are up to you if you want to implement, just suggestions.

tritone (approved):
Couple minor nits, otherwise LGTM

spanner/spanner_snippets/spanner/spanner_update_database.go
harshachinta: That is a good suggestion @tritone.
But, I feel printing `updateDatabase.NewDatabaseAdminClient` will help us quickly understand which sample is throwing error when the tests run. 
`database.NewDatabaseAdminClient` could be used in multiple samples and we might need to go through the entire stack trace to understand which sample is the error occurring in.

Also this has been followed for the few existing samples https://github.com/GoogleCloudPlatform/golang-samples/blob/main/spanner/spanner_snippets/spanner/spanner_create_database_with_default_leader.go#L70

harshachinta:
@tritone 
I have made the suggested changes and all the build pass as expected.
Can you please help me with merging the PR if there are no other questions?

rahul2393 (approved):

rahul2393 (approved):

PR #2809, opened 2023-01-12
feature: allow Automatic IAM Authentication without a password

LukeSchlangen:

PR #2862, opened 2023-02-14
docs: add comments to explain environment variables

LukeSchlangen:
Creating a new Pull Request based on conversation in: https://github.com/GoogleCloudPlatform/golang-samples/pull/2809

This PR adds clearer descriptions around the impact of environment variable in `deployment.yaml` and how that determines how the code will run.

### DIFF
--- a/cloudsql/mysql/database-sql/cloudsql.go
+++ b/cloudsql/mysql/database-sql/cloudsql.go
@@ -156,15 +156,11 @@ func mustConnect() *sql.DB {
 
 	// Use the connector when INSTANCE_CONNECTION_NAME (proj:region:instance) is defined.
 	if os.Getenv("INSTANCE_CONNECTION_NAME") != "" {
-		if os.Getenv("DB_USER") == "" && os.Getenv("DB_IAM_USER") == "" {
-			log.Fatal("Warning: One of DB_USER or DB_IAM_USER must be defined")
-		}
-		// DB_IAM_USER takes precedence if both environment variables are
-		// defined.
-		if os.Getenv("DB_IAM_USER") != "" {
-			db, err = connectWithConnectorIAMAuthN()
-		} else {
+		// Use IAM Authentication (recommended) if there is no password
+		if os.Getenv("password") != "" {
 			db, err = connectWithConnector()
+		} else {
+			db, err = connectWithConnectorIAMAuthN()
 		}
 		if err != nil {
 			log.Fatalf("connectConnector: unable to connect: %s", err)
--- a/cloudsql/mysql/database-sql/cloudsql.go
+++ b/cloudsql/mysql/database-sql/cloudsql.go
@@ -156,8 +156,8 @@ func mustConnect() *sql.DB {
 
 	// Use the connector when INSTANCE_CONNECTION_NAME (proj:region:instance) is defined.
 	if os.Getenv("INSTANCE_CONNECTION_NAME") != "" {
-		// Use IAM Authentication (recommended) if there is no password
-		if os.Getenv("password") != "" {
+		// Use IAM Authentication (recommended) if password is not set
+		if os.Getenv("DB_PASS") != "" {
 			db, err = connectWithConnector()
 		} else {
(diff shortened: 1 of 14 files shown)


cloudsql/postgres/database-sql/cloudsql.go
```diff
@@ -156,15 +156,11 @@ func mustConnect() *sql.DB {
...
-		if os.Getenv("DB_USER") == "" && os.Getenv("DB_IAM_USER") == "" {
-			log.Fatal("Warning: One of DB_USER or DB_IAM_USER must be defined")
-		}
-		// DB_IAM_USER takes precedence if both environment variables are
-		// defined.
-		if os.Getenv("DB_IAM_USER") != "" {
-			db, err = connectWithConnectorIAMAuthN()
-		} else {
+		// Use IAM Authentication (recommended) if there is no password
+		if os.Getenv("password") != "" {
```
jackwotherspoon, line 160:
```suggestion
		// Use IAM Authentication (recommended) if password is not set
		if os.Getenv("DB_PASS") != "" {
```

jackwotherspoon (changes requested):
The env variable for password is `DB_PASS` so i think we should be checking if it is set instead of `password`?

telpirion (approved):

cloudsql/mysql/database-sql/cloudsql.go
```diff
@@ -156,15 +156,11 @@ func mustConnect() *sql.DB {
...
-		if os.Getenv("DB_USER") == "" && os.Getenv("DB_IAM_USER") == "" {
-			log.Fatal("Warning: One of DB_USER or DB_IAM_USER must be defined")
-		}
-		// DB_IAM_USER takes precedence if both environment variables are
-		// defined.
-		if os.Getenv("DB_IAM_USER") != "" {
-			db, err = connectWithConnectorIAMAuthN()
-		} else {
+		// Use IAM Authentication (recommended) if password is not set
+		if os.Getenv("DB_PASS") != "" {
```
enocom, line 160: I worry a bit about users forgetting to set a password and then getting an error about IAM Authn. It's the difference between an implicit and explicit API.

LukeSchlangen: I think that's a fair perspective. The reason I changed it was because I had the opposite experience getting started. I thought I needed a password even when using IAM Authentication.

When I was trying to get this repository working with IAM, I had to read through the code to realize that I needed to change the environment variable. I also spent a while thinking, "Even though this is an IAM user, maybe I still need to include the database instance password?"

That seems totally backwards now, but it took me a while to figure out.

cloudsql/postgres/database-sql/connect_connector_iam_authn.go
```diff
@@ -41,7 +41,7 @@ func connectWithConnectorIAMAuthN() (*sql.DB, error) {
 	// Cloud Secret Manager (https://cloud.google.com/secret-manager) to help
 	// keep secrets safe.
 	var (
-		dbUser                 = mustGetenv("DB_IAM_USER")              // e.g. 'sa-name@project-id.iam'
+		dbUser                 = mustGetenv("DB_IAM_ USER")             // e.g. 'sa-name@project-id.iam'
```
jackwotherspoon, line 44:
```suggestion
		dbUser                 = mustGetenv("DB_IAM_USER")             // e.g. 'sa-name@project-id.iam'
```

jackwotherspoon (changes requested):

jackwotherspoon (approved):
Thanks Luke! This looks great 😄

cloudsql/mysql/database-sql/cloudsql.go
LukeSchlangen: Per your recommendation, I added it back.

I was able to update `deployment.yaml` to the point where I think I'm happy with the result, and I think having a separate `DB_IAM_USER` allows us to clearly split out which environment variable controls IAM and which variable controls the method of connection. Hope it looks good!

PR #2873, opened 2023-02-21
feat(appengine): Add rest of Flex preview samples

jahjahbin:
Same idea as #2868  and #2869 . The pattern is to copy the existing GAE Flex samples to a new folder, update the license years, add a tiny bit more config in each `app.yaml`, and introduce `go.mod` and `go.sum` files.

I figured since the pattern is the same among these PRs, it would be less tedious if I batched the last five sample updates into one PR. Let me know if you still prefer splitting this one up.


jahjahbin:
Is it OK to not fix these particular failures for `snippet-bot` and `Go / Root tests`? The former seems irrelevant bc it's not a region tag, and I don't see a clear reason from the test why a JSON file is "bad". The format in the json file looks normal to me...

telpirion:
It looks like there are several different samples in this PR. Would you please break this into 2 or 3 separate PRs? ~500 lines per PR is ideal.

jahjahbin:
> It looks like there are several different samples in this PR. Would you please break this into 2 or 3 separate PRs? ~500 lines per PR is ideal.

I can do this. I wonder if I can get a simpler git diff, which would make this PR simpler.

PR #2889, opened 2023-03-02
fix: tidy up some vet failures seen in PR #2887

muncus:
Thanks to @enocom for bringing these to our attention.


telpirion (approved):

telpirion:
It looks like these tests are now failing. Maybe we should move the individual test cases into their own methods rather than using `t.Run()`?

jahjahbin:
If I understand correctly, these test failures are blocking *everything* from being merged.

I don't think there's any point in leaving this in the current state. What's the best way to unblock us right now?

enocom:
@muncus I think we can just do this: https://github.com/GoogleCloudPlatform/golang-samples/pull/2887/commits/6dd8f513b74466f8679f8a75eda68c51cac81334

muncus:
I've adopted @enocom 's solution for the vet fix, and found several tests broken by the fix. These tests will be skipped (see #2897 )
I'll continue this fix over in #2899 to get everything else unblocked asap.

PR #2890, opened 2023-03-02
feat(appengine): Move new helloworld sample out of preview

jahjahbin:
Goals of this PR:
1) Move new `helloworld` sample from preview to the main folder.
2) Move old `helloworld` sample into a new `legacy` folder -- we want to preserve this sample for folks who don't want to use Go 1.18 and above.
3) Explain the difference in a readme. I'm open to suggestions on the wording there.

GCP's docs already point to the `appengine_flexible/helloworld` folder, so they will immediately point to the new sample once this is in place.


telpirion (approved):

telpirion (approved):

jahjahbin:
So, after all this work, our team ended up not wanting to do the change this way. I'll send a different PR to re-attempt this. Sorry for the confusion.

PR #2898, opened 2023-03-08
fix(dataflow) Fix go build command in instruction

Abacn:
Fix the go build command in the instruction of dataflow go flex-template sample.

The original command results in error `malformed import path "-o": leading dash`


telpirion (approved):
Thank you, @Abacn , for contributing!

PR #2906, opened 2023-03-10
feat(appengine): Copy existing non-preview samples

jahjahbin:
This is a quick PR to copy all the existing samples (except previews) to a new staging directory. It should make our process easier for updating public docs and pointing to both old and new samples. No changes to the code/config.


appengine_flexible/go115_and_earlier/go.mod
```diff
@@ -0,0 +1,37 @@
+module github.com/GoogleCloudPlatform/golang-samples/appengine_flexible
```
tritone, line 1: I'm not sure if it would work to keep this module path the same if the copy is still present at the original path.

appengine_flexible/go115_and_earlier/pubsub/sample_message.json
```diff
@@ -0,0 +1,5 @@
+{
```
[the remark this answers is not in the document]
jahjahbin: Fixed. I wasn't aware of badfiles_test.go and the allowlist there.

appengine_flexible/go115_and_earlier/go.mod
jahjahbin: That's fair. I've updated the module path now. In my brief testing, it seemed like I always ended up deleting the `go.mod` files in this and the parent directory, to get `gcloud app deploy` to work (gopath mode).

tritone (approved):

PR #2908, opened 2023-03-12
feat(appengine): Migrate helloworld flex to new path

jahjahbin:
Take 2. Let's do the following:

1) Update existing helloworld Flex sample to use new style for specifying higher versions.
2) Delete preview helloworld sample, not needed anymore.


appengine_flexible/helloworld/app.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
```
tritone, line 1: Copyright dates should not be updated on existing files.

tritone (approved):

appengine_flexible/helloworld/app.yaml
jahjahbin: Undid the copyright updates.

jahjahbin:
We do have another internal change we'd like to make before this goes in -- I'll followup this week.

(it's not the end of the world if this gets merged before then, to be honest...)

jahjahbin:
OK, fast followup :) This is ready to merge. We can remove the label.

PR #2913, opened 2023-03-16
chore: cleanup un-wanted region tags

msampathkumar:
Old region tags are replaced with product specific region tags. Please check b/272509882 for more details.

New Region tags were added using https://github.com/GoogleCloudPlatform/golang-samples/pull/2904


msampathkumar:
Update: cl/517138087 is recently finished.

tritone (approved):

PR #2914, opened 2023-03-16
chore(functions/functionsv2/tips/infinite_retries): New cloudevent sample from background functions

kenneth-rosario:


functions/functionsv2/tips/infinite_retries/infinite_retries.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	// Add your message processing logic.
+	return processTheMessage(msg)
+}
+
+// [END functions_cloudevent_tips_infinite_retries]
+
+// processTheMessage is a stand-in for the domain logic of your function.
+func processTheMessage(m MessagePublishedData) error {
```
telpirion, line 69: question: I like what you've done here. It makes sense to have some message processing placeholder here that throws a not implemented error.

However, I'm not sure that it conforms to Google Cloud samples style. As written, the code isn't copy-paste-runnable.
https://googlecloudplatform.github.io/samples-style-guide/#introduction

I would instead put some sort of simple placeholder message here rather than throwing an error.

kenneth-rosario: This is a gen 2 version of https://github.com/GoogleCloudPlatform/golang-samples/blob/main/functions/tips/infinite_retries/infinite_retries.go . Should we leave it as close as possible to the original sample or should we Diverge? Let me know what you think

telpirion: Diverging from the original is perfectly fine. It's evolution!

telpirion (approved):

functions/functionsv2/tips/infinite_retries/infinite_retries.go
kenneth-rosario: Sounds good will follow up with this suggestion in a seperate PR once all samples are updated.

PR #2927, opened 2023-03-22
feat(firestore): adds COUNT aggregate query sample

telpirion:


firestore/aggregate_query_count.go
```diff
@@ -0,0 +1,58 @@
...
+	client, err := firestore.NewClient(ctx, projectID)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	collection := client.Collection("users")
+	query := collection.Where("born", ">", 1850)
+
+	// `alias` provides a key for accessing the aggregate query results
```
tritone, line 41: Is alias "all" in this case? I'm a little confused about what alias values are allowed based on the snippet plus reading the firestore pkg.go.dev docs; however if the dev docs clarify things then fine to leave as is.

firestore/aggregate_query_count_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"bytes"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestCreateCountQuery(t *testing.T) {
+	projectID := os.Getenv("GOLANG_SAMPLES_FIRESTORE_PROJECT")
+	var bytes bytes.Buffer
+	err := createCountQuery(&bytes, projectID)
```
tritone, line 27: Is this snippet dependent upon a collection already existing, or can it run on an empty project? If there are dependencies would be good to note.

tritone (approved):
two minor questions, otherwise lgtm

firestore/aggregate_query_count.go
telpirion: Good question. I've added some additional clarifying comments.

firestore/aggregate_query_count_test.go
telpirion: It is dependent on a query already existing. We have special projects just for Firestore (because you can't have Datastore & Firestore in the same project).

I've added a clarifying note.

PR #2928, opened 2023-03-24
chore(functions/functionsv2/firebase/hellofirestore): new firestore trigger sample for gcf gen2

kenneth-rosario:


functions/functionsv2/firebase/hellofirestore/hellofirestore.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// HelloFirestore is triggered by a change to a Firestore document.
+func HelloFirestore(ctx context.Context, event event.Event) error {
+	dataAny := &anypb.Any{}
+	if err := proto.Unmarshal(event.Data(), dataAny); err != nil {
+		log.Fatal("proto.Unmarshal: ", err)
+	}
+
+	data := &firestoredata.DocumentEventData{}
```
telpirion, line 43: quibble: I think you need only declare this using `var`.

```
var data &firestoredata.DocumentEventData
```

functions/functionsv2/firebase/hellofirestore/hellofirestore_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// create a CloudEvent with the Firestore document event data
+	ce := event.New("1.0")
+	ce.SetSource("test-source")
+	ce.SetType("google.cloud.firestore.document.v1.created")
+	ce.SetDataContentType("application/protobuf")
+	ce.SetData(ce.DataContentType(), dataBytes)
+
+	// capture the output of fmt.Println
+	oldStdout := os.Stdout
```
telpirion, line 59: question: We typically pass an `io.Writer` as an argument to the sample function. I suspect that this is not possible given that the function satisfy the argument type for `functions.CloudEvent()`. Is that correct?

telpirion (approved):

functions/functionsv2/firebase/hellofirestore/hellofirestore.go
kenneth-rosario: Not sure if that is valid. I could do
```
var data *firestoredata....
```
But then it would not be initialized If I am not mistaken.

kenneth-rosario:
@telpirion I will need help merging pull request (if everything looks good) since I do not have write access to the repository.

kenneth-rosario:
Converting to draft. Some changes are needed.

functions/functionsv2/firebase/hellofirestore/hellofirestore.go
telpirion: It initializes to the nil value, which is the same as what you've initialized it to here.

kenneth-rosario:
Converting back to ready for review. The use of Any protobuf is no longer needed.

muncus (approved):

PR #2939, opened 2023-03-30
chore(functions/functionsv2/firebase/firebase/upper): New cloudevent sample for firestore native triggers on gen2

kenneth-rosario:


muncus (approved):
minor comments below.

PR #2944, opened 2023-04-03
feat(bigquery): add example of writing to default stream

shollyman:


bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
```diff
@@ -0,0 +1,199 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+// generateExampleMessages generates a slice of serialized protobuf messages using a statically defined
+// and compiled protocol buffer file, and returns the binary serialized representation.
+func generateExampleDefaultMessages(numMessages int) ([][]byte, error) {
+	msgs := make([][]byte, numMessages)
+	for i := 0; i < numMessages; i++ {
+
+		random := rand.New(rand.NewSource(time.Now().UnixNano()))
```
telpirion, line 38: nit: This line is a bit smooshed together. You might consider putting in a line break here to make it more readable.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
```diff
@@ -0,0 +1,199 @@
...
+
+		// Our example data embeds an array of structs, so we'll construct that first.
+		sList := make([]*exampleproto.SampleStruct, 5)
```
telpirion, line 41: comment: `sList` doesn't tell me enough about what this variable does. I recommend either using a more evocative name--or using something shorter if it's just a momentary container of sorts.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
```diff
@@ -0,0 +1,199 @@
...
+	"cloud.google.com/go/bigquery/storage/managedwriter"
+	"cloud.google.com/go/bigquery/storage/managedwriter/adapt"
+	"github.com/GoogleCloudPlatform/golang-samples/bigquery/snippets/managedwriter/exampleproto"
+	"google.golang.org/protobuf/proto"
+)
+
+// generateExampleMessages generates a slice of serialized protobuf messages using a statically defined
+// and compiled protocol buffer file, and returns the binary serialized representation.
+func generateExampleDefaultMessages(numMessages int) ([][]byte, error) {
+	msgs := make([][]byte, numMessages)
```
telpirion, line 35: nit: I don't think you need to create an array here. You can create a variable-length slice and append to it later on. You don't even need to use `make()`!

```
var msgs [][]byte

// ... stuff happens
msgs.append(msgs, b)
```

Totally up to you.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+
+		m := &exampleproto.SampleData{
+			BoolCol:    proto.Bool(true),
+			BytesCol:   []byte("some bytes"),
+			Float64Col: proto.Float64(3.14),
+			Int64Col:   proto.Int64(123),
+			StringCol:  proto.String("example string value"),
+
+			// These types require special encoding/formatting to transmit.
```
telpirion, line 55: praise: I really appreciate these comments. I like how you explain why you're using each data type and why they require special consideration.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
```diff
@@ -0,0 +1,199 @@
...
+			// TIME also uses literal format.
+			TimeCol: proto.String("12:13:14.000000"),
+
+			// TIMESTAMP uses microseconds since Unix epoch.
+			TimestampCol: proto.Int64(time.Now().UnixNano() / 1000),
+
+			// Int64List is an array of INT64 types.
+			Int64List: []int64{2, 4, 6, 8},
+
+			// This is a required field, and thus must be present.
```
telpirion, line 81: question: Is this required by BigQuery?

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
telpirion: Okay, now I see where it's used. I would rename this as `structList` or `sl`: either go much smaller or much bigger.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
```diff
@@ -0,0 +1,199 @@
...
+func appendToDefaultStream(w io.Writer, projectID, datasetID, tableID string) error {
+	// projectID := "myproject"
+	// datasetID := "mydataset"
+	// tableID := "mytable"
+
+	ctx := context.Background()
+	// Instantiate a managedwriter client to handle interactions with the service.
+	client, err := managedwriter.NewClient(ctx, projectID)
+	if err != nil {
+		return fmt.Errorf("managedwriter.NewClient: %w", err)
```
telpirion, line 113: praise: thank you for using `%w`!!

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	// Close the client when we exit the function.
+	defer client.Close()
+
+	// We need to communicate the descriptor of the protocol buffer message we're using, which
+	// is analagous to the "schema" for the message.  Both SampleData and SampleStruct are
+	// two distinct messages in the compiled proto file, so we'll use adapt.NormalizeDescriptor
+	// to unify them into a single self-contained descriptor representation.
+	m := &exampleproto.SampleData{}
```
telpirion, line 122: nit: you can also just declare this variable with `var`, which has the effect of initializing it to its nil value, for example:

```
var m *exampleproto.SampleData
```

bigquery/snippets/managedwriter/integration_test.go
```diff
@@ -84,8 +84,16 @@ func TestPendingStream(t *testing.T) {
 		t.Fatalf("failed to create destination table(%q %q): %v", testDatasetID, testTableID, err)
 	}
 
-	if err := appendToPendingStream(ioutil.Discard, tc.ProjectID, testDatasetID, testTableID); err != nil {
-		t.Errorf("appendToPendingStream(%q %q): %v", testDatasetID, testTableID, err)
-	}
+	t.Run("PendingStream", func(t *testing.T) {
```
telpirion, line 87: comment: I'd like to move away from having nested  calls to `t.Run()` inside of a test case. That said, we don't yet have an official set of guidance around this, so I leave it up to you.

telpirion (approved):
I really like the descriptive explanations in the comments for this sample. I found that understanding the code was SO much easier with them.

I only have a few quibbles and questions. Feel free to accept/ack as you see fit.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
shollyman: went with sl

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
shollyman: split across lines

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
shollyman: It's a required field of the schema (aka nulls not allowed).

shollyman: augmenting the comment to try to make that clearer.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
shollyman: its a minor difference, but in this case we know exactly the size we need so we avoid unneeded allocations by specifying explicitly.

bigquery/snippets/managedwriter/integration_test.go
shollyman: The value here is that we share a common setup (reduces test time), but can more accurately pinpoint the error based on the generated test name, so the scope of failure is clearer if we do encounter failures.

bigquery/snippets/managedwriter/bigquerystorage_append_rows_default.go
shollyman: switched

PR #2947, opened 2023-04-04
testing: remove IoT tests

shollyman:
IoT core is deprecated.  This PR removes testing, but leaves the samples in place for the time being.  We can remove the remaining assets in Q4 after full turndown.


shollyman:
punting on this, closing the PR

PR #3019, opened 2023-05-08
chore(appengine): move region tags to include the full function

iennae:
assessed https://cloud.google.com/appengine/docs/legacy/standard/go111/capabilities/ and it doesn't appear to require additional changes to the documentation there.

## Description

Fixes b/281116380

## Checklist


tritone (approved):
I'm not sure this really adds much to the sample given that it still doesn't include imports and is not testable. We would also want to direct users to newer App Engine runtimes than Go 1.11. However, feel free to move the region tag if you like.

tritone (approved):
I'm not sure this really adds much to the sample given that it still doesn't include imports and is not testable. We would also want to direct users to newer App Engine runtimes than Go 1.11. However, feel free to move the region tag if you like.

PR #3044, opened 2023-05-15
chore(fixit): Errors update for Cloud Run grpc-server-streaming

grayside:
Update Cloud Run grpc-server-streaming sample for wrapped errors


grayside:
Looking closer, the %v in this change is not associated with an error, but an intentional way the sample is logging timestamp data. Changing that here is out of scope.

PR #3047, opened 2023-05-15
fix: functions error visibility

kweinmeister:
Fixit: Go Error Visibility for functions/*


kweinmeister:
rsamborski@ already has these assigned

PR #3068, opened 2023-05-16
feat(fixit): Adding privateca/ca_pool samples

rsamborski:
## Description

Fixes b/281691685, b/281692014

## Checklist
- [x] These samples need a new **API enabled** in testing projects to pass (let us know which ones)
- [X] This sample adds a new sample directory, and I updated the [CODEOWNERS file](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/.github/CODEOWNERS) with the codeowners for this sample
- [X] This sample adds a new **Product API**, and I updated the [Blunderbuss issue/PR auto-assigner](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/.github/blunderbuss.yml) with the codeowners for this sample


privateca/ca_pool_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+func TestCaPoolTests(t *testing.T) {
+	var r *rand.Rand = rand.New(
+		rand.NewSource(time.Now().UnixNano()))
+	tc := testutil.SystemTest(t)
+	projectId := tc.ProjectID
+	location := "us-central1"
+	caPoolId := fmt.Sprintf("test-ca-pool-%v-%v", time.Now().Format("2006-01-02"), r.Int())
+
+	buf := &bytes.Buffer{}
```
telpirion, line 35: style: we typically just declare an empty buffer like so:

```
var buf *bytes.Buffer
```

privateca/ca_pool_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if err := createCaPool(buf, projectId, location, caPoolId); err != nil {
+		t.Errorf("createCaPool got err: %v", err)
```
telpirion, line 38: issue:  replace `t.Errorf` with `t.Fatal`. If this test has failed, the rest of them will too. Also just print the error--there's no need to format it.

privateca/ca_pool_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	expectedResult := "CA Pool created"
+	if got := buf.String(); !strings.Contains(got, expectedResult) {
+		t.Errorf("createCaPool got %q, want %q", got, expectedResult)
+	}
+
+	buf.Reset()
+
+	if err := deleteCaPool(buf, projectId, location, caPoolId); err != nil {
```
telpirion, line 48: issue: create a separate test case for `deleteCaPool()`. We should have a 1-to-1 mapping of test cases to sample methods.

privateca/ca_pool_test.go
```diff
@@ -0,0 +1,56 @@
...
+		t.Errorf("deleteCaPool got err: %v", err)
```
telpirion, line 49: Same as above--replace `t.Errorf()` with `t.Fatal()`. If there is an err when the sample is called, the rest of sample will fail too.

privateca/ca_pool_test.go
telpirion: Also, using samples to teardown other samples often backfires. I would recommend using the API directly in a dedicated teardown function.

privateca/create_ca_pool.go
```diff
@@ -0,0 +1,69 @@
...
+		return fmt.Errorf("NewCertificateAuthorityClient creation failed: %w", err)
+	}
+	defer caClient.Close()
+
+	caPool := &privatecapb.CaPool{
+		// Set the tier (see: https://cloud.google.com/certificate-authority-service/docs/tiers).
+		Tier: privatecapb.CaPool_ENTERPRISE,
+	}
+
+	locationPath := "projects/" + projectID + "/locations/" + location
```
telpirion, line 46: nit: use `fmt.Sprintf()` instead here.

```
locationPath := fmt.Sprintf("projects/%s/locations/%s", projectID, location)
```

privateca/create_ca_pool.go
```diff
@@ -0,0 +1,69 @@
...
+
+	req := &privatecapb.CreateCaPoolRequest{
+		// See https://pkg.go.dev/cloud.google.com/go/security/privateca/apiv1/privatecapb#CreateCaPoolRequest.
```
telpirion, line 49: Move comment above this line.

privateca/delete_ca_pool.go
```diff
@@ -0,0 +1,62 @@
...
+	// caPoolId := "ca-pool-id"		// A unique id/name for the ca pool.
+
+	ctx := context.Background()
+	caClient, err := privateca.NewCertificateAuthorityClient(ctx)
+	if err != nil {
+		return fmt.Errorf("NewCertificateAuthorityClient creation failed: %w", err)
+	}
+	defer caClient.Close()
+
+	fullCaPoolName := "projects/" + projectID + "/locations/" + location + "/caPools/" + caPoolId
```
telpirion, line 41: See previous--use `fmt.Sprintf()` here instead.

privateca/delete_ca_pool.go
```diff
@@ -0,0 +1,62 @@
...
+
+	req := &privatecapb.DeleteCaPoolRequest{
+		// See https://pkg.go.dev/cloud.google.com/go/security/privateca/apiv1/privatecapb#CreateCaPoolRequest.
```
telpirion, line 44: Move comment above line of code. Is this supposed to link to the `CreateCaPoolRequest` message?

rsamborski: Good catch!

PR #3070, opened 2023-05-16
refactor(firestore): Split firestore/save.go samples to separate file, widen region tags

grayside:
## Description

As advertised in the title, this change splits the **17 samples** in `firestore/save.go` into separate files. This enables narrowly scoped import statements per sample and keeps region tags simple. As part of this change, `save.go` has been removed, but `save_test.go` has been retained: this can be refactored in the future, but for now minimizes scope.

## Per sample notes in this change

note: Samples prefixed with ⚠️ will not be copy-paste runnable because required database setup is not shown in docs. Fixing that requires establishing a convention on how database setup should be shown in samples. I will add brief notes on that below but it's outside scope of this PR.

* **firestore_data_batch_writes**: Noted - Using deprecated firestore `client.Batch`
* **firestore_data_set_id_random_document_ref**: included previously hidden hard-coded data for operation. Unclear docs impact but seems safe.
* **firestore_data_delete_field**: Using `START_EXCLUDE` to hide TODO comment
* ⚠️ **firestore_data_set_field**: Using `START_EXCLUDE` to hide database setup. This was an unclear call because the code isn't actually runnable without seeding the database, but seeding the database feels like it may distract from the purpose of the code sample. Probably more docs coordination is needed. Fixed error declaration to use preErr from other samples, fixing existing runnability bug.
* ⚠️ **firestore_data_set_nested_fields**: Moved region tags, but this move doesn't change that the sample is confusing: initialData map is initialized within the region tag, but only used in hidden code. It seems like maybe the sample should be split or more heavily commented around initial database setup so the context is more clear. As written, the included code will not be runnable regardless of this region tag move.
* **firestore_data_set_server_timestamp**: Exclude database setup code and add comment about why
* ⚠️ **firestore_transaction_document_update:** Moved region tags, excluded database setup, added comment. The exclude prevents this sample from being copy-paste runnable.
* **fs_update_multiple_fields**: Added region tag `firestore_update_multiple_fields`, keeping the previous tag in case of sample browser or other use cases.

## Handling Database Seeding: Proposal for future discussion

Firestore samples need a consistent way to showcase database seeding to enable samples to be copy-paste runnable, and also to ensure developer attention in the sample is directed to what they are trying to learn.

I suggest this is done by having a setup function that is called from the top of the sample, and included in the region tag below the sample. That way the context is available to the reader, and to run the code, but is not in the way.

## Additional Metadata

## Checklist


firestore/save_data_batch_writes.go
```diff
@@ -0,0 +1,56 @@
...
+// [START firestore_data_batch_writes]
+
+import (
+	"context"
+	"log"
+
+	"cloud.google.com/go/firestore"
+)
+
+func batchWrite(ctx context.Context, client *firestore.Client) error {
```
telpirion, line 26: issue: I think that we'll want to construct `Context` objects and the `Client` inside of the snippet. For the client, in particular, we want to demonstrate how to close the client.

grayside: I'm unwilling to expand scope of this PR to include net new code that isn't focused on ensuring code inside region tags will compile.

I'm happy to undo the region tag expansion, we can keep the sample move in place so later change to region tags won't require coordinating docs update. Would that be preferrable to leaving this as-is?

telpirion: No, what you have is fine. I think it's fair to separate refactoring the snippet from a simple move & region tag expansion. Thanks!

grayside:
I temporarily added back the do not merge label while evaluating snippetbot's comment, and I've concluded we are safe to proceed: sample browser can handle moves.

telpirion:
Once all the other statuses are green, I'll force-merge. Thank you @grayside !

telpirion (approved):

PR #3125, opened 2023-05-19
chore(simplelog): refactor samples to be standalone

muncus:
separate the region tags into their own files, so they stand alone as samples

this allows for the inclusion of imports in each sample, to align with the style
guide.

note: these region tags are currently frozen in the docs, so these changes are safe.


telpirion (approved):
I have no notes 😄

PR #3126, opened 2023-05-19
feat(iam): add samples for set_policy and get_policy

yucentao:
## Description

Fixes b/281691650 b/281691651

## Checklist


iam/snippets/policy_get.go
```diff
@@ -0,0 +1,50 @@
...
+	"io"
+
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+// getPolicy gets IAM policy for a project.
+func getPolicy(w io.Writer, projectID string) (*cloudresourcemanager.Policy, error) {
+	ctx := context.Background()
+
+	crmService, err := cloudresourcemanager.NewService(ctx)
```
telpirion, line 30: issue: I would leave a comment about why you don't need to close the `Service` object like a client.

iam/snippets/policy_test.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		t.Fatalf("getPolicy(%q) failed: %v", tc.ProjectID, err)
+	}
+	want := "bindings"
+	if !strings.Contains(buf.String(), want) {
+		t.Errorf("getPolicy got %q, want to contain %q", buf.String(), want)
+	}
+
+	buf.Reset()
+	err = setPolicy(buf, tc.ProjectID, policy)
```
telpirion, line 38: issue: don't test two samples in one test case. Make a separate test case for `setPolicy()`.

iam/snippets/policy_set.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+// setPolicy sets IAM policy for a project.
+func setPolicy(w io.Writer, projectID string, policy *cloudresourcemanager.Policy) error {
```
telpirion, line 27: issue: construct a `Policy` object here so that devs can see how it's done.

telpirion (changes requested):

iam/snippets/policy_set.go
yucentao: Why make it different from all other languages https://cloud.google.com/iam/docs/samples/iam-set-policy? From the fixit instruction, the goal is to add consistent samples.

iam/snippets/policy_test.go
yucentao: Can you suggest what the separate test case looks like? This is a real e2e test, which always need to call getPolicy before setPolicy to not break IAM settings of the project.

iam/snippets/policy_get.go
yucentao: I don't see it necessary explain an non-exist method. It just makes the code more complicated and confuses users.

iam/snippets/policy_set.go
```diff
@@ -0,0 +1,52 @@
...
+
+package snippets
+
+// [START iam_set_policy]
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"google.golang.org/api/cloudresourcemanager/v1"
```
rsamborski, line 23: @Sita04 PTAL and let us know if this is the correct library or if we should use the GAPIC one (https://github.com/googleapis/google-cloud-go/tree/main/resourcemanager).

iam/snippets/policy_get.go
telpirion: Okay, that's fair.

iam/snippets/policy_set.go
telpirion: Let's try to improve the samples gradually over time. The sample here is to set a Policy, which implies that we should show how to create a Policy message.

Also see the Sample Guidelines:
https://googlecloudplatform.github.io/samples-style-guide/#minimal-arguments

iam/snippets/policy_test.go
telpirion: You can use a `TestMain()` to set up your resources. Since these are set & get methods, the set up & teardown should be pretty easy.

telpirion (changes requested):

PR #3129, opened 2023-05-19
chore(spanner): update region tags

Sita04:
## Description

See internal links

Update sample region tags to match other languages.


telpirion:
Please don't publish links to internal resources 😄

telpirion (approved):

PR #3143, opened 2023-05-25
feat(speech): add transcription snippets

Sita04:
## Description

Add 3 snippets
1. Transcribe Model Selection GCS
2. Transcribe Diarization GCS
3. Transcribe Diarization


speech/snippets/transcribe_diarization.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return fmt.Errorf("error running recognize %v", err)
+	}
+
+	response, err := operation.Wait(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Speaker Tags are only included in the last result object, which has only one alternative.
```
telpirion, line 75: nit (non-blocking): consider inserting a newline after "one."

speech/snippets/transcribe_diarization.go
```diff
@@ -0,0 +1,99 @@
...
+	alternative := response.Results[len(response.Results)-1].Alternatives[0]
+
+	wordInfo := alternative.GetWords()[0]
+	currentSpeakerTag := wordInfo.GetSpeakerTag()
+
+	var speakerWords strings.Builder
+
+	speakerWords.WriteString(fmt.Sprintf("Speaker %d: %s", wordInfo.GetSpeakerTag(), wordInfo.GetWord()))
+
+	for i := 1; i < len(alternative.Words); i++ {
```
telpirion, line 85: question: did you intend to start on the second alternative?

If not, you can tighten this code up just a bit:
```
for _, word := range alternative.Words {
    ...
}

```

Sita04:
Go/Root test fails because of bad mime type. But the local audio file is required for testing.
Kindly refer to equivalent resource files in Java here: https://github.com/GoogleCloudPlatform/java-docs-samples/tree/6161d0f7e1881dcab71cf2275126f32fed21b224/speech/resources

speech/snippets/transcribe_diarization.go
```diff
@@ -0,0 +1,99 @@
...
+		wordInfo := alternative.Words[i]
+		if currentSpeakerTag == wordInfo.GetSpeakerTag() {
+			speakerWords.WriteString(" ")
+			speakerWords.WriteString(wordInfo.GetWord())
+		} else {
+			speakerWords.WriteString(fmt.Sprintf("\nSpeaker %d: %s", wordInfo.GetSpeakerTag(), wordInfo.GetWord()))
```
telpirion, line 91: nit: insert a new line after the format string.

speech/snippets/transcribe_diarization_beta.go
```diff
@@ -0,0 +1,104 @@
...
+	wordInfo := alternative.GetWords()[0]
+	currentSpeakerTag := wordInfo.GetSpeakerTag()
+
+	var speakerWords strings.Builder
+
+	speakerWords.WriteString(fmt.Sprintf("Speaker %d: %s", wordInfo.GetSpeakerTag(), wordInfo.GetWord()))
+
+	// For each word, get all the words associated with one speaker, once the speaker changes,
+	// add a new line with the new speaker and their spoken words.
+	for i := 1; i < len(alternative.Words); i++ {
```
telpirion, line 90: See previous about starting on the second alternative.

speech/snippets/transcribe_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Errorf("transcribe_diarization got %q, expected %q", got, "Speaker")
+	}
+}
+
+func TestTranscribeDiarizationGcsBeta(t *testing.T) {
+	testutil.SystemTest(t)
+
+	var buf bytes.Buffer
+	if err := transcribe_diarization_gcs_beta(&buf, gcsDiarizationAudioPath); err != nil {
+		t.Errorf("error in transcribe diarization gcs %v", err)
```
telpirion, line 65: issue: replace call to `t.Errorf` with a call to `t.Fatalf`.

speech/snippets/transcribe_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Errorf("transcribe_model_selection_gcs got %q, expected %q", got, "Transcript:")
+	}
+}
+
+func TestTranscribeDiarizationBeta(t *testing.T) {
+	testutil.SystemTest(t)
+
+	var buf bytes.Buffer
+	if err := transcribe_diarization(&buf, recognitionAudioFile); err != nil {
+		t.Errorf("error in transcribe diarization %v", err)
```
telpirion, line 53: issue: replace call to `t.Errorf` with a call to `t.Fatalf`.

speech/snippets/transcribe_diarization.go
Sita04: This is intended.. since we already obtain the zeroth alternative before the loop.
I used [Java sample](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/6161d0f7e1881dcab71cf2275126f32fed21b224/speech/src/main/java/com/example/speech/RecognizeBeta.java#L142-L205) as the baseline for this.

I believe the loop was constructed this way because the Speaker number in the transcript starts from 1. Please lmk if you'd like me to modify this.

speech/snippets/transcribe_diarization_beta.go
Sita04: ditto. Same comment as above.

telpirion (approved):

PR #3144, opened 2023-05-25
fix(dialogflow): migrate existing go sample from internal reference

iennae:
current sample that is missing from GitHub.

see https://cloud.google.com/dialogflow/es/docs/fulfillment-webhook#webhook-go

updates region tag

needs to be updated with functions gen 2, but need to have a placeholder file in GH. I recognize this is missing a test. will add the test in the next commit.


iennae:
Copyright was missing from the original file, but based on the history it's from 2020 so I think the copyright year should be 2020.

telpirion:
We tend to set the copyright to the year that the sample was added to the repo. That said, the header check isn't required for this sample--we can go with whatever year you think is appropriate.

PR #3182, opened 2023-06-26
feat(dlp): added a sample for dlp reidentify free text with fpe using surrogate

AarshD-crest:
## Description

- Added a sample for [dlp reidentify free text with fpe using surrogate](https://cloud.google.com/dlp/docs/samples/dlp-reidentify-free-text-with-fpe-using-surrogate) with a unit test case.

## Checklist


annadata (approved):

dlp/snippets/deid/deid_test.go
```diff
@@ -347,6 +347,37 @@ func TestDeIdentifyDeterministic(t *testing.T) {
...
+
+	inputStr := "My phone number is 1234567890"
+	infoType := "PHONE_NUMBER"
+	surrogateType := "PHONE_TOKEN"
+	unwrappedKey, err := getUnwrappedKey(t)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := deidentifyFreeTextWithFPEUsingSurrogate(&buf, tc.ProjectID, inputStr, infoType, surrogateType, unwrappedKey); err != nil {
```
telpirion, line 362: nit: it's preferred that you do not exercise more than one sample in a test case. If you need this sample method call to deid the test data, I recommend that you call the service directly.

telpirion (changes requested):

dlp/snippets/deid/deid_test.go
AarshD-crest: Thank You @telpirion for your review, totally got your point, and it is perfectly valid, but I have a suggestion since this sample is using the unwrapped key and it's required to use the same key in the re-identification sample that is used in the de-identification sample, so that part becomes a bit tricky, so to handle that What I have thought that Should we combine the test case of deid and reid both the sample? Could you please share some thoughts on that.

annadata: @telpirion 
The re-identify sample requires text/string that has de-identified data. If the same wrapped key is used for both operations, the asserts can check for expected original value.  "123..90" in this case.
If we use only re-identify, we can only assert that output != input.
Please let us know if you feel we should decouple them nevertheless.

telpirion: Yes, I think they should be decoupled. If the `deidentifyFreeTextWithFPEUsingSurrogate()` sample begins failing, then this test fails. Someone will need to investigate WHICH of the two samples are failing--that's an extra step of diagnostics. It's much better to have this test hermetically independent of any other sample.

AarshD-crest: @telpirion Thank you for your comment. I agree with your point. But from the development side, I will have to include de-identifying of the string before re-identifying; with that said I will have to include the code to de-identify in my test file as well. Which in a way is similar to calling both the samples in a test case. The only difference would be we will have two test cases with this approach: a) One for testing de-identify sample code and b) the Other for testing re-identify sample code along with caller to de-identify function written in the test file. Let me know if that works, I will update the PR. Thank you!

AarshD-crest: Another option is, I can hard code the unwrapped key as well as a de-identified string which will be used as input to re-identify sample. Let me know which approach would work.

telpirion: Since this is a test, you might as well hard-code the unwrapped key as a de-identified string.

AarshD-crest: Updated it with hard-coded values.

telpirion (approved):

PR #3190, opened 2023-06-28
feat(pubsub): add cloud storage subscription sample

hongalex:


pubsub/subscriptions/create_cloud_storage_subscription.go
```diff
@@ -0,0 +1,60 @@
...
+	defer client.Close()
+
+	sub, err := client.CreateSubscription(ctx, subID, pubsub.SubscriptionConfig{
+		Topic: topic,
+		CloudStorageConfig: pubsub.CloudStorageConfig{
+			Bucket:         bucket,
+			FilenamePrefix: "log_events_",
+			FilenameSuffix: ".avro",
+			OutputFormat:   &pubsub.CloudStorageOutputFormatAvroConfig{WriteMetadata: true},
+			MaxDuration:    5 * time.Minute,
```
lahuang4, line 48: Nit: could you change MaxDuration to 1 minute? 5 minutes will be the default, so I wanted to suggest a different value. Also, I suspect that many folks will just copy from your samples :) and a shorter max duration will result in smoother processing for many workloads.

pubsub/subscriptions/subscription_test.go
```diff
@@ -746,7 +747,7 @@ func TestCreateWithFilter(t *testing.T) {
 	}
 }
 
-func TestCreateBigQuerySubscription(t *testing.T) {
+func TestBigQuerySubscription(t *testing.T) {
```
lahuang4, line 750: Nit: why change this test name? CreateBigQuerySubscription seems to accurately describe what it's doing.

pubsub/subscriptions/subscription_test.go
```diff
@@ -779,6 +780,36 @@ func TestCreateBigQuerySubscription(t *testing.T) {
 	}
 }
 
+func TestCreateCloudStorageSubscription(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	tc := testutil.SystemTest(t)
+	client := setup(t)
+	defer client.Close()
+	storageSubID := subID + "-storage"
```
lahuang4, line 789: s/storage/cloud-storage (just for clarity)

lahuang4 (approved):

pubsub/subscriptions/subscription_test.go
hongalex: I think I wanted to make this test do more than "create" but I'll revert for now

pubsub/subscriptions/create_cloud_storage_subscription.go
hongalex: Thanks for the context, changed

pubsub/subscriptions/create_cloud_storage_subscription.go
```diff
@@ -0,0 +1,60 @@
...
+	"time"
+
+	"cloud.google.com/go/pubsub"
+)
+
+// createCloudStorageSubscription creates a Pub/Sub subscription that exports messages to Cloud Storage.
+func createCloudStorageSubscription(w io.Writer, projectID, subID string, topic *pubsub.Topic, bucket string) error {
+	// projectID := "my-project-id"
+	// subID := "my-sub"
+	// topic of type https://godoc.org/cloud.google.com/go/pubsub#Topic
```
tritone, line 31: Passing in a complex type like this seems confusing. Could you supply a topic ID instead and construct or create it in the sample?

pubsub/subscriptions/create_cloud_storage_subscription.go
```diff
@@ -0,0 +1,60 @@
...
+	// note bucket should not have the gs:// prefix
```
tritone, line 32: Don't think it's necessary to state this (or it can be done in the docs surrounding the sample block).

pubsub/subscriptions/subscription_test.go
```diff
@@ -934,3 +965,23 @@ func deleteBigQueryDataset(projectID, datasetID string) error {
 	}
 	return nil
 }
+
+func createOrGetStorageBucket(projectID, bucketID string) error {
```
tritone, line 969: Consider using `testutil.CreateTestBucket` which handles cleanup as well.

pubsub/subscriptions/subscription_test.go
```diff
@@ -779,6 +780,36 @@ func TestCreateBigQuerySubscription(t *testing.T) {
...
+	defer client.Close()
+	storageSubID := subID + "-cloud-storage"
+
+	topic, err := getOrCreateTopic(ctx, client, topicID)
+	if err != nil {
+		t.Fatalf("CreateTopic: %v", err)
+	}
+	var buf bytes.Buffer
+
+	// Use the same bucket across test instances. This
```
tritone, line 797: IMO for storage tests we prefer to create/cleanup buckets in each test run; it's not expensive and having dependency on shared resources can lead to flakiness if someone accidentally deletes something.

tritone (commented):
Overall looks good, a few comments.

pubsub/subscriptions/create_cloud_storage_subscription.go
hongalex: We could, though this is the current pattern for [other subscription samples](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/pubsub/subscriptions/create.go). I'm inclined to keep it consistent and revisit this once we transition to GAPIC client and redoing the samples then.

pubsub/subscriptions/create_cloud_storage_subscription.go
hongalex: I was trying to capture the docs from [the proto](https://github.com/googleapis/googleapis/blob/64d54ff78cfe7cdc2f4b95717dc9afa3ef32a29a/google/pubsub/v1/pubsub.proto#L1057), but I think you're right that it's a bit redundant since we already give the example.

pubsub/subscriptions/subscription_test.go
hongalex: We used to do this with BigQuery subscriptions (using bq datasets/tables), but sometimes when the test invocations crash, cleanup doesn't happen properly and it leads to orphaned resources. This would require us to have a cleanup method that runs prior to sample invocation to delete these resources. I was proposing this change since our samples doesn't really depend on the resource other than it existing, and we guarantee it exists by calling `createOrGet` prior to using it.

pubsub/subscriptions/create_cloud_storage_subscription.go
tritone: Yeah I just mention it because it's not something we generally note in the GCS docs; `gs://` is more of a gsutil construct and not used in clients or API calls elsewhere at all.

pubsub/subscriptions/subscription_test.go
tritone: There is also `testutil.DeleteExpiredBuckets` which can be used to handle this automatically, but ultimately up to you

PR #3206, opened 2023-07-12
feat: Add go.work file

muncus:
Add a `go.work` file that lists all (current) modules in the repo.

This is an experiment to reduce unnecessary testing when modifying a single
module/package.


muncus:
I'm going to close this, since I do not think this is costing us much beyond slightly messy go.mod files.

PR #3207, opened 2023-07-13
fix(eventarc): update e2e tests to send proper payload

muncus:
the service this tests is now using the CloudEvent SDK, which does more
payload validation, and caused the failures in #3188


eventarc/testing/audit_storage.e2e_test.go
```diff
@@ -49,6 +66,8 @@ func TestAuditStorageSinkService(t *testing.T) {
 	fmt.Printf("client.Do: %s %s\n", req.Method, req.URL)
 
 	if got := resp.StatusCode; got != http.StatusOK {
+		b, _ := ioutil.ReadAll(resp.Body)
+		fmt.Println(b)
```
grayside, line 70: suggestion: Rather than just print b, should we add a test case around the response body?

grayside (approved):
LGTM, left a non-blocking suggestion about test depth.

eventarc/testing/audit_storage.e2e_test.go
muncus: I added this to give us a better message than "want 200 got 400" in the case of errors.
There is a test for body content in `../audit_storage/main_test.go`, which i think is sufficient.

PR #3227, opened 2023-07-24
docs(samples): Add batch mode sample and test

irataxy:
b:288111390

Please merge on approval.


media/transcoder/create_job_from_preset_batch_mode.go
```diff
@@ -0,0 +1,67 @@
...
+// createJobFromPresetBatchMode creates a batch mode job based on a given preset
+// template. See
+// https://cloud.google.com/transcoder/docs/how-to/create-batch-jobs
+// for more information.
+func createJobFromPresetBatchMode(w io.Writer, projectID string, location string, inputURI string, outputURI string, preset string) error {
+	// projectID := "my-project-id"
+	// location := "us-central1"
+	// inputURI := "gs://my-bucket/my-video-file"
+	// outputURI := "gs://my-bucket/my-output-folder/"
+	// preset := "preset/web-hd"
```
tritone, line 36: Preset doesn't seem necessary as an arg given that the test seems to set it to the same thing. Can you remove it and use a local var in the snippet instead?

media/transcoder/transcoder_test.go
```diff
@@ -99,6 +100,8 @@ func TestJobTemplatesAndJobs(t *testing.T) {
 	t.Logf("\nwriteTestGCSFiles() completed\n")
 	testJobFromPreset(t, projectNumber, inputURI, outputURIForPreset)
 	t.Logf("\ntestJobFromPreset() completed\n")
+	testJobFromPresetBatchMode(t, projectNumber, inputURI, outputURIForPresetBatchMode)
+	t.Logf("\ntestJobFromPresetBatchMode() completed\n")
```
tritone, line 104: Don't have to change it in this PR but it would probably be nicer to use subtests or fully separated tests rather than these test functions.

media/transcoder/transcoder_test.go
```diff
@@ -324,6 +327,80 @@ func testJobFromPreset(t *testing.T, projectNumber string, inputURI string, outp
...
+			r.Errorf("getJob got err: %v", err)
+		}
+		if got := buf.String(); !strings.Contains(got, jobName) {
+			r.Errorf("getJob got\n----\n%v\n----\nWant to contain:\n----\n%v\n----\n", got, jobName)
+		}
+	})
+	buf.Reset()
+
+	// Get the job state, which should be succeeded. If the job is still running on the last attempt, pass the test.
+	testutil.Retry(t, 3, 30*time.Second, func(r *testutil.R) {
```
tritone, line 365: Are these timings confirmed with the API team that the job is expected to complete within these bounds?

tritone (commented):
Overall looks good, a few minor things and questions.

media/transcoder/transcoder_test.go
irataxy: These test jobs were intentionally made to complete quickly and have been running for about 2 years now.

tritone: Sounds good, just wanted to check.

media/transcoder/create_job_from_preset_batch_mode.go
irataxy: Done

media/transcoder/transcoder_test.go
irataxy: I looked into refactoring this, but the `testutil/storage_cleaner.go` functions take a parameter of `t *testing.T`. As far as I can tell, this parameter isn't available in a `TestMain(m *testing.M)` function. I need to do some GCS bucket setup in `TestMain()`.

tritone (approved):
Discussed test refactoring/cleanup offline; this is not required and could happen separately.

PR #3240, opened 2023-08-04
docs(samples): Add assets and pools samples and tests

irataxy:
b:264616218
Please **merge** this PR for me once it is approved


media/livestream/livestream_test.go
```diff
@@ -60,6 +64,12 @@ func TestLiveStream(t *testing.T) {
 
 	testChannelEvents(t, outputURI)
 	t.Logf("\ntestChannelEvents() completed\n")
+
+	testAssets(t, assetURI)
```
muncus, line 68: These tests should probably be in their own test function (e.g. `TestAssets`), if they do not depend on the previous tests in this method. same for `testPools()`.

This will make them run faster, and ensure that the Asset tests still run if there are problems with other livestream tests.

media/livestream/livestream_test.go
```diff
@@ -28,13 +28,16 @@ const (
...
 	deleteInputResponse        = "Deleted input"
+	deleteAssetResponse        = "Deleted asset"
 	startChannelResponse       = "Started channel"
 	stopChannelResponse        = "Stopped channel"
 	location                   = "us-central1"
 	inputID                    = "my-go-test-input"
 	backupInputID              = "my-go-test-backup-input"
 	channelID                  = "my-go-test-channel"
 	eventID                    = "my-go-test-channel-event"
+	assetID                    = "my-go-test-asset"
```
muncus, line 39: please consider using unique identifiers, to avoid resource contention between multiple concurrent test runs. This could be:

```
var assetID = fmt.Sprintf("my-go-test-asset-%s", uuid.New().String())
```

there's no need to change the existing patterns, but it will produce better outcomes, and it means there's no need to delete the existing objects *before* testing.

(except for pool, which you have documented only allows one per location. Thanks for that!)

media/livestream/livestream_test.go
irataxy: Done

media/livestream/livestream_test.go
irataxy: Done

muncus (approved):

PR #3251, opened 2023-08-18
feat(run): add filename to run/testing log output

muncus:
- switch cloudrunci to use 'log' instead of 'fmt.Print*'
- feat(run/testing): include filename in log prefix


run/testing/markdown_preview_renderer.e2e_test.go
```diff
@@ -52,31 +58,35 @@ func TestRendererService(t *testing.T) {
...
-		}
+			if got := resp.StatusCode; got != http.StatusOK {
+				r.Errorf("response status: got %d, want %d", got, http.StatusOK)
+			}
 
-		out, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			t.Fatalf("ioutil.ReadAll: %v", err)
-		}
+			out, err := ioutil.ReadAll(resp.Body)
```
telpirion, line 82: nit: I see that current Go style is to use `io` rather than `ioutil`.

See:
https://github.com/golang/go/issues/42026

telpirion (approved):

muncus:
This was an effort to make some of the logs more readable, but it did not prove to be as helpful as i imagined. Closing.

PR #3266, opened 2023-08-25
feat(aiplatform): adds create dataset sample for Vertex

telpirion:
This PR adds the region tag `aiplatform_create_dataset`.

### DIFF
--- a/aiplatform/snippets/create_dataset_test.go
+++ b/aiplatform/snippets/create_dataset_test.go
@@ -1,0 +1,49 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


aiplatform/snippets/create_dataset_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	m.Run()
+
+	deleteDataset()
+}
+
+func TestCreateDataset(t *testing.T) {
+	tc := testutil.SystemTest(t)
+	datasetID = "my-image-dataset"
```
muncus, line 57: Can we add a dynamic component to this dataset, like some UUID or something? It will help reduce resource contention.

muncus (approved):
LG, with a minor comment for resource contention. please address before merging.

aiplatform/snippets/create_dataset_test.go
telpirion: That is a reasonable request ... unfortunately, the developer doesn't control the actual dataset ID. It is automatically assigned by the service. I will rename this variable to reflect that behavior.

PR #3269, opened 2023-08-28
feat(aiplatform): adds delete dataset sample

telpirion:
This PR adds a Go code sample for deleting datasets.


aiplatform/snippets/create_dataset_test.go
```diff
@@ -65,10 +46,13 @@ func TestCreateDataset(t *testing.T) {
...
+	output := got
+	teardown(output, t)
 }
 
-func deleteDataset() {
+func teardown(output string, t *testing.T) {
+	t.Helper()
+
 	// parse dataset name--we cannot predict the dataset ID at creation time.
 	tmp := strings.Split(output, "\n")
```
muncus, line 57: optional: consider strings.Cut instead of Split. it has a nicer error condition, IMO.

aiplatform/snippets/delete_dataset_test.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+func setup(t *testing.T) {
+	t.Helper()
+
+	// Create a new dataset
+	ctx := context.Background()
+	tc := testutil.SystemTest(t)
+
+	projectID = tc.ProjectID
```
muncus, line 45: If there are multiple tests in parallel, they'll fight over the package-level `datasetName` and `datasetID` variables if `setup()` is called more than once.

can we make setup() return those values instead?

aiplatform/snippets/create_dataset_test.go
telpirion: If I'm reading the docs for [`Cut()`](https://pkg.go.dev/strings#Cut) correctly, it looks like it splits a string into two strings on the first instance of the separator. For this resource name, there are 5 instances of the separator and I need the 6th (and final) item.

aiplatform/snippets/delete_dataset_test.go
muncus: if a test suite calls `t.Parallel()`, with multiple tests, it could happen. But now that's *2* things that have to go different for this to cause an issue. its probably safe enough as-is.

muncus (approved):

aiplatform/snippets/create_dataset_test.go
telpirion: Never mind, I was looking at the wrong place in the code. Done!

PR #3283, opened 2023-09-08
feat(cloudrunci): Add retry by default to Service.Request

muncus:
Functional options pattern allows customizing retry behavior

Also updates several tests to use this new helper.


run/testing/helloworld.e2e_test.go
```diff
@@ -36,30 +34,17 @@ func TestHelloworldService(t *testing.T) {
 	}
 	defer service.Clean()
 
-	req, err := service.NewRequest("GET", "/")
+	resp, err := service.Request("GET", "/")
+	out, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		t.Fatalf("service.NewRequest: %v", err)
+		t.Errorf("ioutil.ReadAll: %v", err)
```
telpirion, line 40: nit: replace use of `ioutil` with `io`.

telpirion (approved):
This looks really good to me! I would just make sure that we have usage of this pattern well documented (somewhere).

run/testing/helloworld.e2e_test.go
muncus: is this safe to do while we still (nominally) support versions < 1.18, or should we wait?

muncus: We discussed this OOB, and decided to wait on updating the package here until 1.18 is our minimum supported version.

PR #3291, opened 2023-09-13
fix: refactor error reporting code sample

minherz:
align code sample for error reporting with code in other languages (b/275730159). use an explicit error with the "fixed" message. populate the service context with the name of the service (module) and the version.

### DIFF
--- a/errorreporting/errorreporting_quickstart/main.go
+++ b/errorreporting/errorreporting_quickstart/main.go
@@ -44,7 +44,7 @@ func main() {
 		ServiceName:    "errorreporting_quickstart",
 		ServiceVersion: "0.0.0",
 		OnError: func(err error) {
-			log.Printf("Could not repport the error: %v", err)
+			log.Printf("Could not report the error: %v", err)
 		},
 	})
 	if err != nil {
@@ -52,7 +52,7 @@ func main() {
 	}
 	defer errorClient.Close()
 
-	err = errors.New("Something went wrong")
+	err = errors.New("something went wrong")
 	if err != nil {
 		logAndPrintError(err)
 		return
@@ -60,7 +60,8 @@ func main() {
 }
 
 func logAndPrintError(err error) {
-	// error context is autopopulated
+	/// Client autopopulates the error context of the error. For more details about the context see:
+	/// https://cloud.google.com/error-reporting/reference/rest/v1beta1/ErrorContext
 	errorClient.Report(errorreporting.Entry{
 		Error: err,
 	})
--- a/errorreporting/errorreporting_quickstart/main.go
+++ b/errorreporting/errorreporting_quickstart/main.go
@@ -21,8 +21,9 @@ package main
 
 import (
 	"context"
+	"errors"
 	"log"
-	"net/http"
+	"os"
 
 	"cloud.google.com/go/errorreporting"
 )
@@ -32,14 +33,18 @@ var errorClient *errorreporting.Client
 func main() {
 	ctx := context.Background()
 
-	// Sets your Google Cloud Platform project ID.
-	projectID := "YOUR_PROJECT_ID"
+	// TODO: Sets your Google Cloud Platform project ID via environment or explicitly
+	projectID := os.Getenv("GOOGLE_PROJECT_ID")
+	if projectID == "" {
+		projectID = "YOUR_PROJECT_ID"
+	}
 
 	var err error
 	errorClient, err = errorreporting.NewClient(ctx, projectID, errorreporting.Config{
-		ServiceName: "myservice",
+		ServiceName:    "errorreporting_quickstart",
+		ServiceVersion: "0.0.0",
 		OnError: func(err error) {
-			log.Printf("Could not log error: %v", err)
+			log.Printf("Could not repport the error: %v", err)
 		},
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


errorreporting/errorreporting_quickstart/main.go
```diff
@@ -32,30 +33,34 @@ var errorClient *errorreporting.Client
 func main() {
 	ctx := context.Background()
 
-	// Sets your Google Cloud Platform project ID.
-	projectID := "YOUR_PROJECT_ID"
+	// TODO: Sets your Google Cloud Platform project ID via environment or explicitly
+	projectID := os.Getenv("GOOGLE_PROJECT_ID")
```
tritone, line 37: Why add this? Also, if you are going to add it, `GOOGLE_CLOUD_PROJECT` is the standard environment variable, not `GOOGLE_PROJECT_ID`.

errorreporting/errorreporting_quickstart/main.go
```diff
@@ -32,30 +33,34 @@ var errorClient *errorreporting.Client
...
+	}
 
 	var err error
 	errorClient, err = errorreporting.NewClient(ctx, projectID, errorreporting.Config{
-		ServiceName: "myservice",
+		ServiceName:    "errorreporting_quickstart",
+		ServiceVersion: "0.0.0",
 		OnError: func(err error) {
-			log.Printf("Could not log error: %v", err)
+			log.Printf("Could not repport the error: %v", err)
```
tritone, line 47: nit: s/repport/report

errorreporting/errorreporting_quickstart/main.go
```diff
@@ -32,30 +33,34 @@ var errorClient *errorreporting.Client
...
+	err = errors.New("Something went wrong")
 	if err != nil {
 		logAndPrintError(err)
 		return
 	}
-	log.Print(resp.Status)
 }
 
 func logAndPrintError(err error) {
+	// error context is autopopulated
```
tritone, line 63: nit: use complete sentences for comments.

errorreporting/errorreporting_quickstart/main.go
minherz: We try to align samples across _all_ languages to exhibit the same behavior. The options are:

* Environment variable
* Explicit setup
* Fallback to project ID implicitly configured at client initialization

Unfortunately I did not have time to dig for implementation of the last bullet in Golang. I will appreciate it if you can point me to the code that "retrieves" project ID from the client settings.

errorreporting/errorreporting_quickstart/main.go
minherz: I guess it is not a problem then, correct? Although the difference is small the message is more precise because there is a clear notion of "logging errors".

errorreporting/errorreporting_quickstart/main.go
minherz: English is not my mother tongue. To the best of my knowledge it is a complete sentence. Can you provide your version to demonstrate what you mean?

errorreporting/errorreporting_quickstart/main.go
NimJay:
```suggestion
			log.Printf("Could not report the error: %v", err)
```

errorreporting/errorreporting_quickstart/main.go
NimJay: How about 
```suggestion
	// The context of the error is auto-populated.
```

errorreporting/errorreporting_quickstart/main.go
minherz: My apologies. Indeed, I misunderstood the comment. Thank you, Nim.

errorreporting/errorreporting_quickstart/main.go
tritone: Yes, sorry, I just meant that go style is to capitalize and use a period. See https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

minherz: Thanx

errorreporting/errorreporting_quickstart/main.go
NimJay: +1 to changing to `GOOGLE_CLOUD_PROJECT`
```suggestion
	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
```
... primarily because the rest of this repository uses `GOOGLE_CLOUD_PROJECT` ([see search results](https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fgolang-samples+%22GOOGLE_CLOUD_PROJECT%22&type=code)).
There's no use of `GOOGLE_PROJECT_ID`.

tritone: I am not sure autodetecting the project ID is supported in all clients for Go; see discussion on googleapis/google-cloud-go#1294 . @codyoss please correct me if I'm wrong.

In snippet samples we generally pass the project ID in via an arg (see https://googlecloudplatform.github.io/samples-style-guide/#minimal-arguments ). As this is a quickstart, I think it's fine to try to detect it from the environment. However I do think we should standardize on `GOOGLE_CLOUD_PROJECT` rather than `GOOGLE_PROJECT_ID` given that is what is supported across tools.

NimJay (approved):
Thanks for working on these updates related to 

Approved (since the remaining `GOOGLE_CLOUD_PROJEC` is a nit, IMO).

Looks like the `Go / Lint` check is passing.
It sounds like we just need to run:
```
goimports -w . && find . -name go.mod -execdir go mod tidy \;
```

tritone (approved):

errorreporting/errorreporting_quickstart/main.go
minherz: Changes made to address this while allowing users to provide the project ID via argument.

PR #3307, opened 2023-09-20
feat(cloudrunci): Service.Do() method, to support automatic retries in more tests

muncus:
- feat: move retry logic to Do(), have Request() call that.
- feat: move run tests to use the service.Do method


internal/cloudrunci/cloudrunci.go
```diff
@@ -177,6 +171,15 @@ func (s *Service) Request(method string, path string, opts ...func(*RetryOptions
 	return resp, lastSeen
 }
 
+// Request issues an HTTP request to the deployed service.
```
grayside, line 174: question: I want to recap my understanding:

* Existing method: `NewRequest()` which is a request factory that allows the caller to customize the request before calling `Do()`
* New method: `Do()` which runs an HTTP request, returns a response, and manages retries.
* New method: `Request()` which wraps `NewRequest()` and `Do()` for simple HTTP requests.

We expect Request() to work for many use cases.

internal/cloudrunci/cloudrunci.go
```diff
@@ -154,16 +154,10 @@ func (s *Service) Request(method string, path string, opts ...func(*RetryOptions
...
-			lastSeen = err
-			continue
-		}
 		defaultClient := &http.Client{}
 
-		resp, err = defaultClient.Do(req)
-		if err != nil {
-			lastSeen = err
+		resp, lastSeen = defaultClient.Do(req)
+		if lastSeen != nil {
```
grayside, line 160: suggestion: I realize this retry code was already in place, but this is the first I recall seeing it and we're talking about rolling it out to more use. I find with retries it's often really valuable to have some logging:

1. What's the action
2. What the current attempt we're on
3. The total number of attempts we might accept.

That might look like:

```
HTTP Request [attempt #2/5] POST http://example.com
```

Why is this valuable?

1. It shows you what to expect if you are watching the logs in real time
2. It allows familiarity with the logs to surface that you need a longer delay before making the first request, or that you might benefit from increasing the number of retries, without as much time staring at the logs

internal/cloudrunci/cloudrunci.go
```diff
@@ -154,16 +154,10 @@ func (s *Service) Request(method string, path string, opts ...func(*RetryOptions
 	var lastSeen error
 	resp := &http.Response{}
 	for i := 0; i < options.MaxAttempts; i++ {
-		req, err := s.NewRequest(method, path)
-		if err != nil {
-			lastSeen = err
-			continue
-		}
 		defaultClient := &http.Client{}
```
grayside, line 157: suggestion: The HTTP client we are removing from the tests uses a 10 second timeout, should we move that timeout here? (Or keep a timeout but make it a more conservative 30 seconds to avoid needing another option?)

internal/cloudrunci/cloudrunci.go
```diff
@@ -154,16 +154,10 @@ func (s *Service) Request(method string, path string, opts ...func(*RetryOptions
...
 
-		resp, err = defaultClient.Do(req)
-		if err != nil {
-			lastSeen = err
+		resp, lastSeen = defaultClient.Do(req)
+		if lastSeen != nil {
 			continue
 		}
 		if options.ShouldAccept(resp) {
```
grayside, line 163: suggestion: GitHub UI is fighting with me, but here's another bit of review on previously existing code: a simplification opportunity.

```go
if options.ShouldAccept(resp) {
    return resp, nil
}
time.Sleep(options.Delay)
```

1. No need for the else clause because all code will return before reaching it or execute the clause.
2. No need for continue since all code will continue if it reaches the sleep.

internal/cloudrunci/cloudrunci.go
muncus: `Request()` is an existing method. I updated it last week to do retries, and to make more tests use it.

internal/cloudrunci/cloudrunci.go
muncus: Good call, thanks!

internal/cloudrunci/cloudrunci.go
muncus: Do you know why that timeout was set?
It does not appear in most tests I've encountered, so without context it does not look to be structural enough to justify adding the flexibility.
If it does become necessary, we can add an http.Client to the Service object.

internal/cloudrunci/cloudrunci.go
muncus: If we had a leveled logger, i'd include this as a Debug log, but I'm doubtful that the volume of logging is useful in the vast majority of cases. (happy to revisit this when `log/slog` is an option).
I have wrapped the returned error to indicate retries, so that calling code will always know how many retries were done.

internal/cloudrunci/cloudrunci.go
grayside: I believe in many cases it's not being set is an oversight. In this case, it is for local test execution and the notion that if the HTTP request takes more than 10 seconds to respond, that's a category of failure that something's wrong in the service responsiveness.

I'm okay deferring to your preference on this, we can document local test runs using `-timeout` and leave service responsiveness to an investigation in overall test runner timeouts in the rare event they come up.

grayside (approved):

grayside:
It looks like the fail is from flaky tests, triggered as part of this PR because the change is part of common test infrastructure.

muncus:
The spanner test failures in the kokoro CI build are not related to this change.

PR #3308, opened 2023-09-21
WIP OTel-Go code sample

punya:
### Design principles

*   Keep GCP-specific logic out of the application binary
*   Configure things at runtime, ideally using OTel env variable conventions, rather than baking decisions into the binary
*   Use slog to stdout for structured logging rather than OTLP (to align with the convention in containers)

### To do

*   [ ] Fix trace-log correlation


punya:
Superseded by #3312 .

PR #3312, opened 2023-09-25
feat(trace): add opentelemetry instrumentation example

dashpole:
## Description

Supersedes #3308

This example demonstrates instrumentation for an example http client and server.  It uses OpenTelemetry libraries for Traces and Metrics, and `slog` for logging.  It writes logs in json format to stdout, and writes traces and metrics to a local OpenTelemetry collector using OTLP.  The application does not include any GCP-specific dependencies.

The collector is configured to send to GMP for metrics, Cloud Logging for logs, and Cloud Trace for traces.

This example shows how to correlate logs with traces:

Screenshot 2023-09-25 at 1 36 29 PM

It does not correlate metrics with traces (through exemplars), since OpenTelemetry-go does not support exemplars yet: https://github.com/open-telemetry/opentelemetry-go/issues/559

Running instructions:

```sh
gcloud auth application-default login
export GOOGLE_APPLICATION_CREDENTIALS=$HOME/.config/gcloud/application_default_credentials.json
export PROJECT_ID=my-project
docker-compose up --abort-on-container-exit
```

## Checklist


dashpole:
@punya

opentelemetry/instrumentation/otel-collector-config.yaml
```diff
@@ -0,0 +1,68 @@
+receivers:
+  otlp:
+    protocols:
+      http:
+        endpoint: "0.0.0.0:4318"
+  filelog:
```
dashpole, line 6: one downside of using the filelog receiver, and relying on it for translation (e.g. parsing trace-related fields), is that users running on managed services (cloud run, GKE, GCE) who copy the application logic won't get the expected behavior.  We could, instead, make the example send the gcp-specific fields (e.g. trace with project prefix), but that would go against the goal of not including gcp-specific behavior.

punya: Is it okay if we leave it this way for simplicity, and have a larger discussion within the team to find the best way forward on managed runtimes?

cc @aabmass @jsuereth

opentelemetry/instrumentation/app/server.go
```diff
@@ -0,0 +1,51 @@
...
+			http.Error(w, err.Error(), http.StatusBadGateway)
+			return
+		}
+	}
+
+	fmt.Fprintln(w, "ok")
+}
+
+func runServer() error {
+	http.Handle("/single", otelhttp.NewHandler(http.HandlerFunc(handleSingle), "/single"))
```
dashpole, line 48: TODO: consider using WithRouteTag to make this produce better metrics.

dashpole:
Note: I'm having issues getting CI to pass.  I believe I need to update CI to 1.21 in order to use slog for this new example, but it doesn't look like there is a kokoro image for 1.21 yet (gcr.io/golang-samples-tests/go121).

opentelemetry/instrumentation/app/server.go
dashpole: Done.  It doesn't appear to add the dimension to metrics, though.

opentelemetry/instrumentation/otel-collector-config.yaml
dashpole: I'm definitely in-favor of leaving it this way for simplicity.

dashpole:
Any suggestions on how to test something that requires docker-compose?

opentelemetry/instrumentation/otel-collector-config.yaml
aabmass: SGTM

opentelemetry/instrumentation/app/client.go
```diff
@@ -0,0 +1,36 @@
...
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+func callSingle(ctx context.Context) error {
```
telpirion, line 24: issue: provide comment for function that describes usage, purpose, etc.

opentelemetry/instrumentation/app/client.go
```diff
@@ -0,0 +1,36 @@
...
+	res, err := otelhttp.Get(ctx, "http://localhost:8080/single")
+	if err != nil {
+		return fmt.Errorf("error invoking /single: %w", err)
```
telpirion, line 27: comment: you can just return error, no need to format.

opentelemetry/instrumentation/app/client.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	err = res.Body.Close()
+	if err != nil {
+		return fmt.Errorf("error closing response body: %w", err)
```
telpirion, line 32: Same as previous.

opentelemetry/instrumentation/app/client.go
telpirion: issue: don't pass in `Context` unless really necessary. If it is necessary, provide a comment explaining why.

opentelemetry/instrumentation/app/logger.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+// handerWithTraceContext adds attributes from the trace context
+func handerWithTraceContext(handler slog.Handler) *traceContextLogHandler {
+	return &traceContextLogHandler{Handler: handler}
+}
+
+type traceContextLogHandler struct {
```
telpirion, line 29: issue: provide comment for type.

opentelemetry/instrumentation/app/logger.go
```diff
@@ -0,0 +1,44 @@
...
+	slog.Handler
+}
+
+func (t *traceContextLogHandler) Handle(ctx context.Context, record slog.Record) error {
```
telpirion, line 33: issue: provide comment for struct method.

opentelemetry/instrumentation/app/logger.go
```diff
@@ -0,0 +1,44 @@
...
+	if s := trace.SpanContextFromContext(ctx); s.IsValid() {
```
telpirion, line 34: comment: this is a really dense wall of code. I would move the calls to `slog.Any()` onto their own lines. I would also insert a few more line breaks and comments.

opentelemetry/instrumentation/app/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+package main
+
+import (
+	"context"
+	"log/slog"
+	"os"
+)
+
+func main() {
```
telpirion, line 23: issue: provide a comment for this app. It needs some explanation.

opentelemetry/instrumentation/app/main.go
```diff
@@ -0,0 +1,30 @@
...
+	ctx := context.Background()
+	slog.InfoContext(ctx, "server starting...")
```
telpirion, line 25: comment: insert line breaks here. This is hard to read.

telpirion (changes requested):
There are a few global issues that need to be addressed. PTAL, address the issues, and I'll take another pass on this PR.

Issues:
+ It looks like these samples are missing region tags. Please add.
+ There are almost no comments explaining this sample. Please add.
+ The code in these files is really dense, with few line breaks. It's hard to read. Please insert a few more breaks.
+ Consider keeping each method call on its own line, rather than nesting them. Keeping them separate improves readability.

opentelemetry/instrumentation/app/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang:1.21
+WORKDIR /usr/src/app
+COPY go.mod go.sum ./
+RUN go mod download && go mod verify
+COPY *.go ./
+RUN go build -v -o /usr/local/bin/app .
+CMD sh -c "app | tee /var/log/app.log"
```
aabmass, line 7: I think this will lose stderr logs, not sure if the example makes any

opentelemetry/instrumentation/app/setup.go
```diff
@@ -0,0 +1,66 @@
...
+	texporter, err := autoexport.NewSpanExporter(ctx)
+	if err != nil {
+		return err
+	}
+	tp := trace.NewTracerProvider(trace.WithBatcher(texporter))
+	defer tp.Shutdown(ctx)
+	otel.SetTracerProvider(tp)
+
+	// Configure Metric Export to send metrics as OTLP
+	// TODO(https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4131) - use env-based configuration instead of hardcoding
```
aabmass, line 45: This applies to trace exporter setup too right?

opentelemetry/instrumentation/docker-compose.yaml
```diff
@@ -0,0 +1,29 @@
...
+    volumes:
+      - logs:/var/log:rw
+    depends_on:
+      - "otelcol"
+  otelcol:
+    image: otel/opentelemetry-collector-contrib:0.85.0
+    volumes:
+      - ./otel-collector-config.yaml:/etc/otelcol-contrib/config.yaml:ro
+      - $GOOGLE_APPLICATION_CREDENTIALS:/tmp/keys/gcp-credentials.json:ro
+      - logs:/var/log:rw
```
aabmass, line 18: nit could be `ro`?

opentelemetry/instrumentation/docker-compose.yaml
```diff
@@ -0,0 +1,29 @@
...
+    environment:
+      - GOOGLE_APPLICATION_CREDENTIALS=/tmp/keys/gcp-credentials.json
+      - PROJECT_ID=$PROJECT_ID
```
aabmass, line 21: Is this required or optional with ADC fallback?

opentelemetry/instrumentation/otel-collector-config.yaml
aabmass, line 1: it would be nice to add some comments in here to explain what each part does for those unfamiliar with collector conig.

aabmass (approved):
This is a nice sample, like how easy it is to run with compose

opentelemetry/instrumentation/app/client.go
dashpole: done.  Context is required, and I added why.

opentelemetry/instrumentation/app/client.go
dashpole: done

opentelemetry/instrumentation/app/client.go
dashpole: done

opentelemetry/instrumentation/app/logger.go
dashpole: done

opentelemetry/instrumentation/app/logger.go
dashpole: done

opentelemetry/instrumentation/app/logger.go
dashpole: done

opentelemetry/instrumentation/app/main.go
dashpole: done

opentelemetry/instrumentation/app/main.go
dashpole: done

opentelemetry/instrumentation/app/setup.go
dashpole: No.  It uses autoexport for traces.

opentelemetry/instrumentation/docker-compose.yaml
dashpole: When I do `gcloud auth application-default login`, it doesn't appear to set my project.  I'm not sure if that is something special with my setup, or if I need to change the login command...

opentelemetry/instrumentation/otel-collector-config.yaml
dashpole: done

opentelemetry/instrumentation/docker-compose.yaml
aabmass: I think it depends if we ADC gets auth/projectID from the metadata server vs credentials file in your home dir.

If `PROJECT_ID` is required in Cloud Shell (target env for quickstarts), can we use [mandatory environment variables](https://docs.docker.com/compose/compose-file/12-interpolation/#:~:text=Similarly%2C%20the%20following%20syntax%20allows%20you%20to%20specify%20mandatory%20variables%3A)? It prints out a nice message if the variable is unset

dashpole: made both environment variables required

opentelemetry/instrumentation/app/Dockerfile
dashpole: fixed by adding `2>&1`

dashpole:
> * It looks like these samples are missing region tags. Please add.

Done.

> * There are almost no comments explaining this sample. Please add.

Done.

> * The code in these files is really dense, with few line breaks. It's hard to read. Please insert a few more breaks.

Done.

> * Consider keeping each method call on its own line, rather than nesting them. Keeping them separate improves readability.

Done.  I did keep some calls nested, but separated ones I thought were less readable.

dashpole:
This now includes an integration test that checks that metrics, logs, and traces are successfully generated.

dashpole:
@muncus @telpirion Please let me know if I missed any requirements

telpirion (approved):

PR #3335, opened 2023-10-04
fix(functions): remove old region tag

rogerthatdev:
Follow up to #3324 

This PR removes the old region tag reserved for the [2nd gen version of the sample](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/functions/functionsv2/helloworld/hello_world.go)


telpirion:
@rogerthatdev Do you want to leave this as a Draft? All the tests are green and the change is pretty small.

PR #3547, opened 2023-10-27
feat(spanner): add samples for BatchWrite

sunnsing-google:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


sunnsing-google:
Note that this needs a new release of https://github.com/googleapis/google-cloud-go. Once the release happens, I'll update go.mod in this PR.

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	singerColumns := []string{"SingerId", "FirstName", "LastName"}
+	albumColumns := []string{"SingerId", "AlbumId", "AlbumTitle"}
+	mutationGroups := make([]*spanner.MutationGroup, 2)
+
+	group1 := []*spanner.Mutation{
```
harshachinta, line 40: nit: rename this to mutationGroup1 for better readability

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,67 @@
...
+		spanner.InsertOrUpdate("Singers", singerColumns, []interface{}{16, "Scarlet", "Terry"}),
+	}
+	mutationGroups[0] = &spanner.MutationGroup{Mutations: group1}
+
+	group2 := []*spanner.Mutation{
+		spanner.InsertOrUpdate("Singers", singerColumns, []interface{}{17, "Marc", ""}),
+		spanner.InsertOrUpdate("Singers", singerColumns, []interface{}{18, "Catalina", "Smith"}),
+		spanner.InsertOrUpdate("Albums", albumColumns, []interface{}{17, 1, "Total Junk"}),
+		spanner.InsertOrUpdate("Singers", singerColumns, []interface{}{18, 2, "Go, Go, Go"}),
```
harshachinta, line 49:
```suggestion
		spanner.InsertOrUpdate("Albums", albumColumns, []interface{}{18, 2, "Go, Go, Go"}),
```

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	mutationGroups[1] = &spanner.MutationGroup{Mutations: group2}
+
+	iter := client.BatchWrite(ctx, mutationGroups)
+	doFunc := func(response *sppb.BatchWriteResponse) error {
+		if ts := response.GetCommitTimestamp(); ts == nil {
```
harshachinta, line 55: Can we do the same status validation that we used in Java sample to maintain consistency?

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,67 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spanner
+
+// [START spanner_batch_write]
```
harshachinta, line 17: I am worried about the sample tag that is used here.
We used `spanner_batch_write_at_least_once` for Java, and this should be consistent across the client libraries as this tag is mapped in the public documentation. We cannot have it as `spanner_batch_write` and most importantly `spanner_batch_write_at_least_once` tag does not properly suit.
Any idea why did we suffix it `with at_least_once`?

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,67 @@
...
+	doFunc := func(response *sppb.BatchWriteResponse) error {
+		if ts := response.GetCommitTimestamp(); ts == nil {
+			return fmt.Errorf("invalid commit timesamp")
+		}
+		return nil
+	}
+	if err = iter.Do(doFunc); err != nil {
+		return err
+	}
+	fmt.Fprintf(w, "BatchWrite successful")
```
harshachinta, line 63: fmt.Fprint should work here.

spanner/spanner_snippets/spanner/spanner_batch_write.go
sunnsing-google: The Java library has `write()` and `writeAtLeastOnce()` methods for `Commit`, whereas the golang library has a single method `Apply()` with `ApplyAtLeaseOnce()` as an option.

Since the current batch write API is non-replay protected, it is named `batchWriteAtLeastOnce()` in Java, and the hence the region tag is named such. In golang, we'd have an option in `BatchWriteOptions` instead of a new method.

While it is technically feasible to have different region tags, it is not ideal as you mentioned. Updated this PR to use the same region tag here as well.

spanner/spanner_snippets/spanner/spanner_batch_write.go
sunnsing-google: Thanks for catching.

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,66 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/spanner"
+	sppb "cloud.google.com/go/spanner/apiv1/spannerpb"
+	"google.golang.org/grpc/status"
+)
+
+func batchWrite(w io.Writer, db string) error {
```
telpirion, line 29: issue: provide a function comment that explains what `batchWrite()` does and what the arguments are.

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,66 @@
...
+	ctx := context.Background()
+	client, err := spanner.NewClient(ctx, db)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	singerColumns := []string{"SingerId", "FirstName", "LastName"}
```
telpirion, line 37: question: what database schema is used here? What schema is assumed? Maybe explain in a comment.

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,66 @@
...
+		spanner.InsertOrUpdate("Singers", singerColumns, []interface{}{18, "Catalina", "Smith"}),
+		spanner.InsertOrUpdate("Albums", albumColumns, []interface{}{17, 1, "Total Junk"}),
+		spanner.InsertOrUpdate("Albums", albumColumns, []interface{}{18, 2, "Go, Go, Go"}),
+	}
+	mutationGroups[1] = &spanner.MutationGroup{Mutations: mutationGroup2}
+
+	iter := client.BatchWrite(ctx, mutationGroups)
+	doFunc := func(response *sppb.BatchWriteResponse) error {
+		if err = status.ErrorProto(response.GetStatus()); err == nil {
+			fmt.Fprintf(w, "Mutation group indexes %v have been applied with commit timestamp %v", response.GetIndexes(), response.GetCommitTimestamp())
```
telpirion, line 57: issue: break this into multiple lines. It's hard to parse. Readability should be key, especially when understanding an error.

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,66 @@
...
+	mutationGroup2 := []*spanner.Mutation{
+		spanner.InsertOrUpdate("Singers", singerColumns, []interface{}{17, "Marc", ""}),
+		spanner.InsertOrUpdate("Singers", singerColumns, []interface{}{18, "Catalina", "Smith"}),
+		spanner.InsertOrUpdate("Albums", albumColumns, []interface{}{17, 1, "Total Junk"}),
+		spanner.InsertOrUpdate("Albums", albumColumns, []interface{}{18, 2, "Go, Go, Go"}),
+	}
+	mutationGroups[1] = &spanner.MutationGroup{Mutations: mutationGroup2}
+
+	iter := client.BatchWrite(ctx, mutationGroups)
+	doFunc := func(response *sppb.BatchWriteResponse) error {
```
telpirion, line 55: issue: provide code comment that explains what this function does. Does it trigger for each iteration of `BatchWrite()`, regardless of an error? Right now it seems to assume that this function will only provide a behavior if there is an error.

It also looks like an error is never returned to the caller, as the only return statement here returns nil.

telpirion (changes requested):
I think that the callback function in `iter.Do()` needs more explanation at the very least.

harshachinta:
> Note that this needs a new release of https://github.com/googleapis/google-cloud-go. Once the release happens, I'll update go.mod in this PR.

Hi @sunnsing-google,
The Batch Write changes have been released in version `v1.52.0`. https://github.com/googleapis/google-cloud-go/releases/tag/spanner%2Fv1.52.0
Can you update the go.mod file with this version and run `go mod tidy` and commit those changes?

sunnsing-google:
> > Note that this needs a new release of https://github.com/googleapis/google-cloud-go. Once the release happens, I'll update go.mod in this PR.
> 
> Hi @sunnsing-google, The Batch Write changes have been released in version `v1.52.0`. https://github.com/googleapis/google-cloud-go/releases/tag/spanner%2Fv1.52.0 Can you update the go.mod file with this version and run `go mod tidy` and commit those changes?

Thanks Harsha. Done.

spanner/spanner_snippets/spanner/spanner_batch_write.go
sunnsing-google: Done. Also added a link to go package documentation for BatchWrite.

spanner/spanner_snippets/spanner/spanner_batch_write.go
sunnsing-google: Added a comment.

spanner/spanner_snippets/spanner/spanner_batch_write.go
sunnsing-google: It is not an error. Broke this into multiple lines.

spanner/spanner_snippets/spanner/spanner_batch_write.go
sunnsing-google: Linked to go package documentation for `iter.Do()`

spanner/spanner_snippets/spanner/spanner_batch_write.go
harshachinta: I guess all the samples use schema that is mentioned in start of the public documentation attached below.
https://cloud.google.com/spanner/docs/getting-started/go#create_a_database

Instead of adding the schema here, should we add the link to this schema instead? This makes sample look neat.
@telpirion Can you suggest if this is better?

harshachinta (approved):

spanner/spanner_snippets/spanner/spanner_batch_write.go
sunnsing-google: Harsha. This looks neater. Updated. @telpirion - PTAL.

spanner/spanner_snippets/spanner/spanner_batch_write.go
```diff
@@ -0,0 +1,73 @@
...
+		if err = status.ErrorProto(response.GetStatus()); err == nil {
+			fmt.Fprintf(w, "Mutation group indexes %v have been applied with commit timestamp %v",
+				response.GetIndexes(), response.GetCommitTimestamp())
+		} else {
+			fmt.Fprintf(w, "Mutation group indexes %v could not be applied with error %v",
+				response.GetIndexes(), err)
+		}
+		// Return an actual error as needed.
+		return nil
```
harshachinta, line 68: We are always returning `nil` which will always succeed the `runSample` method even in case of failures in batch write sample. We should return the error instead to indicate that the test failed.
Currently it might fail the assertion but ideally the sample invocation should fail.

spanner/spanner_snippets/spanner/integration_test.go
```diff
@@ -142,6 +142,8 @@ func TestSample(t *testing.T) {
 	mustRunSample(t, createDatabase, dbName, "failed to create a database")
 	runSample(t, createClients, dbName, "failed to create clients")
 	runSample(t, write, dbName, "failed to insert data")
+	out = runSample(t, batchWrite, dbName, "failed to write data using BatchWrite")
+	assertNotContains(t, out, "could not be applied with error")
```
harshachinta, line 146: nit: can we instead check for the success case `have been applied with commit timestamp`?

spanner/spanner_snippets/spanner/spanner_batch_write.go
sunnsing-google: Batch write allows for partial failure and thus did it this way - process all response items and collect successes and failures. Partial failure may not mean failure for the client app.

spanner/spanner_snippets/spanner/integration_test.go
sunnsing-google: Since there can be partial success, the code checks if there was any failure.

spanner/spanner_snippets/spanner/spanner_batch_write.go
telpirion: Thanks!

telpirion (approved):

PR #3553, opened 2023-11-02
chore(deps): combined update of grpc -> 1.56.3

muncus:
resolves ~88 open dependabot alerts


muncus:
the kokoro failure is an unrelated connection reset in a compute test. i'll merge with the current tests.

telpirion (approved):

PR #3563, opened 2023-11-10
feat: Update supported go versions to match cloud SDK

muncus:
that's go 1.19 - go 1.21

This also includes a `system_test.sh` fix for UID mismatches between the namespaces used by each of our nested containers.


muncus:
The test failures in Kokoro do not appear related to this change, except the `discoveryengine` test, which ive captured in #3569

telpirion (approved):

PR #3574, opened 2023-11-17
feat(testing): convert Kokoro to use new Makefile-based testing

muncus:
- feat(testing): use gotestsum to write sponge_log.xml
- feat: use 'make' to run tests.

### DIFF
--- a/testing/kokoro/system_tests.sh
+++ b/testing/kokoro/system_tests.sh
@@ -32,6 +32,7 @@ go version
 date
 
 cd "${1:-github/golang-samples}"
+PROJECT_ROOT=$(pwd)
 
 export GO111MODULE=on # Always use modules.
 export GOPROXY=https://proxy.golang.org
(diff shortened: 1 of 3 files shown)


appengine_flexible/websockets/go.mod
telpirion, line 1: Is this change intended for this PR?

testing/kokoro/system_tests.sh
```diff
@@ -32,10 +32,12 @@ go version
 date
 
 cd "${1:-github/golang-samples}"
+PROJECT_ROOT=$(pwd)
 
 export GO111MODULE=on # Always use modules.
 export GOPROXY=https://proxy.golang.org
 TIMEOUT=60m
+export GOLANG_SAMPLES_E2E_TEST=
```
telpirion, line 40: It looks like this is missing an assignment on the right-hand side.

muncus: well, its explicitly set to an empty string, so that `testutil` considers this value false. But adding an explicit empty string is clearer, so i've done that.

telpirion (approved):

PR #3576, opened 2023-11-21
fix: kokoro jobs for system tests should bypass the diff check

muncus:


codyoss (approved):

muncus:
The kokoro test failures are the usual flaky culprits. 😞 
They are not related to this change, so i'll merge anyway.

PR #3583, opened 2023-11-27
docs(bigquery): add more output to migration quickstart

shollyman:
Quickstart testing is failing with no output.  This change adds more status updates, as it's likely we're dealing with the occasional slow workflow and this should yield more concrete details when it recurs.

Fixes: https://github.com/GoogleCloudPlatform/golang-samples/issues/3567


muncus (approved):
Thanks for looking into this failing test!

PR #3585, opened 2023-11-28
chore: Update App Engine samples to the go121 runtime

nick-oconnor:
## Description

`gcloud` throws deprecation warnings when deploying an app with the  `go115` runtime.

## Checklist

### DIFF
--- a/appengine/go.mod
+++ b/appengine/go.mod
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/golang-samples/appengine
 
-go 1.19
+go 1.21
 
 require (
 	cloud.google.com/go/cloudtasks v1.11.1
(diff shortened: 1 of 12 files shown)


appengine/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/golang-samples/appengine
 
-go 1.19
```
muncus, line 3: Will these samples still run on 1.19? If so, we should leave the go.mod version at 1.19.

nick-oconnor: I successfully rebuilt each project. I deployed the `helloworld` example and it works. I'll deploy the others and report back.

nick-oconnor: I ran `go mod tidy` for each module. I would expect it to catch most of the issues and upgrade incompatible dependencies, but I agree that it's a good idea to check. Testing this will require creating a cloudsql instance, a pub sub queue, and a cloud tasks instance.

muncus: Ok, if `mod tidy` made the update, there must be some dependency that requires 1.21.
Since this is tied to the appengine runtime environment, this seems totally reasonable.
Thanks for your efforts, Nick!

muncus (approved):

appengine/go.mod
nick-oconnor: To clarify, I updated the go to 1.21 to match the most recent AE runtime, then executed `mod tidy` to refresh the dependency tree.

PR #3617, opened 2023-12-15
feat(profiler): add ListProfiles API sample

psx95:
## Description

Adds a sample to showcase the use of ListProfiles API via the client libraries.

## Checklist
samples/blob/main/CONTRIBUTING.md#testing))


profiler/export_profiles/main.go
```diff
@@ -0,0 +1,129 @@
...
+		if profilesIterator.PageInfo().Remaining() == 0 {
+			// This signifies that the client will make a new API call internally
+			fmt.Printf("next page token: %v\n", profilesIterator.PageInfo().Token)
+		}
+	}
+	return nil
+}
+
+func main() {
+	flag.Parse()
```
aabmass, line 122: IIUC, project_id flag is mandatory. Consider either

1. validate that it is set and give a helpful error
2. try to use the default project from ADC if its not set

profiler/export_profiles/main.go
```diff
@@ -0,0 +1,129 @@
...
+// This function reads profiles for a given project and stores them into locally created files.
+// The profile metadata gets stored into a 'metdata.csv' file, while the individual pprof files
+// are created per profile.
+func run(ctx context.Context) error {
+	client, err := cloudprofiler.NewExportClient(ctx)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+	fmt.Printf("Attempting to fetch %v profiles with a pageSize of %v for %v\n", *maxProfiles, *pageSize, *project)
```
aabmass, line 49: Consider using `log.Printf` instead of `fmt.Printf` throughout

profiler/export_profiles/main.go
```diff
@@ -0,0 +1,129 @@
...
+			log.Println("Read all available profiles")
+			break
+		}
+		if err != nil {
+			return fmt.Errorf("error reading profile from response: %v", err)
+		}
+		profileCount++
+
+		filename := fmt.Sprintf("profiles/profile%06d.pprof", profileCount)
+		err = os.WriteFile(filename, profile.ProfileBytes, 0)
```
aabmass, line 91: What does permission `0` do here?

profiler/export_profiles/main.go
```diff
@@ -0,0 +1,129 @@
...
+
+		if err != nil {
+			fmt.Printf("Error writing %s, skipping: %v\n", filename, err)
```
aabmass, line 94: I imagine the most likely errors would be permission denied or no space on device. It might be better to just fail the whole script at this point

profiler/export_profiles/main.go
```diff
@@ -0,0 +1,129 @@
...
+		if err == iterator.Done {
+			log.Println("Read all available profiles")
+			break
+		}
+		if err != nil {
+			return fmt.Errorf("error reading profile from response: %v", err)
+		}
+		profileCount++
+
+		filename := fmt.Sprintf("profiles/profile%06d.pprof", profileCount)
```
aabmass, line 90: If you restarted the script with `-page_token=...` I think it would overwrite the previous files in the directory

psx95: No, the script fails - because it tries to recreate an existing directory in the beginning. Seeing as it is just a sample showcasing one potential way to use the API, I feel ok with this.

profiler/export_profiles/main.go
psx95: I think it represents `ModeDIr`.

profiler/export_profiles/main.go
psx95: Done, added validation check for missing project ID flag.

profiler/export_profiles/main.go
muncus: This arg is usually something like `0644` for setting the usual permission bits on a file.
`0` is likely to create files with no permissions set (so the owner will need to `chmod` them first), and `fs.ModeDir` will try to create a directory, which I do not expect will work.

aabmass: What permission do files end up with?

muncus (commented):
This tool should have a corresponding test that covers at least the profile API calls being made.

Given the nature of this tool, it may be easier to split the fetching of profiles from the writing of files, so that at least the profile fetching code can be tested.

profiler/export_profiles/main.go
psx95: It's just a sample - you could remove the files from the location and re-run the script again with the last page token.

I could add additional handling in the code to check for existence of the folder - but I feel its not relevant to how ListProfiles API work.

profiler/export_profiles/main.go
psx95: Yeah, checked again - its no permissions. Will update it with `0666`.

aabmass (approved):

psx95:
> This tool should have a corresponding test that covers at least the profile API calls being made.
> 
> Given the nature of this tool, it may be easier to split the fetching of profiles from the writing of files, so that at least the profile fetching code can be tested.

@muncus this is mostly a sample script (kind of a "quickstart") which helps the users understand how to leverage the ListProfiles API to download profiles. This example will mostly be cited in the documentation only.

Regarding the testing, do you mean an integration test to verify that files are actually being created when this script is run ? 

If its unit testing, IMO - this script does not own much business logic that can be tested - its merely calling the APIs from auto-generated client libraries.

profiler/export_profiles/main.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Initial request for the ListProfiles API
+	request := &pb.ListProfilesRequest{
+		Parent:    fmt.Sprintf("projects/%s", *project),
+		PageSize:  int32(*pageSize),
+		PageToken: *pageToken,
+	}
+
+	// create a folder for storing profiles & metadata
+	profilesDirName := fmt.Sprintf("profiles_%v", time.Now())
```
aabmass, line 60: nit i tried this out and it made pretty verbose pathnames that require escaping in the shell e.g. `profiles_2023-12-18 22:12:18.807095513 +0000 UTC m=+0.009132339`. Consider something a little more compact like `time.RFC3339` or unix timestamp

psx95: I think, I will just use `time.Unix()`.

muncus (approved):
I discussed testing with @psx95, and we're waiving the testing requirement on this because the profiler API **lacks a delete method**, so tests are not able to clean up after themselves.

PR #3756, opened 2024-01-22
test: enable the 'loopvar' experiment to prepare for 1.22

muncus:
This PR should shake out any tests that need changes prior to the release of
go1.22, which introduces a change to the way loop variables are scoped.

see https://go.dev/wiki/LoopvarExperiment for details.


muncus:
with several pre-submit tests, no unexpected errors appeared with the loopvar experiment enabled.
This gives us significant confidence that loopvar will not cause significant breakage for us. 🎉

PR #3761, opened 2024-01-25
feat(spanner): add samples for spanner opentelemetry instrumentation

harshachinta:
This PR adds OpenTelemetry samples for spanner metrics and traces.


harshachinta:
@surbhigarg92 Please review the samples for OpenTelemetry.

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
```diff
@@ -0,0 +1,130 @@
...
+	// db = `projects/<project>/instances/<instance-id>/database/<database-id>`
+	ctx := context.Background()
+
+	// Create a new resource to uniquely identify the application
+	res, err := newResource()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Enable OpenTelemetry traces by setting environment variable GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING=opentelemetry at the start of the application
```
surbhigarg92, line 51: At the start of the application is not the correct wording here. Add something like below

`Enable OpenTelemetry traces by setting environment variable GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING to the case-insensitive
value `opentelemetry` before loading the client library.`

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
```diff
@@ -0,0 +1,130 @@
...
+		log.Fatal(err)
+	}
+	// Register tracer provider as global
+	otel.SetTracerProvider(tracerProvider)
+
+	// Create a new meter provider
+	meterProvider := getOtlpMeterProvider(ctx, res)
+	defer meterProvider.ForceFlush(ctx)
+
+	// Inject meter provider locally via ClientConfig when creating a spanner client.
```
surbhigarg92, line 68: Add the global option in the comment. `// Inject meter provider locally via ClientConfig when creating a spanner client or set via setMeterProvider`

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	stmt := spanner.Statement{SQL: `SELECT SingerId, AlbumId, AlbumTitle FROM Albums`}
+	iter := client.Single().Query(ctx, stmt)
+	defer iter.Stop()
+	for {
+		row, err := iter.Next()
```
surbhigarg92, line 79: This can be removed from opentelemetry usage sample.

spanner/spanner_snippets/spanner/spanner_opentelemetry_capture_query_stats_metric.go
```diff
@@ -0,0 +1,107 @@
+//go:build go1.20
```
surbhigarg92, line 1: We can merge 3 files into one file and have separate methods for gfe and query stats

spanner/spanner_snippets/spanner/spanner_opentelemetry_capture_query_stats_metric.go
```diff
@@ -0,0 +1,107 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spanner
+
+// [START spanner_opentelemetry_capture_query_stats_metric]
+// Ensure that your Go runtime version is supported by the OpenTelemetry-Go compatibility policy before enabling OpenTelemetry instrumentation.
```
surbhigarg92, line 21: Mention this comment in usage sample tag.

spanner/spanner_snippets/spanner/spanner_opentelemetry_capture_query_stats_metric.go
muncus, line 1: please do not use build directives in sample code. they are not visible in the regions displayed to the user, and they require additional maintenance. 
If this sample requires go1.20, create a separate module that requires go1.20 (maybe even in a less redundantly-named directory like `/spanner/opentelemetry` 😄 ).

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
```diff
@@ -0,0 +1,130 @@
...
+		var singerID, albumID int64
+		var albumTitle string
+		if err := row.Columns(&singerID, &albumID, &albumTitle); err != nil {
+			return err
+		}
+		fmt.Fprintf(w, "%d %d %s\n", singerID, albumID, albumTitle)
+	}
+}
+
+func getOtlpMeterProvider(ctx context.Context, res *resource.Resource) *metric.MeterProvider {
```
muncus, line 95: These helper functions are used in the other samples, but are not part of those regions. Was this deliberate?
we generally discourage this, because it is more confusing for readers to have to locate additional helper code.
I'd suggest putting the samples in the same file, and adding the region tags for the other samples around the helpers too (multiple uses of the same region in the same file are concatenated). this way, the helpers will appear alongside all the samples, not just the usage one.

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
```diff
@@ -0,0 +1,130 @@
...
+	otlpExporter, err := otlpmetricgrpc.New(ctx, otlpmetricgrpc.WithInsecure())
```
muncus, line 96: does this require `otlpmetricgrpc.WithInsecure()`? This does not seem like the right path to recommend, and I have not noticed this in other otel samples.

muncus (commented):
Are you aware of `https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport`? it may simplify some of your {Metric,Tracer}Provider initialization. the samples in `/opentelemetry` use it already.

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
harshachinta: Done

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
harshachinta: Done

spanner/spanner_snippets/spanner/spanner_opentelemetry_capture_query_stats_metric.go
harshachinta: Done

spanner/spanner_snippets/spanner/spanner_opentelemetry_capture_query_stats_metric.go
harshachinta: Done

spanner/spanner_snippets/spanner/spanner_opentelemetry_capture_query_stats_metric.go
harshachinta: Thanks for the suggestion. I have created a separate module as suggested.

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
harshachinta: The reason for adding this is because OTLP collector is non-TLS enabled and OTLP exporter is TLS enabled. This mismatch does not make the above sample directly work and need some authentication setup by following OpenTelemetry documentation. Adding this `insecure` will make the exporter non-TLS and customers can directly run the sample without issues.
I have got consensus from OpenTelemetry team on this.

PLMK if this is not the right way. I can remove it if you feel so.

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
harshachinta: Setting up OpenTelemetry for Spanner is same for all the samples. Adding this info in every sample will be redundant and also will increase the size of the samples. We have written our blog in such a way where customers will see setup steps initially and then separate sample for specific use case.

spanner/spanner_snippets/spanner/opentelemetry/spanner_opentelemetry_usage.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}
+}
+
+// [END spanner_opentelemetry_gfe_metric]
+
+// [START spanner_opentelemetry_capture_query_stats_metric]
+func captureQueryStatsMetric(ctx context.Context, mp metric.MeterProvider, client spanner.Client) error {
+	meter := mp.Meter(spanner.OtInstrumentationScope)
+	// Register query stats metric with OpenTelemetry to record the data.
```
surbhigarg92, line 142:
```suggestion
	// Register query stats metric with OpenTelemetry to record the data.
	// This should be done once before start recording the data.
```

spanner/spanner_snippets/spanner/opentelemetry/spanner_opentelemetry_usage.go
```diff
@@ -0,0 +1,182 @@
...
+			return err
+		}
+	}
+}
+
+// [END spanner_opentelemetry_gfe_metric]
+
+// [START spanner_opentelemetry_capture_query_stats_metric]
+func captureQueryStatsMetric(ctx context.Context, mp metric.MeterProvider, client spanner.Client) error {
+	meter := mp.Meter(spanner.OtInstrumentationScope)
```
surbhigarg92, line 141: Where is this OtInstrumentationScope defined ? We can hardcode the value "cloud.google.com/java" here to make it easy for users to understand.

surbhigarg92 (approved):

spanner/spanner_snippets/spanner/opentelemetry/spanner_opentelemetry_usage.go
harshachinta: Yes it is a public variable that can be accessed by customers. This will prevent copy paste errors.

harshachinta:
@muncus Can you please help review this PR on priority? We want to merge these changes by today.

spanner/spanner_snippets/spanner/spanner_opentelemetry_usage.go
harshachinta: I have removed this from the sample.

muncus (approved):
Samples look good.

One suggestion would be to move the files so that they are in the `/spanner/opentelemetry` directory, which will make them easier to link and easier to find by browsing github (and it lets you type 'spanner' a few less times!)

PR #3779, opened 2024-02-14
feat(vertexai): function calling

Deleplace:


Deleplace:
@muncus Hi Marc, could you merge this sample?

Deleplace:
FWIW this is an example of what "Function Calling" would look like in a real webapp.

The end user would ask questions in natural language and get answers in natural language. Transparently (under the hood) some "conversions" are performed, and the webapp backend is calling an external service  (the actual call is made by the webapp backend, not by the model).

image

polong-lin (approved):

JulienBreux (approved):
LGTM

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"os"
+
+	"cloud.google.com/go/vertexai/genai"
+)
+
+func main() {
```
telpirion, line 35: Don't declare a `main()` in golang-samples. Provide a meaningful function name with a docstring.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
```
telpirion, line 36: Pass `projectID` into the function.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+	location := "us-central1"
+	modelName := "gemini-pro"
+
+	prompt := "What's the weather like in Paris?"
+
+	if projectID == "" {
```
telpirion, line 42: Remove this check.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+		log.Fatal("require environment variable GOOGLE_CLOUD_PROJECT")
+	}
+
+	err := weather(os.Stdout, prompt, projectID, location, modelName)
+	if err != nil {
+		log.Fatalf("unable to generate: %v", err)
```
telpirion, line 48: Replace with call to `fmt.Fprintf(writer)`

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// function-calling shows an example of OpenAPI Function Call
+//
+// This Function Call involves 3 messages:
```
telpirion, line 17: Lower case "function call" unless this is a proper noun.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+	location := "us-central1"
+	modelName := "gemini-pro"
+
+	prompt := "What's the weather like in Paris?"
+
+	if projectID == "" {
+		log.Fatal("require environment variable GOOGLE_CLOUD_PROJECT")
+	}
+
+	err := weather(os.Stdout, prompt, projectID, location, modelName)
```
telpirion, line 46: Pass in a writer instead of stdout. We can't read from stdout in tests (unless we do some awkward pipe-ing).

vertexai/function-calling/functioncalling.go
telpirion: If "Function Call" is a proper noun -- that is, if it is a unique feature of Vertex -- provide a link to the documentation on cloud.google.com.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"cloud.google.com/go/vertexai/genai"
+)
+
+func main() {
+	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
+	location := "us-central1"
+	modelName := "gemini-pro"
+
+	prompt := "What's the weather like in Paris?"
```
telpirion, line 40: NIT: pass in prompt to function. That way we can configure inputs/outputs in the tests.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+	ctx := context.Background()
+	client, err := genai.NewClient(ctx, projectID, location)
+	if err != nil {
+		return fmt.Errorf("unable to create client: %v", err)
+	}
+	defer client.Close()
+
+	model := client.GenerativeModel(modelName)
+
+	params := &genai.Schema{
```
telpirion, line 65: Why are we creating a schema in memory? AIUI, developers would open an OpenAPI Swagger file that defines the schema. Opening a schema file from the internet makes more sense here.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	chat := model.StartChat()
+
+	fmt.Fprintf(w, "Question: %s\n", prompt)
+	resp, err := chat.SendMessage(ctx, genai.Text(prompt))
+	if err != nil {
+		return err
+	}
+	if len(resp.Candidates) == 0 || len(resp.Candidates[0].Content.Parts) == 0 {
```
telpirion, line 92: Insert line break before `||`.

vertexai/function-calling/functioncalling.go
telpirion, line 46: Why is has the actual sample (the calls to Gemini) been abstracted out of the main body? I recommend moving the entire sample into a single function.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+		return fmt.Errorf("empty response from model")
+	}
+
+	// The model has returned a function call to the declared
+	// function `getCurrentWeather` with "Paris" as the value for the
+	// argument `location`.
+	rb, _ := json.MarshalIndent(resp.Candidates[0].Content.Parts[0], "", "  ")
```
telpirion, line 99: Use more descriptive variable names in samples. I don't know what 'rb' is unless I look at the ref docs for `json.MarshalIndent()`.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	rb, _ = json.MarshalIndent(funresp, "", "  ")
+	fmt.Fprintf(w, "Function Call response sent to the model:\n%s\n\n", string(rb))
+
+	// And provide the Function Call response to the model
+	resp, err = chat.SendMessage(ctx, funresp)
+	if err != nil {
+		return err
+	}
+	if len(resp.Candidates) == 0 || len(resp.Candidates[0].Content.Parts) == 0 {
```
telpirion, line 118: Insert line break before `||`.

telpirion (changes requested):
This sample requires a test. Look at other functional tests in this repo for examples on how to write a test.

vertexai/function-calling/functioncalling.go
Deleplace: According to [this doc page](https://cloud.google.com/vertex-ai/docs/generative-ai/multimodal/function-calling), lower case should be prefered.

vertexai/function-calling/functioncalling.go
Deleplace: After `||` ?

vertexai/function-calling/functioncalling.go
Deleplace: Unfortunately, I'm not aware of an existing ready-to-use parser from bytes to `genai.Schema`

Deleplace:
I made the requested changes

vertexai/function-calling/functioncalling.go
telpirion: Fair enough. I would put a comment before this block that explicitly states that you're creating an OpenAPI schema in memory.

I do believe that there are a few Go libraries that can parse this -- for example, libopenapi -- but it's reasonable to skip that step for demonstration purposes.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,114 @@
...
+		return err
+	}
+	if len(resp.Candidates) == 0 ||
+		len(resp.Candidates[0].Content.Parts) == 0 {
+		return errors.New("empty response from model")
+	}
+
+	// The model has returned a function call to the declared function `getCurrentWeather`
+	// with a value for the argument `location`.
+	jsondata, _ := json.MarshalIndent(resp.Candidates[0].Content.Parts[0], "", "  ")
```
telpirion, line 82: Don't throw away the error here.

vertexai/function-calling/functioncalling.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// Create a function call response, to simulate the result of a call to a
+	// real service
+	funresp := &genai.FunctionResponse{
+		Name: "getCurrentWeather",
+		Response: map[string]any{
+			"currentWeather": "sunny",
+		},
+	}
+	jsondata, _ = json.MarshalIndent(funresp, "", "  ")
```
telpirion, line 93: Same as above. Be sure to handle the error.

telpirion (approved):
This is much better, thank you!

I will approve this PR now. Be sure to address the comments about throwing away errors.

PR #3800, opened 2024-02-23
feat(pubsub): add create kinesis topic and update samples

hongalex:
This PR adds the samples for the new kinesis ingestion feature.


fayssalmartanigcp (approved):

pubsub/topics/create_topic_kinesis_ingestion.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/pubsub"
+)
+
+func createTopicWithKinesisIngestion(w io.Writer, projectID, topicID, streamARN, consumerARN, awsRoleARN, gcpSA string) error {
+	// projectID := "my-project-id"
+	// topicID := "my-topic"
+	// streamARN := "stream-arn"
```
tritone, line 29: It looks like from the test that all of these are static values; if so create them as local variables rather than args. https://googlecloudplatform.github.io/samples-style-guide/#minimal-arguments

pubsub/topics/update_topic_type.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/pubsub"
+)
+
+func updateTopicType(w io.Writer, projectID, topicID, streamARN, consumerARN, awsRoleARN, gcpSA string) error {
+	// projectID := "my-project-id"
+	// topicID := "my-topic"
+	// streamARN := "stream-arn"
```
tritone, line 29: Same comment here re: extra args.

pubsub/topics/topics_test.go
```diff
@@ -293,3 +295,26 @@ func TestDelete(t *testing.T) {
...
 }
+
+func TestTopicKinesis(t *testing.T) {
+	tc := testutil.SystemTest(t)
+	buf := new(bytes.Buffer)
+
+	// Use the pstest fake with emulator settings since Pub/Sub service expects real AWS Kinesis
+	// resources, which we cannot provide in a samples test.
+	srv := pstest.NewServer()
+	os.Setenv("PUBSUB_EMULATOR_HOST", srv.Addr)
```
tritone, line 306: I believe this permanently sets the env variable which may affect other tests' execution. Recommend using https://pkg.go.dev/testing#T.Setenv instead.

pubsub/topics/create_topic_kinesis_ingestion.go
hongalex: I initialized the values in the sample itself, is this what you were suggesting?

tritone (approved):

PR #3808, opened 2024-02-28
feat(spanner): add transaction timeout sample

olavloite:
Adds a sample for setting a timeout for an entire read/write transaction when using Spanner. This sample is needed for the Spanner documentation pages, as this is a recurring question from Spanner customers.

### DIFF
--- a/spanner/spanner_snippets/spanner/spanner_transaction_timeout.go
+++ b/spanner/spanner_snippets/spanner/spanner_transaction_timeout.go
@@ -1,0 +1,90 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


rahul2393 (approved):

spanner/spanner_snippets/spanner/spanner_transaction_timeout.go
```diff
@@ -0,0 +1,90 @@
...
+	ctx := context.Background()
+	client, err := spanner.NewClient(ctx, db)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	// Create a context with a 60-second timeout and use this context to run a read/write transaction.
+	// This context timeout will be applied to the entire transaction, and the transaction will fail
+	// if it cannot finish within the specified timeout value.
```
grayside, line 41: praise: Providing great context with this comment, thank you.

suggestion: Since the next line is the primary lesson of the sample, it may be worth making it even more explicit that the spanner library owns applying this timeout. Suggesting this because folks less familiar with how Go context & timeouts work might be confused.

grayside (approved):

spanner/spanner_snippets/spanner/spanner_transaction_timeout.go
olavloite: Thanks. Added a couple of clarifying comments for that as well.

PR #3851, opened 2024-03-08
feat(vertexai): Align Function calling chat sample with the Python sample.

Deleplace:
`functionCallsChat` prints:

```
Question: Do you have the Pixel 8 Pro in stock?
function call generated by the model:
  {
	  "Name": "getProductSku",
	  "Args": {
	    "productName": "Pixel 8 Pro"
	  }
	}
function call response sent to the model:
	{
	  "Name": "getProductSku",
	  "Response": {
	    "in_stock": "yes",
	    "sku": "GA04834-US"
	  }
	}

Answer generated by the model:
	"Yes, we have the Pixel 8 Pro in stock. The SKU is GA04834-US."

Question: Do you have the Pixel 8 Pro in stock?
function call generated by the model:
	{
	  "Name": "getStoreLocation",
	  "Args": {
	    "location": "Mountain View, CA"
	  }
	}
function call response sent to the model:
	{
	  "Name": "getStoreLocation",
	  "Response": {
	    "store": "2000 N Shoreline Blvd, Mountain View, CA 94043, US"
	  }
	}

Answer generated by the model:
	"Yes, there is a store in Mountain View, CA that you can visit to try out the Pixel 8 Pro. The store is located at 2000 N Shoreline Blvd, Mountain View, CA 94043, US."
```


Deleplace:
The region tag has changed. This is because Chat and Text modalities used to not have a distinct tag, now they do. In Go we are currently able to provide a sample for Chat modality, but not for Text modality.

grayside (approved):

grayside (changes requested):
Since the region tag is in place in docs, in order to merge the docs need to be revised first or both region tags should be left in place for a transitional period.

Deleplace:
This makes sense, thank you. Now keeping both region tags for a transitional period.

I will ask separately for the docs to be revised.

grayside (approved):

PR #3961, opened 2024-03-14
feat(pubsub): add otel tracing samples

hongalex:
Adds samples for enabling tracing of Pub/Sub client with OpenTelemetry and Cloud Trace.


telpirion (changes requested):
Looks good --- just needs tests.

telpirion (approved):
Just fix that linter error and SHIP IT!

PR #3966, opened 2024-03-15
fix: address the use of InsecureSkipVerify in sample

iennae:
## Description

Addresses https://github.com/GoogleCloudPlatform/golang-samples/security/code-scanning/5

## Checklist

in order for the sample to work, requires this to be set to true. Not a pattern we'd want to encourage in use so adding a comment to help encourage change when copied.

### DIFF
--- a/cloudsql/mysql/database-sql/connect_tcp.go
+++ b/cloudsql/mysql/database-sql/connect_tcp.go
@@ -60,6 +60,8 @@ func connectTCPSocket() (*sql.DB, error) {
 	// For deployments that connect directly to a Cloud SQL instance without
 	// using the Cloud SQL Proxy, configuring SSL certificates will ensure the
 	// connection is encrypted.
+	// Change InsecureSkipVerify to false in production environments to
+	// verify the certificate and host name in the certificate.
 	if dbRootCert, ok := os.LookupEnv("DB_ROOT_CERT"); ok { // e.g., '/path/to/my/server-ca.pem'
 		var (
 			dbCert = mustGetenv("DB_CERT") // e.g. '/path/to/my/client-cert.pem'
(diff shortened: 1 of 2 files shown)


enocom (commented):
This is actual a symptom of Cloud SQL's public key infrastructure. By default, the Go TLS stack will attempt to verify the server's identity by inspecting the server's certificate. Because the server's certificate has a non-standard common name (e.g., `myinstance:myproject`), the verification will fail. Technically there is a pseudo DNS name in the server's certificate, but customers have to inspect their server certificate to know that value. If they knew that value, they could remove `InsecureSkipVerify` and add a `ServerName` field in the TLS config.

iennae:
@enocom can you recommend a suggested update to the comment? :)

enocom:
How about we move this comment to just above `InsecureSkipVerify` and say:

InsecureSkipVerify and a custom VerifyPeerCertificate function is required to handle Cloud SQL's custom certificates.
As an alternative it's also possible to inspect the server certificate and extract the SAN field and use that a `ServerName`
while removing InsecureSkipVerify and VerifyPeerCertificate.

It's not a great situation that Cloud SQL is working on fixing. Until then, the workaround is a headache.

muncus:
this PR has been idle for a while, so i'm converting it to draft. Feel free to un-draft it when it is ready for review.

enocom:
FWIW I think we should just change the proposed comment to something like what I proposed above.

cc @jackwotherspoon who may have additional ideas.

subfuzion:
@enocom @iennae PTAL

subfuzion:
Kokoro failing on: cloudsql/mysql/database-sql/cmd/functions
module jackc/pgx/v5@v5.6.0 requires go 1.20
```
/go/pkg/mod/github.com/jackc/pgx/v5@v5.6.0/pgconn/pgconn.go:153:100: undefined: errors.Join
/go/pkg/mod/github.com/jackc/pgx/v5@v5.6.0/pgconn/pgconn.go:159:57: undefined: errors.Join
note: module requires Go 1.20
```
https://btx.cloud.google.com/invocations/b5ed80ba-d1de-437c-a7f3-3c6e5b161777/targets/cloud-devrel%2Fgo%2Fgolang-samples%2Fpresubmit%2Fearliest-version;config=default/log

enocom (approved):

PR #3972, opened 2024-03-18
test(storagetransfer): add retries to all storagetransfer tests

JesseLovelace:
This adds retries to all the storagetransfer tests, which should fix the flake issues we've been having

Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/3782
Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/3763
Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/3714
Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/3712
Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/3593
Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/3274
Fixes https://github.com/GoogleCloudPlatform/golang-samples/issues/3736


storagetransfer/storagetransfer_test.go
```diff
@@ -151,12 +151,14 @@ func TestQuickstart(t *testing.T) {
...
-	defer cleanupSTSJob(resp, tc.ProjectID)
+	testutil.Retry(t, 5, time.Second, func(r *testutil.R) {
+		resp, err := quickstart(buf, tc.ProjectID, gcsSourceBucket, gcsSinkBucket)
+		defer cleanupSTSJob(resp, tc.ProjectID)
 
-	if err != nil {
-		t.Errorf("quickstart: %#v", err)
-	}
+		if err != nil {
+			t.Errorf("quickstart: %#v", err)
```
tritone, line 159: Here and elsewhere, use `r` rather than `t` when inside the inner func.

storagetransfer/storagetransfer_test.go
```diff
@@ -151,12 +151,14 @@ func TestQuickstart(t *testing.T) {
 	tc := testutil.SystemTest(t)
 
 	buf := new(bytes.Buffer)
```
tritone, line 153: I think the buffer alloc and check should also go inside the retry loops (here and elsewhere). This means that we retry if the output is unexpected as well.

tritone (commented):
Two changes to make sure we are fully covered here; otherwise looks good.

storagetransfer/storagetransfer_test.go
```diff
@@ -150,91 +150,90 @@ func TestMain(m *testing.M) {
...
 		}
-	})
 
-	got := buf.String()
-	if want := "transferJobs/"; !strings.Contains(got, want) {
-		t.Errorf("quickstart: got %q, want %q", got, want)
-	}
+		got := buf.String()
+		if want := "transferJobs/"; !strings.Contains(got, want) {
+			t.Errorf("quickstart: got %q, want %q", got, want)
```
tritone, line 165: These checks also need to trigger `r.Errorf` rather than t (here and elsewhere)

tritone (approved):
One more small thing, otherwise lgtm

tritone (approved):

PR #4001, opened 2024-04-19
test(testutil): add new function for creating a test bucket

BrennaEpp:
Part 1 to bucket creation cleanup.
Part 2 will be a PR replacing all references to CleanBucket with TestBucket or CreateTestBucket.
Part 3 will remove the CleanBucket function.

Deprecates CleanBucket; a followup PR will remove all references to this function. CleanBucket can run into problems of availability due to fast creation after deletion. Instead, unique randomly generated bucket names should be used.

To replace CleanBucket, TestBucket provides a simple way to create a bucket for testing, creating this random name and taking charge of the cleanup.

## Checklist


internal/testutil/storage_cleaner.go
```diff
@@ -28,15 +28,43 @@ import (
...
+		if err := DeleteBucketIfExists(ctx, client, bucketName); err != nil {
+			log.Printf("Bucket.Delete(%q): %v", bucketName, err)
+		}
+	})
+	return bucketName
 }
 
 // CleanBucket creates a new bucket. If the bucket already exists, it will be
 // deleted and recreated.
+// Deprecated: use TestBucket or CreateTestBucket instead.
```
tritone, line 67: Just FYI, there are a bunch of uses of this outside of GCS samples -- those will have to be replaced as well.

internal/testutil/storage_cleaner.go
```diff
@@ -28,15 +28,43 @@ import (
...
+		t.Fatalf("storage.NewClient: %v", err)
+	}
+	t.Cleanup(func() { client.Close() })
+	return CreateTestBucket(ctx, t, client, projectID, prefix)
+}
+
+// CreateTestBucket creates a new bucket with the given prefix and registers a
+// cleanup function to delete the bucket and any objects it contains.
+// It is equivalent to TestBucket but allows Storage Client re-use.
+func CreateTestBucket(ctx context.Context, t *testing.T, client *storage.Client, projectID, prefix string) string {
```
tritone, line 48: Maybe re-name these to indicate what the difference is (e.g. CreateBucket vs CreateBucketWithClient)? Also is it actually important to have both functions available? Creating an extra client in a test is not much overhead.

storage/buckets/buckets_test.go
```diff
@@ -702,8 +702,6 @@ func TestRPO(t *testing.T) {
 		t.Fatalf("createBucketTurboReplication: %v", err)
 	}
 
-	testutil.WaitForBucketToExist(ctx, t, bucket)
```
tritone, line 705: Is this change relevant to this PR?

internal/testutil/storage_cleaner.go
BrennaEpp: Yes, I have a follow up ready to replace those.

internal/testutil/storage_cleaner.go
BrennaEpp: So I don't want to change the signature of the `CreateTestBucket` function since it will break tests. But I agree that we could just create the extra client instead.

tritone (approved):

PR #4111, opened 2024-04-25
feat(vertexai): multimodal audio

Deleplace:
Go version of the Python sample at https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/generative_ai/gemini_audio.py

Region tags `generativeaionvertexai_gemini_audio_summarization` and `generativeaionvertexai_gemini_audio_transcription`


gericdong (approved):

vertexai/multimodal-audio/multimodalaudio.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// [END generativeaionvertexai_gemini_audio_transcription]
+// [END generativeaionvertexai_gemini_audio_summarization]
+
+// [START generativeaionvertexai_gemini_audio_summarization]
+// summarizeAudio generates a response into w, based upon the prompt
+// and audio provided.
+// audio is a Google Cloud Storage path starting with "gs://"
+func summarizeAudio(w io.Writer, prompt, audio, projectID, location, modelName string) error {
```
jba, line 38: Too many string args. See my comment on #4111.

vertexai/multimodal-audio/multimodalaudio.go
```diff
@@ -0,0 +1,123 @@
...
+	ctx := context.Background()
+
+	client, err := genai.NewClient(ctx, projectID, location)
+	if err != nil {
+		return fmt.Errorf("unable to create client: %v", err)
+	}
+	defer client.Close()
+
+	model := client.GenerativeModel(modelName)
+	model.SetTemperature(0.4)
```
jba, line 56: Is there a reason for this particular value? For changing the default temperature at all?

vertexai/multimodal-audio/multimodalaudio.go
```diff
@@ -0,0 +1,123 @@
...
+	return nil
+}
+
+// [END generativeaionvertexai_gemini_audio_summarization]
+
+// [START generativeaionvertexai_gemini_audio_transcription]
+// transcriptAudio generates a response into w, based upon the prompt
+// and audio provided.
+// audio is a Google Cloud Storage path starting with "gs://"
+func transcriptAudio(w io.Writer, prompt, audio, projectID, location, modelName string) error {
```
jba, line 84: transcribeAudio

jba (changes requested):

vertexai/multimodal-audio/multimodalaudio.go
Deleplace: No good reason, thanks for finding this!

My understanding is the default temp in the client lib model is `nil`, which means the endpoint model's default will be used (it depends on the model name).  In this case, not setting the temperature is the right thing to do, and it does not mean "temp=0".

jba: Correct. IIRC the default is like 0.9.

grayside: In an earlier sample, temperature was included as a parameter and I suggested hard-coding it. This doesn't need to happen everywhere, but it does surface that the toggle exists while keeping the distraction from temperature minimal. I think including this is up to the overall design/scope of the sample set, and is ideally consistent across languages.

grayside: @gericdong

jba: I just don't want anyone to get the idea that 0.4 is magic in some way. I'm okay with showing that it can be set, as long as there is a comment that prevents something thinking that the exact number is relevant.

jba (approved):

grayside (approved):

PR #4121, opened 2024-05-03
feat(vertexai): system instruction

Deleplace:
Go version of the Python sample at https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/generative_ai/gemini_system_instruction.py

Region tag `generativeaionvertexai_gemini_system_instruction`

Note: using `cloud.google.com/go/vertexai v0.7.2-0.20240430000329-84e3236355de` as there is no tag yet that includes `model.SystemInstruction`

**Edit:** now using `cloud.google.com/go/vertexai v0.8.0` — thank you jba


jba (approved):

jba (approved):

vertexai/system-instruction/systeminstruction.go
```diff
@@ -0,0 +1,70 @@
...
+	// The System Instruction is set at model creation
+	model := client.GenerativeModel(modelName)
+	model.SystemInstruction = &genai.Content{
+		Parts: []genai.Part{genai.Text(instruction)},
+	}
+
+	chat := model.StartChat()
+	res, err := chat.SendMessage(ctx, genai.Text(prompt))
+	if err != nil {
+		return err
```
grayside, line 59: suggestion: Other samples prefix this error, is it not needed here?

```go
return fmt.Errorf("unable to generate contents: %w", err)
```

vertexai/system-instruction/systeminstruction.go
```diff
@@ -0,0 +1,70 @@
...
+	defer client.Close()
+
+	// The System Instruction is set at model creation
+	model := client.GenerativeModel(modelName)
+	model.SystemInstruction = &genai.Content{
+		Parts: []genai.Part{genai.Text(instruction)},
+	}
+
+	chat := model.StartChat()
+	res, err := chat.SendMessage(ctx, genai.Text(prompt))
```
grayside, line 57: Most of the other Go samples use `model.GenerateContent`. The canonical python sample for this region tag uses the python equivalent of `model.generate_content`.

question: Should this be switched for consistency?
suggestion: If there's a technical reason to use this instead of GenerateContent, please add an explanation comment.

grayside (commented):
Commenting with pending question on `model.GenerateContent`.

vertexai/system-instruction/systeminstruction.go
Deleplace: Good poijt, thanks

grayside (approved):

PR #4149, opened 2024-05-23
feat(otel-instrumentation): Add custom instrumentation 

psx95:
## Description
Updates the opentelemetry instrumentation example to showcase custom instrumentation - metrics & traces, along with exemplars.

Generated metrics also contain exemplars.


opentelemetry/instrumentation/app/work.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"math/rand"
+	"time"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+const name = "work"
```
dashpole, line 27: Should we make this a full name? `github.com/GoogleCloudPlatform/golang-samples/opentelemetry/instrumentation/app`

opentelemetry/instrumentation/app/server.go
```diff
@@ -28,11 +27,9 @@ import (
 // the number of milliseconds slept as its response.
 // [START opentelemetry_instrumentation_handle_single]
 func handleSingle(w http.ResponseWriter, r *http.Request) {
-	sleepTime := time.Duration(100+rand.Intn(100)) * time.Millisecond
+	elapsedTime := doWork(r.Context(), r.Host)
```
dashpole, line 30: nit: maybe just pass in `r`?

opentelemetry/instrumentation/app/server.go
```diff
@@ -28,11 +27,9 @@ import (
 // the number of milliseconds slept as its response.
 // [START opentelemetry_instrumentation_handle_single]
 func handleSingle(w http.ResponseWriter, r *http.Request) {
```
dashpole, line 29: I wonder if we should put this on the handleMulti endpoint so we only add one extra span?

opentelemetry/instrumentation/app/work.go
```diff
@@ -0,0 +1,53 @@
...
+	meter         = otel.Meter(name)
+	tracer        = otel.Tracer(name)
+	workHistogram metric.Int64Histogram
+)
+
+func init() {
+	var err error
+	workHistogram, err = meter.Int64Histogram("example.histogram",
+		metric.WithDescription("Sample histogram"),
+		metric.WithUnit("1"))
```
dashpole, line 39: It would be nice for these to feel a tiny bit more real.  E.g. make up something that this is measuring, with an appropriate unit

dashpole: Or, if this is moved to the /multi endpoint, make this a histogram of the number of subrequests.

dashpole: If it is kept at the single endpoint, make it a histogram of the time slept.

opentelemetry/instrumentation/app/work.go
```diff
@@ -0,0 +1,53 @@
...
+
+	if err != nil {
+		panic(err)
+	}
+}
+
+// doWork simulates a some job being triggerred in response to an API call to the server.
+// This function computes 10 random values and records them into a histogram which can be
+// later visualized as a distribution.
+func doWork(ctx context.Context, host string) time.Duration {
```
dashpole, line 49: Lets come up with a better name (e.g. randomSleep() or computeSubrequests()) that says what it is actually doing

opentelemetry/instrumentation/app/work.go
```diff
@@ -0,0 +1,53 @@
...
+	start := time.Now()
+	hostValue := attribute.String("host.value", host)
+
+	// simulate the overall work by sleeping
+	sleepTime := time.Duration(100+rand.Intn(100)) * time.Millisecond
+	time.Sleep(sleepTime)
+
+	// wrap the random number generation in a span - to better visualize the time spent in this part
+	traceCtx, span := tracer.Start(ctx, "doWork")
```
dashpole, line 58: it is convention to overwrite the existing context
```suggestion
	ctx, span := tracer.Start(ctx, "doWork")
```

opentelemetry/instrumentation/app/work.go
```diff
@@ -0,0 +1,53 @@
...
+	for i := 0; i < 10; i++ {
```
dashpole, line 59: Do we need the loop here?  Or do we not get enough data in the histogram without it?

opentelemetry/instrumentation/app/work.go
```diff
@@ -0,0 +1,53 @@
...
+		randomNum := rand.Intn(100)
+		workHistogram.Record(traceCtx, int64(randomNum), metric.WithAttributes(hostValue))
+	}
+	span.End()
+
+	elapsedTime := time.Since(start)
```
dashpole, line 65: why not return sleepTime?

opentelemetry/instrumentation/app/work.go
psx95: I think the loop can be removed, we are generating enough exemplars.

opentelemetry/instrumentation/app/work.go
psx95: I tried to kept the implementation close to what we already had, we were computing the the elapsed time. So the `doWork` function was representing fictional work.

However, based on the comments, I think we would be moving in a different direction which would probably cause this function to be `randomSleep` - then it would be appropriate to return the sleep time.

opentelemetry/instrumentation/app/work.go
psx95: Is that the convention in Go ? I was following suit from [rolldice example](https://github.com/open-telemetry/opentelemetry-go/blob/main/example/dice/rolldice.go).

opentelemetry/instrumentation/app/work.go
psx95: I can re-purpose the histogram to "measure" something. The rationale behind this was that we already have metrics like `server_duration_millis` that cover this use-case.

So this kind of work is actually a different use-case. Those random numbers being generated could represent different user-interactions in a production-app and the histogram shows the distribution for each of those interactions.

opentelemetry/instrumentation/app/server.go
psx95: Based on the position of this comment, I'm not sure what you're referring to. 

Are you suggesting to remove custom span generation from `doWork` to `handleMulti` ?

opentelemetry/instrumentation/app/work.go
aabmass, line 1: I think you need region tags or the doc write up

opentelemetry/instrumentation/app/work.go
aabmass: +1 I would put the span around the sleep

opentelemetry/instrumentation/app/work.go
psx95: Yup, wanted thoughts on this PR to finalize the code before doc write-up. Can deal with region tags in a different PR.

opentelemetry/instrumentation/app/work.go
psx95: Both your suggestions seem valid to me though.

I can do both and depending on how the doc write-up goes we can choose whichever one is more appropriate.

opentelemetry/instrumentation/app/server.go
psx95: Done

opentelemetry/instrumentation/app/work.go
psx95: Put span around sleep & removed the loop.

opentelemetry/instrumentation/app/work.go
psx95: FYI: For /multi endpoint the histogram tracking the number of requests will not always have a good distribution since the range of number of requests is quite low - 3 to 7. Even after explicit buckets, with sensible boundaries, there weren't many exemplars. But I feel its WAI.

dashpole: You can set the bucket boundaries if you want, rather than using the default.  E.g. buckets of [1, 2, 3, 4, 5, 6, 7, 8, 9, 10].

opentelemetry/instrumentation/app/server.go
dashpole: If you look at the example trace, IIRC, it currently can have up to 1 server span for `/multi `+ 14 spans for single (client + server).  So adding a new span to `/single` will add up to 7 new spans to the trace, which will make it visually much larger.  If you add a new span to `/multi`, it will only add a single span to the trace.

I'm wondering if up to 22 spans will make the trace seem cluttered.

opentelemetry/instrumentation/app/work.go
dashpole: Yes, it is the convention: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-go-contrib%20scopeName&type=code

I opened an issue in the otel-go repo as well to update the rolldice example

psx95: I will make this change then.

opentelemetry/instrumentation/app/server.go
psx95: I see, I can certainly remove the custom span from `/single` to keep things simple. I think a single custom span in `/multi` should be good enough to demonstrate custom instrumentation for tracing.

opentelemetry/instrumentation/app/work.go
psx95: I thought these bucket sizes might become too fine grained, the current ones I have are [2,4,6,8,10] *(which are still a little fine-grained)*. 

Are they setting a good example for users?

dashpole: I think it is fine for now.  We can remove it once GMP supports exponential histograms

opentelemetry/instrumentation/app/work.go
```diff
@@ -0,0 +1,87 @@
...
+// visualized as a distribution.
+func randomSleep(r *http.Request) time.Duration {
+	hostValue := attribute.String("host.value", r.Host)
+
+	// simulate the work by sleeping
+	sleepTime := time.Duration(100+rand.Intn(100)) * time.Millisecond
+	time.Sleep(sleepTime)
+
+	// record time slept
+	sleepHistogram.Record(r.Context(), sleepTime.Seconds(), metric.WithAttributes(hostValue))
```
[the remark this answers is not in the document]
dashpole: great

dashpole (commented):
just needs location tags

dashpole (approved):

PR #4171, opened 2024-06-07
feat(compute): samples/compute_disk_list sample

irudykss:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


compute/disks/list.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// listDisks lists disks in a project zone.
+func listDisks(w io.Writer, projectID, zone, filter string) error {
+	// projectID := "your_project_id"
+	// zone := "us-central1-a"
+	// filter := ""
+
+	// Formatting for filters:
+	//https://cloud.google.com/python/docs/reference/compute/latest/google.cloud.compute_v1.types.ListDisksRequest
```
muncus, line 34: please use a golang link here, instead of python.

compute/disks/disks_test.go
```diff
@@ -465,4 +465,40 @@ func TestComputeDisksSnippets(t *testing.T) {
 		// Cannot clean up the disk just yet because it must be done after the VM is terminated.
 		// It will be done by deleteInstance function.
 	})
+	t.Run("Create a test disk and list it", func(t *testing.T) {
```
muncus, line 468: nit: this test does not create a disk. a more accurate comment would be "list disks"

compute/disks/disks_test.go
```diff
@@ -465,4 +465,40 @@ func TestComputeDisksSnippets(t *testing.T) {
...
+		var buf bytes.Buffer
+
+		if err := listDisks(&buf, tc.ProjectID, zone, ""); err != nil {
+			t.Fatalf("listDisks got err: %v", err)
```
muncus, line 472:
```suggestion
			t.Errorf("listDisks got err: %v", err)
```
while it is true this test fails on an error here, we can still continue. so this should be `Errorf` instead. The same is true for the two Fatalfs below.

muncus (commented):
minor changes needed, but overall this looks good, thanks!

compute/disks/disks_test.go
irudykss: fixed

compute/disks/disks_test.go
irudykss: fixed

compute/disks/list.go
irudykss: fixed

compute/disks/list.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// listDisks lists disks in a project zone.
+func listDisks(w io.Writer, projectID, zone, filter string) error {
+	// projectID := "your_project_id"
+	// zone := "us-central1-a"
+	// filter := ""
+
+	// Formatting for filters:
+	// https://cloud.google.com/php/docs/reference/cloud-compute/latest/V1.ListDisksRequest
```
muncus, line 34: this is still the wrong link.

muncus:
Please avoid force-pushing to PR branches, as it makes it much harder to identify which parts of the PR have been changed since the last review.

doing a `git pull` prior to your push should avoid the errors that prevented a non-force push.

compute/disks/list.go
irudykss: lol, missed that, thanks 👍 
added inline explanation 👍

compute/disks/list.go
```diff
@@ -31,7 +31,7 @@ func listDisks(w io.Writer, projectID, zone, filter string) error {
...
...u can use filtering on nested fields to filter based on resource labels. To filter on multiple expressions, provide each separate expression within parentheses. For example: ``` (scheduling.automaticRestart = true) (cpuPlatform = "Intel Skylake") ``` By default, each expression is an `AND` expression. However, you can include `AND` and `OR` expressions explicitly. For example: ``` (cpuPlatform = "Intel Skylake") OR (cpuPlatform = "Intel Broadwell") AND (scheduling.automaticRestart = true) ``` If you want to use a regular expression, use the `eq` (equal) or `ne` (not equal) operator against a single un-parenthesized expression with or without quotes or against multiple parenthesized expressions. Examples: `fieldname eq unquoted literal` `fieldname eq 'single quoted literal'` `fieldname eq "double quoted literal"` `(fieldname1 eq literal) (fieldname2 ne "literal")` The literal value is interpreted as a regular expression using Google RE2 library syntax. The literal value must match the entire field. For example, to filter for instances that do not end with name "instance", you would use `name ne .*instance`. You cannot combine constraints on multiple fields using regular expressions.
```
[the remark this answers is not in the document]
irudykss: NP 👍  Changed to python link

muncus (approved):

PR #4214, opened 2024-06-25
feat(testing): run tests in modified packages only

muncus:
current testing logic also ensures modules *under* these packages are
also tested.


Deleplace:
I tried to execute the steps but encountered these problems:

- `GO_CHANGED_PKGS` would contain a list of go files, not a list of folders
- `GO_CHANGED_MODULES` does not contain the modified modules

testing/kokoro/system_tests.sh
```diff
@@ -54,15 +54,16 @@ SIGNIFICANT_CHANGES=$(git --no-pager diff --name-only main..HEAD | grep -Ev '(\.
...
 
 # List all modules in changed directories.
 # If running on main will collect all modules in the repo, including the root module.
 # shellcheck disable=SC2086
-GO_CHANGED_MODULES="$(find ${CHANGED_DIRS:-.} -name go.mod)"
-# If we didn't find any modules, use the root module.
-GO_CHANGED_MODULES=${GO_CHANGED_MODULES:-./go.mod}
-# Exclude the root module, if present, from the list of sub-modules.
-GO_CHANGED_SUBMODULES=${GO_CHANGED_MODULES#./go.mod}
+GO_CHANGED_MODULES="$(find ${GO_CHANGED_PKGS:-.} -name go.mod | xargs --no-run-if-empty dirname)"
```
Deleplace, line 62: I tried to execute the steps but encountered these problems:

- `GO_CHANGED_PKGS` would contain a list of go files, not a list of folders
- `GO_CHANGED_MODULES` does not contain the modified modules

Deleplace (changes requested):

testing/kokoro/system_tests.sh
Deleplace: Maybe `dirname` could help:
```
$ dirname vertexai/chat/chat.go
vertexai/chat
```

muncus: Thanks for taking a look at this!

I've added dirname to this pipeline, and to make future changes easier, i've allowed us to set $GIT_CHANGES locally to enable better testability here.

So running `GIT_CHANGES="functions/functionsv2/firebase/hellofirestore/go.mod compute/instances/foo.go" sh testing/kokoro/system_tests.sh` will show (among other things): `GO_CHANGED_PKGS='compute/instances
functions/functionsv2/firebase/hellofirestore'`

the script stops when it tries to read the nonexistant kokoro credentials, but this is still an improvement.

testing/kokoro/system_tests.sh
```diff
@@ -44,25 +44,27 @@ export GOLANG_SAMPLES_E2E_TEST=""
...
   echo "No diffs detected. This is unexpected - check above for additional error messages."
   exit 2
 fi
-SIGNIFICANT_CHANGES=$(git --no-pager diff --name-only main..HEAD | grep -Ev '(\.md$|^\.github)' || true )
+SIGNIFICANT_CHANGES=$(echo $GIT_CHANGES | grep -Ev '(\.md$|^\.github)' || true )
 # CHANGED_DIRS is the list of significant top-level directories that changed,
 # but weren't deleted by the current PR.
 # CHANGED_DIRS will be empty when run on main.
 CHANGED_DIRS=$(echo "$SIGNIFICANT_CHANGES" | tr ' ' '\n' | grep "/" | cut -d/ -f1 | sort -u | tr '\n' ' ' | xargs --no-run-if-empty ls -d 2>/dev/null || true)
+GO_CHANGED_PKGS=$(echo "$SIGNIFICANT_CHANGES" | tr ' ' '\n' | grep "/" | tr '\n' ' ' | xargs --no-run-if-empty dirname | xargs --no-run-if-empty ls -d 2>/dev/null || true)
```
grayside, line 58: question: Not recognizing the logic indicating these directories are necessarily Go packages. What am I missing?

grayside (commented):
Overall this looks good to me. I wasn't able to get some of the specific script elements working locally on Mac but it makes sense.

testing/kokoro/system_tests.sh
muncus: A Package is just a directory that contains go code. a Module is a grouping of packages (see https://go.dev/ref/mod#modules-overview).

We are not looking for the module that *contains* the modified packages, we are just testing the modified packages themselves. We believe this is safe due to the rarity of inter-package dependencies, and the use of exhaustive nightly testing.

if we do experience breakage, we can move toward a "find the containing module" approach, but that would be more complex.

Deleplace (dismissed):
Marc, I tested the modified changes in bash (macOS) and now I can confirm the env var values look good 👍

Before LGTM, I'd want to understand if `GO_CHANGED_MODULES` is supposed to be used at all?

testing/kokoro/system_tests.sh
```diff
@@ -44,25 +44,27 @@ export GOLANG_SAMPLES_E2E_TEST=""
 # allow files to be owned by a different user than our current uid.
 # Kokoro runs a double-nested container, and UIDs may not match.
 git config --global --add safe.directory $(pwd)
-GIT_CHANGES=$(git --no-pager diff --name-only main..HEAD)
+# Allow $GIT_CHANGES to be set in the env, enabling local testing of the change detection below.
+GIT_CHANGES=${GIT_CHANGES:-$(git --no-pager diff --name-only main..HEAD)}
```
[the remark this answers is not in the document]
muncus: `main` here is using the current status of the `main` branch in your checkout - which is likely missing some of the newest changes to `origin/main`. you can make those changes disappear by doing a `git pull` on your `main` branch, then switching back to the PR branch.

This is not a problem in CI, where we're using a fresh clone of the repo.

We may be able to address this by using `origin/main` instead of `main`, but that's probably unnecessary for our usage, and i'd like to minimize the behavior change in this PR. If you think this is an important change, you're welcome to file an issue for it.

testing/kokoro/system_tests.sh
muncus: That's correct. the module-based approach was never fully implemented, so these variables are informational only.

muncus:
@grayside, @Deleplace Ping. This PR has been awaiting re-review for a week. Could one of you take a look so we can get this speed improvement delivered?

testing/kokoro/system_tests.sh
grayside, line 48: future/suggestion: Would be nice to have a usage example of this, maybe in CONTRIBUTING?

testing/kokoro/system_tests.sh
grayside, line 58: question: What is `xargs ls -d` meant to be doing? Some of these chained shellscripts seem like they are handling some undocumented edge cases.

grayside (approved):

testing/kokoro/system_tests.sh
muncus: its ensuring that the directory still exists at HEAD (i.e. was not deleted). Tests dont run on deleted directories.
a better approach would be to use the output of `git diff --porcelain` to exclude deletions, but that can be done later as needed.

PR #4222, opened 2024-06-27
samples(discoveryengine): Update Vertex AI Search sample to use regional endpoint if needed

holtskinner:
## Description

Partial fix for b/349640878

### DIFF
--- a/discoveryengine/search.go
+++ b/discoveryengine/search.go
@@ -40,7 +40,7 @@ func search(projectID, location, searchEngineID, query string) error {
 	}
 	client, err := discoveryengine.NewSearchClient(ctx, option.WithEndpoint(endpoint))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Deleplace (changes requested):
Thank you Holt

I think we want to return the error (not print and continue):
```
    return fmt.Errorf("creating Vertex AI Search client: %w", err)
```

holtskinner:
> Thank you Holt
> 
> I think we want to return the error (not print and continue):
> 
> ```
>     return fmt.Errorf("creating Vertex AI Search client: %w", err)
> ```

Done

Deleplace (approved):
LGTM
We're currently lacking a test (the only test is skipped and not ready). This is not your fault and makes it harder to defend against regressions.

PR #4223, opened 2024-06-27
feat(gimmeproj): blocking retry on lease operations

muncus:
- **feat: enable blocking retries on gimmeproj**
- **simplify awkward time math**

stderr will now print something like this:

```
$ gimmeproj lease 10m
2024/06/27 10:41:11 Temporary error: datastore: could not find a free project
2024/06/27 10:41:41 Temporary error: datastore: could not find a free project
2024/06/27 10:42:12 Temporary error: datastore: could not find a free project
Leased! one is yours for 10m0s.
```

stdout is unchanged, and contains only the project id.


testing/gimmeproj/main.go
```diff
@@ -142,7 +145,21 @@ Administrative commands:
...
-		return lease(ctx, flag.Arg(1))
+		// When leasing, keep trying until we reach our configured timeout
+		ctx, cancel := context.WithTimeout(ctx, *waitTime)
+		defer cancel()
+		for ctx.Err() == nil {
+			err := lease(ctx, flag.Arg(1))
+			if err == nil {
+				return err
+			} else if errors.Is(err, ErrNoProjects) {
+				log.Printf("Temporary error: %v\n", err)
```
grayside, line 156: suggestion: It would be good to provide some guidance on what the wait might be, such as "trying every 30 seconds until 30 minutes".

testing/gimmeproj/main.go
```diff
@@ -33,10 +34,12 @@ import (
 var (
 	metaProject = flag.String("project", "", "Meta-project that manages the pool.")
 	format      = flag.String("output", "", "Output format for selected operations. Options include: list")
+	waitTime    = flag.Duration("timeout", 30*time.Minute, "maximum wait time for leasing a project")
```
grayside, line 37: question: Where does 30 min come from? Should this be a smaller number? The larger the number the more resources we are locking down for resource polling.

testing/gimmeproj/main.go
```diff
@@ -142,7 +145,21 @@ Administrative commands:
...
+		// When leasing, keep trying until we reach our configured timeout
+		ctx, cancel := context.WithTimeout(ctx, *waitTime)
+		defer cancel()
+		for ctx.Err() == nil {
+			err := lease(ctx, flag.Arg(1))
+			if err == nil {
+				return err
+			} else if errors.Is(err, ErrNoProjects) {
+				log.Printf("Temporary error: %v\n", err)
+				time.Sleep(30 * time.Second)
```
grayside, line 157: future: Is 30 seconds the right delay? My experience is that Go testing currently takes longer than a minute. Thinking this through, this is probably a fairly cheap poll.

If we wanted to make it smarter, getting a retry-after direction from the lease function would be interesting, where we can add estimated test runtime to the oldest runner lease time.

testing/gimmeproj/main.go
```diff
@@ -199,7 +216,7 @@ func lease(ctx context.Context, duration string) error {
 		var ok bool
 		proj, ok = pool.Lease(d)
 		if !ok {
-			return errors.New("Could not find a free project. Try again soon.")
+			return ErrNoProjects
```
grayside, line 219: praise: Nice refactor.

testing/gimmeproj/main.go
```diff
@@ -238,7 +255,7 @@ func status(ctx context.Context) error {
 		for _, proj := range pool.Projects {
 			exp := ""
 			if !proj.Expired() {
-				secs := proj.LeaseExpiry.Sub(time.Now()) / time.Second * time.Second
+				secs := time.Until(proj.LeaseExpiry)
```
grayside, line 258: praise: Nice simplification!

grayside (approved):
LGTM, left a few comments but nothing that seemed blocking.

testing/gimmeproj/main.go
muncus: This was arbitrarily chosen, based on my personal experience that the shortest test runs seem to take about 15-20m. The longest ones go for 2h+, so that's basically the range for this option.

the polling operation is pretty quick (one datastore API call), so i stuck toward the low end. we can always specify a different time with this flag so 🤷🏼

testing/gimmeproj/main.go
muncus: there's certainly room for improvement here, but i hesitate to add complexity to this tool directly. I think we'd be better served by building a server-side API component for this, which will also give us better visibility into client issues.

grayside: Fair. I should have explained a bit more: I started with a real, optional feedback for right now but switched it to the "future" prefix when I realized I went too far, but still thought the notion worth sharing.

grayside:
Marc and I discussed out of band, better to roll out this change with a plan for active monitoring and rollback than adding a feature flagging mechanism to the tooling.

PR #4245, opened 2024-07-02
fix(datastore): Iterator reused result object

dustinmoris:
Fixed issue in official samples for cursor paging where the same task object was copied into the final array over and over again without clearing its internal state.

## Description

Currently the [official documentation for cursor paging](https://cloud.google.com/datastore/docs/samples/datastore-cursor-paging) contains a little bug, where the `task` object pertains old values between the iterations. For instance if `Task` had an optional pointer property called `Foo *Bar` then `Foo` would pertain its value from the previous iteration if the next data record doesn't override it.

My change proposes a small fix in the sample where each iteration instantiates a new object of type `Task` and hence avoids this bug.

Another alternative would have been to insert `task := Task{}` inbetween the `append` and `it.Next` call inside the `for` loop:

```golang
// Read the tasks.
var tasks []Task
var task Task
it := client.Run(ctx, query)
_, err := it.Next(&task)
for err == nil {
        tasks = append(tasks, task)
        task = Task{}
        _, err = it.Next(&task)
}
if err != iterator.Done {
        log.Fatalf("Failed fetching results: %v", err)
}
```

Having said that I think my modification in this PR makes the code simpler by following the same iterator pattern as it was done in previous examples, with a single `it.Next` inside the `for` loop.


Deleplace:
Hi @dustinmoris 

Thank you for your really great finding, explanation, and suggested fix.

I could reproduce the problem and can confirm that your fix is solving it:

image

```

type Task struct {
	Category        string
	...
	...
	Foo             *Bar
}

type Bar struct {
	Value int
}

func (t Task) String() string {
	if t.Foo == nil {
		return fmt.Sprintf("[%s]", t.Category)
	}
	return fmt.Sprintf("[%s, %d]", t.Category, t.Foo.Value)
}
```

```
func TestSnippetIterator_Cursor(t *testing.T) {
	...
	...
	t.Log(tasks)

	...
```
## Before fix

```
% go test -v -run=TestSnippetIterator_Cursor
=== RUN   TestSnippetIterator_Cursor
    snippet_test.go:550: [[CatC, 111] [CatA, 111] [CatB, 111]]
```

Value 111 showing up 3 times, which is incorrect.

## After fix

```
% go test -v -run=TestSnippetIterator_Cursor 
=== RUN   TestSnippetIterator_Cursor
    snippet_test.go:553: [[CatC, 333] [CatA, 111] [CatB]]
```

Correct

dustinmoris:
@Deleplace Done:

image

Deleplace (approved):

PR #4253, opened 2024-07-03
feat: added labels samples for secret manager

vipul7499:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


secretmanager/create_update_secret_label.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"io"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
+	"google.golang.org/genproto/protobuf/field_mask"
+)
+
+// updateSecret updates the metadata about an existing secret.
+func createUpdateSecretLabel(w io.Writer, name, labelKey, labelValue string) error {
```
muncus, line 29: the comment and function name do not match, please correct that.

secretmanager/create_secret_with_labels.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
+)
+
+// createSecretWithLabels creates a new secret with the given name and labels.
+// A secret is a logical wrapper around a collection of secret versions.
```
muncus, line 28: The comments around secret vs secret version is not needed here. samples appear in documentation, which will cover this concept.

secretmanager/create_secret_with_labels.go
```diff
@@ -0,0 +1,69 @@
...
+	"io"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
+)
+
+// createSecretWithLabels creates a new secret with the given name and labels.
+// A secret is a logical wrapper around a collection of secret versions. 
+// Secret versions hold the actual secret material.
+func createSecretWithLabels(w io.Writer, parent, id, labelKey, labelValue string) error {
```
muncus, line 30: you can simplify this function by setting fixed values for labelKey and labelValue in the code, rather than taking it as a parameter.

This would align with the [Samples Style Guide](http://googlecloudplatform.github.io/samples-style-guide).

secretmanager/secretmanager_test.go
```diff
@@ -188,6 +191,26 @@ func TestCreateSecret(t *testing.T) {
...
 
+func TestCreateSecretWithLabels(t *testing.T) {
+	tc := testutil.SystemTest(t)
+
+	secretID := "createSecretWithLabels"
+	labelKey := "secretmanager"
+	labelValue := "rocks"
+
+	parent := fmt.Sprintf("projects/%s", tc.ProjectID)
+	defer testCleanupSecret(t, fmt.Sprintf("projects/%s/secrets/%s", tc.ProjectID, secretID))
```
muncus, line 202: this defer should happen only *after* the secret has been successfully created, so move this down to 208.
As written, the cleanup will fail any time the create fails, because it cannot delete a nonexistent secret.

muncus (dismissed):
Label key/value should be specified inline in the sample for all samples in this PR. consider using `example-label` and `example-value`, since they will be part of the sample and presented alongside formal documentation.

secretmanager/create_secret_with_labels.go
vipul7499: I dont think so adding hardcoded values for label key and value would be a good idea, because this is a sample to show how to create a secret with labels, and if we hardcode that value, it might confuse the users. 
Also, if we hardcode for go-lang, it will be inconsistent with all the other languages.
Which part of the guide refers to this?

vipul7499:
Hey @muncus any update on this?

secretmanager/create_secret_with_labels.go
vipul7499: @muncus any updates on this?

subfuzion: @vipul7499 Recognizing that a degree of inconsistency has creeped into various samples over time, we have started a conversation about updating the style guide that's currently still ongoing. The general principle is [minimal arguments](https://googlecloudplatform.github.io/samples-style-guide/#minimal-arguments), but we plan to add more clarification (such as for testing).

For the time being, however, the main point is that snippets should be standalone because they are used for our documentation; while tests can be instructional, their primary purpose here is to ensure samples work. The preference is that required arguments are present in the sample and optional arguments are also present as comments.

I committed an update to get tests passing, so if you could make the change that @muncus requested, I'd be happy to get this merged right away. Thanks!

subfuzion (changes requested):
@vipul7499 Thanks for working on this! See note regarding argument usage so we can get this merged.

secretmanager/secretmanager_test.go
```diff
@@ -327,6 +328,23 @@ func TestCreateSecretWithLabels(t *testing.T) {
...
 		t.Errorf("createSecretWithLabels: expected %q to contain %q", got, want)
+  }
+}
+func TestCreateRegionalSecret(t *testing.T) {
+	tc := testutil.SystemTest(t)
+
+	secretID := "createRegionalSecret"
+	locationID := testLocation(t)
+
+	defer testCleanupRegionalSecret(t, fmt.Sprintf("projects/%s/locations/%s/secrets/%s", tc.ProjectID, locationID, secretID))
```
subfuzion, line 339: Hi @vipul7499 - I'd like to ask a favor of you because our goal with pull requests includes leaving code better than we found it. There are a number of tests in this file that have a `defer` statement preceding the actual resource acquisition. Would you mind not only updating your test but also fixing any of the other tests to help enhance consistency? Move `defer` statements so that they immediately follow the resource acquisition statement (in this specific case, line 346). Thank you, good citizen! :)

subfuzion (dismissed):

subfuzion:
@vipul7499 Please run the following in your sample directory and then commit so lint passes (failed because go.mod/go.sum not up to date):

```
go mod tidy
go get toolchain@none
```

secretmanager/create_secret_with_labels.go
vipul7499: Fixed.

secretmanager/secretmanager_test.go
vipul7499: Fixed.

telpirion (approved):

telpirion:
I think all outstanding comments have been addressed. @subfuzion @muncus with your pardon, I'm going to merge this PR. Thanks!

PR #4259, opened 2024-07-05
docs(compute): Renaming region tag

m-strzelczyk:
Changing the name of the region tag for consistency with Java and Python.


iennae (approved):

muncus:
LG. i've verified that this region tag is not currently in use, so the in-place rename is sufficient.

PR #4267, opened 2024-07-11
feat(vertexai): Context caching

Deleplace:
Adapted from the [Python samples for Context caching](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/main/generative_ai/context_caching)

Region tags:
```
generativeaionvertexai_gemini_create_context_cache]
generativeaionvertexai_gemini_get_context_cache
generativeaionvertexai_gemini_use_context_cache
generativeaionvertexai_gemini_update_context_cache
generativeaionvertexai_gemini_delete_context_cache
```


Deleplace:
As an exception, `createContextCache` returns a string. This is necessary for the test to refer to the cached content that it has just created.

ymotongpoo:
@Deleplace do you mean this is port of these samples? (the link you mentioned is pointing at controlled generation) https://github.com/GoogleCloudPlatform/python-docs-samples/tree/621454ef68f7fbda02060d1a249acb83d5e490d6/generative_ai/context_caching

vertexai/context-caching/contextcaching_delete.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/vertexai/genai"
+)
+
+// deleteContextCache shows how to delete a cached content
+func deleteContextCache(w io.Writer, contentName string, projectID, location string) error {
```
ymotongpoo, line 28: What does `contentName` actually mean here? Though I looked up [the package document](https://pkg.go.dev/cloud.google.com/go/vertexai/genai#Client.DeleteCachedContent), still it's vague to me. Could you elaborate the code comment a bit more around that? (I understand that the original Go package doc should be fixed, but at least for this demo, the comment should help for now).

In the corresponding Python sample, `cache_id` is used, which gives a bit more context.
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/621454ef68f7fbda02060d1a249acb83d5e490d6/generative_ai/context_caching/delete_context_cache.py#L28

vertexai/context-caching/contextcaching_get.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/vertexai/genai"
+)
+
+// getContextCache shows how to retrieve the metadata of a cached content, by its name
+func getContextCache(w io.Writer, contentName string, projectID, location string) error {
```
ymotongpoo, line 28: ditto as `deleteContextCache`

vertexai/context-caching/contextcaching_update.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return fmt.Errorf("GetCachedContent: %w", err)
+	}
+
+	update := &genai.CachedContentToUpdate{
+		Expiration: &genai.ExpireTimeOrTTL{TTL: 2 * time.Hour},
+	}
+
+	_, err = client.UpdateCachedContent(ctx, cachedContent, update)
+	fmt.Fprintf(w, "Update cached content %q", contentName)
```
ymotongpoo, line 50: Why don't you print the content of cached content here, like its corresponding Python sample does?
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/621454ef68f7fbda02060d1a249acb83d5e490d6/generative_ai/context_caching/update_context_cache.py#L35

vertexai/context-caching/contextcaching_use.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/vertexai/genai"
+)
+
+// useContextCache shows how to use an existing cached content, when prompting the model
+func useContextCache(w io.Writer, contentName string, projectID, location, modelName string) error {
```
ymotongpoo, line 29: ditto as `deleteContextCache`

ymotongpoo (commented):
@Deleplace Could you provide your comments for mine?

vertexai/context-caching/contextcaching_delete.go
ianlewis: I assume `cacheName` is effectively a unique cache key that can be used to reference the cached content. While not the best naming choice for the SDK, I personally think it's ok to use "name" here so as to match the function signature of the SDK.

vertexai/context-caching/contextcaching_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"cloud.google.com/go/vertexai/genai"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+// TestContextCaching tests createContextCache, getContextCache, useContextCache,
+// updateContextCache, deleteContextCache.
+func TestContextCaching(t *testing.T) {
+	tc := testutil.SystemTest(t)
+
+	buf := new(bytes.Buffer)
```
ianlewis, line 33: Why not just use stdout? It might be useful to see the output in the event that the test fails.

Deleplace:
> @Deleplace do you mean this is port of these samples? (the link you mentioned is pointing at controlled generation) https://github.com/GoogleCloudPlatform/python-docs-samples/tree/621454ef68f7fbda02060d1a249acb83d5e490d6/generative_ai/context_caching

Thanks! Indeed the link was wrong, I updated it at the top of the PR.

vertexai/context-caching/contextcaching_delete.go
Deleplace: Indeed this is not self-explanatory, I'll add a comment to say that `contentName` is the string ID of the cached content.

vertexai/context-caching/contextcaching_update.go
Deleplace: We only have metadata, like "expire_time". To access it, we need to call `GetCachedContent` again after `UpdateCachedContent`.  This is not the main job of `updateContextCache`, so instead we're doing this consistency check in part (4) of `TestContextCaching`.

vertexai/context-caching/contextcaching_test.go
Deleplace: In the tests we don't use `os.Stdout`, nor `t.Logf`.  We do use `t.Errorf` and `t.Fatalf` though.  See [CONTRIBUTING.md](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#print-to-an-iowriter-for-snippets).

ianlewis: I would have thought that just setting `w` to `io.Stdout` would be fine for tests. I'm not sure I understand the reasoning that we should throw out this output in tests but I'm willing to defer.

ymotongpoo (approved):
LGTM

ianlewis (approved):

vertexai/context-caching/contextcaching_test.go
Deleplace: I guess writing to stdout/stderr/t.output  is not necessary for CI/CD, and may generate noise in the logs

ianlewis: Go's testing module suppresses stdio output unless tests fail so it shouldn't generate noise. It's not that big a deal though. 🤷

rsamborski (approved):

PR #4269, opened 2024-07-11
feat: add v2 API for resource notification config Go client libraries

vijaykanthm:
## Description

Reference to https://b.corp.google.com/issues/339692211
This PR adds v2 API to create, list, get, update, delete, receive notifications of the samples.

## Checklist


securitycenter/notificationsv2/notifications_test.go
```diff
@@ -0,0 +1,265 @@
...
+	securitycenter "cloud.google.com/go/securitycenter/apiv2"
+	"cloud.google.com/go/securitycenter/apiv2/securitycenterpb"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+	"github.com/google/uuid"
+)
+
+func orgID(t *testing.T) string {
+	orgID := os.Getenv("GCLOUD_ORGANIZATION")
+	if orgID == "" {
+		t.Skip("GCLOUD_ORGANIZATION not set")
```
muncus, line 35: What was the reason for using `t.Skip` here?
these seem required to me, so omitting them should be a failure (probably `t.Fatal`, since the test cannot continue). wdyt?

securitycenter/notificationsv2/notifications_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+func pubsubSubscription(t *testing.T) string {
+	pubsubSubscription := os.Getenv("SCC_PUBSUB_SUBSCRIPTION")
+	if pubsubSubscription == "" {
+		t.Skip("SCC_PUBSUB_SUBSCRIPTION not set")
+	}
+	return pubsubSubscription
+}
+
+func addNotificationConfig(t *testing.T, notificationConfigID string) error {
```
muncus, line 64: this looks nearly identical to the createNotification sample. can we reuse that in the test to avoid duplicate code?

securitycenter/notificationsv2/notifications_test.go
```diff
@@ -0,0 +1,265 @@
...
+		if err := createNotificationConfig(buf, orgID(t), pubsubTopic(t), configID); err != nil {
+			r.Errorf("createNotificationConfig failed: %v", err)
+			return
+		}
+
+		if !strings.Contains(buf.String(), "New NotificationConfig created") {
+			r.Errorf("createNotificationConfig did not create.")
+		}
+
+		cleanupNotificationConfig(t, configID)
```
muncus, line 140: for all tests, you can `defer` the cleanup method right after the create, so that the resoruces are deleted no matter how we exit the function (e.g. if there's a nil pointer error). deferred calls happen any time we exit the method, where this is only called if we get to the end.

securitycenter/notificationsv2/notifications_test.go
```diff
@@ -0,0 +1,265 @@
...
+			r.Errorf("updateNotificationConfig did not update.")
+		}
+		cleanupNotificationConfig(t, configID)
+	})
+}
+
+func TestReceiveNotifications(t *testing.T) {
+	testutil.Retry(t, 5, 30*time.Second, func(r *testutil.R) {
+		buf := new(bytes.Buffer)
+		if err := receiveMessages(buf, projectID(t), pubsubSubscription(t)); err != nil {
```
muncus, line 256: this does not look like we're sending a notification. what notifications do you expect this test to receive?

muncus (changes requested):

securitycenter/notificationsv2/doc.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package notifications contains example snippets for working with notification
+// configs and their notifications.
+package notificationsv2
```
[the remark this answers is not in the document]
vijaykanthm: Removed

securitycenter/notificationsv2/notifications_test.go
vijaykanthm: Addressed, updated with t.Fatal

securitycenter/notificationsv2/notifications_test.go
vijaykanthm: Addressed

securitycenter/notificationsv2/notifications_test.go
vijaykanthm: Addressed

securitycenter/notificationsv2/notifications_test.go
vijaykanthm: Addressed. Updated the functionality

securitycenter/notificationsv2/notifications_test.go
```diff
@@ -0,0 +1,326 @@
...
+func pubsubSubscription(t *testing.T) string {
+	pubsubSubscription := os.Getenv("SCC_PUBSUB_SUBSCRIPTION")
+	if pubsubSubscription == "" {
+		t.Fatal("SCC_PUBSUB_SUBSCRIPTION not set")
+	}
+	pubsubSubscription = strings.TrimSpace(pubsubSubscription)
+	return pubsubSubscription
+}
+
+func createTestNotificationConfig(buf *bytes.Buffer, orgID string, pubsubTopic string, configID string) error {
```
muncus, line 69: is this the same as the sample function `createNotificationConfig`?

I dont think this is necessary, since we can just call `createNotificationConfig` in these tests. what was your reasoning for creating this helper method?

the same is true for deleteNotificationConfig and the cleanupNotificationConfig methods.

muncus (changes requested):

securitycenter/notificationsv2/notifications_test.go
```diff
@@ -0,0 +1,326 @@
...
+	_, err = client.CreateNotificationConfig(ctx, req)
+	if err != nil {
+		return fmt.Errorf("Failed to create notification config: %w", err)
+	}
+
+	buf.WriteString("New NotificationConfig created")
+	return nil
+}
+
+func projectIDFromEnv() string {
```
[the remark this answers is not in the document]
vijaykanthm: Thank you for the suggestion, addressed.

securitycenter/notificationsv2/notifications_test.go
vijaykanthm: I thought of creating a sample notification in the test itself for testing purposes and have written the helper function. But you are correct, we can reuse the existing sample function. Removed the function and refactored the file.

muncus (approved):
Thank you!

telpirion (approved):

PR #4305, opened 2024-08-06
fix: moving prompt inside sample

BigBlackWolf:
## Description

- Moved func parameter from tests to sample as local variable
- Added additional call to model, to be consistent with Python samples

## Checklist


vertexai/token-count/tokencount_test.go
```diff
@@ -26,19 +26,18 @@ import (
...
-	err := countTokens(&buf, prompt, tc.ProjectID, location, modelName)
+	err := countTokens(&buf, tc.ProjectID, location, modelName)
 	if err != nil {
 		t.Fatalf("Test_countTokens: %v", err.Error())
 	}
 
 	answer := buf.String()
 	s := strings.TrimPrefix(answer, "Number of tokens for the prompt: ")
-	s = strings.TrimSpace(s)
+	s = strings.Split(s, "\n")[0]
```
subfuzion, line 40: Just curious about the change; are you trying to preserve any other leading or trailing whitespace other than specifically "\n"?

subfuzion (commented):
Thanks, @BigBlackWolf. Looks good; I'm just curious about one change you made.

vertexai/token-count/tokencount_test.go
BigBlackWolf: There was a request to add additional print, therefore buf contains a bit longer string (desired output + new print). To properly adjust test I would need to split by "\n"

FWIW TrimSpace was redundant, cause previous line already removed space before the number 🙂

vertexai/token-count/tokencount.go
```diff
@@ -27,12 +27,12 @@ import (
...
 
 // countTokens returns the number of tokens for this prompt.
-func countTokens(w io.Writer, prompt, projectID, location, modelName string) error {
-	// prompt := "why is the sky blue?"
+func countTokens(w io.Writer, projectID, location, modelName string) error {
 	// location := "us-central1"
 	// modelName := "gemini-1.5-flash-001"
 
 	ctx := context.Background()
+	prompt := genai.Text("why is the sky blue?")
```
rsamborski, line 35: nit: Can we start the prompt with capital letter (why -> Why)?

vertexai/token-count/tokencount_test.go
rsamborski, line 40: suggestion: please add test for the second print added in the sample ("Total number of tokens for the prompt...")

vertexai/token-count/tokencount.go
```diff
@@ -42,13 +42,19 @@ func countTokens(w io.Writer, prompt, projectID, location, modelName string) err
...
 		return err
 	}
 
 	fmt.Fprintf(w, "Number of tokens for the prompt: %d\n", resp.TotalTokens)
 
+	resp2, err := model.GenerateContent(ctx, prompt)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(w, "Total number of tokens for the prompt: %d\n", resp2.UsageMetadata.PromptTokenCount)
```
rsamborski, line 56: question: Does `resp.TotalTokens = resp2.UsageMetadata.PromptTokenCount`? 
nit: I think we should add a comment or update the printed statement to reflect the difference between these two numbers. My understanding is that `resp.TotalTokens` is calculated without any generation, while `resp2.UsageMetadata.PromptTokenCount` is post generation and is the actual billed amount of tokens. Correct?

BigBlackWolf: Yes, the number is the same, cause `CountTokens()` accepts prompt and calculate input number of tokens.
You are right `resp2.UsageMetadata` contains the data about how many tokens were passed - `PromptTokenCount` and number of tokens being generated - `CandidatesTokenCount`. The billed amount for the input is the same, cause in both calls accepts the same value, however output text is also billed, but this number can't be estimated.

I guess this is just a proto attributes naming confusion, cause `CountTokens()` total means tokens just for input, but for `GenerateContent()` total means input + output (`PromptTokenCount + CandidatesTokenCount`)

I will add print for the output part as well, to make it more clear.

subfuzion (approved):

PR #4307, opened 2024-08-06
chore: update kokoro go versions

subfuzion:
Please merge PR once we have a 1.22 image and this is approved.


muncus:
I've looked at the logs from the kokoro failures, and i'm comfortable with this change.

PR #4326, opened 2024-08-19
fix: refactoring aiplatform samples

BigBlackWolf:
## Description

- Refactoring samples to put hidden variables inside samle

## Checklist


BigBlackWolf:
It appears to be two identical samples:
**generativeaionvertexai_sdk_embedding** and **generativeaionvertexai_text_embeddings / aiplatform_text_embeddings** .
Should one of them being removed?

aiplatform/snippets/embeddings.go
```diff
@@ -28,40 +27,34 @@ import (
...
-		location = match[1]
-	}
 	endpoint := fmt.Sprintf("projects/%s/locations/%s/publishers/google/models/%s", project, location, model)
 	instances := make([]*structpb.Value, len(texts))
 	for i, text := range texts {
 		instances[i] = structpb.NewStructValue(&structpb.Struct{
 			Fields: map[string]*structpb.Value{
 				"content":   structpb.NewStringValue(text),
-				"task_type": structpb.NewStringValue(task),
+				"task_type": structpb.NewStringValue("QUESTION_ANSWERING"),
```
telpirion, line 49: nit: I would keep the task_type as a string literal initialized outside of this object. Up to you!

aiplatform/snippets/embeddings.go
```diff
@@ -28,40 +27,34 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-func embedTexts(
-	project, location string, texts []string) ([][]float32, error) {
+func embedTexts(project, location, model string) ([][]float32, error) {
```
telpirion, line 30: issue: since we're refactoring this sample, let's remove the return type.

See:
https://googlecloudplatform.github.io/samples-style-guide/#result

aiplatform/snippets/embeddings.go
```diff
@@ -28,40 +27,34 @@ import (
...
 	ctx := context.Background()
 
 	apiEndpoint := fmt.Sprintf("%s-aiplatform.googleapis.com:443", location)
-	model := "text-embedding-004"
-	task := "QUESTION_ANSWERING"
-	customOutputDimensionality := 5
+	dimensionality := 5
+	texts := []string{"banana muffins? ", "banana bread? banana muffins?"}
```
telpirion, line 35: question: why are we passing in the model as a string but hard-coding the prompt in the sample?

aiplatform/snippets/embeddings.go
telpirion, line 30: issue: provide a comment that explains each parameter and/or gives an example.

aiplatform/snippets/embeddings_preview.go
```diff
@@ -27,38 +26,34 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-func embedTextsPreview(
-	apiEndpoint, project, model string, texts []string,
-	task string, dimensionality *int) ([][]float32, error) {
+func embedTextsPreview(project, location, model string) ([][]float32, error) {
```
telpirion, line 29: See previous comments:
+ Pass texts in as parameter
+ Provide comments explaining parameters
+ Remove return type (error is okay)

aiplatform/snippets/embeddings_preview.go
```diff
@@ -27,38 +26,34 @@ import (
...
-		location = match[1]
-	}
 	endpoint := fmt.Sprintf("projects/%s/locations/%s/publishers/google/models/%s", project, location, model)
 	instances := make([]*structpb.Value, len(texts))
 	for i, text := range texts {
 		instances[i] = structpb.NewStructValue(&structpb.Struct{
 			Fields: map[string]*structpb.Value{
 				"content":   structpb.NewStringValue(text),
-				"task_type": structpb.NewStringValue(task),
+				"task_type": structpb.NewStringValue("QUESTION_ANSWERING"),
```
telpirion, line 48: See previous around hard-coding this string literal into the object initialization.

aiplatform/snippets/embeddings.go
telpirion: This sample also needs to accept a `io.Writer` object as its first parameter.

telpirion (changes requested):

telpirion:
> It appears to be two identical samples in _embeddings.go_ and _embeddings_preview.go_: **generativeaionvertexai_sdk_embedding** and **generativeaionvertexai_text_embeddings / aiplatform_text_embeddings**. Should one of them being removed?

Yes, If the two samples are identical, we should remove one of them. If I had to guess, one was written when the feature was in a preview state.

aiplatform/snippets/embeddings.go
BigBlackWolf: Thanks, for your comments! 
In this case I don't have strong preference, but received comments in previous PRs, to not create variables, which used only once. Therefore I am trying to follow consistent pattern

aiplatform/snippets/embeddings.go
BigBlackWolf: Regarding `io.Writer` - function doesn't print anything.

I think this is that rare case, when return is neat

> Methods should avoid a return type whenever possible

Cause otherwise there is a need to print two-dimensional array into buffer and later on somehow parse it back from the string. It seems much harder, with additional complexity.
Probably, I am missing something, do you any suggestion about that?

aiplatform/snippets/embeddings.go
BigBlackWolf: Good catch! Just noticed that in those samples used 2 different models in test, which is not straightforward.
I will move it into the sample and leave a comment

aiplatform/snippets/embeddings.go
telpirion: Okay, fair enough. I'll chat with the other Golang reviewers about what our preferred style should be.

telpirion (commented):
Almost done!

telpirion (approved):
Thank you!

PR #4334, opened 2024-08-21
fix(pubsub): add locational endpoint for exactly once sample

hongalex:
## Description

Exactly once delivery is best set with a locational endpoint. This adds the `WithEndpoint` option to the exactly once sample, and fixes the comments for ordering samples.

---

When you run your application within Google Cloud, it typically connects to the Pub/Sub endpoint in the same region by default. Therefore, running your application in a single region within Google Cloud generally ensures you are interacting with a single region.

If you are running your publisher application outside of Google Cloud or in multiple regions, you can guarantee you are connecting to a single region using a [locational endpoint](https://cloud.google.com/pubsub/docs/reference/service_apis_overview#pubsub_endpoints) when configuring your Pub/Sub client. All location endpoints for Pub/Sub point to single regions. For a list of all locational endpoints for Pub/Sub, see [List of locational endpoints](https://cloud.google.com/pubsub/docs/reference/service_apis_overview#list_of_locational_endpoints).

### DIFF
--- a/pubsub/subscriptions/pull_exactly_once_delivery.go
+++ b/pubsub/subscriptions/pull_exactly_once_delivery.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/pubsub"
+	"google.golang.org/api/option"
 )
 
 // receiveMessagesWithExactlyOnceDeliveryEnabled instantiates a subscriber client.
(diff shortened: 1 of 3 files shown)


telpirion (approved):

hongalex:
Thanks for the review!

PR #4347, opened 2024-08-28
fix: resolve breaking build

subfuzion:
Beginning with Go 1.21, `go mod tidy` updates `go.mod` with Go version in `x.y.z` format. Our CI builds currently tests with Go 1.20 as the earliest version, which expects `x.y` format. 

This mismatch creates a conflict because our lint/vet action currently uses 1.21 (and thus fail after after detecting a change after running `go mod tidy` if using the `x.y` format) or else fail in the Kokoro build on Go 1.20 if using the `x.y.z` format.

This original version of this PR attempted to remove the mismatch by (temporarily) downgrading the version of go used for linting, but this caused go vet to fail. The latest version of this PR simply cherry picks the last commit that caused this to break with the last working version of main rebased on HEAD (instead of the default merge that GitHub did to bring HEAD up to date), with go mod tidy run manually.

### DIFF
--- a/run/go.mod
+++ b/run/go.mod
@@ -4,4 +4,4 @@ go 1.21
 
 require cloud.google.com/go/compute/metadata v0.5.0
 
-require golang.org/x/sys v0.22.0 // indirect
+require golang.org/x/sys v0.24.0 // indirect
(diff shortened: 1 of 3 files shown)


telpirion (approved):

telpirion:
Thank you @subfuzion -- Just to confirm that a quorum of Gophers (you, me, and @grayside ) met and agreed this was the best path forward.

PR #4360, opened 2024-09-03
chore(build): bump earliest kokoro build version to go121

subfuzion:
Bumping earliest Kokoro test go version to `go121`. Bumping latest version to `go123` will be submitted in a separate, forthcoming PR.

This should solve the breakage we're seeing from the `go.mod` version format change between `1.20` and `1.21`. For example:

```
go: errors parsing go.mod:
/[tmpfs/src/github/golang-samples/opentelemetry/instrumentation/app/go.mod:3](https://cs.corp.google.com/piper///depot/google3/tmpfs/src/github/golang-samples/opentelemetry/instrumentation/app/go.mod?l=3):
invalid go version '1.21.0': must match format 1.23
```


iennae (approved):

telpirion (approved):
I think we have the requisite approvals, but I'll add mine just in case.

muncus:
This change was made in a different PR #4392 , so i think its safe to close this one.
Tony, thanks for investigating this and helping us find the solution!

PR #4370, opened 2024-09-09
samples: adds sample for AWS provider

aeitzman:


auth/authenticate_with_aws_provider.go
```diff
@@ -0,0 +1,115 @@
...
+	"io"
+
+	"cloud.google.com/go/auth/credentials/externalaccount"
+	"cloud.google.com/go/storage"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+)
+
+type customAwsProvider struct {
```
telpirion, line 30: issue: remove this struct, since it only has one field (and its a string).

auth/authenticate_with_aws_provider.go
```diff
@@ -0,0 +1,115 @@
...
+	awsRegion string
+}
+
+func (acp customAwsProvider) AwsRegion(ctx context.Context, opts *externalaccount.RequestOptions) (string, error) {
```
telpirion, line 34: issue: remove this function, since the error always returns nil.

auth/authenticate_with_aws_provider.go
```diff
@@ -0,0 +1,115 @@
...
+	return acp.awsRegion, nil
+}
+
+func (acp customAwsProvider) AwsSecurityCredentials(ctx context.Context, opts *externalaccount.RequestOptions) (*externalaccount.AwsSecurityCredentials, error) {
```
telpirion, line 38: issue: don't pass around context objects, unless you setting a timeout callback.

auth/authenticate_with_aws_provider.go
```diff
@@ -0,0 +1,115 @@
...
+		SecretAccessKey: val.SecretAccessKey,
+		SessionToken:    val.SessionToken,
+	}
+
+	return awsCredentials, nil
+}
+
+// authenticateExplicitWithAdc uses Application Default Credentials
+// to print storage buckets.
+func authenticateExplicitWithAwsProvider(w io.Writer) error {
```
telpirion, line 61: suggestion: move the code from the helper functions into this single code snippet function.

Ideally, code-snippets should be "copy/paste runnable" with few (if any) helper functions.

See: https://googlecloudplatform.github.io/samples-style-guide/#introduction

auth/authenticate_with_aws_provider.go
```diff
@@ -0,0 +1,115 @@
...
+	ctx := context.Background()
+
+	// Set the active AWS region, i.e. "us-east-2"
+	awsRegion := "us-east-2"
+	awsCredentialsProvider := customAwsProvider{awsRegion: awsRegion}
```
telpirion, line 66: issue: pass this in as an anonymous struct.

telpirion (changes requested):
This is a good start. Can we include an integration test with this sample as well? See https://googlecloudplatform.github.io/samples-style-guide/#testing.

Thanks!

auth/authenticate_with_aws_provider.go
aeitzman: It seems like implementing an interface (which is required here) with an anonymous struct is a bit messy, would look something like this unless I'm missing something:

```
awsCredentialsProvider := struct {
    region string // Could also get rid of this and have the struct be completely empty.
}{
    region: "us-east-2",
}

func (provider struct {
    region string
}) AwsRegion(ctx context.Context, *opts RequestOptions) (string, error) {
    return region, nil
}

func (provider struct {
    region string
}) AwsSecurityCredentials(ctx context.Context, *opts RequestOptions) (AwsSecurityCredentials, error) {
  // Logic here
}

```
It doesn't seem to me like making the struct anonymous makes the code any simpler or clearer, and IMO makes it a little less obvious that a interface is being implemented. Thoughts? This seems like an example where having a helper struct/functions makes sense since in general we would expect customers to create a struct that implements the Aws credentials provider rather than use an anonymous struct.

telpirion: Couldn't you just do this?

```
	awsCredentialsProvider := struct {
		awsRegion string
	}{
		awsRegion: "us-east2",
	}
```

aeitzman: Yeah that would create an anonymous struct, but the struct needs to implement the AwsSecurityCredentialsProvider interface to be used to create a new external account struct (line 80) by having a defintion for the AwsRegion and AwsSecurityCredentials functions. If I remove the customAwsProvider struct and change it to be anonymous, then I have to have functions that act on an anonymous struct as well, leading to needing two functions set up like this

```
func (provider struct {
    region string
}) AwsRegion(...) (string, error) {
    return region, nil
}
```

vs having the explicit struct which would allow

```
func(c customAwsProvider) AwsRegion (...) (string, error) {
   return c.region, nil
}
```

the helper struct and functions are there to be the most basic working implementation of the interface that we would expect users to add on to, apologies if that wasn't clear, probably need to add comments.

telpirion: Oh, I didn't catch the pointer receivers in those two functions. I see.

telpirion: I think I would still move this struct and its methods inside of the sample function.

auth/authenticate_with_aws_provider.go
```diff
@@ -0,0 +1,115 @@
...
+	// https://www.googleapis.com/auth/cloud-platform, because IAM is used to provide fine-grained
+	// permissions for Cloud.
+	// For more information on scopes to use,
+	// see: https://developers.google.com/identity/protocols/oauth2/scopes
+	scopes := []string{"https://www.googleapis.com/auth/cloud-platform"}
+
+	// Set external account credential options.
+	options := externalaccount.Options{
+		Audience:                       "//iam.googleapis.com/projects/PROJECT_NUMBER/locations/global/workloadIdentityPools/WORKLOAD_POOL/providers/WORKLOAD_PROVIDER", // Set the Workload Identity audience, replacing PROJECT_NUMBER, WORKLOAD_POOL, and WORKLOAD_PROVIDER.
+		SubjectTokenType:               "urn:ietf:params:aws:token-type:aws4_request",                                                                                   // Set the AWS subject token type.
```
telpirion, line 79: issue: Comments should go above the line that they refer to.

See https://googlecloudplatform.github.io/samples-style-guide/#comments -- these comments aren't very readable as-written.

telpirion:
Hello @aeitzman -- please follow-up with me directly if you still need this sample. I am going to close for now, since it has been ~1 month since I left a review. Thank you!

PR #4378, opened 2024-09-13
feat: create shared reservation

BigBlackWolf:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist


telpirion (commented):
question: why are we mocking the client?

BigBlackWolf:
Hey! It was a request to do in this way, because without mocking it would be needed to set up one more project in CI
cc: @m-strzelczyk

compute/reservations/create_shared_reservation.go
```diff
@@ -0,0 +1,87 @@
...
+	"fmt"
+	"io"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "cloud.google.com/go/compute/apiv1/computepb"
+	gax "github.com/googleapis/gax-go/v2"
+	"google.golang.org/protobuf/proto"
+)
+
+type ClientInterface interface {
```
rsamborski, line 29: q: Is this used only for `mocking` the client? Is there a way to make a sample fully standalone (it should initialise the client), but still mock the client when running tests?

See my example in Python (different sample): https://github.com/GoogleCloudPlatform/python-docs-samples/pull/12598

BigBlackWolf: Yeah, it's done this way to allow for either real or mocked client. I tried to do it in a mocking way without modifying the sample, but in Go it couldn't be done as in Python or Java...
Go doesn't support object replacement in runtime, so the typical solution for that is to make an abstraction (interface) that matches the same signature. I tried to cover the gap in comments to explain how it can be done by the developer, who will run this sample, but agree that it doesn't seem ideal.
Perhaps I am missing something - I am open to any suggestions on how to improve it

telpirion (approved):

PR #4383, opened 2024-09-16
chore: update samples for new auth

codyoss:
- Remove a couple of old unused samples
- Switch apiary examples to use NewService variant or constructs that don't require users to pass a client in
- Update all refs of old auth library to new


grayside:
Thanks for updating these samples Cody! Did you intend to remove several region tags?

codyoss:
@grayside Yeah, should have called that out in the commit message. It is from the two files I deleted that were non-code and not associated with any code. I verified these relics where not being used anywhere :)

auth/downscoping/credential_initialization.go
```diff
@@ -20,31 +20,35 @@ package downscopedoverview
...
 	if err != nil {
-		return fmt.Errorf("failed to generate rootSource: %w", err)
+		return fmt.Errorf("failed to generate rootCreds: %w", err)
 	}
 
-	// downscope.NewTokenSource constructs the token source with the configuration provided.
-	dts, err := downscope.NewTokenSource(ctx, downscope.DownscopingConfig{RootSource: rootSource, Rules: accessBoundary})
+	// downscope.NewCredentials constructs the credentials with the
+	// configuration provided.
+	downscopedCreds, err := downscope.NewCredentials(&downscope.Options{
```
telpirion, line 44: observation: this is interesting that it's no longer necessary to provide a context object for new credentials.

auth/id_token_from_impersonated_credentials.go
```diff
@@ -32,34 +31,35 @@ func getIdTokenFromImpersonatedCredentials(w io.Writer, scope, targetAudience, i
 	// scope := "https://www.googleapis.com/auth/cloud-platform"
 	// targetAudience := "http://www.example.com"
 	// impersonatedServiceAccount := "name@project.service.gserviceaccount.com"
-
 	ctx := context.Background()
 
-	// Construct the GoogleCredentials object which obtains the default configuration from your
+	// Construct Credentials which obtains the default configuration from your
```
telpirion, line 36: nit: put in lower case. I don't think this needs to be a proper noun.

telpirion (approved):
No notes! I always learn so much from reading your Go code.

auth/downscoping/credential_initialization.go
codyoss: The context for the old libraries has one thing they were used for: a way to shove an http.Client into the context that was later used during token exchanges. This is now an explicit param on the options struct. Everything in auth is evaluated lazy so you don't need a context when creating objects as no network calls are made.

auth/id_token_from_impersonated_credentials.go
codyoss: I can switch it to lower. The reason I has this is upper as it is the struct type: auth.Credentials.

endpoints/getting-started/client/main.go
```diff
@@ -16,37 +16,32 @@
...
+	"cloud.google.com/go/auth/credentials"
 )
 
 var (
-	url                 = flag.String("url", "", "The request url. Required.")
-	audience            = flag.String("audience", "", "The audience for the JWT. equired")
-	serviceAccountFile  = flag.String("service-account-file", "", "Path to service account JSON file. Required.")
-	serviceAccountEmail = flag.String("service-account-email", "", "Path email associated with the service account. Required.")
+	url                = flag.String("url", "", "The request url. Required.")
+	audience           = flag.String("audience", "", "The audience for the JWT. equired")
```
iennae, line 33: I see that this was already an issue but I think this is supposed to be Required instead of equired.

endpoints/getting-started/client/main.go
```diff
@@ -16,37 +16,32 @@
...
+	serviceAccountFile = flag.String("service-account-file", "", "Path to service account JSON file. Required.")
```
iennae, line 34: I'm confused why we are dropping the serviceAccountEmail but still leaving it in what is line 40 below.

endpoints/getting-started/client/main.go
```diff
@@ -16,37 +16,32 @@
...
 )
 
 func main() {
 	flag.Parse()
 
-	if *audience == "" || *url == "" || *serviceAccountFile == "" || *serviceAccountEmail == "" {
+	if *audience == "" || *url == "" || *serviceAccountFile == "" {
 		fmt.Println("requires: --url, --audience, --service-account-file, --service-account-email")
```
iennae, line 41: I suspect you might have wanted to drop , --service-account-email" from this Println based on the removal of the serviceAccountEmail

amcolin:
Closing this PR as part of our repository cleanup. This PR has been inactive for a prolonged period and contains merge conflicts. If this feature or fix is still relevant, please rebase against the latest main branch and open a new PR. Thank you for your contribution!

PR #4392, opened 2024-09-18
chore: update min version to 1.21

codyoss:
Noticed this was still needed in #4383. Note we should also switch latest to 1.23, but I did not in this PR because the image has not been built yet.


quartzmo (approved):

muncus:
I've reviewed the failed tests, and none appear related to the version update (capacity or server-side quota issues)

PR #4407, opened 2024-09-26
feat: consume reservation samples

BigBlackWolf:
## Description

Samples, which demonstrates how to create reservation, which can be consumed by:
- any instance
- instance, which should specify exact reservation

doc link: https://cloud.google.com/compute/docs/instances/reservations-consume

## Checklist


compute/reservations/consume_any_reservation.go
```diff
@@ -0,0 +1,127 @@
...
+	"fmt"
+	"io"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "cloud.google.com/go/compute/apiv1/computepb"
+	"google.golang.org/protobuf/proto"
+)
+
+// Creates the reservation from given template in particular zone
+func createConsumableReservation(w io.Writer, projectID, zone, reservationName, sourceTemplate string) error {
```
telpirion, line 29: issue: it looks like you're duplicating the same code, `createConsumableReservation()` twice. Let's not do that. If the snippet needs this to run, then put all of this code into the test file.

compute/reservations/consume_specific_reservation.go
```diff
@@ -0,0 +1,131 @@
...
+	"fmt"
+	"io"
+
+	compute "cloud.google.com/go/compute/apiv1"
+	computepb "cloud.google.com/go/compute/apiv1/computepb"
+	"google.golang.org/protobuf/proto"
+)
+
+// Creates the reservation from given template in particular zone
+func createSpecificConsumableReservation(w io.Writer, projectID, zone, reservationName, sourceTemplate string) error {
```
telpirion, line 29: issue: same as previous. If this function is required as a snippet, it should be in its own code file.

See: https://googlecloudplatform.github.io/samples-style-guide/#one-per-file

Also, it looks very similar to `createConsumableReservation()`. Is there any code that makes it unique compared to the other function.

telpirion (changes requested):
There are a couple core issues with this PR:

1) There is more than one function per region tag
2) It looks like there is a lot of duplication with existing samples.

See:
https://github.com/GoogleCloudPlatform/golang-samples/blob/main/compute/reservations/create_reservation.go

compute/reservations/consume_any_reservation.go
BigBlackWolf: I've done it in this way, to be as close as possible to the [gcloud example in documentation](https://cloud.google.com/compute/docs/instances/reservations-consume#consuming_a_specific_single-project_reservation). 
There it is done in 2 steps:
1) shows how to create a reservation (yeah, in this case, consuming any reservation, it just duplicates existing sample)
2) actual consuming

Therefore, I thought it would be consistent to keep 2 functions. Do you think it's better to remove first and keep only **consumeAnyReservation()**?

compute/reservations/consume_specific_reservation.go
BigBlackWolf: Thanks for pointing out, but here is the same situation. In [gcloud example](https://cloud.google.com/compute/docs/instances/reservations-consume#consuming_a_specific_single-project_reservation) it is separated into 2 steps, but this is still a single sample. 

> Is there any code that makes it unique compared to the other function.

Yes, there is additional field, which makes the difference to consuming 
`SpecificReservationRequired: proto.Bool(true)`

I can make all actions being made in a single function, will it be better?

compute/reservations/consume_any_reservation.go
telpirion: Yes. The major issue here is maintenance. 

Consider this hypothetical: we have >1 samples that demonstrate how to create a reservation. Then we get a breaking change to the underlying API. That's more than 1 potential point of failure that we need to resolve.

If the docs want to show both creating and consuming a reservation, then they can import each atomic sample individually.

BigBlackWolf: I see your point. 
Adjusted samples and instead of additional function, left a note in comment to highlight a requirement for reservation, which is expected to be consumed.

BigBlackWolf:
> Hello @BigBlackWolf -- please resolve the merge conflicts and I will merge this PR. Thank you!

Hi @telpirion , thanks for pointing out. I’ve synced branch with main and now it’s up to date 🙌

telpirion (approved):

PR #4419, opened 2024-10-02
feat(securitycenter): add sample for creating v2 clients with regional and default endpoints

hegemonic:
## Description

This sample shows how to create a Security Command Center client that uses a regional endpoint, along with a client that uses the default endpoint. That's because Security Command Center's regional endpoints support only a limited subset of Security Command Center resources that are subject to data residency controls: https://cloud.google.com/security-command-center/docs/data-residency-support#scc-regional-urls-rest

As a result, it's likely that customers will need to create two clients: one for their regional endpoint, to call the methods it supports, and another for the default endpoint, to call all other methods.

This sample will be used here: https://cloud.google.com/security-command-center/docs/data-residency-support#regional-urls

## Checklist


securitycenter/clientv2/create_client_with_endpoint.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+	"fmt"
+
+	securitycenter "cloud.google.com/go/securitycenter/apiv2"
+	"google.golang.org/api/option"
+)
+
+// createClientWithEndpoint creates a Security Command Center client for a
+// regional endpoint, along with another client for the default endpoint.
+func createClientWithEndpoint(repLocation string) error {
```
hegemonic, line 28: For both clients, it would be nice if this code could print the endpoint that each client is using to an io.Writer, so that the test could verify that the endpoints are correct. However, it looks like the Go client doesn't expose the endpoint that it's currently using.

If I'm wrong, please let me know!

telpirion, line 28: issue: pass in `io.Writer` object.

See [CONTRIBUTING](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#print-to-an-iowriter-for-snippets).

securitycenter/clientv2/create_client_with_endpoint.go
```diff
@@ -0,0 +1,50 @@
...
+	// Instantiate client for default endpoint.
+	ctx := context.Background()
+	client, err := securitycenter.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("securitycenter.NewClient: %w", err)
```
telpirion, line 33: nit: it's not necessary to wrap this error in a new error--you can just return `err`.

securitycenter/clientv2/create_client_with_endpoint.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	defer client.Close() // Closing the client safely cleans up background resources.
```
telpirion, line 35: issue: remove comment. Calling `defer client.Close()` is a common Go pattern.

telpirion: Here and below.

telpirion (commented):
Thank you for contributing to the Go samples, @hegemonic !

I think, given what you're attempting to demonstrate, that you need only demonstrate how to create a new client with the regional endpoint. We should have many samples that demonstrate creating a global client.

securitycenter/clientv2/create_client_with_endpoint.go
hegemonic: Please see my previous comment:

> For both clients, it would be nice if this code could print the endpoint that each client is using to an io.Writer, so that the test could verify that the endpoints are correct. However, it looks like the Go client doesn't expose the endpoint that it's currently using.
>
> If there's a way to get a client's endpoint, please let me know!

hegemonic:
> Thank you for contributing to the Go samples, @hegemonic !
> 
> I think, given what you're attempting to demonstrate, that you need only demonstrate how to create a new client with the regional endpoint. We should have many samples that demonstrate creating a global client.

For a couple of reasons, I'd like the sample to show how to create both clients:

* Security Command Center's regional endpoint supports a limited subset of methods in the API. If you're using a regional endpoint, you'll also have to use the global endpoint.

    This behavior is pretty unusual—I'm not aware of any other Google Cloud service that requires customers to do this. Given how unusual this is, I'd like the code sample to make it very clear that customers must create two clients.

* I'd like the Go code sample to match the code samples for other languages.

Does that work for you?

securitycenter/clientv2/create_client_with_endpoint.go
hegemonic: Done.

(Many other code samples in this repo have similar comments. Please consider removing them globally.)

telpirion:
> For a couple of reasons, I'd like the sample to show how to create both clients:
> 
> * Security Command Center's regional endpoint supports a limited subset of methods in the API. If you're using a regional endpoint, you'll also have to use the global endpoint.
>   This behavior is pretty unusual—I'm not aware of any other Google Cloud service that requires customers to do this. Given how unusual this is, I'd like the code sample to make it very clear that customers must create two clients.
> * I'd like the Go code sample to match the code samples for other languages.
> 
> Does that work for you?

Your reasoning here makes sense. Please add a comment that explains ^^this point so that it's clear why the customer needs to create two clients.

hegemonic:
> Your reasoning here makes sense. Please add a comment that explains ^^this point so that it's clear why the customer needs to create two clients.

Done.

securitycenter/clientv2/create_client_with_endpoint.go
hegemonic: @telpirion Please advise.

telpirion (approved):

securitycenter/clientv2/create_client_with_endpoint.go
telpirion: I don't believe there is an easy way to do that. I'll experiment on my own to check.

PR #4438, opened 2024-10-15
feat(storage): add gRPC quickstart sample

frankyn:
## Description

Fixes b/373643411

## Checklist


storage/storage_grpc_quickstart/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"github.com/GoogleCloudPlatform/golang-samples/internal/testutil"
+)
+
+const (
+	testPrefix      = "test-gcs-grpc-team"
```
tritone, line 30: Do we still need an allowlisted prefix now that this is GA? Let's just say storage-grpc-quickstart-test maybe?

storage/storage_grpc_quickstart/main.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// Creates a gRPC enabled client.
+	client, err := storage.NewGRPCClient(ctx)
+	if err != nil {
+		log.Fatalf("Failed to create client: %v", err)
+	}
+	defer client.Close()
+
+	// Creates a Bucket instance.
+	bucket := client.Bucket(*bucketName)
```
tritone, line 47: This doesn't need to be a separate line of code and comment; can just do client.Bucket(*bucketName).Create() on L52 instead.

tritone (commented):
Few minor comments, otherwise lgtm

storage/storage_grpc_quickstart/main_test.go
frankyn: Yes it's still required; fails otherwise.

storage/storage_grpc_quickstart/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+		if err := testutil.DeleteExpiredBuckets(client, tc.ProjectID, testPrefix, bucketExpiryAge); err != nil {
+			log.Printf("DeleteExpiredBuckets: %v", err)
+		}
+	})
+
+	bucketName := testutil.UniqueBucketName(testPrefix)
+
+	stdOut, stdErr, err := m.Run(nil, time.Minute, "--project", tc.ProjectID, "--bucket", bucketName)
+
+	if err != nil {
```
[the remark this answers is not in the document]
frankyn: Warning wasn't implemented into the new auth code path introduced in recent releases and didn't see that until now. filed issue: https://github.com/googleapis/google-cloud-go/issues/11001

storage/storage_grpc_quickstart/main_test.go
tritone: Hmm when is the allowlist dropped?

Let's use the same prefix we use in the storage client integration tests ("golang-grpc-test").

storage/storage_grpc_quickstart/main_test.go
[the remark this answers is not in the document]
tritone: My point was that we probably don't want to be getting these warnings, right? We want to quickstart to be able to run without warnings if the project has the correct permissions.

storage/storage_grpc_quickstart/main_test.go
frankyn: Updated as it's not longer required.

storage/storage_grpc_quickstart/main_test.go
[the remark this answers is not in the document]
frankyn: Failed to enable client metrics isn't logged with existing tests. However, current implementation is emitting a warning that the Resource Detector detected a different project compared to the project used by the Go client. Can talk through this with you when you're back on.

telpirion (approved):

tritone (approved):

storage/storage_grpc_quickstart/main.go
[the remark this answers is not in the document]
frankyn: Package names need to match if stored in the same directory which means we can't use main() for both. I need to put it into a different directory at least in Golang.

https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#MismatchedPkgName

Moving into separate directory to unblock PR.

PR #4449, opened 2024-10-25
feat: add livestream channel clip samples and tests

irataxy:
Fixes b:341726237

Important: FFmpeg (https://www.ffmpeg.org/) must be installed in the test environment.

```
sudo apt install ffmpeg
```

## Checklist


media/livestream/livestream_test.go
```diff
@@ -517,6 +525,64 @@ func TestChannelEvents(t *testing.T) {
...
+
+	// Create a new channel clip.
+	clipOutputUri := fmt.Sprintf("%sclips", outputURI)
+	testutil.Retry(t, 3, 2*time.Second, func(r *testutil.R) {
+		clipName := fmt.Sprintf("projects/%s/locations/%s/channels/%s/clips/%s", tc.ProjectID, location, channelID, clipID)
+		if err := createChannelClip(buf, tc.ProjectID, channelID, clipID, clipOutputUri); err != nil {
+			r.Errorf("createChannelClip got err: %v", err)
+		}
+		if got := buf.String(); !strings.Contains(got, clipName) {
+			r.Errorf("createChannelClip got\n----\n%v\n----\nWant to contain:\n----\n%v\n----\n", got, clipName)
```
telpirion, line 547: nit: I'd prefer to not have the test output formatted in this way. It makes it harder to read on GitHub.

The test input can be as simple as:
```go
r.Errorf("createChannelClip: want: %v, got %v", clipName, got)
```

media/livestream/livestream_test.go
```diff
@@ -517,6 +525,64 @@ func TestChannelEvents(t *testing.T) {
 		}
 	})
 
+	// Tests for clips
+	// Send a test stream for the clip.
+	cmd := exec.Command("ffmpeg", "-re", "-f", "lavfi", "-t", "45", "-i",
```
telpirion, line 530: issue: I have concerns about running arbitrary processes from within tests. @muncus PTAL.

1. Let's isolate this test into its own file. I want to keep this test hermeneutically sealed from other tests as best as possible.
2. What exactly is the point of executing this command? The bytes returned from the output are thrown away.

telpirion (changes requested):

media/livestream/livestream_test.go
muncus: IIUC, this command is generating a video stream. Can we do this with a static file instead?
I'd prefer to avoid adding `ffmpeg` to our testing images, as it has a history of vulnerabilities and could increase ongoing maintenance requirements.

telpirion:
Closing as @irataxy  no longer works at Google. Thank you for all your contributions over the years!

Issue #4228, opened 2024-06-28
Prepare for container registry shutdown


muncus:
Golang testing is using Container Registry, which is deprecated and scheduled for shutdown in March 2025.

golang-samples-tests hosts the images we use to run our testing pipeline, which will need to be moved to Artifact Registry (and the image names updated in the kokoro configs).

grayside:
Do you have a list of current image URIs?

PR #4462, opened 2024-11-01
feat: migrate earliest-version off of Container Registry

muncus:
## Description

Switch kokoro's 'earliest-version' builds to use the 1.21 image from Artifact
Registry instead of container registry, which is in turndown.

Part of #4228


telpirion (approved):

muncus:
I have verified that the earliest-version failures on this build are unrelated to this change, and i'm comfortable merging in this state.

PR #4514, opened 2024-11-06
feat(gen ai): add samples for aiplatform evaluation api

Valeriy-Burlaka:
## Description

Adds code samples for the Vertex AI `model-reference/evaluation` [page](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/evaluation) sections:

- "Evaluate an output" [section](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/evaluation#evaluate_an_output)
- "Evaluate an output: pairwise summarization quality" [section](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/evaluation#evaluate_an_output_pairwise_summarization_quality)
- "Get rouge score" [section](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/evaluation#get_rouge_score)

**Notes:**

- All 3 samples are using the lower-level `cloud.google.com/go/aiplatform` package, not `cloud.google.com/go/vertexai`, as `vertexai` doesn't have this functionality yet.
- It seems that the release of [aiplatform](https://github.com/googleapis/google-cloud-go/releases/tag/aiplatform%2Fv1.68.0) is significantly lagging behind the googleapis/aiplatform releases (e.g., [this commit](https://github.com/googleapis/googleapis/commit/634d59e1f565397be708af825014c2e115abc648#diff-7f387506c04f9acb791f98486b9430bb828ecd55b0f71b66d0f224ac8bd10497R125) from Aug is still not released), — I will try to bring up this issue to the team.

## Checklist
- [x] I have followed [Contributing Guidelines from CONTRIBUTING.MD](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md)
- [x] **Tests** pass:   `go test -v ./...`
- [x] **Code formatted**:   `go fmt`
- [x] **Vetting** pass:   `go vet` 
- [x] Please **merge** this PR for me once it is approved

PR #4568, opened 2024-11-13
feat(gen ai): add parameter example to evaluation samples

Valeriy-Burlaka:
## Description

Adds example value for the `location` argument to all evaluation samples (I missed to add this in the scope of https://github.com/GoogleCloudPlatform/golang-samples/pull/4514)

## Checklist
- [x] **Tests** pass:   `go test -v ./..`
- [x] **Code formatted**:   `gofmt`
- [x] **Vetting** pass:   `go vet`


vertexai/evaluation/evaluate_output.go
```diff
@@ -29,6 +29,7 @@ import (
 // evaluateModelResponse evaluates the output of an LLM for groundedness, i.e., how well
 // the model response connects with verifiable sources of information
 func evaluateModelResponse(w io.Writer, projectID, location string) error {
+	// location  := "us-central1"
```
Valeriy-Burlaka, line 32: @telpirion , all existing Go samples have this line, but I wonder if it shouldn't be a _re-assignment_ (`=`) instead of a fresh _assignment_ (`:=`) instead? i.e.:

```go
// location  = "us-central1"
```

With the current version, if I copy a sample and un-comment this line, I will immediately have `no new variables on left side of := ...` compiler error. While with the `=` version I'll immediately have a working version of the sample.

What do you think?

telpirion: Yeah, what you say makes sense. We had written some of the other samples with the thought that users would copy the func, delete the args (or most of them), and then uncomment the variable assignments in the func body.

We haven't received any feedback on this from users (that I'm aware of), so this may be a trivial issue.

telpirion (approved):
I'm fine with the changes to the comments that you recommend. Thank you for checking!

PR #4570, opened 2024-11-13
fix: Ignore unknown fields when unmarshaling GCS Events

kramarz:
GCS in currently adding a new field and that would break any code that is based on these samples.

## Description

internal: b/378854932


iennae (changes requested):
@kramarz thanks for the PR. @telpirion is currently investigating this and we'll provide more info directly.

telpirion:
Thank you for the contribution!

We actually have another fix in mind for this issue, so this PR is unnecessary.

PR #4758, opened 2024-12-03
fix(eventarc): sets protojson to DiscardUnknown

telpirion:
## Description

This PR sets all of the calls to `protojson.Unmarshal()` to `DiscardUnknown` fields. This will reduce the likelihood of customers encountering errors in their deployed functions. This PR also adds a comment explaining the usage of the `DiscardUnknown` field.

### DIFF
--- a/eventarc/storage_handler/handler_test.go
+++ b/eventarc/storage_handler/handler_test.go
@@ -96,31 +96,3 @@ func TestHelloStorage_NotCloudEvent(t *testing.T) {
 		t.Errorf("got %q, want %q", got, wantBody)
 	}
 }
-
-func TestHelloStorage_NotStorageObjectData(t *testing.T) {
-	jsondata := []byte(`{"arbitrary": "value"}`)
-
-	ce := cloudevents.NewEvent()
-	ce.SetID("sample-id")
-	ce.SetSource("//sample/source")
-	ce.SetType("google.cloud.storage.object.v1.finalized")
-	ce.SetData(*cloudevents.StringOfApplicationJSON(), jsondata)
-
-	w := httptest.NewRecorder()
-	r, err := cloudevents.NewHTTPRequestFromEvent(context.Background(), "http://localhost", ce)
-	if err != nil {
-		t.Fatalf("cloudevents.NewHTTPRequestFromEvent: %v", err)
-	}
-	HelloStorage(w, r)
-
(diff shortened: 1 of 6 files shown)


functions/functionsv2/firebase/upper/upper.go
```diff
@@ -65,7 +65,15 @@ func init() {
...
-	if err := proto.Unmarshal(e.Data(), &data); err != nil {
+
+	// If you omit `DiscardUnknown`, protojson.Unmarshal returns an error
+	// when encountering a new or unknown field.
+	options := proto.UnmarshalOptions{
+		DiscardUnknown: true,
+	}
+	err := options.Unmarshal(e.Data(), &data)
+
+	if err != nil {
```
NimJay, line 76: Thanks for also moving `if err != nil` into its own line.

eventarc/storage_handler/handler_test.go
```diff
@@ -96,31 +96,3 @@ func TestHelloStorage_NotCloudEvent(t *testing.T) {
...
-	if got := w.Result().StatusCode; got != wantStatus {
-		t.Errorf("got %q, want %q", got, wantStatus)
-	}
-
-	// Ensure failure on malformed storage data.
-	wantBody := "Bad Request: expected Cloud Storage event\n"
-	if got := w.Body.String(); got != wantBody {
-		t.Errorf("got %q, want %q", got, wantBody)
-	}
-}
```
NimJay, line 126: **Question:** What was the exact error that prompted the removal of this test case?

NimJay (approved):
Thanks for doing this! Looks good to me. Left a comment about the removed test.

eventarc/storage_handler/handler_test.go
telpirion: That's a great question. The removed test checked that the sample would throw an error if the handler received a field that it didn't recognize. Unfortunately, this very issue (throwing an error on an unknown field) has been the source of some problems for customers.

functions/functionsv2/firebase/upper/upper.go
telpirion: 👍🏻

telpirion:
> Thanks for doing this! Looks good to me. Left a comment about the removed test.
> 
> Feel free to merge.

Thank you so much, @NimJay !

PR #4887, opened 2024-12-18
chore(pubsub): copy v1 samples to new directory

hongalex:
In preparation for the pubsub v2.0.0 migration, we should duplicate our samples into an old directory. I created region tags for these old samples so they will show up in our sample browser. Once the v2.0 changes are ready, we will replace the samples in the current golang-samples/pubsub directory with v2 samples, which negates the need to change large swaths of our documentation.

This PR has a lot of changes but is effectively doing two things:
1. Copy all samples into new `v1samples` directory
2. Replace all region tags of `pubsub_*` with `pubsub_old_version_*`


telpirion:
Hey @hongalex -- I've converted this PR to a draft. Please convert it back to an open PR when you're ready to submit these changes. 

Also, please investigate the Kokoro failures -- let's make sure that we have a clean build before we consider merging this PR. Thanks!

hongalex:
@telpirion thanks for taking a look. I fixed the header warning, though I'm unsure why CI is complaining about these files, considering these are just copied and pasted from the existing files.

I'll keep investigating, but hope to get this out soon.

hongalex:
@telpirion seems to be good now. I'm still not entirely sure what caused it, but it was related to the fact that I added a new directory without any .go files in it. I removed the extra resources directory and pointed the tests to the existing one in `golang-samples/pubsub/schemas/resources` and it seems to pass now.

telpirion (approved):

PR #4108, opened 2024-04-25
feat(spanner): add change streams transaction exclusion sample

ShuranZhang:
Add sample for how to configure change streams transaction exclusion in various write APIs.

Note: the current latest release of spanner go client library does not contain the source code change in https://github.com/googleapis/google-cloud-go/pull/9779 yet, so there will be build failures in CI/CD. However I have verified that all IT tests passed after pointing the go client dependency to my newest local cloned copy. 

This pr will not be merged until the latest go client release contains the necessary source code change, just uploading it now to kick off the review process earlier.

PR #5071, opened 2025-01-13
feat(spanner): add change streams transaction exclusion sample

ShuranZhang:
Add sample for how to configure change streams transaction exclusion option for a read write transaction.

I don't have access to re-open https://github.com/GoogleCloudPlatform/golang-samples/pull/4108, this is basically a duplicate.


spanner/spanner_snippets/spanner/spanner_change_streams_txn_exclusion.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/spanner"
+)
+
+// rwTxnExcludedFromChangeStreams executes the insert and update DMLs on Singers table excluded from allowed tracking change streams
```
telpirion, line 27: issue: rewrite comment for clarity. The final clause, "excluded from allowed tracking change streams," isn't clear. What is an "allowed tracking change stream"?

spanner/spanner_snippets/spanner/spanner_change_streams_txn_exclusion.go
```diff
@@ -0,0 +1,63 @@
...
+func rwTxnExcludedFromChangeStreams(w io.Writer, db string) error {
```
telpirion, line 28: issue: use complete words for function names. Customers don't intuitively know what a 'rwTxn' is.

spanner/spanner_snippets/spanner/spanner_change_streams_txn_exclusion.go
```diff
@@ -0,0 +1,63 @@
...
+	// db = `projects/<project>/instances/<instance-id>/database/<database-id>`
+	ctx := context.Background()
+	client, err := spanner.NewClient(ctx, db)
+	if err != nil {
+		return fmt.Errorf("rwTxnExcludedFromChangeStreams.NewClient: %w", err)
+	}
+	defer client.Close()
+
+	_, err = client.ReadWriteTransactionWithOptions(ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
```
telpirion, line 37: question: why is the return value from the method thrown away?

telpirion (changes requested):

spanner/spanner_snippets/spanner/spanner_change_streams_txn_exclusion.go
ShuranZhang: Because all we want to verify here is the DML executed successfully without error. This is not a select query so no row iter will be returned. See similar example [here](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/spanner/spanner_snippets/spanner/spanner_read_write_transaction_with_tag.go#L56)

spanner/spanner_snippets/spanner/spanner_change_streams_txn_exclusion.go
ShuranZhang: Added clarification on what is "allowed change streams"

spanner/spanner_snippets/spanner/spanner_change_streams_txn_exclusion.go
ShuranZhang: Changed to readWriteTxn

telpirion (approved):

PR #5091, opened 2025-01-16
fix(testing): remove version checking in system_test.sh

muncus:
this check is redundant as of 1.21, and the go.mod specified version is
a *requirement* already checked by the go tooling.

It is being removed now because the shell script check is inaccurate has
no knowledge of patch versions used in the build images.


telpirion (approved):

muncus:
I've reviewed the failed kokoro tests, and they appear to be the usual suspects of flakey tests.

PR #5150, opened 2025-02-14
chore: try heterogenous Go versions

telpirion:
This PR attempts to demonstrate how to support different sub packages in the repo that have different minimum supported versions.

The "canary" package has a minimum support version of 1.23, due to its use of iterators (introduced in 1.23). It also declares its toolchain, which might allow for some flexibility when building on Go 1.22.x. We shall see if this approach is sufficient for our CI/CD.

### DIFF
--- a/go.work
+++ b/go.work
@@ -94,6 +94,7 @@ use (
 	./healthcare
 	./iam
 	./iap
+	./internal/canary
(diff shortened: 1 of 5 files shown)


telpirion:
This experiment didn't work out as intended; closing.

Issue #4789, opened 2024-12-06
storage: samples for Soft Delete (objects)


JesseLovelace:
Please add samples for object soft delete:

- storage_disable_soft_delete
- storage_get_soft_delete_policy
- storage_set_soft_delete_policy
- storage_list_soft_deleted_objects
- storage_list_soft_deleted_object_versions
- storage_restore_object

Reference Java Implementation: https://github.com/googleapis/java-storage/pull/2754

PR #5231, opened 2025-03-19
feat(storage): add samples and tests for bucket soft delete (get, set, disable)

madisonhall38:
## Description

Part 1 of #4789. storage_list_soft_deleted_objects, storage_list_soft_deleted_object_versions, storage_restore_object will be done in a follow up.

## Checklist


storage/buckets/disable_soft_delete_policy.go
```diff
@@ -0,0 +1,63 @@
...
+// [START storage_disable_soft_delete_policy]
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"cloud.google.com/go/storage"
+)
+
+func disableSoftDeletePolicy(bucketName string) error {
```
tritone, line 26: Functions should have a godoc comment https://tip.golang.org/doc/comment#func

storage/buckets/disable_soft_delete_policy.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	// Setting retention duration to 0 disables soft delete.
+	attrs.SoftDeletePolicy = &storage.SoftDeletePolicy{
+		RetentionDuration: time.Duration(0),
+	}
+
+	// Update the bucket with zero retention duration.
+	_, err = bucket.Update(ctx, storage.BucketAttrsToUpdate{
+		SoftDeletePolicy: attrs.SoftDeletePolicy,
```
tritone, line 53: I would just create a new SoftDeletePolicy in this line rather than modifying `attrs`. Although, as below, I think the bucket.Attrs() request is probably superfluous.

storage/buckets/disable_soft_delete_policy.go
tritone, line 26: Functions should pass `w io.Writer` as the first arg. Then, use this to print the result. This allows you to capture the output for testing purposes. See https://googlecloudplatform.github.io/samples-style-guide/#result

storage/buckets/get_soft_delete_policy.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Get the current bucket attributes.
+	attrs, err := bucket.Attrs(ctx)
+	if err != nil {
+		return fmt.Errorf("bucket.Attrs: %w", err)
+	}
+
+	// Check if soft delete policy is set.
+	softDeletePolicy := attrs.SoftDeletePolicy
+	if softDeletePolicy == nil {
```
tritone, line 49: Is this actually a possible scenario?

I think we can probably just print the value of the field  rather than having if/else cases (use %+v to include the key values in the output struct).

storage/buckets/set_soft_delete_policy.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+
+	bucket := client.Bucket(bucketName)
+
+	// Get the current bucket attributes.
+	attrs, err := bucket.Attrs(ctx)
```
tritone, line 41: I think you don't need to do this? Unless you're setting a metageneration condition on the request.

storage/buckets/buckets_test.go
```diff
@@ -783,3 +783,77 @@ func TestCreateBucketObjectRetention(t *testing.T) {
...
+
+	if err := setSoftDeletePolicy(bucketName); err != nil {
+		t.Fatalf("setSoftDeletePolicy: %v", err)
+	}
+
+	attrs, err := client.Bucket(bucketName).Attrs(ctx)
+	if err != nil {
+		t.Fatalf("Bucket(%q).Attrs: %v", bucketName, err)
+	}
+	if got, want := (attrs.SoftDeletePolicy), (&storage.SoftDeletePolicy{RetentionDuration: 10 * 24 * time.Hour}); got == nil || got.RetentionDuration != 10*24*time.Hour {
```
tritone, line 808: You need to do a deep comparison here. cmp.Diff is a good tool for this. See https://pkg.go.dev/github.com/google/go-cmp/cmp#example-Diff-Testing

storage/buckets/buckets_test.go
```diff
@@ -783,3 +783,77 @@ func TestCreateBucketObjectRetention(t *testing.T) {
...
+		t.Errorf("Attrs.SoftDeletePolicy: got %v, want %v", got, want)
+	}
+}
```
tritone, line 811: Also check that the output was printed as expected (using the io.Writer passed in to print to a buffer).

storage/buckets/buckets_test.go
BrennaEpp, line 808: Small suggestion: use the `want` value defined instead of repeating 10*24*hour

storage/buckets/get_soft_delete_policy.go
madisonhall38: I'm not sure. Wanted to be extra safe but it is probably unnecessary. Done

storage/buckets/set_soft_delete_policy.go
madisonhall38: Done

storage/buckets/disable_soft_delete_policy.go
madisonhall38: Done

storage/buckets/disable_soft_delete_policy.go
madisonhall38: Done

storage/buckets/buckets_test.go
madisonhall38: Done

madisonhall38: reworked this to not duplicate 1024hour

storage/buckets/buckets_test.go
madisonhall38: Done

madisonhall38:
Thx for the comments, updated and added a new commit @BrennaEpp @tritone

storage/buckets/disable_soft_delete_policy.go
```diff
@@ -18,12 +18,14 @@ package buckets
...
 	"context"
 	"fmt"
+	"io"
 	"time"
 
 	"cloud.google.com/go/storage"
 )
 
-func disableSoftDeletePolicy(bucketName string) error {
+// Disables the soft delete policy for a bucket by setting retention duration to 0. Returns an error if the operation fails.
```
tritone, line 27: The statement about the error is unnecessary

storage/buckets/disable_soft_delete_policy.go
```diff
@@ -37,26 +39,17 @@ func disableSoftDeletePolicy(bucketName string) error {
...
-	attrs.SoftDeletePolicy = &storage.SoftDeletePolicy{
-		RetentionDuration: time.Duration(0),
-	}
-
-	// Update the bucket with zero retention duration.
+	// Update the bucket with zero retention duration to disable soft delete.
 	_, err = bucket.Update(ctx, storage.BucketAttrsToUpdate{
-		SoftDeletePolicy: attrs.SoftDeletePolicy,
+		SoftDeletePolicy: &storage.SoftDeletePolicy{
+			RetentionDuration: time.Duration(0),
```
tritone, line 45: Nit: I think this can just be represented as an empty soft delete policy because time.Duration is an int64 type, so it defaults to zero.

tritone (approved):
Two more minor changes, otherwise looks good.

telpirion (changes requested):
This PR looks really good, despite the several comments. There are just a handful of refinements to make.

storage/buckets/disable_soft_delete_policy.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"cloud.google.com/go/storage"
+)
+
+// Disables the soft delete policy for a bucket by setting retention duration to 0.
```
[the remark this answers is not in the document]
madisonhall38: Done

storage/buckets/get_soft_delete_policy.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"cloud.google.com/go/storage"
+)
+
+// Retrieves and outputs the soft delete policy for a bucket.
```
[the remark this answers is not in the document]
madisonhall38: Done

storage/buckets/buckets_test.go
```diff
@@ -783,3 +785,104 @@ func TestCreateBucketObjectRetention(t *testing.T) {
...
+
+	// Verify the output was printed as expected.
+	gotOutput := buf.String()
+	wantOutput := fmt.Sprintf("Soft delete policy for %s was set to a 10-day retention period\n", bucketName)
+	if gotOutput != wantOutput {
+		t.Errorf("Output mismatch: got %q, want %q", gotOutput, wantOutput)
+	}
+
+	// Verify the bucket's attributes.
+	attrs, err := client.Bucket(bucketName).Attrs(ctx)
```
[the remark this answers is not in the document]
madisonhall38: Assuming this is referring to the checks on 816-818. Agreed but I need attrs on line 824 and I don't see another way to fetch the attrs other than doing this, and if I don't use err or don't capture err I get compilation errors

storage/buckets/buckets_test.go
```diff
@@ -783,3 +785,104 @@ func TestCreateBucketObjectRetention(t *testing.T) {
...
+
+	bucketName := testutil.CreateTestBucket(ctx, t, client, tc.ProjectID, testPrefix)
+	defer testutil.DeleteBucketIfExists(ctx, client, bucketName)
+
+	buf := new(bytes.Buffer)
+	if err := disableSoftDeletePolicy(buf, bucketName); err != nil {
+		t.Fatalf("disableSoftDeletePolicy: %v", err)
+	}
+
+	attrs, err := client.Bucket(bucketName).Attrs(ctx)
```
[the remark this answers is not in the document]
madisonhall38: commented on other thread, same idea applies here

storage/buckets/disable_soft_delete_policy.go
```diff
@@ -0,0 +1,54 @@
...
+func disableSoftDeletePolicy(w io.Writer, bucketName string) error {
+	// bucketName := "bucket-name"
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("storage.NewClient: %w", err)
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
```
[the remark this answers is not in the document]
madisonhall38: This is the pattern we're following in storage as we want the customer to set a timeout or else infinite retries will occur

storage/buckets/get_soft_delete_policy.go
```diff
@@ -0,0 +1,52 @@
...
+func getSoftDeletePolicy(w io.Writer, bucketName string) error {
+	// bucketName := "bucket-name"
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("storage.NewClient: %w", err)
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
```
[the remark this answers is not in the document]
madisonhall38: responded on main thread

storage/buckets/set_soft_delete_policy.go
```diff
@@ -0,0 +1,56 @@
...
+func setSoftDeletePolicy(w io.Writer, bucketName string) error {
+	// bucketName := "bucket-name"
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("storage.NewClient: %w", err)
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
```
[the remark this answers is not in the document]
madisonhall38: responded on main thread

storage/buckets/buckets_test.go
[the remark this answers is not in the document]
telpirion: The test checks the printed output from the string on lines 809-810. That is sufficient for a code sample test case. Everything afterwards is unnecessary: the code sample tests need only check that the _sample_ has correct behavior. If the service API doesn't behave as intended, that's not the fault of the sample. Ideally, if the call from the code sample is bad, then the client library returns an err -- which we check for on line 803.

storage/buckets/buckets_test.go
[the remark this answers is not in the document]
telpirion: My previous comment stands.

telpirion (changes requested):

storage/buckets/buckets_test.go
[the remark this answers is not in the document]
madisonhall38: Gotcha, thanks for explaining. Removed

storage/buckets/buckets_test.go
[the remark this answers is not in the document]
madisonhall38: Done

storage/buckets/disable_soft_delete_policy.go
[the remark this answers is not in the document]
madisonhall38: I don't think I can use gax.WithTimeout -- storage.NewClient only accepts ctx and optional ClientOption (https://pkg.go.dev/google.golang.org/api/option) args. Can't use gax.WithTimeout on the Update call as an argument either.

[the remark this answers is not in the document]
madisonhall38: RE infinite retries, I consulted with my colleague @BrennaEpp as I am new to the team. From her, "We override the linked timeout policies, so we do indeed do infinite retries and encourage customers to set their own timeouts. Constructing the client with a default timeout is a behaviour change we'd have to implement for all customers, or provide a way to pass in a timeout to the client that applies to each request as we do in Transfermanager https://pkg.go.dev/cloud.google.com/go/storage@v1.43.0/transfermanager#WithPerOpTimeout".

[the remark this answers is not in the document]
tritone: To provide some historical context -- I believe we added context timeouts to all storage samples as part of a PM AI a few years back due to a customer experiencing hangs/infinite retries in some cases.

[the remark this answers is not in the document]
telpirion: Okay, fair enough. I just wanted to make sure that we don't accidentally communicate some sort of unnecessary uncertainty around the client's behavior. That said, y'all are the Storage client experts, so I will defer to you!

PR #5243, opened 2025-04-03
feat(storage): add samples and tests for object soft delete (list, list versions, restore)

### LINKED ISSUE #4789: storage: samples for Soft Delete (objects)

madisonhall38:
Part 2 - samples for object soft delete (part 1 in https://github.com/GoogleCloudPlatform/golang-samples/pull/5231)

Fixes #4789 

## Checklist


storage/objects/objects_test.go
```diff
@@ -677,3 +677,89 @@ func TestObjectRetention(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Object(%q).Attrs: %v", objectName, err)
+	}
+	generation := attrs.Generation
+	// Simulate soft deletion by deleting the object.
+	if err := obj.Delete(ctx); err != nil {
+		t.Fatalf("Object(%q).Delete: %v", objectName, err)
+	}
+
+	t.Run("listSoftDeletedObjects", func(t *testing.T) {
```
telpirion, line 721: nit: move each individual test case into its own function. I would prefer that sample : test function is 1:1. It's harder to read when test cases are nested in the test function--especially when debugging a broken test! 

The Samples Style Guide (implicitly) recommends this practice: https://googlecloudplatform.github.io/samples-style-guide/#dedicated-testing-per-sample

telpirion (approved):
Looks really good! I just have some concerns about the maintainability of the samples tests.

storage/objects/objects_test.go
madisonhall38: Done

storage/objects/list_soft_deleted_versions_of_object.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return fmt.Errorf("storage.NewClient: %w", err)
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+
+	it := client.Bucket(bucket).Objects(ctx, &storage.Query{
+		SoftDeleted: true,
```
tritone, line 43: This needs to include Query.Versions to get all generations: https://pkg.go.dev/cloud.google.com/go/storage#Query.Versions

storage/objects/restore_soft_deleted_object.go
```diff
@@ -0,0 +1,52 @@
...
+		return fmt.Errorf("storage.NewClient: %w", err)
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+
+	// Restore the soft-deleted object.
+	objHandle := client.Bucket(bucketName).Object(objectName).Generation(generation)
+	attrs, err := objHandle.Restore(ctx, &storage.RestoreOptions{})
```
tritone, line 44: I think you can just pass nil as the second param?

tritone (commented):
Sorry for the late review -- Two minor comments; you can resolve in a follow up PR.

storage/objects/list_soft_deleted_versions_of_object.go
madisonhall38: If I include both SoftDeleted and Versions in the query, I get "googleapi: Error 400: Cannot request both versions and soft_deleted, invalid"

I tried just doing query with Versions then check that Deleted timestamp is non zero, but from the docs, soft deleted results are by default excluded unless SoftDeleted is in the query, so that won't work. 

So I'm not sure what to do other than keep it as-is and beef up the test to have 2 objects like the Java sample test does (https://github.com/googleapis/java-storage/blob/main/samples/snippets/src/test/java/com/example/storage/ITObjectSnippets.java#L484)

PR to extend test: https://github.com/GoogleCloudPlatform/golang-samples/pull/5269

PR #5290, opened 2025-04-14
feat(pubsub): add message transform samples

hongalex:


pubsub/v1samples/subscriptions/create_message_transforms.go
```diff
@@ -0,0 +1,61 @@
...
+// [START pubsub_old_version_create_subscription_with_smt]
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/pubsub"
+)
+
+func createSubscriptionWithSMT(w io.Writer, projectID, subID string, topic *pubsub.Topic) error {
```
telpirion, line 26: nit: I've started asking contributors to provide a short comment describing what the sample function does and what each of the arguments is.

telpirion (approved):

pubsub/v1samples/subscriptions/create_message_transforms.go
hongalex: Yeah that makes sense. I added a comment to both new sample functions.

PR #5291, opened 2025-04-15
feat(cloudrun): add sample for service-to-service auth

OremGLG:
## Description
Add sample for service-to-service authentication.

Fixes b/409549637

## Checklist
- [~] Please **merge** this PR for me once it is approved


run/service-auth/receive.go
```diff
@@ -0,0 +1,65 @@
...
+	if len(strings.Split(authHeader, " ")) != 2 {
+		http.Error(w, "Malformed Authorization header", http.StatusBadRequest)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+
+	token := strings.Split(authHeader, " ")[1]
+
+	// Verify and decode the JWT
+	v, err := idtoken.NewValidator(r.Context(), option.WithHTTPClient(http.DefaultClient))
```
telpirion, line 50: nit: provide a more meaningful variable name for `v`.

telpirion, line 50: issue: move the instantiation of the `CallOption` object to its own line.

```go
callOption := option.WithHTTPClient(http.DefaultClient)
validator, err := idtoken.NewValidator(r.Context(), callOption)
```

telpirion (changes requested):
Some global comments and concerns about this PR:

+ Usage of `exec.Command()`: we should discourage this pattern and instead rely on Go client libraries. Replace as many of these `Command()` calls as possible with the corresponding client calls.
+ Deploying the service: I don't know whether deploying the service is a required step for this sample. Can we use the `httptest` package instead?

OremGLG:
> Some global comments and concerns about this PR:
> 
> * Usage of `exec.Command()`: we should discourage this pattern and instead rely on Go client libraries. Replace as many of these `Command()` calls as possible with the corresponding client calls.
> * Deploying the service: I don't know whether deploying the service is a required step for this sample. Can we use the `httptest` package instead?

I tried to follow the published [sample in python](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/run/service-auth/receive_test.py#L50-L67). Should the Python sample be updated as well?

OremGLG:
> Some global comments and concerns about this PR:
> 
> * Usage of `exec.Command()`: we should discourage this pattern and instead rely on Go client libraries. Replace as many of these `Command()` calls as possible with the corresponding client calls.
> * Deploying the service: I don't know whether deploying the service is a required step for this sample. Can we use the `httptest` package instead?

Answering the second question.

To follow the sample's "Authenticate service to service" duty, we need to deploy the application in CloudRun. This won't be possible if we run the service locally. Let me know if that answers your question, or if you need more context.

run/service-auth/receive.go
```diff
@@ -0,0 +1,63 @@
...
+	
+	if r.Method != http.MethodGet{
+		http.Error(w, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
+		return
+	}
+	
+	// Attempt to retrieve and validate the Authorization header.
+	authHeader := r.Header.Get("Authorization")
+	if authHeader == "" {
+		w.Write([]byte("Hello, anonymus user\n"))
```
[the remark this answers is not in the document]
OremGLG: Thanks for the correction

telpirion:
Splitting responses here: with regards to use of `exec.Command()`, we want customers to use the client libraries. We want to discourage invoking the `gcloud` cli programmatically. If the Python sample invokes the `gcloud` CLI programmatically, then yes, we want to refactor that.

telpirion:
> Answering the second question.
> 
> To follow the sample's "Authenticate service to service" duty, we need to deploy the application in CloudRun. This won't be possible if we run the service locally. Let me know if that answers your question, or if you need more context.

Okay, if that is the samples intention, then deploying it makes sense.

OremGLG:
> Splitting responses here: with regards to use of `exec.Command()`, we want customers to use the client libraries. We want to discourage invoking the `gcloud` cli programmatically. If the Python sample invokes the `gcloud` CLI programmatically, then yes, we want to refactor that.

@eapl-gemugami, the python sample should be updated as well.

eapl-gemugami:
@telpirion I found that replacing the command `gcloud run deploy --source . --allow-unauthenticated` with a Client Library requires many steps (compressing the source code, creating and uploading an image, overriding defaults, deploying the container, setting up IAM to `allUsers`). There are a few other commands, for example to delete the service, although `tun deploy` could be the most difficult one.

After that research, internal discussion with @OremGLG and later with @iennae, I'd say that as this deployment is made in a Unit Test rather than in a sample, it's not worth the refactor in this specific case. Do you agree?

PR #5309, opened 2025-04-30
fix(pubsub): update message transform samples

hongalex:
There was a bug in the existing samples where a pointer to a JavascriptUDF was passed instead of the value itself. This meant the library wasn't properly converting these and the message transform wasn't being created.

Also removes the extra quotation in the message transform functions.


telpirion (approved):

hongalex:
Thanks for pushing this through @telpirion

PR #5343, opened 2025-07-01
docs(pubsub): add comments for publisher instantiation

hongalex:


pubsub/topics/publish.go
```diff
@@ -34,9 +34,12 @@ func publish(w io.Writer, projectID, topicID, msg string) error {
 	}
 	defer client.Close()
 
-	// Make sure to reuse this publisher across publishes.
-	p := client.Publisher(topicID)
-	result := p.Publish(ctx, &pubsub.Message{
+	// client.Publisher can be passed a topic ID (e.g. "my-topic") or
+	// a fully qualified name (e.g. "projects/my-project/topics/my-topic").
+	// If a topic ID is provided, the project ID from the client is used.
+	// Make sure to reuse this publisher for all publish calls.
```
quartzmo: > Make sure to reuse this publisher for all publish calls.

Do you think you could add an explanation for this advice? For example: Make sure to reuse this publisher for all publish calls, since each publisher holds its own ... bla bla"

hongalex: Yeah I'll switch to something like "Make sure to reuse this publisher for all publish calls to make sure messages are batched properly".

PR #5352, opened 2025-07-22
feat(pubsub/v2): add single message transform samples

hongalex:
Migrates the [existing v1 SMT samples](https://github.com/GoogleCloudPlatform/golang-samples/pull/5290) to use the v2 library.


pubsub/subscriptions/create_message_transforms.go
```diff
@@ -0,0 +1,68 @@
...
+	"context"
+	"fmt"
+	"io"
+
+	"cloud.google.com/go/pubsub/v2"
+	"cloud.google.com/go/pubsub/v2/apiv1/pubsubpb"
+)
+
+// createSubscriptionWithSMT creates a subscription with a single message transform function applied.
+func createSubscriptionWithSMT(w io.Writer, projectID, topicID, subID string) error {
```
briandorsey, line 28: The example function signature is different than the [V1 version](https://github.com/GoogleCloudPlatform/golang-samples/blob/d3bdadadac8b9ff046d23249d3ce19681f8e1061/pubsub/v1samples/subscriptions/create_message_transforms.go#L27). Is that by design?

pubsub/subscriptions/subscription_test.go
```diff
@@ -731,3 +733,34 @@ func createOrGetStorageBucket(projectID, bucketID string) error {
 
 	return nil
 }
+
+func TestCreateSubscriptionWithSMT(t *testing.T) {
```
briandorsey, line 737: If (and only if) you decide to pass a Topic struct to createSubscriptionWithSMT, it would probably make sense to update the test to the same format I used in https://github.com/GoogleCloudPlatform/golang-samples/pull/5381. (Avoids a race condition when setting the topic variable.) Not needed if createSubscriptionWithSMT stays as is.

briandorsey (approved):
LGTM (with one question to double check the example function's format).

PR #5368, opened 2025-08-19
feat(cloudrun): create sample service for cloud run readiness probe and service health feature

mysterytony:
## Description

Create a sample service for Cloud Run to demonstrate the upcoming readiness probe & service health feature

## Checklist
  - ran this command from `run/testing` directory `GOLANG_SAMPLES_E2E_TEST=true GOLANG_SAMPLES_PROJECT_ID=lihenan-s8s-networking-test go test -run ^TestServiceHealth$ -v .`


mysterytony:
```
To fix this check, run "goimports -w . && find . -name go.mod -execdir go mod tidy \;"
HEAD detached at pull/5368/merge
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   functions/functionsv2/imagemagick/go.mod

no changes added to commit (use "git add" and/or "git commit -a")
```

go lint is complaining about go.mod from functions directory

mysterytony:
https://btx.cloud.google.com/invocations/c0a39662-c209-4994-bbc2-387355fb9dcb/targets is not failing run/testing tests 🎉

run/service-health/main.go
```diff
@@ -0,0 +1,579 @@
...
+	go func() {
+		for {
+			time.Sleep(1 * time.Second)
+			mutex.Lock()
+			if err := cache(); err != nil {
+				log.Printf("%+v", err)
+			}
+			mutex.Unlock()
+		}
+	}()
```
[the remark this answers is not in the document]
mysterytony: performance is not a too big of a concern, blocking the rootRequestHandler is necessary so it will wait for the first GCS read before returning the http response

run/service-health/main.go
```diff
@@ -0,0 +1,604 @@
...
+			} `json:"spec"`
+		} `json:"template"`
+	} `json:"spec"`
+}
+
+func init() {
+	var err error
+
+	var longRegion string
+	if longRegion, err = queryMetadataServer("/computeMetadata/v1/instance/region"); err != nil {
```
muncus, line 104: please use the metadata client from https://pkg.go.dev/cloud.google.com/go/compute/metadata - it has some builtin retry logic that makes it more robust than raw http, and avoids hardcoding many of these strings.

run/service-health/main.go
```diff
@@ -0,0 +1,604 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	accessToken, err := getAccessToken()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	apiURL := fmt.Sprintf(
```
muncus, line 133: again, you should be using a client library. https://pkg.go.dev/cloud.google.com/go/run

this will greatly simplify this code by removing the hardcoded urls, the structs for returned objects, and the authentication wrangling.

run/service-health/main.go
```diff
@@ -0,0 +1,604 @@
...
+	go func() {
+		for {
+			time.Sleep(time.Second)
+			if err := refreshReadinessConfig(); err != nil {
+				log.Print(err)
+			}
+		}
+	}()
+
+	go func() {
```
muncus, line 189: spawning goroutines in samples is generally frowned upon unless there's some concurrent processing necessary.

these seem to be tidying up the GCS cache? what is the significance of the GCS cache to this specific sample? it seems to be adding a **lot** of complexity here and distracting from readiness probes entirely.

internal/cloudrunci/gcs.go
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cloudrunci
+
+import (
+	"os/exec"
+)
+
+func DeleteBucketAndContents(bucketName string) ([]byte, error) {
```
muncus: Similar functionality exists already: 
https://github.com/GoogleCloudPlatform/golang-samples/blob/0f9d83276109eb453cf6bd6ca044bd9b99528545/internal/testutil/storage_cleaner.go#L65

internal/cloudrunci/gcs_test.go
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+package cloudrunci
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestDeleteBucketAndContents(t *testing.T) {
```
muncus: this test does not actually validate the intended behavior of your method. remove it.

run/service-health/go.mod
```diff
@@ -0,0 +1,57 @@
+module service_health
```
muncus: (minor) Go does not usually have separators in package or directory names. directory name and package name should match.
see the style guide https://google.github.io/styleguide/go/decisions#package-names

muncus (changes requested):
When structuring samples, please consider exactly what you're intending to demonstrate. You may want to read the [samples style guide](https://googlecloudplatform.github.io/samples-style-guide/) for additional advice.

This sample can be simplified a lot by removing the caching, and using appropriate API clients instead of rolling your own.

run/service-health/main.go
mysterytony: the go client doesn't have this "readiness probe" field in the container type yet, but it is in v1 API

run/service-health/main.go
mysterytony: Seems like Cloud Run metadata server doesn't exactly have the same interface as GCE metadata server. For example "computeMetadata/v1/instance/region" not present in GCE metadata server.

run/service-health/go.mod
mysterytony: oh the style guide said it should be "short and contain only lowercase letters."

but I see other examples like `module github.com/GoogleCloudPlatform/golang-samples/run/hello-broken`

should I use just `module servicehealth` or `module github.com/GoogleCloudPlatform/golang-samples/run/service-health`?

run/service-health/main.go
mysterytony: image

The UI looks like this. The main reason to leverage GCS is to manipulate the "readiness" state for other Cloud Run instances. If enough readiness status are reporting unhealthy, and if the Cloud Run service is attached to a load balancer, that load balancer will failover the traffic to a different region. This is why each running instance has to periodically sync from GCS.

mysterytony:
@briandorsey This PR should be ready for review, thank you in advance!

mysterytony:
> Could I ask you to make a follow up PR to update this sample code to use the client libraries once they do have the support you need?

Yes I will do that.

briandorsey:
I've taken over this review from @muncus, so I'm going to go head and merge.

PR #5450, opened 2026-01-16
feat(spanner): add sample for setting spanner read lock mode

shobhitsg:
## Checklist
- [X] **Tests** pass:   `go test -v ./..` (see [Testing](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#testing)) - `go test -v -run TestSample`

### DIFF
--- a/spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
+++ b/spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
@@ -32,7 +32,7 @@ import (
 func writeWithTransactionUsingReadLockMode(w io.Writer, db string) error {
 	ctx := context.Background()
 
-    // Client-level configuration: Applies to all read-write transactions
+	// Client-level configuration: Applies to all read-write transactions
 	// for this client. OPTIMISTIC mode avoids locks during reads and
 	// verifies changes during the commit phase.
 	cfg := spanner.ClientConfig{
@@ -46,7 +46,7 @@ func writeWithTransactionUsingReadLockMode(w io.Writer, db string) error {
 	}
 	defer client.Close()
 
-    // Transaction-level options take precedence over client-level
+	// Transaction-level options take precedence over client-level
 	// configuration. PESSIMISTIC mode is used here to override the
 	// client-level setting and ensure immediate locking during reads.
 	txnOpts := spanner.TransactionOptions{
@@ -54,7 +54,7 @@ func writeWithTransactionUsingReadLockMode(w io.Writer, db string) error {
 	}
 
 	_, err = client.ReadWriteTransactionWithOptions(ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
-        // In PESSIMISTIC mode with SERIALIZABLE isolation, the transaction
+		// In PESSIMISTIC mode with SERIALIZABLE isolation, the transaction
 		// acquires a shared lock during this read.
 		key := spanner.Key{1, 2}
 		row, err := txn.ReadRow(ctx, "Albums", key, []string{"AlbumTitle"})


spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
```diff
@@ -0,0 +1,90 @@
...
+"context"
+"fmt"
+"io"
+
+"cloud.google.com/go/spanner"
+pb "cloud.google.com/go/spanner/apiv1/spannerpb"
+)
+
+func writeWithTransactionUsingReadLockMode(w io.Writer, db string) error {
+	ctx := context.Background()
```
iennae, line 33: based on the type of operation, does this need a timeout?

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// The read lock mode specified at the client-level will be applied
```
iennae: I'm not sure I'm reading this comment correctly. Does ReadWrite_OPTIMISTIC not acquire a lock on read? "read lock mode" applied to all RW transactions is unclear to me. 

It might be helpful to specify what kind of workloads this is good for in case someone copies and pastes this code as is.

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	client, err := spanner.NewClientWithConfig(ctx, db, cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create client: %v", err)
+	}
+	defer client.Close()
+
+	// The read lock mode specified at the transaction-level takes
+	// precedence over the read lock mode configured at the client-level.
+	// PESSIMISTIC is used here to demonstrate overriding the client-level setting.
```
iennae: nit: request improved phrasing here something like the following. As it is, it's confusing as it's not explaining the what (or are we just showing how to override the client level setting and not the why you'd do it?)

// PESSIMISTIC: acquires locks immediately upon reading. Prevents other transactions 
// from modifying the data until this transaction commits.
// Best for high-contention workloads.

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
```diff
@@ -0,0 +1,90 @@
...
+	txnOpts := spanner.TransactionOptions{
+		ReadLockMode: pb.TransactionOptions_ReadWrite_PESSIMISTIC,
+	}
+
+	_, err = client.ReadWriteTransactionWithOptions(ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
+		// Read the current album title
```
iennae: Should we state that it's now PESSIMISTIC lock mode, so the lock is acquired immediately?

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
```diff
@@ -0,0 +1,90 @@
...
+		key := spanner.Key{1, 2}
+		row, err := txn.ReadRow(ctx, "Albums", key, []string{"AlbumTitle"})
+		if err != nil {
+			return fmt.Errorf("failed to read album: %v", err)
```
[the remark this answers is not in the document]
iennae: This is a valid and current recommendation.

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		var title string
+		if err := row.Column(0, &title); err != nil {
+			return fmt.Errorf("failed to get album title: %v", err)
```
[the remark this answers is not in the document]
iennae: This is another valid and current recommendation.

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
```diff
@@ -0,0 +1,90 @@
...
+				WHERE SingerId = @SingerId AND AlbumId = @AlbumId`,
+			Params: map[string]interface{}{
+				"SingerId":   1,
+				"AlbumId":    1,
+				"AlbumTitle": "New Album Title",
+			},
+		}
+		count, err := txn.Update(ctx, stmt)
+		if err != nil {
+			return fmt.Errorf("failed to update album: %v", err)
```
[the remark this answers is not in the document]
iennae: As earlier, this is valid and current best practice

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
```diff
@@ -0,0 +1,90 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spanner
+
+// [START spanner_read_lock_mode]
```
iennae, line 17: I'm confused about the intent here with this region tag naming. Is the goal to show ReadLockMode: options? is it to show overrides? can you add a little more context (might be helpful to have a comment such as 

// writeWithTransactionUsingReadLockMode demonstrates how to set the ReadLockMode
// for Cloud Spanner transactions. It shows how to configure a default mode at the
// client level and how to override it for specific transactions.

as a brief explainer. 

Or you might want to add a wrapper around the transaction part to explicitly call it out with a secondary region tag? something like 
spanner_transaction_read_lock_mode

iennae (commented):
This looks to be a good addition for the Spanner samples and it's almost ready to go.

I’ve left some specific line comments regarding the documentation and intent, but at a high level, there are two main areas to address:

- Modern Go Practices: Please update the error handling to use %w instead of %v in your fmt.Errorf calls. This ensures the errors are wrapped correctly, which is our standard for these repos.
- Clarity on Locking Logic: I've suggested some expanded comments around PESSIMISTIC vs OPTIMISTIC modes. Since these are samples, we want to make sure the user understands the 'why' (e.g., contention handling) and not just the 'how.'

I also raised a question about the region tag naming—let’s make sure those tags align with how we want to surface this in the documentation.

Once we tighten up those explanations and the error wrapping, I’m happy to LGTM!"

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
shobhitsg: The goal is to show both - various ways to pass the read lock mode option and their order of precedence. I've added the comment as well.
region tag is as per the requirement and will be used in our upcoming docs as well (similar to [this](https://docs.cloud.google.com/spanner/docs/use-repeatable-read-isolation#client-libraries)).

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
shobhitsg: The purpose here is to explain various ways and how one setting overrides the other. We cannot effectively discuss how to use it without first covering the isolation level. e.g. with Repeatable read isolation level, PESSIMISTIC won't acquire the locks upon reading except if we use `for update`. These examples will be used in our official docs (similar to [this](https://docs.cloud.google.com/spanner/docs/use-repeatable-read-isolation#client-libraries)).

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
shobhitsg: There are three transaction types: ReadOnly (RO), PartitionedDml (PD), and ReadWrite (RW). 
Isolation level and Read lock mode configurations apply exclusively to ReadWrite (RW) transactions. e.g., these configurations are still applicable even if RW transaction only involves reads, though they won't significantly impact the outcome.
I'm considering that customers will go through our official doc before starting to use these in their workloads. Since the combination of isolation level and read lock mode is highly specific to Spanner, understanding these concepts is crucial before usage. These are discussed in detail in our upcoming official docs and same docs  will point to these code samples (using `spanner_read_lock_mode` tag).

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
shobhitsg: Not necessarily; it's a simple transaction shape and should complete quickly. I'm assuming this based on how other similar samples are written.

spanner/spanner_snippets/spanner/spanner_read_lock_mode.go
shobhitsg: Updated

PR #5582, opened 2026-05-19
feat: add client side metrics for redis

fosky94:
## Description

Added a new sample code to log client side metrics and traces for redis. this will be added to the public docs once this sample code is posted

client libraries that need to be installed:

go get github.com/gomodule/redigo/redis@latest
go get go.opentelemetry.io/otel
go get go.opentelemetry.io/otel/sdk/trace
go get go.opentelemetry.io/otel/sdk/metric
go get github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace
go get github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric

### DIFF
--- a/opentelemetry/instrumentation/docker-compose.yaml
+++ b/opentelemetry/instrumentation/docker-compose.yaml
@@ -21,7 +21,7 @@ services:
       - GOOGLE_CLOUD_QUOTA_PROJECT
       - GOOGLE_CLOUD_PROJECT=${GOOGLE_CLOUD_PROJECT?}
   loadgen:
-    image: golang:1.22
+    image: golang:1.25
     command: ["go", "run", "github.com/rakyll/hey@latest", "-c=2", "-q=1", "http://app:8080/multi"]
     depends_on:
       - "app"
--- a/opentelemetry/instrumentation/otel-collector-config.yaml
+++ b/opentelemetry/instrumentation/otel-collector-config.yaml
@@ -10,7 +10,21 @@ receivers:
     include:
     - "/var/log/app.log"
     operators:
+      - type: router
+        id: log_router
+        routes:
+          - expr: 'body matches "^\\{"'
+            output: json_parser
+        default: bypass_non_json
+
+      - type: add
+        id: bypass_non_json
+        field: attributes.is_json
+        value: false
+        output: end_of_pipeline
+
       - type: json_parser
+        id: json_parser
         parse_to: body
         timestamp:
           parse_from: body.timestamp
@@ -29,47 +43,76 @@ receivers:
             fatal: critical
             fatal3: alert
             fatal4: emergency
+        output: check_trace_context
+
+      # Check if the parsed JSON actually contains trace context before extracting
+      - type: router
+        id: check_trace_context
+        routes:
+          - expr: '"logging.googleapis.com/trace" in body'
+            output: add_trace_flags
+        default: end_of_pipeline
 
       # set trace_flags to SAMPLED if GCP attribute is set to true
       - type: add
+        id: add_trace_flags
         field: body.trace_flags
         value: "01"
         if: body["logging.googleapis.com/trace_sampled"] == true
+        output: regex_parser
 
       # parse the trace context fields from GCP attributes
       - type: regex_parser
+        id: regex_parser
         parse_from: body["logging.googleapis.com/trace"]
         parse_to: body
         regex: (?P<trace_id>.*)
         trace:
           span_id:
             parse_from: body["logging.googleapis.com/spanId"]
+        output: remove_timestamp
 
       # Remove fields that are redundant from translation above
       - type: remove
+        id: remove_timestamp
         field: body.timestamp
+
       - type: remove
+        id: remove_trace_id
         field: body.trace_id
+
       - type: remove
+        id: remove_trace_flags
         field: body.trace_flags
+
       - type: remove
+        id: remove_severity
         field: body.severity
+
       - type: remove
+        id: remove_gcp_trace
         field: body["logging.googleapis.com/trace"]
+
       - type: remove
+        id: remove_gcp_spanid
         field: body["logging.googleapis.com/spanId"]
+
       - type: remove
+        id: remove_gcp_trace_sampled
         field: body["logging.googleapis.com/trace_sampled"]
 
+      - type: add
+        id: end_of_pipeline
+        field: attributes.processed_by_collector
+        value: true
+
 exporters:
-  # Export logs and traces using the standard googelcloud exporter
+  # Export logs using the standard googlecloud exporter
   googlecloud:
     project: ${GOOGLE_CLOUD_PROJECT}
     log:
       default_log_name: "opentelemetry.io/collector-exported-log"
-  # Export metrics to Google Managed service for Prometheus
-  googlemanagedprometheus:
-    project: ${GOOGLE_CLOUD_PROJECT}
+  # Export telemetry using standard OTLP HTTP exporter to Google Cloud
   otlphttp:
     encoding: proto
     endpoint: https://telemetry.googleapis.com
@@ -95,7 +138,8 @@ processors:
       - { key: "cluster", value: "no-cluster", action: "upsert" }
       - { key: "namespace", value: "no-namespace", action: "upsert" }
       - { key: "job", value: "us-job", action: "upsert" }
-      - { key: "instance", value: "us-instance", action: "upsert" }
+      - { key: "instance", from_attribute: "service.instance.id", action: "upsert" }
+      - { key: "instance", value: "us-instance", action: "insert" }
   # If running on GCP (e.g. on GKE), detect resource attributes from the environment.
   resourcedetection:
     detectors: ["env", "gcp"]
@@ -118,12 +162,12 @@ service:
   pipelines:
     traces:
       receivers: ["otlp"]
-      processors: ["batch", "resourcedetection", resource/gcp_project_id]
+      processors: ["batch", "resourcedetection", "resource/gcp_project_id"]
       exporters: ["otlphttp"]
     metrics:
       receivers: ["otlp"]
-      processors: ["batch", "resourcedetection", "resource"]
-      exporters: ["googlemanagedprometheus"]
+      processors: ["batch", "resourcedetection", "resource/gcp_project_id", "resource"]
+      exporters: ["otlphttp"]
     logs:
       receivers: ["filelog"]
       processors: ["batch", "resourcedetection"]


fosky94 (commented):
added region tags

fosky94 (commented):
feat: add client side metrics for redis

This commit introduces functionality to capture and expose client-side metrics for Memorystore for Redis interactions.

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,165 @@
...
+
+	metricExporter, err := gcpmetric.New()
+	if err != nil {
+		log.Fatalf("Failed to create metric exporter: %v", err)
+	}
+	mp := sdkmetric.NewMeterProvider(sdkmetric.WithReader(sdkmetric.NewPeriodicReader(metricExporter, sdkmetric.WithInterval(10*time.Second))))
+	otel.SetMeterProvider(mp)
+	meter := mp.Meter("redigo.metrics")
+
+	rttHist, _ = meter.Float64Histogram("redis_client_rtt", metric.WithUnit("ms"))
```
amcolin: Please do not ignore errors in Go, especially during initialization. If a metric name is malformed, OpenTelemetry will return an error, and your dashboard will be silently empty. Use `if err != nil` instead 🙏

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,165 @@
...
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/trace"
+
+	gcpmetric "github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric"
+	gcptrace "github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+var (
```
amcolin: suggestion: Change this var to a struct definition 
```suggestion
type MetricClient struct {
	tracer           trace.Tracer
	rttHist          metric.Float64Histogram
	clientBlockHist  metric.Float64Histogram
	appBlockHist     metric.Float64Histogram
	retryCounter     metric.Int64Counter
	connErrorCounter metric.Int64Counter
}
```

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,174 @@
...
+			if attempt >= maxRetries {
+				break
+			}
+			sleep(time.Duration(100<<attempt) * time.Millisecond)
+			continue
+		}
+
+		appStart := time.Now()
+		_ = fmt.Sprintf("%v", reply)
+		appBlockHist.Record(ctx, float64(time.Since(appStart).Milliseconds()), metricOpts)
```
amcolin: Since Redis operations often execute under 1ms, using `time.Since().Milliseconds()` will truncate those values to 0. To retain fractional milliseconds, we should use `Microseconds() / 1000.0` instead. Also, since this time conversion happens in multiple places, let's extract it into a quick helper function to keep the code clean.
```golang
// sinceMs calculates the time elapsed since the given start time 
// and returns it in fractional milliseconds (e.g., 1.5ms).
func sinceMs(start time.Time) float64 {
	return float64(time.Since(start).Microseconds()) / 1000.0
}
```

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,174 @@
...
+	span.SetAttributes(attribute.String("redis.command", commandName))
+	defer span.End()
+
+	maxRetries := 3
+	attempt := 0
+	metricOpts := metric.WithAttributes(attribute.String("operation", operationName))
+
+	for attempt < maxRetries {
+		poolStart := time.Now()
+		conn := pool.Get()
```
amcolin: I think we need to use the `pool.GetContext(ctx)` function here instead to avoid blocking indefinitely. Plus, we can handle any errors with the connection.
```suggestion
		conn, err := pool.GetContext()
```

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,174 @@
...
+			attempt++
+			if attempt >= maxRetries {
+				break
+			}
+			sleep(time.Duration(100<<attempt) * time.Millisecond)
+			continue
+		}
+
+		reqStart := time.Now()
+		reply, err := conn.Do(commandName, args...)
```
amcolin: I believe we also need to pass the context here. Can we investigate if we can replace this with `redis.DoContext` or `redis.ConnWithContext`

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,174 @@
...
+			continue
+		}
+
+		appStart := time.Now()
+		_ = fmt.Sprintf("%v", reply)
+		appBlockHist.Record(ctx, float64(time.Since(appStart).Milliseconds()), metricOpts)
+
+		return reply, nil
+	}
+	return nil, fmt.Errorf("max retries reached for %s", operationName)
```
amcolin: suggestion: Include the redis error here
```suggestion
	return nil, fmt.Errorf("max retries reached for %s: %w", operationName, lastErr)
```

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,165 @@
...
+	for attempt < maxRetries {
+		poolStart := time.Now()
+		conn := pool.Get()
+		clientBlockHist.Record(ctx, float64(time.Since(poolStart).Milliseconds()), metricOpts)
+
+		if err := conn.Err(); err != nil {
+			conn.Close()
+			connErrorCounter.Add(ctx, 1, metricOpts)
+			retryCounter.Add(ctx, 1, metricOpts)
+			span.RecordError(err) // Attach error to trace
```
amcolin: question: Don't we also need to flag the span as failed?
```suggestion
			span.RecordError(err) // Attach error to trace
			span.SetStatus(codes.Error, err.Error()) // Set span status
```

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,174 @@
...
+				break
+			}
+			sleep(time.Duration(100<<attempt) * time.Millisecond)
+			continue
+		}
+
+		appStart := time.Now()
+		_ = fmt.Sprintf("%v", reply)
+		appBlockHist.Record(ctx, float64(time.Since(appStart).Milliseconds()), metricOpts)
+
```
amcolin: Following up on the status question above: if the retry eventually succeeds, I think we should reset the status.
```golang
// Reset span status to Ok if a retry eventually succeeds
		span.SetStatus(codes.Ok, "")
```

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,174 @@
...
+func smartRedisCall(ctx context.Context, pool *redis.Pool, operationName string, commandName string, args ...interface{}) (interface{}, error) {
+	// Create a dedicated child span for the Redis command
+	ctx, span := tracer.Start(ctx, operationName)
+	span.SetAttributes(attribute.String("redis.command", commandName))
+	defer span.End()
+
+	maxRetries := 3
+	attempt := 0
+	metricOpts := metric.WithAttributes(attribute.String("operation", operationName))
+
```
amcolin: Store the error at this point so we can return it on the later
```golang
var lastErr error
```

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,174 @@
...
+		rttHist.Record(ctx, float64(time.Since(reqStart).Milliseconds()), metricOpts)
+		conn.Close()
+
+		if err != nil {
+			retryCounter.Add(ctx, 1, metricOpts)
+			span.RecordError(err) // Attach error to trace
+			attempt++
+			if attempt >= maxRetries {
+				break
+			}
```
amcolin:
```suggestion
			}
			lastErr = err
```

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,174 @@
...
+			sleep(time.Duration(100<<attempt) * time.Millisecond)
+			continue
+		}
+
+		appStart := time.Now()
+		_ = fmt.Sprintf("%v", reply)
```
amcolin: Can we use the `time.sleep` here instead the `Sprintf`?
```suggestion
time.Sleep(2 * time.Millisecond)
```

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,174 @@
...
+	maxRetries := 3
+	attempt := 0
+	metricOpts := metric.WithAttributes(attribute.String("operation", operationName))
+
+	for attempt < maxRetries {
+		poolStart := time.Now()
+		conn := pool.Get()
+		clientBlockHist.Record(ctx, float64(time.Since(poolStart).Milliseconds()), metricOpts)
+
+		if err := conn.Err(); err != nil {
```
amcolin: Adding a comment here might be useful for devs to understand the flow
```suggestion
       // Check if the connection is dead
		if err := conn.Err(); err != nil {
```

memorystore/redis/client_side_metrics/main_test.go
```diff
@@ -0,0 +1,233 @@
...
+	// Call GET which should trigger the retry logic upon seeing the conn.Err()
+	getReply, err := smartRedisCall(ctx, pool, "get_user", "GET", "user:123")
+	if err != nil {
+		t.Fatalf("smartRedisCall GET failed on retry: %v", err)
+	}
+	if s, ok := getReply.(string); !ok || s != "active" {
+		t.Errorf("GET reply got %v, want 'active'", getReply)
+	}
+
+	t.Logf("RetryOnConnError: dialCalls=%d, closeCalls=%d", sharedConn.dialCalls, sharedConn.closeCalls)
```
amcolin: I think we need and assert that we prove the retry happened
```golang
if sharedConn.dialCalls != 2 {
    t.Errorf("expected 2 Dial calls (1 fail, 1 retry), got %d", sharedConn.dialCalls)
}
```

amcolin (changes requested):
Excellent addition! Thank you for implementing this change. I have provided some feedback that could enhance the experience of working with this sample.

Also curious whether a table driven test implementation would suite better this case here is a brief example:
```golang
func TestSmartRedisCall(t *testing.T) {
	// Initialize our test mocks once
	initTestTelemetry()

	// Define our table of test cases
	tests := []struct {
		name            string
		mock            *mockRedisConn
		poolMaxIdle     int
		opName          string
		command         string
		args            []interface{}
		expectedVal     interface{}
		expectedErrStr  string 
		expectedDials   int
		expectedDoCalls int
	}{
```
Which follows this patter
https://github.com/GoogleCloudPlatform/golang-samples/blob/main/bigtable/reads/reads_test.go#L32

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: you are right! thanks! done!

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main.go
fosky94: done

memorystore/redis/client_side_metrics/main_test.go
fosky94: done

fosky94 (commented):
Thank you so much for your review!! It was really helpfull. I have applied all the changes. Thanks!

memorystore/redis/client_side_metrics/main.go
```diff
@@ -0,0 +1,235 @@
...
+// sinceMs calculates the time elapsed since the given start time
+// and returns it in fractional milliseconds (e.g., 1.5ms).
+func sinceMs(start time.Time) float64 {
+	return float64(time.Since(start).Microseconds()) / 1000.0
+}
+
+func initTelemetry(ctx context.Context) (*MetricClient, func()) {
+	traceExporter, err := gcptrace.New()
+	if err != nil {
+		log.Fatalf("Failed to create trace exporter: %v", err)
```
amcolin: One thing I forgot to mention in my last review is that we should follow the [contribution guidelines to return errors](https://github.com/GoogleCloudPlatform/golang-samples/blob/main/CONTRIBUTING.md#return-errors) here instead `log.Fatalf` 
```suggestion
		return fmt.Errorf("gcptrace.New: %w", err)
```

amcolin (commented):
Great work on these fixes! The code is looking really solid, and I've tested the GCP metrics integration which works like a charm. There's just one final piece of feedback to look at before we're good to go.

memorystore/redis/client_side_metrics/main.go
fosky94: done! I have changed the rest of the errors. thanks!

fosky94 (commented):
fixed error messages as instructed in the review. Thanks!

PR #5603, opened 2026-06-20
docs(vision): fix broken "Go Code" link for label detection

manumishra12:
The label-detection section of `vision/README.md` links to `label`, but there is no `label/` directory. The label-detection sample lives in `detect/` (`detectLabels` in `detect/detect.go`, under the `vision_label_detection` region — the sibling "Detect Content" section already links to `detect`). This points the link at `detect` so it resolves.

Docs-only, one line.


amcolin (approved):
Thanks for fixing the broken link 👍

PR #5675, opened 2026-08-13
feat(storage): support delete source objects on compose

nidhiii-27:
Update the Go GCS compose sample and tests to support the deleteSourceObjects option.


storage/objects/objects_test.go
```diff
@@ -274,17 +275,60 @@ func TestObjects(t *testing.T) {
...
+		for _, src := range []string{src1Temp, src2Temp} {
+			if _, err := bkt.Object(src).Attrs(ctx); !errors.Is(err, storage.ErrObjectNotExist) {
+				t.Errorf("Source object %q should have been deleted, but got err: %v", src, err)
+			}
+		}
+
+		// Clean up dstObjTemp
+		if err := bkt.Object(dstObjTemp).Delete(ctx); err != nil {
+			t.Errorf("failed to clean up destination object %q: %v", dstObjTemp, err)
+		}
```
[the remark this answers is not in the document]
nidhiii-27: Done

_Co-authored by AI Agent_

storage/objects/compose_object.go
```diff
@@ -47,11 +48,17 @@ func composeFile(w io.Writer, bucket, object1, object2, toObject string) error {
...
+	composer.DeleteSourceObjects = deleteSourceObjects
+	_, err = composer.Run(ctx)
 	if err != nil {
 		return fmt.Errorf("ComposerFrom: %w", err)
 	}
-	fmt.Fprintf(w, "New composite object %v was created by combining %v and %v\n", toObject, object1, object2)
+	if deleteSourceObjects {
+		fmt.Fprintf(w, "New composite object %v was created by combining %v and %v and the source objects were deleted.\n", toObject, object1, object2)
+	} else {
+		fmt.Fprintf(w, "New composite object %v was created by combining %v and %v\n", toObject, object1, object2)
```
cpriti-os: Should we clarify here as well that source objects were retained?

storage/objects/objects_test.go
```diff
@@ -274,17 +275,60 @@ func TestObjects(t *testing.T) {
...
+
+		// Verify source objects are deleted
+		for _, src := range []string{src1Temp, src2Temp} {
+			if _, err := bkt.Object(src).Attrs(ctx); !errors.Is(err, storage.ErrObjectNotExist) {
+				t.Errorf("Source object %q should have been deleted, but got err: %v", src, err)
+			}
+		}
+
+		// Clean up dstObjTemp
```
cpriti-os: Normal english would be better here like destination object?

storage/objects/objects_test.go
```diff
@@ -274,17 +275,60 @@ func TestObjects(t *testing.T) {
...
+
+		// Verify source objects still exist
+		for _, src := range []string{object1, object2} {
+			if _, err := bkt.Object(src).Attrs(ctx); err != nil {
+				t.Errorf("Source object %q should still exist, but got err: %v", src, err)
+			}
+		}
+
+		// Test with deleteSourceObjects = true
+		src1Temp := "compose-src-1-temp.txt"
```
cpriti-os: I could not understand the reason of temp suffix here.

storage/objects/objects_test.go
```diff
@@ -274,17 +275,60 @@ func TestObjects(t *testing.T) {
...
+		}
+		if err := uploadFile(io.Discard, bucket, src2Temp); err != nil {
+			t.Fatalf("uploadFile(%q): %v", src2Temp, err)
+		}
+
+		if err := composeFile(io.Discard, bucket, src1Temp, src2Temp, dstObjTemp, true); err != nil {
+			t.Errorf("composeFile(deleteSourceObjects=true): %v", err)
+		}
+
+		// Verify destination object was created
```
cpriti-os: Nit: As per Go readability, all sentence like comments should end with a period, fix it here and this could be a good tip for the agent for future PRs in Go!

cpriti-os (changes requested):

storage/objects/compose_object.go
nidhiii-27: Updated the print output to explicitly mention that the source objects were not deleted when `deleteSourceObjects` is set to false.

_Co-authored by AI Agent_

storage/objects/objects_test.go
nidhiii-27: Done. Renamed it to "temporary destination object".

_Co-authored by AI Agent_

storage/objects/objects_test.go
nidhiii-27: We use distinct source objects for the `deleteSourceObjects = true` test case to prevent deleting the shared test objects (`object1` and `object2`) that subsequent tests in the `TestObjects` suite rely on.

I've renamed these variables to `composeDeleteSrc1`, `composeDeleteSrc2`, and `composeDeleteDst` and added an explanatory comment.

_Co-authored by AI Agent_

storage/objects/objects_test.go
nidhiii-27: Done. Added periods to the end of sentence-like comments.

_Co-authored by AI Agent_

storage/objects/compose_object.go
```diff
@@ -47,11 +48,17 @@ func composeFile(w io.Writer, bucket, object1, object2, toObject string) error {
 
 	// ComposerFrom takes varargs, so you can put as many objects here
```
cpriti-os: nit: unrelated but good to fix, we can only have 32 source objects in a single compose https://docs.cloud.google.com/storage/quotas

cpriti-os (approved):

storage/objects/compose_object.go
nidhiii-27: Done. Updated the comment to clarify that ComposerFrom takes up to 32 source objects.

_Co-authored by AI Agent_

storage/objects/compose_object.go
```diff
@@ -45,13 +46,18 @@ func composeFile(w io.Writer, bucket, object1, object2, toObject string) error {
...
-	_, err = dst.ComposerFrom(src1, src2).Run(ctx)
+	// ComposerFrom takes varargs, so you can put up to 32 source objects.
+	composer := dst.ComposerFrom(src1, src2)
+	composer.DeleteSourceObjects = deleteSourceObjects
+	_, err = composer.Run(ctx)
 	if err != nil {
 		return fmt.Errorf("ComposerFrom: %w", err)
 	}
-	fmt.Fprintf(w, "New composite object %v was created by combining %v and %v\n", toObject, object1, object2)
+	if deleteSourceObjects {
```
amcolin, line 57: A small nitpick to make this a bit more Go-idiomatic: we can drop the else block. One option is to add an early return nil inside the if.

Another approach (which I'd recommend to avoid duplicating) is to extract the second sentence into a variable. Something like this:

```golang
    status := "Source objects were not deleted."
	if deleteSourceObjects {
		status = "The source objects were deleted."
	}

	fmt.Fprintf(w, "New composite object %v was created by combining %v and %v. %s\n", toObject, object1, object2, status)
	return nil
```

amcolin (changes requested):

storage/objects/compose_object.go
nidhiii-27: Done. Extracted the status message into a variable to avoid duplication and removed the else block.

_Co-authored by AI Agent_

amcolin (approved):
LGTM

